PR #29, opened 2016-01-30
WIP: Stored Value and Offline Payments In Stellar

nullstyle:
This PR presents a presently in progress design for a system to facilitate offline payments within a local economy.


cloudspeech:
I don't see how double spending or overdraft can be avoided in a fully offline situation (i.e. customer and merchant are both offline), since 1 customer can visit N merchants while offline and reset his balance everytime. Hence, I find it difficult to see how merchants can ever develop sufficient trust in the proposed offline scheme. Can these problems be addressed explicitly?

jedmccaleb:
I'm cleaning up the repo. This would make a great SEP. but closing till someone does that!

PR #42, opened 2017-08-04
SEP: /tx_status

jedmccaleb:


bartekn:
I think we should move `Compliance protocol` to this repo and make `/tx_status` part of it. BTW, we should also move other protocols here, probably in this order:
```
sep-0001: Stellar.toml
sep-0002: Federation protocol
sep-0003: Stellar Attachment Convention
sep-0004: Compliance protocol
```

jedmccaleb:
Yeah I agree they should be here. I don't think the numbering matters though.

PR #45, opened 2017-08-22
SEP-0001 fixes

bartekn:
I added a small fixes to `SEP-0001` but I wanted to outline some bigger issues:
* `recv_code` is not explained in a doc. Is it a code required to pickup cash?
* Tx Status endpoint does not include any authorization mechanism what means that everyone can check tx status. I think that `recv_code` shouldn't be public and `msg` can contain some private data.
* Not sure why current steps of Compliance protocol are included in a doc. This seems redundant and should be a separate SEP.

Also, can we change this repo settings to require at least one approval from other team member before merging? I think this SEP wasn't ready to be merged.

### DIFF
--- a/ecosystem/sep-0001.md
+++ b/ecosystem/sep-0001.md
@@ -1,11 +1,11 @@
 ## Preamble
 
 ```
-SEP: <to be assigned>
+SEP: 0001
 Title: Tx Status Endpoint
-Author: <list of authors' names and optionally, email addresses>
+Author: Jed McCaleb
 Status: Draft
-Created: <date created on, in ISO 8601 (yyyy-mm-dd) format>
+Created: 2017-08-05
 ```
 
 ## Simple Summary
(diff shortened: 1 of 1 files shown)


jedmccaleb:
recv_code is arbitrary extra info. It depends on the FIs involved and the status code

access to the endpoint should be handled elsewhere with http access password or white list IP or whatever else. 

It shouldn't have been given a number but it was fine to merge. The process says to merge before the SEP is given a number and finalized.

ecosystem/sep-0001.md
```diff
@@ -14,37 +14,46 @@ Simple way to get the status of a payment that has been sent through the complia
...
+```
+GET AUTH_SERVER/tx_status?id=tx_id
+```
+
+* `id` The Stellar transaction ID of of the payment we are interested in.
+
+### Response
+
+HTTP status code:
+* `404` if tx with a given ID has not been found.
```
jedmccaleb: I don't think we should 404. I think it should return `status` of `unknown`

bartekn:
> recv_code is arbitrary extra info. It depends on the FIs involved and the status code

OK, it should be explained in the doc. Why isn't it called `extra_info` but `recv_code`?

> access to the endpoint should be handled elsewhere with http access password or white list IP or whatever else.

Then if several FIs send transactions they can see statuses of others. I think this also should be standardized. Otherwise developers would need to additionally implement server/client code for authentication rather than simply pluging-in existing tools.

jedmccaleb:
I'm closing this. this is now SEP-0004.md if you want to make changes there.

PR #55, opened 2017-10-24
Add existing protocols and errors to Compliance protocol

bartekn:
* Moving existing protocols here:
  * `SEP-0001` - Stellar.toml
  * `SEP-0002` - Federation protocol
  * `SEP-0003` - Compliance protocol
* Existing `SEP`s were moved to `SEP-0004` and `SEP-0005`.
* Added errors in Compliance protocol (https://github.com/stellar/bridge-server/issues/70).
* Added more examples and reference implementations.

Questions:
* Do we have a list of allowed fields in stellar.toml? This should be in: `<<List of fields here>>`.

### DIFF
--- a/ecosystem/sep-0001.md
+++ b/ecosystem/sep-0001.md
@@ -2,74 +2,164 @@
 
 ```
 SEP: <to be assigned>
-Title: Tx Status Endpoint
+Title: stellar.toml
 Author: <list of authors' names and optionally, email addresses>
 Status: Draft
 Created: <date created on, in ISO 8601 (yyyy-mm-dd) format>
 ```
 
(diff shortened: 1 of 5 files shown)


jedmccaleb:
why renumber them? if you really want to let's get that PR/discussion for SEP-2 in first.

bartekn:
SEP-0001 (0004 in this PR) is mentioning Compliance protocol and SEP-0002 (0005 here) is talking about `stellar.toml`. I thought it's strange that they reference SEPs with higher ID.

jedmccaleb:
I did this with https://github.com/stellar/stellar-protocol/pull/57

PR #63, opened 2017-11-09
SEP-0005 Key Derivation Methods for Stellar Accounts

bartekn:
Links in "Implementation" section are currently in PRs:
* https://github.com/stellar/go/pull/78
* https://github.com/stellar/go/pull/150

@lenondupe would you mind checking if your Ledger app gives the same results as in "Test Cases" section? I'll try to get a test Ledger device by the end of next week.

@lenondupe @zulucrypto let me know if you want me to change your usernames to real names.


irisli:
Parity recommends users to not use a mnemonic as a password and I think the reasoning can be applied here too. Lets apply murphy's law. If mnemonic was used as a passphrase and a user may think they could just think of a new mnemonic password and use it. We need a way to prevent this from happening.

Off the top of my head, a error detection word (checksum) would prevent users from creating new passwords, and a error correction word would help mitigate issues with mistyping. Of course, this issue has been thought of a lot and I'm sure many cryptography UX experts have thought about this.

The key here is to use this only as a RECOVERY method. Additionally, more focus on the key derivation being DETERMINISTIC.

bartekn:
Yes, the mnemonic code is certainly not a password. The only context I used this word ("Test Cases" section) was a BIP-0039 _passphrase_. I will change it to _passphrase_. As you noticed, mnemonic codes should be used mainly to recovery and to move keys between wallets (also kind of recovery 😉).

When it comes to checksums, it's implemented in BIP-0039 [Generating the mnemonic](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki#generating-the-mnemonic) section.

irisli:
Okay, yeah i only just commented on the mnemonic stuff. I'm sure the mnemonics themselves are secure, but I think we need to put more emphasis recommending wallet creators to use mnemonics in a responsible way.

lenondupe:
@bartekn 
I could successfully validate tests 2 & 3, but Ledger won't let me enter a 15 word mnemonic. Only 12 and 18. So, so far seems to work brilliantly 👍

jedmccaleb:
bartek I changed the process for these SEPs a bit. The idea is that first they go into /drafts and are moved over to /ecosystem once they are approved.  I think this will be less confusing

bartekn:
@lenondupe that's a great news! Added additional test with 12 words.

@jedmccaleb yes, I've seen a new process, but I decided to make it "Accepted" directly because 1) it doesn't invent anything new, it's more like defining which standard to use in Stellar, 2) we don't have much room for changes/improvement here (maybe except derivation path: `m/44'/148'/x'`) because changing anything will make it not compatible with hardware wallets. But I can follow the process if you want. Let me know.

zulucrypto:
This is great! Thanks a lot for the work on this and to both you and @lenondupe for getting this standard figured out so quickly.

I'll use this same implementation in my PHP library and on the Trezor.

lenondupe:
12 word mnemonic works fine

JeremyRubin:
post-merge: Spec looks good to me, didn't look at the implementation.

darocha:
This is great! Is this already available on the js-stellar-sdk?

bartekn:
@darocha not sure if it will be ever added to js-stellar-sdk. There's already Node JS package for this: https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0005.md#other-implementations

darocha:
Hi Bartek, thanks for answering my question and for the link to the Node JS
package.

2018-03-02 13:59 GMT-05:00 Bartek Nowotarski <notifications@github.com>:

PR #64, opened 2017-11-12
Add PHP implementation

zulucrypto:


bartekn:
Thanks for this. Can we change it to something similar to BIPs, like:
```md
## Reference Implementation
* Console HD wallet: https://github.com/stellar/go/tree/master/tools/stellar-hd-wallet
* Go package for `ed25519` key derivation: https://github.com/stellar/go/tree/master/exp/crypto/derive

## Other Implementations
* PHP Stellar SDK: https://github.com/zulucrypto/stellar-api (See [Example](https://github.com/zulucrypto/stellar-api/blob/master/examples/mnemonic-address.php))
```

It looks good but I haven't used PHP for ages so I can't check thoroughly and I don't want readers to think this is a reference implementation they should test their code against.

zulucrypto:
@bartekn Sure thing, updated!

PR #67, opened 2017-11-15
accept sep-0006

jedmccaleb:


efritze:
No major issues with the **/deposit** endpoint.

I wish to propose changes for withdrawals.

1. *amount* should be a compulsory parameter and
2. *priority* an optional parameter. (I know that the *dest_extra* field exists.) 

------
Amount is such an important parameter it shouldn't be made optional. Some crypto currencies directly or indirectly determine transaction fees based on amount sent. 

(E.g BTC. Technically, BTC txn fees depend on the nature of your inputs, but the selected inputs depend on the amount you want to send).

(Traditional banks may also have different ways they handle withdrawal requests based on the amount requested).

------

The current draft proposes withdrawals should be handled with (forward) federation. I'm proposing a separate **/withdraw** endpoint, just like there's a separate /deposit endpoint. 


I am currently experiencing a blocking issue trying to implement federation for withdrawals as specified in SEP-0006. My withdrawal process uses a different backend (different server, different DB) from the normal federation service I provide. I do this so that I can scale my anchor service independently from my federation service. I don't see any easy way to use federation in this case. One solution will be to do REST calls from my federation service to my anchor service. Complicating the code.  /withdraw solves this issue beautifully, and doesn't make it any harder for services that do federation and withdrawals on the same server.

istrau2:
I have a bone to pick with the `deposit` endpoint.

I don't like how the `how` parameter depends on the type of asset. IMO `how` should ALWAYS be a human readable string. In the case of cryptocurrencies (or any other asset), specifying an address should be done through the `extra_info` field. This way clients that help with deposits of any asset don't need to differentiate between the kind of asset (not sure how to even go about doing that).

efritze:
I still think the withdrawal api needs an "amount" field, and it should be made mandatory.

The current withdrawal api design is useless for assets where withdrawals fees depend on the amount being withdrawn. It is also useless for assets where withdrawal instructions depend on the amount being withdrawn. 

I'm very familiar with two mobile money operators in Africa, and all two have withdrawal fees that depend on the amount being withdrawn.

I'm also familiar with around three banks in Africa, and all three have separate counters and instructions for small, medium and high withdrawal amounts. This could also be extended in the crypto world where there is one distribution account that handles tiny BTC withdrawals, and another distribution account that handles heavy BTC withdrawals.

PR #69, opened 2017-11-20
Add additional test vector to SEP-0005

bartekn:
This commit adds additional test vector for SEP-0005 generated using 16 bytes of entropy (filled with `0` bits). Some BIP-0039 packages have problems with 0-padded entropy and this vector will check this out.

@zulucrypto you may want to check it out in your PHP library.


zulucrypto:
@bartekn Thanks for the mention, I've added the test case

PR #95, opened 2018-04-16
draft for SEP on URI Scheme

nikhilsaraf:
This is the draft for the SEP for the URI Scheme. It may be easier to view the actual file to see the rendered markdown, [here](https://github.com/nikhilsaraf/stellar-protocol/blob/draft_uri_scheme/drafts/draft-0002.md)

__Simple Summary__: This Stellar Ecosystem Proposal introduces a URI Scheme that can be used to generate a URI that will serve as a request to sign a transaction. The URI (request) will typically be signed by the user’s trusted wallet where she stores her secret key(s).

__RFC__ - Please add any comments to this PR on how this specification can be improved to better serve the community.


drafts/draft-0002.md
```diff
@@ -0,0 +1,118 @@
...
+7. `msg` (optional) - There can be an optional `msg` query param to indicate any additional information that the website or application wants to show the user in her wallet. The value for this query param should be URL-encoded as well and should not be longer than 300 characters before the URL-encoding. Note that the `msg` field is _different_ from the `memo` field that is included in a transaction. The `msg` field will not be put on-chain, but the `memo` field will be put on-chain.
```
nullstyle: This is my least favorite portion of this proposal.  Generic messaging systems are always usually dangerous (see memo spam for an example in our own system) and the distinction between `msg` and `memo` will be very thin for users who aren't very experienced with stellar jargon.

nullstyle:
I'd love to see some discussion around the security implications of using a URL scheme for payment links.  For example, I'd like to see some discussion around scheme hijacking and if/how that's an issue for wallet developers.

drafts/draft-0002.md
jedmccaleb: I agree it is confusing but it seems like the system that generates the transaction needs someway to give the URI handler some context of what is going on.

drafts/draft-0002.md
```diff
@@ -0,0 +1,118 @@
...
+
+The scheme name will be `stellar` and should always be followed by a colon and two forward slashes to look like this: `stellar://`. The syntax for the URI will look like this: `stellar://<operation>/?<param1>=<value1>&<param2>=<value2>`, where `operation`, and the options for the query params are defined below.
+
+### Operation `tx`
+The `tx` operation represents a request to sign a specific XDR `Transaction`. The parameters for the `tx` op are as follows:
+1. `xdr` (required) -  A Stellar transaction in XDR format that is base64 encoded and then URL-encoded. If the existing source account uses the Master Account behind the network passphrase with an empty string (i.e. `keypair.Master("").Address()`) then the interpreting code will replace the source account and sequence number on this XDR `Transaction` with its own source account and sequence number before it signs it. If the source account is not the sentinel value and the sequence number is 0 then the interpreting code should replace only the sequence number before signing. See the reference implementation for implementation details.
```
bartekn: Using master key for the network as a sentinel value seems overcomplicated. Why not use byte slice filled with zeros instead?

drafts/draft-0002.md
```diff
@@ -0,0 +1,118 @@
...
+
+### Operation `pay`
+The `pay` operation represents a request to pay a specific address with a specific asset, regardless of the source asset used by the payer. If the payer decides to use a different source asset then the wallet should leverage the [path payment operation](https://www.stellar.org/developers/guides/concepts/list-of-operations.html#path-payment) to achieve this. The parameters for the `pay` operation are as follows:
+1. `destination` (required) - A valid account ID or payment address
+2. `amount` (required) - Amount that destination will receive
+3. `asset_code` (optional) - Asset code (XLM if not present) destination will receive
+4. `asset_issuer` (optional) - Account ID of asset issuer (XLM if not present) destination will receive
+5. `memo` (optional) - Can be a memo to be included in the payment / path payment. This should follow the normal guidelines for memos. This should also be URL-encoded.
```
bartekn: We should also have a param for `memo_type` and agree on encoding of `memo_hash`: hex or base64?

bartekn:
Other than my comments I think this SEP should have following features:
* There should be `network_passphrase` or just `network` param (with two predefined values: `public` and `testnet`). Without it, it will be hard to perform end-to-end testing.
* Payment requests should be signed. Without it, this solution doesn't prevent from phishing as any maliciousbrowser extension can easily change the params in-flight. Here's how I think it could work:
  * Add `origin` param to the request. If it's possible for an external app to check the origin of the request (domain) use this value and ignore the `origin` param.
  * Add `sig` param to the request which is a signature generated using `SIGNING_KEY` in stellar.toml file for a given domain.
  * Signing app should display origin domain to a user and check if the signature is valid.

drafts/draft-0002.md
poliha: The sequence number being updated here is that of the source account correct? This should stated clearly.

poliha: Agree with bartek here, this was a bit complicated to me when I read it at first. Can this be part of the parameters?  E.g. when `updateSource = true` the source should be updated, or something along those lines.

drafts/draft-0002.md
```diff
@@ -0,0 +1,121 @@
...
+6. `memo_type` (optional) - One of `MEMO_TEXT`, `MEMO_ID`,`MEMO_HASH`, `MEMO_RETURN`. See [transaction guide](https://www.stellar.org/developers/guides/concepts/transactions.html#memo) for a description of these values. 
+7. `submit` (optional) - If this value is omitted then the URI handler should create and submit the proper transaction to the network. If the value is present then it should be interpreted as a URL-encoded callback. The URL-encoded callback will be prefixed with its own namespace to denote whether this is a `url` callback or some other form of callback. In the case where it is a url callback (denoted by `url:`) the URI handler should send the signed XDR to this url in a `POST` request with `Content-Type` set to `application/x-www-form-urlencoded` with the data fields `xdr` containing the signed XDR (URL-encoded). If there are any query params specified in the URL callback then those should be included in the URL when submitting. For now only `url` callback types are supported.
```
tomerweller: I'm concerned about only allowing the `url` callback type. We need to have a good story for how a client side app communicates with the signer app without setting up a backend. Specifically, I'm thinking about single page apps communicating with browser extensions (cryptokitties + metamask style).

One way to do so is to have the single page app register a [custom protocol handler](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/registerProtocolHandler) and have the submission callback refer to that. That can work but will create an awkward user experience (the browser will ask the user if it approves this webpage to handle "Stellar Response" links or something like that)

Usually, webpage <-> extension communication happen through `postMessage()`, but that requires the website to know of a specific extension in advance ([example](https://gist.github.com/johnschimmel/fad562a4352e93353a91)) rather than being able to communicate with any extension that supports the protocol.

I wonder if I'm missing another option here. @nullstyle @bartekn ?

jedmccaleb:
@tomerweller how would "whois" work? how would the URI handler tell the webpage?

tomerweller:
@jedmccaleb the `whois` operation will have a `submit` callback option, similar to the one in the other operations. 

This will provide a mechanism for "stellar enabled" websites to present personalized information for the user. Similarly, websites that have the metamask sdk can show users (that have the sdk installed) information about their account by asking the extension for the address.

johansten:
I'll add this as a comment, for "whois", since I'm doing the same thing(?) in QR.

https://galactictalk.org/d/33-log-in-via-qr-code-challenge-response

drafts/draft-0002.md
```diff
@@ -0,0 +1,121 @@
...
+1. `xdr` (required) -  A Stellar transaction in XDR format that is base64 encoded and then URL-encoded. If the source account in the xdr is all zeros then the URI handler should replace the source account and sequence number with the user's source account and sequence number before it signs it. If the source account is set and the sequence number is 0 then the URI handler should replace only the sequence number before signing. 
+2. `submit` (optional) - If this value is omitted then the signed XDR should be submitted to the network. If the value is present then it should be interpreted as a URL-encoded callback. The URL-encoded callback will be prefixed with its own namespace to denote whether this is a `url` callback or some other form of callback. In the case where it is a url callback (denoted by `url:`) the URI handler should send the signed XDR to this url in a `POST` request with `Content-Type` set to `application/x-www-form-urlencoded` with the data fields `xdr` containing the signed XDR (URL-encoded). If there are any query params specified in the URL callback then those should be included in the URL when submitting. For now only `url` callback types are supported.
```
pselden: Should it actually be defined what the behavior is if submit is not defined? Why is the default to submit to the network? I know it says _should_ -- but is that actually a requirement?

Consider a multisig account -- the application handling the signing request should be the one deciding how to coordinate the rest of the signatures if submit is not present. In its current state it seems like they would be required to submit a transaction that may not have the proper signatures. The note on multisig wallets seems to conflict with this parameter.

To provide a real world use case with StellarGuard:

1) Application submits request for payment.
2) Wallet handles the URI and builds and signs the transaction with one signature.
3) Wallet checks if it is a StellarGuard-protected multisig account and submits it to StellarGuard for authorization.
4) Post-authorization, StellarGuard can either sign and submit it to the network, or return the transaction as another URI for further handling.

Step 3 would fall apart if submit was not present and the wallet were required to submit it to the network.

It seems like this is probably covered by the multisig note (meaning does not need to be submitted), but likely just needs some clarifying wording around this part.

jedmccaleb: The URI handler can really do whatever it wants. The application is just giving the handler the info that it has that is presumably the user's intention. If the URI handler is managing a multi-sig account than yeah it would need to gather all the signatures before submitting to the network. I'll reword a bit to try to clarify

jedmccaleb:
Let's leave the whois to a later extension for this or another mechanism

drafts/draft-0002.md
pselden: Understood, and the change to "callback" improves the clarity. My concern was with the language "If this value is omitted then the signed XDR should be submitted to the network." which is not true in the multisig case (and maybe not even true in the normal case -- it's probably up to the wallet/user to decide to submit it or not).

MisterTicot:
Hi,

I'm implementing a similar protocol and have a working alpha version, see:

> https://galactictalk.org/d/1267-cosmic-link-alpha-release

The present draft include some features I had in my initial design that I removed after more serious research and implementation attempts.

I'll try to explain my choices here.

-----

I dropped the `/operation/?arguments...` semantic in favor of `/?operation&arguments...`. This is better because this can be handled straightforwardly as a query string. This doesn't need several pages or redirection for webservice to handle.

I ask that you match the semantic used in cosmic-lib, or at least express every operation using a query string in order to ease implementation.

-----

I dropped custom URI scheme in favor of universal links. Meaning, instead of 'stellar://' the link should start with any website URL. I provided 'cosmic.link' as a default but services such as exchanges and wallet could directly link to authenticator URL. Now this seems wrong as it centralize the process, but this simply was the less bad design. This obviously makes cosmic.link a target of choice, but I found interesting means to make the node secure enough. I'm open to write an article about those possibilities.

Custom URI scheme, however, are insecure by design as they may be hijacked by any program or application you install. Meaning, I can simply take the code of a wallet such as stargazer and include it in my new fancy program, then subscribe to 'stellar://' link handling on installation and go ahead for phishing attempt, like prompring for backup restoration. How are you going to prevent this ?

Another big drawback is that custom URI scheme can't be handled by webwallets at all, putting stellar port, stronghold, stroopy and such definitely out of the game.

Additionaly, clicking on such links without installed authenticator lead to an error, while universal links elegantly handle this case.

Custom URI scheme is being progressively deprecated by Apple for those reasons.

I ask that more time is given explorating seriously those issues, and that a solution to phishing attack is found if stellar:// have to be made standard.


-----

The statement that not all operation should have their own semantic because it may be complicated to handle new operations an so on is simply untrue. If you take a look at cosmic-lib you'll see that it is a matter of few lines to add/remove/change operations.

There may be cases where we may loose backward compatibility and broke links. By example, if an asset disappear or if an operation doesn't accept the same parameters anymore, or need a new one to be given. In any of those cases, XDR would break as well and it is library implementator's job to write some compatibility layer. 

There's a true value in having writeable/readable links and that's why I support federated address in cosmic link protocol. Think public adoption.

I ask that we adopt the semantic proposed in cosmic-lib, which mimic js-stellar-sdk and rules out any individual implementator preferences.

I ask for federated addresses to be supported in the standard.

-----

I've been rulling out the message feature. It sure seems nice, but it boils down to allowing an untrusted source to display a message in a trusted interface. This is very bad. Just think about what scammers are going to do with such a gift. Really, they will.

So, I'd say it's up to the transaction emitter to explain the transaction in its own interface. As protocol implentators, we could eventually give additional description to operations, such as displaying warning when master key is going to be set to 0, but that's it.

I ask that this feature is removed from the standard.

-----

I know I'm being fairly critic here. I've rewrote cosmic-lib several times in an attempt to find the most satisfying form for this protocol. Not to say there's nothing to improve, as there's nice ideas in this standard proposal I'll be happy to grab. My point is what looks good on the paper may appear like wrong at implementation time or after extensive searches about the actual possibilities. 

Please acknowledge that I'm only trying to have us finding the best way to implement this thing. We'll have to support it forever then, so better think twice about the design.

pselden:
https://github.com/stellar/stellar-protocol/pull/95#issuecomment-383369876 -- those are a lot of great points you've made and I agree that this decision should not be taken lightly, specifically the scheme hijacking issue that you and @nullstyle both mentioned.

From my reading of the Universal URL implementations for both iOS and Android -- it seems like you cannot use it as a general purpose "I just want something to be able to handle this" type of thing like you could with Deep Links. This sort of destroys the original use case that this is designed to solve, wouldn't it -- the origination point of the payment request will almost never know ahead of time how the user wants to handle the final signing/submitting.

I think we should figure out some concrete use-cases/stories that this SEP is aimed at supporting, and verify that what we've come up with satisfies it.

MisterTicot:
You are perfectly right.

This is why cosmic.link platform has the sole functionality of redirecting the link toward user defined authenticator. This setting is stored locally for each user and redirection happens automatically if requested by user. This is really safe only on smartphones were disk space is properly segregated, though.

This is for the use case where user's account or authenticator is not known ahead.

In most cases however, or for most big services, the user will have an account or some identification process and settings. Here the authenticator may be asked to the user, or derived from an authentification process involving the authenticator.

The game here is shortcutting the cosmic.link platform as much as possible to mitigate centralization.

Please note that nothing prevent the existence of several such platform, and a rising payment system may, by example, setup its own link redirection service.

Such design require to harden the redirecting service to the point that a successful attack would reach only a micro portion of the users. The most basic implementation of this requirement can be achieved using long-period user caching. A most evolved solution involve a mix of caching, service workers and update signing. I'll release my plans to solve this once I'll have a reliable proof-of-concept.

johansten:
@jedmccaleb 

If you're doing raw TXs it's quite helpful to know who the source accounts are.
Having a way of querying that info from the user is what makes that useful.
I'd prefer a generalized message signing functionality, but getting a signed pubkey is a start

johansten:
re: hijacking

You can redirect the URI handler all you want, but nothing gets signed without a private key.

jedmccaleb:
@johansten "f you're doing raw TXs it's quite helpful to know who the source accounts are.
Having a way of querying that info from the user is what makes that useful."  what do you mean?

The URI handler will know what the source accounts are since it will have the secret keys for them right?

@MisterTicot I think having a central service in the middle isn't really a tenable solution. What are the security concerns of doing it the way we proposed? Sure some other app can register as the URI handler but it doesn't have your secret key.

nikhilsaraf:
@pselden 

> "I think the most obvious one that comes to mind is some sort of phishing attempt. A seemingly innocuous app does an update and registers to handle stellar:// URIs"

A web page can only _request_ to be a handler for a specific URI scheme. The user will still need to explicitly allow that specific web page to handle links for that URI scheme.

pselden:
https://github.com/stellar/stellar-protocol/pull/95#issuecomment-384088206

On Android (at least previous versions, not sure what the latest version does), if multiple apps match the scheme in their manifests then a disambiguation dialog pops up that asks you what you want to open it in. This attack scenario involves a normal app (let's say a boring Calculator app) doing an update that changes its name/icon and registers a handler for stellar://. Next time the user clicks on a stellar link they get to choose between two "Lobstr Wallet" apps or whatever it decided to spoof.

Yes, the user still does have to choose to open the link in this new app, and this should be firing off alarm bells everything... but heck even today people clicked through MyEtherWallet SSL error pages and got tons of ETH stolen, so it's not out of the realm of possibility that it happens.

All that said, I don't think this should derail this proposal -- we just need to evaluate things and see what the risk vs reward is, and whether there can be any ways to mitigate this at the protocol-level or whether it just needs to be handled at the app level.

mfontanini:
@pselden in that example, if the Boring Calculator ends up handling the link, what can it do? It's not like the user will type down the secret key in these sort of apps, the key is handled by it transparently. At most the app could know what transaction you were trying to make, but it wouldn't be able to impersonate the user.

MisterTicot:
@mfontanini

The app could embark the UI of a well known open source wallet, and complain for a database update failure. Then ask for restoration from keystone or secret key.

Attacking using a known interface is powerfull because you inherit the confidence placed on such app, this how phishing works.

On a platform such as Android this would lead to a quick ban, and work only on a portion of user using the given app, and naive enough.

On platform such as Mac/Linux installanlr apps are kind of restricted as well, so we could assume that this kind of attack may not hit too hard at once.

On Windows however, lot's of doors are open ... Well I guess there's no much to do there except educating people to sign from phones using QR. This is the weak point.

MisterTicot:
Another possible attack would be changing the destination address and passing the link to the legitimate authenticator.

As someone else stated this attack is also possible using a corrupted browszr extension. Here also, this wouldn't be a mass attack and some prevention may helps, as well as the mechanism asking confirmation when sending to new address.

nikhilsaraf:
Updated the proposal to include suggestions about making it work for web wallets (`web+stellar` scheme name) and mitigating MITM attacks by defining a signature scheme for the URI requests.

johansten:
@jedmccaleb "The URI handler will know what the source accounts are since it will have the secret keys for them right?"

But the server will not. I'm thinking big here. Multi-party transactions.
Use whois/challenge-response/message signing to "login" to a service, so the server know who it is dealing with.

nikhilsaraf:
@johansten 
> But the server will not. I'm thinking big here. Multi-party transactions.
Use whois/challenge-response/message signing to "login" to a service, so the server know who it is dealing with.

The proposal includes provisions to support multiparty transactions. One of the bigger use cases (but not the only use case) for including the _callback_ is to support multisig transactions. Here is one way in which multi-party signature collection can be implemented:

The user clicks on a URI request and wants to complete the transaction using an account that requires 2+ signers. _The user needs to open the link with a wallet that supports orchestration of multi-sig URI request signing._ The user will already have registered with this multi-sig service by providing his/her account ID, or would hand over the signed transaction to this multi-sig coordination service without logging in. In either case the multi-sig coordination service has enough information to extract the user's account since the signed transaction would include the _source account_ which can be used by the the multi-sig coordination service. This multi-sig coordination service will register the user's account and the original URI request to start the process of multi-sig coordination.

As part of starting the process of multi-sig coordination, the coordination service will generate a new URI request with a `callback` endpoint of its _own_ server so it can _collate_ signatures before responding to the original URI request. This new URI request should also be signed by the multi-sig coordination service's `signature` and `origin_domain` to reflect it being a new request. The coordination service should include a `msg` denoting the original `origin_domain` so that the signers can correctly recognize the transaction's original source. Since these new URI requests will also be `tx` or `pay` requests they can easily be handled by any wallet that supports this URI Scheme. It is the user's responsibility to distribute these new URI requests, although the multi-sig coordination service could add some additional value here in helping in this process.

I hope this explanation is helpful. We can add this as a new section in the proposal if that would be useful.

johansten:
@nikhilsaraf 

I'm just pushing for a way the user can get their account id over to the server without copy/pasting. The less people have to copy/paste private keys or account ids, the better. Preferably, generic message signing would be included.

I get everything you're saying about multisig. I run one of few apps that even *has* support for that.

johansten:
OK, use case

Atomic swaps. I had a little PoC before for showing off p2p swaps

A pair of QR codes are displayed.

You scan one, and get a popup on you wallet that the site wants you to sign a message.

Your friend does the same with the other one.

QR codes disappear, only to display forms that lets you select an asset, and an amount.

Both of you do.

You press a 'swap' button, and on the backend the server sends signing requests to both of you.

Both of you get notifications in your wallet, you review the transaction, and you sign.

Tada. Atomic swaps.

Now do you see where having a way of getting an account ID over to a server is *very* useful?

johansten:
@nikhilsaraf 

After talking about standardizing payments on galactictalk,
I'd like to get your take on multi-sig accounts and the *callback* on `pay`.

I see people going for the low-hanging fruit, and relying on the callback only, and then we can't have nice things anymore.

pselden:
Just curious - is there any prior art for this from other cryptocurrencies? How do they handle it?

kuyawa:
Operations should be strictly named as they are in Stellar so different ops can be used like /setOptions /manageOffer etc making it future proof if new ops are added.

A basic example is setting inflation to lumenaut pool so the web page shows a link or QR with uri as:

web+stellar://setOptions/?inflationDest=GA2345...NAUT

nikhilsaraf:
@pselden 
> Just curious - is there any prior art for this from other cryptocurrencies? How do they handle it?

Here's how [bitcoin](https://github.com/bitcoin/bips/blob/master/bip-0021.mediawiki) and [ethereum](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-681.md) do it. This SEP is very close to the syntax used by them with our additions being `callback` to help close the loop with a response and `signature` for added security. Neither bitcoin nor ethereum provide a way to "respond" to the request with a signed transaction the way we do with the `callback`. I'll be updating the SEP to include references to their specs.

---

@kuyawa 
> Operations should be strictly named as they are in Stellar so different ops can be used like /setOptions /manageOffer etc making it future proof if new ops are added.
>
> A basic example is setting inflation to lumenaut pool so the web page shows a link or QR with uri as:
>
> web+stellar://setOptions/?inflationDest=GA2345...NAUT

Future-proofness is one of our [goals](https://github.com/nikhilsaraf/stellar-protocol/blob/draft_uri_scheme/drafts/draft-0002.md#motivation) with this SEP. The way we achieve this is by supporting the `tx` namespace which allows including any `xdr` into the request --this ensures that all current and future operations that are supported by stellar will be compatible with this URI, and wallets would automatically be able to handle these requests without a code change. You can read more about this in the [Rationale section](https://github.com/nikhilsaraf/stellar-protocol/blob/draft_uri_scheme/drafts/draft-0002.md#rationale).

The `pay` operation is a special case that helps to support [path payments](https://www.stellar.org/developers/horizon/reference/resources/path.html) because it is not possible for a website to know which asset a payer would like to use --but they can define the asset and amount they would like to accept which gives the payer the freedom to decide what asset to use when making the payment and also the path they want to use-- so this required a special namespace.

MisterTicot:
Hi @nikhilsaraf 

> Future-proofness is one of our goals with this SEP. The way we achieve this is by supporting the tx namespace which allows including any xdr into the request --this ensures that all current and future operations that are supported by stellar will be compatible with this URI, and wallets would automatically be able to handle these requests without a code change. You can read more about this in the Rationale section.

This is only true if operation are never modified. If we change any parameter in the operation semantic, XDR won't be compatible anymore. On the other hand, human-writted links may retains compatibility if protocol implementators update the translation layer carefully. Does Stellar design include that operations will never be modified ?

-----

About domain signatures, I suggest that the domain public key is kept in a DNS TXT record, as for Stellar DKIF protocol (https://galactictalk.org/d/789-domainkeys-identified-federation-dkif/29). With the current design, if an attacker get access to the server storage, she could change the links and the public key at the same time.

MisterTicot:
@nikhilsaraf
Those are good points. I didn't catch the specificity of the pay operation, and XDR indeed is easier to handle for wallet operators.

However I must insist about the query string. My proposal still include to keep operation in first position. This is
> web+stellar:?operation&parameters...

Like in
> web+stellar:?pay&parameters...

And
> web+stellar:?xdr=....&parameters..

Here the operations are `pay` and `xdr`. I dropped tx as it doesn't offer much here but if you want to keep it for consistency it could be :

> web+stellar:?tx&xdr...

It's as easy to parse but it won't require web wallet to create several different pages or special redirection. So at the end it's way more convenient, even if it doesn't look as intuitive.

Speaking of position and parsing, it could be nice to specify that signature should be in last position.

-----

My proposition about the public key is not valid for the other stellar.toml fields and is only about signing keys, which could authenticate stellar.toml by the way.

nikhilsaraf:
@MisterTicot 
> It's as easy to parse but it won't require web wallet to create several different pages or special redirection. So at the end it's way more convenient, even if it doesn't look as intuitive.

I'm not sure I understand why the web wallet would need to _"create several different pages or special redirection"_. Can you elaborate on this please?

---
> Speaking of position and parsing, it could be nice to specify that signature should be in last position.

We have already implied that the `signature` field should be in the last position in the [Request Signing section](https://github.com/nikhilsaraf/stellar-protocol/blob/draft_uri_scheme/drafts/draft-0002.md#request-signing) (point # 5) but I will update the SEP to make this explicit.

---
> My proposition about the public key is not valid for the other stellar.toml fields and is only about signing keys, which could authenticate stellar.toml by the way.

The `stellar.toml` already includes a `SIGNING_KEY` field that is used for compliance, [link in docs](https://www.stellar.org/developers/guides/concepts/stellar-toml.html#stellartoml-example). Additionally it also includes public keys for the issuer accounts which if changed by an attacker can have a significant impact to the issuer's assets. I think these are equally sensitive fields.

How would the `URI_REQUEST_SIGNING_KEY` be used to authenticate the `stellar.toml` file?

The SEP requires wallets to take [specific security precautions](https://github.com/nikhilsaraf/stellar-protocol/blob/draft_uri_scheme/drafts/draft-0002.md#uri-handler-compliance) to be fully compliant, which includes the pinning of the `URI_REQUEST_SIGNING_KEY` against the `origin_domain` and to notify the user if the `URI_REQUEST_SIGNING_KEY` ever changes. This will provide some protection as well.

MisterTicot:
@nikhilsaraf

I assume that

> web+stellar:tx?xdr=...

Get converted into

> wallet.org/tx?xdr...

And that two different pages will be required to handle `tx` in one and `pay` in another. This is the reason why I moved operations inside the query.

pselden:
Thank you for the clarification. I will mention that many people feel that the signature part of oauth 1 was the worst part about it, but I have no better alternative to offer at this time.

MisterTicot:
@nikhilsaraf

I understand better your point about stellar.toml. I can see my lack of knowledge about how everything works together. Pinning the public keys for signature checking is indeed a good way to do.

Signing the stellar.toml would be a way to guarantee its content. You could either sign the whole content, or preferably sign its shasum. Obviously the public key would have to be stored in another location. I'm saying this only to answer your question now that I mentioned it and I'm aware I'm out of scope there.

nikhilsaraf:
@MisterTicot 
> I assume that
> 
>> web+stellar:tx?xdr=...
> 
> Get converted into
> 
>> wallet.org/tx?xdr...
> 
> And that two different pages will be required to handle tx in one and pay in another. This is the reason why I moved operations inside the query.

`registerProtocolHandler` allows you to set a handler that takes in the URI as a `string`. It is up to you to decide how to parse it and what it should be converted to. See the [Syntax for registerProtocolHandler](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/registerProtocolHandler#Syntax) and the sample [webHandler.html](https://gist.github.com/nikhilsaraf/ff3ae46116b6ae6dbdcd1743ad9495ec#file-webhandler-html) included in the SEP for a PoC on how this can work.

---
@pselden
> Thank you for the clarification. I will mention that many people feel that the signature part of oauth 1 was the worst part about it, but I have no better alternative to offer at this time.

I understand your concerns and agree that the signatures would be tricky to get right. This is why it would be important to have the right libraries to handle this --something we would depend on the community to help with. The security provided by having signatures outweighs the cost of having to use these libraries, especially when libraries would likely be used to construct these URIs anyway.

---
@MisterTicot 
> I understand better your point about stellar.toml. I can see my lack of knowledge about how everything works together. Pinning the public keys for signature checking is indeed a good way to do.
>
> Signing the stellar.toml would be a way to guarantee its content. You could either sign the whole content, or preferably sign its shasum. Obviously the public key would have to be stored in another location. I'm saying this only to answer your question now that I mentioned it and I'm aware I'm out of scope there.

This is a great suggestion and should be included as a future SEP to strengthen the `stellar.toml`.

MisterTicot:
Hi @nikhilsaraf,

> registerProtocolHandler allows you to set a handler that takes in the URI as a string. It is up to you to decide how to parse it and what it should be converted to. See the Syntax for registerProtocolHandler and the sample webHandler.html included in the SEP for a PoC on how this can work.

I'm a bit lost here. Isn't this protocol supposed to allow sending transaction *between* services? How are we going to do if we don't specify how it should be done?

We need a standard way to write a query encoding a transaction, in such a way that if my wallet is Stellar Port I can sign at 
> https://stellarport.io/auth?{operation}&{parameters...}

And if my wallet is stroppy then 
> https://stroopy.org/sign?{operation}&{parameters...}

And a transaction editor could offer a service at
 > https://stellar.org/editor/?{operation}&{parameters...}

As well as a signature collector would allow to register a new transaction to synchronise at a given URL and so on.

This would allow each users to authenticate with different services without having to talk each service language each time. I mean, isn't that the point of this protocol?

I think you're getting my point that this would be easier using a query string so everybody speak the same language and can work together. I would be a shame to miss such an opportunity by ending saying : " ok each platforms issue links for itself and handle it as it want."

That's why I felt a bit perplex when cosmic link propocol have been said to be centralized; while the cosmic.link website itself is a service and the protocol is a syntax available to everybody.

I think something may have been overlooked here. Or I missed something again.

nikhilsaraf:
@MisterTicot 
> I'm a bit lost here. Isn't this protocol supposed to allow sending transaction between services?

This protocol is not meant _specifically_ to allow sending transaction between services --it can certainly be used for that but is **not the intended use**.

The ideal use case is to go to a shopping website (or exchange or hotel website etc.) to make a payment which generates a URI request that you can sign in your **_trusted_** desktop/web wallet. Quoting from the [Motivation section](https://github.com/nikhilsaraf/stellar-protocol/blob/draft_uri_scheme/drafts/draft-0002.md#motivation) of the SEP below:
```
Non-wallet applications want a way to have their users sign a transaction
that is generated by it without requiring it to ever see the user’s secret
key in any form. Users already have their preferences for a trusted wallet
where they are storing their secret keys.

This leaves room for a common protocol that can be implemented by both
applications that want transactions to be signed as well as wallets that can
sign transactions.
```

---
> if my wallet is Stellar Port I can sign at
>
>> https://stellarport.io/auth?{operation}&{parameters...}

If your wallet is stellarport then stellarport would generate a URI request for a payment (or change trust etc.) like `web+stellar:pay?...`. When you click this link it will open up the wallet that you have registered to handle the `web+stellar` scheme and you can complete the payment. Sellarport [checks to see if you've sent the payment](https://github.com/nikhilsaraf/stellar-protocol/blob/draft_uri_scheme/drafts/draft-0002.md#transaction-submission) and then proceeds in the workflow to reflect this payment. I've added a section on [User Experience](https://github.com/nikhilsaraf/stellar-protocol/blob/draft_uri_scheme/drafts/draft-0002.md#user-experience) to further clarify how this should work.

---
> This would allow each users to authenticate with different services without having to talk each service language each time. I mean, isn't that the point of this protocol?

This sounds more like [Single Sign-On](https://en.wikipedia.org/wiki/Single_sign-on), which is **not** the point of this protocol.

---
> That's why I felt a bit perplex when cosmic link propocol have been said to be centralized; while the cosmic.link website itself is a service and the protocol is a syntax available to everybody.
> 
> I think something may have been overlooked here. Or I missed something again.

There can certainly be a protocol that provides a standard way to send payment requests to any domain/service. If it were to exist I think it would be useful to have it specified in the `stellar.toml` file to be standardized in the same way that the `stellar.toml` hosts a link to the [`FEDERATION_SERVER`](https://www.stellar.org/developers/guides/concepts/stellar-toml.html#stellartoml-example).

In my opinion, such a standard should go through the SEP process and is independent from this SEP. I can see how such a SEP would use the URI Scheme specified in this SEP by passing around a URL-encoded URI Scheme so the new SEP would not need to reinvent the concept of transaction request. I hope this is useful in clarifying the goals of this SEP and how it's different from what cosmic.link is trying to achieve.

MisterTicot:
@nikhilsaraf

Ok then I got it.

nikhilsaraf:
You can find the SEP here: https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0007.md

johansten:
You never replied to me about tx and callback in regards to multisig.

MisterTicot:
@nikhilsaraf Please don't let such an ugly security hole as an example ;-)
```
--- webHandler1.html	2018-05-08 11:09:40.056256219 +0200
+++ webHandler2.html	2018-05-08 11:18:29.989358747 +0200
@@ -29,13 +29,15 @@
                 var url = new URL(url_string);
                 var req_url_string = url.searchParams.get("req");
                 console.log(req_url_string)
-                document.getElementById("req").innerHTML = req_url_string;
+                var req_url_node = document.createTextNode(req_url_string);
+                document.getElementById("req").appendChild(req_url_node);
 
                 // extract the xdr from the payment request
                 var req_url = new URL(req_url_string);
                 var xdr_string = req_url.searchParams.get("xdr");
                 console.log(xdr_string)
-                document.getElementById("xdr").innerHTML = xdr_string;
+                var xdr_node = document.createTextNode(xdr_string);
+                document.getElementById("xdr").appendChild(xdr_node);
             }
         </script>
     </head>
```

johansten:
There's so much wrong with this I don't even know where to start.

Everything relating to multisig coordination for starters. It would have been nice to have been listened to.

QR codes *can't* use the same encoding, since they will become too large, and if they are too large they can't be scanned using entry level camera phones.

MisterTicot:
@nikhilsaraf I tried the webHandler.html as a first step to implement the protocol and I got pretty bad results. I may have made an error, or something may be wrong somewhere. 

Anyway the webhandler.html is available for now at https://testing.stellar-authenticator.org/test.html

Lubuntu 16.04
* Firefox : Works like a charm. Register webhandler and cross-domain links works.
* Chomium/Opera/Webkits browsers: Doesn't register webhandler. Try to open URI with xdr-open but ungracely fails

Android 7:
* I tried chrome, brave, opera, firefox, firefox focus and edge. Got nothing at all. Clicking at the link does nothing. It seems the webhandler doesn't register.

Windows:
* Chrome: clicking on the link does nothing
* Internet Explorer 11: error page, ask to install an application
I could not install other browsers as it was not my own machine.

All in all the problem seems to be that no protocol handler seems to get registered.
A quick look at Can I Use confirmed the situation:
https://caniuse.com/#feat=registerprotocolhandler
Global support is said to be 35% with no clue that it could go better. Despite being flagged as supported by Chrome on desktops I couldn't get it to work. No support at all for phone & tablets.

I can't tell if the links would have been trigger had I installed a native application handler. I suppose that yes. Anyway web wallets support for this protocol will  be poor to say the least.

For this reason I question whether it's of any use I implement it in cosmic-lib which's a javascript library. I know some people are using javascript to code phone apps so it could make sense anyway. Maybe I'd like to hear about them.

I'd also like to know if there's any support at all for this protocol inside iOs/macOs as I heard Apple is gradually dropping custom URI schemes in iOs in favor of universal links.

It would be nice to have a better view of how it performs on other platforms and how it behaves with native programs/applications. I assume this have been checked before as the protocol just got accepted, so I guess I'm missing something again.

johansten:
@MisterTicot 

macOS:
Firefox and Opera manages to register the handler, Chrome doesn't.
Chrome opens up Opera when registered though.

MisterTicot:
@johansten Thank you Johan. I add the same behavior for chrome here on Linux.
Could you please check if it does works cross-domain?
https://markdownshare.com/view/9118c815-6bb0-4662-8a96-7c8e6b6d3d39

PR #98, opened 2018-04-20
added type=cancel

tempofr:
add cancellation of transactions


bartekn:
I think this functionality should be in a separate endpoint. In general, actionable endpoint should be a `POST` request. So something like this would work:
```
POST AUTH_SERVER/cancel
Request data:
id=tx_id
```
Since it's not part of `tx_status` we should make a new SEP.

tempofr:
Ok will do..

PR #100, opened 2018-04-24
SEP-0006: add transaction history endpoint

tomquisel:
Add a transaction history endpoint to SEP-0006. This allows wallets to display status of pending deposit/withdrawals, as well as a history of transactions involving the anchor for a given Stellar account.

### DIFF
--- a/ecosystem/sep-0006.md
+++ b/ecosystem/sep-0006.md
@@ -139,7 +139,7 @@ Name | Type | Description
 `asset_code` | string | The code of the asset of interest. E.g. BTC,ETH,USD,INR,etc
 `account` | string | The stellar account ID involved in the transactions
 `no_older_than` | UTC ISO 8601 string | (optional) The response should contain transactions starting on or after this date & time
-`n` | int | (optional) the response should contain at most `n` transactions
+`limit` | int | (optional) the response should contain at most `limit` transactions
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0006.md
```diff
@@ -110,8 +116,90 @@ Example:
...
+Name | Type | Description
+-----|------|------------
+`id` | string | Unique, anchor-generated id for the deposit/withdrawal
+`kind` | string | `deposit` or `withdrawal`
+`status` | string | Processing status of deposit/withdrawal
+`status_eta` | number | (optional) Estimated number of seconds until a status change is expected
+`amount` | float | (optional) Amount of deposit/withdrawal
+`started_at` | UTC ISO 8601 string | (optional) start date and time of transaction
+`completed_at` | UTC ISO 8601 string | (optional) completion date and time of transaction
+`stellar_operation_id` | string | (optional) operation_id on Stellar network of transfer that either completed the deposit or started the withdrawal
```
bartekn: `operation_id` is specific to Horizon. I think the better way would be to use something like `[transaction_id]-[operation_id_in_transaction]`, ex. `3389e9f0f1a65f19736cacf544c2e825313e8447f569233bb8db39aa607c8889-0`.

jedmccaleb: Its probably fine to just use txid here right?

PR #104, opened 2018-04-30
improve transactions endpoint

s-a-y:
In general I can raise many questions for discussion:
- why mixing dates and paging_id, 
- why there is no `order` parameter when querying transactions. 
- shouldn't we be able to fetch transaction by specific `stellar_transaction_id` or `external_transaction_id`
- what is `operation_id` parameter in withdrawal response and why is it not documented?

But in my current PR I tried to be productive and included only things that are a MUST.

I believe that [@fritz](https://stellar-public.slack.com/messages/@fritz/) also had an opinion on that endpoint, you should check with him if he has something to add to the conversation.

### DIFF
--- a/ecosystem/sep-0006.md
+++ b/ecosystem/sep-0006.md
@@ -156,7 +156,7 @@ Name | Type | Description
 `kind` | string | `deposit` or `withdrawal`
 `status` | string | Processing status of deposit/withdrawal
 `status_eta` | number | (optional) Estimated number of seconds until a status change is expected
-`amount` | float | (optional) Amount of deposit/withdrawal
+`amount` | string | (optional) Amount of deposit/withdrawal as a string with up to 7 decimals
 `started_at` | UTC ISO 8601 string | (optional) start date and time of transaction
 `completed_at` | UTC ISO 8601 string | (optional) completion date and time of transaction
 `stellar_transaction_id` | string | (optional) transaction_id on Stellar network of the transfer that either completed the deposit or started the withdrawal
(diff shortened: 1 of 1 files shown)


tomquisel:
Thanks for those changes @s-a-y!

To your points:

* Showing the last 30 (or N) days of transactions seems like a common operation. Adding a date filter allows a client to efficiently get that list without having to paging past that cutoff date and then throw out results that are too old.
* I think returning results in time-descending order as specified is most useful. Can you think of a good reason to fetch oldest first?
* What's a good use case for fetching a transaction by a specific id? Clients won't know the `external_transaction_id` in the deposit case, so I figured they can handle both deposit and withdrawal status monitoring in the same way: by polling the transaction endpoint filtering for transactions after the time that the SEP6 deposit/withdraw information was initially fetched.
* `operation_id` in the example is a mistake. Good catch! I'll fix it up.

s-a-y:
1. I get why you may want to have dates instead of ledgers, it's easier to work with, but confusing part is that you have lower limit by date and upper limit by paging_id. If you want to fetch all transactions since beginning of times, then you might want to go month by month or day by day and with current setup it doesn't look possible
2. Same, as 1. Imagine you have 1000 transactions to show for 2018, wallet will pull first batch with limit 100, then next batch, then next, etc I guess that's where your paging_id is useful, but it only allows you to go back in history, you can't say make paginated interface, because then user would click last page and it's really hard to calculate
3. They do know external_transaction_id in deposit case. These are support queries I'm answering all the times: somebody sends me ETH transaction hash and asks why he haven't got anything on stellar yet. Use cases are: search box on website, instead of trying to go through 100 records to find the one you need you might want to enter hash and find out the status. Of course can be imitated by the wallet, if it fetches whole history, but that might negatively affect performance in case of big transaction volume.

tomquisel:
Cool, those are all great points. Feel free to submit a PR or add to this one with those changes.

Issue #36, opened 2017-06-11
Offers do not affect asset balances.


istrau2:
What I mean by this is that when I create an offer with a selling asset of ABCD, my account's asset balance for ABCD is not diminished (I am referring to the asset balance returned by Horizon for the account, I assume this is the same balance that exists in the db). 
 
This results in an individual being able to create multiple offers that sum to a greater balance than is in their account.
 
I tested this on the test network as follows:
 
1)  Create an account and fund with 5000 XLM (actually, get 10000 XLM from friendbot and burn 5000 of them).
2) Create two offers:
a. Bid 4000 XLM for 50 BTC (price of 80 XLM/BTC)
b. Bid 4050 XLM for 50 BTC (price of 81 XLM/BTC)
As mentioned above, because the offers do not affect the account balance, offers can be created with a higher total value that is in the account.
3) I accepted the offers from another account (with 100 BTC). What happened is that only 4960 XLM worth of BTC was executed and an ask offer created for the rest of the BTC.
 
I think this behavior is buggy for several reasons:
 
1) It is VERY not intuitive for a user to see offers for 100 BTC, accept the offer, and have it only partially execute because some of the offers were vapor offers. Which leads me to point number 2)…
2) This allows users to manipulate markets. A user can create fake buy or sell walls (and even pop them themselves later) without the funds necessary by simply repeatedly creating offers with a value that is less than their account balance.

I think that an effect of manageOffer should be an adjustment to the account's asset balance.

MonsieurNicolas:
yes, this is by design. It allows for people to sell an asset in exchange for any combination of other assets. 

For example, if I want to sell my 100 Bitstamp BTCs for some ETH (and I have accounts with Bitfinex, Poloniex and Kraken), I would create an account with 100 BTC.Bitstamp and the following offers:

What I sell | What I buy
------------| -----------------
100 BTC.Bitstamp | ETH.Bitfinex
100 BTC.Bitstamp | ETH.Poloniex
100 BTC.Bitstamp | ETH.Kraken

Then, even though I have 300 BTC.Bitstamp up for sale, the total amount will be limited by the account to 100 and will get some combination of ETH.Bitfinex/Poloniex/Kraken.

istrau2:
@MonsieurNicolas Yes, that advantage was explained to me by @jedmccaleb.

[...]

PR #116, opened 2018-05-17
Asset backed offers (draft-0002)

jonjove:
Implements discussion from #36


MonsieurNicolas:
please resubmit as cap-0003 (per cap process we do not use the draft folder to track cap drafts, and instead use the "Status" field in the doc and cap readme)

PR #118, opened 2018-05-18
Actually specify the process followed for submiting CAPs

MonsieurNicolas:
## This PR specifies the process followed through the life cycle of a CAP.

This is a first step towards having all significant changes going into the core network go through an open process.
 
It's mostly the same thing than the EIP process except that I had to gut some of the governance as we are not there yet (we need more contributors to reach critical mass).

## Out of scope for now
I didn't modified or tried to specify the process used for SEPs as it doesn't have the same requirements.

### DIFF
--- a/core/readme.md
+++ b/core/readme.md
@@ -1,0 +1,33 @@
+# Contributing CAPs
+
+ 1. Review [CAP-0001](cap-0001.md).
+ 2. Fork the repository by clicking "Fork" in the top right.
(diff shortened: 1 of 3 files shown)


core/readme.md
```diff
@@ -0,0 +1,33 @@
+# Contributing CAPs
+
+ 1. Review [CAP-0001](cap-0001.md).
```
bartekn: This will be "Bump Sequence" right? By "review" do you mean check how regular CAP looks like?

PR #127, opened 2018-06-13
Add instructions on `recipient_info`

ramontayag:
When the receiving FI is a forwarder, it does not have information on the recipient. The sending FI needs to provide this. This commit describes where this should be.

### DIFF
--- a/ecosystem/sep-0003.md
+++ b/ecosystem/sep-0003.md
@@ -59,6 +59,7 @@ A few things to note:
 
 - *memo* value of `tx` is a sha256 hash of the attachment
 - *nonce* is unique to each individual transaction. When a transaction on the receiving end is `pending`, the same `nonce` should be sent to fetch updates about the `pending` transaction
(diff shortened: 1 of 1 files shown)


bartekn:
I think it's incorrect. Sender shouldn't send `recipient_info`, it just sends request for recipient information (`"need_info": true`). If recipient/receiving organization decides to share this information, it's send in the response (`dest_info` field).

ramontayag:
@bartekn there are instances, like in ours, where we do not know or have a relationship with the recipient. We get this information from the sender (through the sending FI).

bartekn:
Could you send this information in [federation forward](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0002.md#federation-request)?

ramontayag:
Yes, I suppose you could. Would it be appropriate to attach all information there, like: full name, address, mobile #?

bartekn:
Yes, other partners are using federation forward to send recipient information when sending to bank account.

ramontayag:
Ok, thanks. Will close this.

ramontayag:
See https://github.com/stellar/stellar-protocol/pull/128

PR #136, opened 2018-07-07
Extend SEP-6

tomquisel:
Proposed [SEP-6](https://github.com/stellar/stellar-protocol/blob/sep0006-expansion/ecosystem/sep-0006.md).
New [SEP-9](https://github.com/stellar/stellar-protocol/blob/sep0006-expansion/ecosystem/sep-0009.md): defines standard AML / KYC fields
New [SEP-10](https://github.com/stellar/stellar-protocol/blob/sep0006-expansion/ecosystem/sep-0010.md): defines Stellar web authentication protocol

Major changes:

* `/customer` endpoint added to allow upload of KYC information
* Added web authentication flow
* Deposit and withdrawal endpoints modified to add new kinds of responses:
  * non-interactive customer information needed response: Issuer needs more information about user before deposit or withdrawal information can be given
  * interactive customer information needed response: Issuer needs user to go through web flow that they host before deposit or withdrawal information can be given
  * customer information status response: once the customer information has uploaded, this responds with the status of that process

### DIFF
--- a/ecosystem/sep-0006.md
+++ b/ecosystem/sep-0006.md
@@ -26,9 +26,7 @@ This proposal defines a standard protocol enabling the following features direct
 
 To support this protocol an anchor acts as a server and implements the specified REST API endpoints, while a wallet implements a client that consumes the API implemented by anchors. The goal is interoperability, so a wallet implements a single client according to the protocol, and will be able to interact with any compliant anchor. Similarly, an anchor that implements the API endpoints according to the protocol will work with any compliant wallet.
 
-## Specification
-
-### Anchor Servers
+## Anchor Servers
 
 This protocol requires an anchor to implement two servers:
 
@@ -39,20 +37,22 @@ They can have the same URL, or they can be completely independent. They are sepa
 
 An anchor must define the location of their deposit and federation servers in its [`stellar.toml`](./sep-0001.md) with the `DEPOSIT_SERVER` and `FEDERATION_SERVER`. This is how a wallet knows where to find the anchor's servers.
 
-### API Endpoints
+## API Endpoints
 
 * [`GET /deposit`](#deposit): required
 * [`GET /withdrawal`](#withdrawal): required
-* [`GET /transactions`](#transactions): optional, but strongly recommended
-* [`POST /customer`](#customer): optional, but necessary if anchor desires any information about the customer for KYC or other purpose
+* [`GET /transactions`](#transaction-history): optional, but strongly recommended
+* [`GET /nonce`](#nonce): optional, but necessary for authentication (step 1)
+* [`GET /token`](#token): optional, but necessary for authentication (step 2)
+* [`POST /customer`](#customer): optional, but necessary if anchor desires any information about the customer for KYC or other purposes. Requires authentication (`/nonce` and `/token`)
 
-### Deposit
+## Deposit
 
 A deposit is when a user sends an external token (BTC via Bitcoin, USD via bank transfer, etc...) to an address held by an anchor. In turn, the anchor sends an equal amount of tokens on the Stellar network (minus fees) to the user's Stellar account.
 
 The deposit endpoint allows a wallet to get deposit information from an anchor, so a user has all the information needed to initiate a deposit. It also lets the issuer specify additional information (if desired) that the user must submit via the `/customer` endpoint to be able to deposit.
 
-#### Request
+### Request
 
 ```
 GET DEPOSIT_SERVER/deposit
@@ -70,22 +70,22 @@ Name | Type | Description
 Example:
 
 ```
-https://apay.io/api/deposit?asset_code=ETH&account=GACW7NONV43MZIFHCOKCQJAKSJSISSICFVUJ2C6EZIW5773OU3HD64VI
+GET https://apay.io/api/deposit?asset_code=ETH&account=GACW7NONV43MZIFHCOKCQJAKSJSISSICFVUJ2C6EZIW5773OU3HD64VI
 ```
 
 If the given Stellar `account` doesn't exist yet then the anchor will fund and thus create the account with at least enough `XLM` for the minimum reserve and a trust line. It is suggested that the anchor take some of the asset that is sent in to pay for the `XLM`. The anchor doesn't have the user account's secret key so the user must create a trust line to the anchor's asset before the anchor can send the remaining asset tokens to the user's account. The anchor should listen to see when the user has established this trust line. Once the trust line is there the anchor should send the asset tokens to the account in Stellar to complete the deposit.
 
-If the anchor won't create new accounts for users, it should return an error if the  `account` doesn't exist yet.
+If the anchor won't create new accounts for users, it should return a `400 Bad Request` error if the  `account` doesn't exist yet.
 
-#### Response
+### Response
 
 There are four possible kinds of response, depending on whether the issuer needs more information about the user, how it should be sent to the issuer, and if there are any errors.
 
-##### 1. Success: no additional information needed
+#### 1. Success: no additional information needed
 
 Response code: `200 OK`
 
-This is the correct response if the issuer is able to accept the deposit and needs no additional information about the user.
+This is the correct response if the issuer is able to accept the deposit and needs no additional information about the user. It should also be used if the issuer requires information about the user, but the information has previously been submitted and accepted.
 
 The response body should be a JSON object with the following fields:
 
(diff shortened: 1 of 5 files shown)


s-a-y:
1 vote for separation of concerns.
I believe that `type=forward` served well, filling that gap before, but doesn't make the process easy for anchors and users anymore.
Should eventually be removed from federation protocol (when everyone stops using it). Federation should be about name resolution within the network.

istrau2:
Sorry, if I misunderstand the PR. Are we back to hardcoding fields in the non_interactive response type? What is the plan for dealing with fields that don't happen to be in this PR? What about optional fields? is the plan interactive? Does that mean the every time a user wants to withdraw ripple via some anchor, the wallet needs to pop up the anchor's website to collect a tag?

Why not just a variant of the `/form` endpoint I suggested?

tomquisel:
@istrau2 this is a variant of the `/form` endpoint. The `non_interactive_customer_info_needed` response specifies a list of fields that need to be filled in, just like `/form`. Any subset of the fields specified in the protocol can be requested. The goal is to have an extensive list of possible fields specified in the protocol. Please feel free to add any that are missing to the PR, like the Ripple tag. It's important that possible list of fields are defined in the protocol so that wallets and issuers know how to handle the data correctly without user interaction.

The `interactive_customer_info_needed` response specifies an interactive URL if unusual fields are needed or an interactive process (such as signing up) is necessary.

istrau2:
@tomquisel Yes, I understand. However, I think specifying individual fields is not a good standard, we will end up with 100+ fields that wallets/clients will have to cater to using giant switch statements. 

I'd much rather switch on a type (control type). For example, lots of the proposed fields, are of the same type, just need different labels. That is what I was trying to accomplish with the `/form` endpoint. Basically, where we hardcode as few types as possible.

tomquisel:
@istrau2 I agree with what you're saying in general, although I think we have different mental models about the goals. Hear me out.

I believe the `/form` endpoint was intended for interactive use cases. An issuer specifies and explains information that they want from the user, the wallet generates a form with that information, shows it to the user, the user fills it out, and the wallet sends the info back to the issuer. Having a hard-coded set fields makes no sense in this case. As you point out, types (string, image, etc...) is all a wallet needs to know to gather info and send it back. That case is handled in this proposal via the interactive url that an issuer supplies, so I believe all that functionality from `/form` is covered.

Now let's think about the non-interactive use case. This is where a user enters their info into their wallet once, and then the wallet automatically transmits that info to each new issuer the user wants to interact with. In the future I suspect most tokens will be `auth_required`, meaning a user will need to KYC 10 times to hold 10 tokens. Much better than making the user go through 10 slightly different KYC forms is to collect the info once in a standard way, and transmit it automatically when it is needed.

As you can probably intuit by now, a pre-defined list of fields is the only way to make this happen reasonably. If there's no list, issuer A will ask for "firstName" and issuer B will ask for "given_name" and the (French) wallet will have collected something called "prenom", and the wallet will never be able to automatically match the fields requested by issuers to the fields it has collected from the user.

The pre-defined list is there to _enable_ automatic KYC exchange, not restrict interactive KYC (which is handled by a separate interactive path that doesn't involve the pre-defined list).

I hope that helps clarify.

ecosystem/sep-0006.md
```diff
@@ -6,55 +6,102 @@ Title: Anchor/Client interoperability
...
 
-If the anchor won't create new accounts for users then it should return an error if the given account doesn't exist yet.
+```
+GET https://api.example.com/deposit?asset_code=ETH&account=GACW7NONV43MZIFHCOKCQJAKSJSISSICFVUJ2C6EZIW5773OU3HD64VI
+```
 
-On success the endpoint should return `200 OK` HTTP status code and a JSON object with the following fields:
+If the given Stellar `account` doesn't exist yet then the anchor will fund and thus create the account with at least enough `XLM` for the minimum reserve and a trust line. It is suggested that the anchor take some of the asset that is sent in to pay for the `XLM`. The anchor doesn't have the user account's secret key so the user must create a trust line to the anchor's asset before the anchor can send the remaining asset tokens to the user's account. The anchor should listen to see when the user has established this trust line. Once the trust line is there the anchor should send the asset tokens to the account in Stellar to complete the deposit.
+
+If the anchor won't create new accounts for users, it should return a `400 Bad Request` error if the  `account` doesn't exist yet.
```
jedmccaleb: I think we need to tell the client why the request failed

istrau2:
@tomquisel 
I see. So, then it seems that the `/customer` endpoint should be for customer creation exclusively (i.e. KYC info upload) with the issuer.

Can you expand on the the `/nonce` and '/token` endpoints. I don't really understand what kinds of attacks that are trying to prevent. I mean, it would seem that one of the purposes is to prevent anybody from being able to change customer info related to a stellar account without having the secret key to that account. Is that the idea?

tomquisel:
@istrau2 yep, that's exactly right.

That's right, `/nonce` and `/token` are primarily there to prevent third parties from changing or uploading false KYC information for a user's account.

It's necessary to have a two step process (1. request nonce 2. sign & transmit it back) to prevent replay attacks. If we remove the `/nonce` endpoint, and allow a user to send any signed nonce they want to `/token`, an attacker can eavesdrop on a user signing some other nonce, and then can submit that same nonce & signature to the `/token` endpoint to upload fake data.

What's nice is that since this same authentication mechanism can be used for any of the endpoints, in case an issuer wants to have authenticated deposits or withdrawals.

tomquisel:
I've made all the planned major changes. Please take a look. I plan to merge in the next few days if there are no further issues.

ecosystem/sep-0006.md
```diff
@@ -6,55 +6,95 @@ Title: Anchor/Client interoperability
...
+Example:
 
-If the anchor won't create new accounts for users then it should return an error if the given account doesn't exist yet.
+```
+GET https://api.example.com/deposit?asset_code=ETH&account=GACW7NONV43MZIFHCOKCQJAKSJSISSICFVUJ2C6EZIW5773OU3HD64VI
+```
 
-On success the endpoint should return `200 OK` HTTP status code and a JSON object with the following fields:
+If the given Stellar `account` doesn't exist yet then the anchor will fund and thus create the account with at least enough `XLM` for the minimum reserve and a trust line. It is suggested that the anchor take some of the asset that is sent in to pay for the `XLM`. The anchor should not list this as a fee, as the user is still receiving the value of the `XLM`. The anchor doesn't have the user account's secret key so the user must create a trust line to the anchor's asset before the anchor can send the remaining asset tokens to the user's account. The anchor should listen to see when the user has established this trust line. Once the trust line is there the anchor should send the asset tokens to the account in Stellar to complete the deposit.
```
rice2000: This paragraph is pretty dense.  There's a lot to process.  Maybe split it up into at least 2 parts?  Something like:

If the given Stellar `account` has balance of 0, the anchor will fund and thus technically create the account with at least enough `XLM` for the minimum reserve and a trust line. It is suggested that the anchor take some of the asset that is sent in to pay for the `XLM`.  The anchor should not list this minimal funding as a fee because the user still receives the value of the `XLM` in their account. 

Since the anchor doesn't have the user account's secret key, the *user* must create a trust line to the anchor's asset before the anchor can actually send the remaining asset tokens to the user's account. The anchor should listen to see when the user has established this trust line. Once the trust line is there, the anchor should send the asset tokens to the account in Stellar to complete the deposit.

istrau2:
@tomquisel The nonce and jwt seems like a complex way to accomplish the same thing that signing the body of the upload request with the account's secret key doesn't it? I mean, instead of fetching a nonce and generating a jwt with account's signature, why not simply sign the body of the '/customer' request (similar to signing a stellar tx)?

Also, I was thinking, putting some version of the `/form` endpoint (as separate from the `/customer` endpoint) is probably desired. I mean, otherwise withdrawing/depositing via some anchors will have to be disrupted for a popup just to collect a simple piece of data (like an XRP tag).

tomquisel:
@istrau2 agreed that nonce and jwt is more complex than ideal. The issue is that `multipart/form-data` encoding is needed for `/customer` in order to support image uploading, and the boundary strings may be generated at a low level that makes it hard for the client to access the body after it's generated but before it's sent. It seemed tricky for clients to get the signing just right, so I opted for a flow that's a little more work but us more straightforward. As a bonus, we get the potential to authenticate any other calls too, like `/deposit`, if the issuer wants.

I totally agree about the `/form` idea! The `/info` endpoint is inspired by form, and allows the issuer to specify `extra_fields` that are needed for deposit or withdrawal. Does that fill the use case?

nebolsin:
Regarding authentication part, we have a similar challenge-response flow implemented in Mobius. The key difference is that we're using plain Stellar transactions to handle the process (those transactions are purposefully created as invalid, so they cannot be submitted to the network). We [have](https://docs.mobius.network/docs/authentication) [some](https://docs.mobius.network/v2.0/reference#challenge) [documentation](https://docs.mobius.network/v2.0/reference#auth-2) about the flow, but in short the process is (I updated our actual process for client-anchor scenario):

1. Client requests a challenge, which is an XDR-encoded Stellar transaction with timebounds and nonce signed by an anchor  `tx(source: anchor_account, seq: invalid_seq, timebounds: { min: now(), max: now() + session_duration }, operations: [ set_data(source: client_account, key: '<anchor name> auth', value: random_nonce()) ], signatures: [anchor_account])`
2. Client verifies that challenge is correctly signed by an anchor
3. Client signs this transaction using their secret key and submits back to anchor
4. Anchor verifies that received response is:
4.1. signed correctly by client's account
4.2. signed correctly by it's own anchor account
4.3. is fresh: `now() - tx.timebounds.min < challenge_expiration_period` (like 10 seconds)
5. Anchor issues a session JWT token, containing client's public key, issuing and expiration times, taken directly from the received response tx.

It looks like this has a few benefits over the proposed scheme:
* it allows using established Stellar libraries for challenge/response generation and verification;
* client can verify authenticity of the challenge (given that it can fetch the anchor's public key from a separate source, like stellar.toml file);
* it doesn't require server-side storage of <client account> - <challenge> association if potential replay over short time period (configurable, but generally < 1 minute) is acceptable (which looks ok  as long as auth flow is performed over HTTPS since replay is only possible if the final response submission is intercepted, and even in that case token expiration time cannot be changed);
* it is compatible with Ledger Nano S and other hardware wallets, since it only deals with signing Stellar transaction.

istrau2:
@tomquisel sure, it should have the ability to specify a field's type, required or optional, label etc.

Out of curiosity, since you work for Lightyear, a for profit company that we have to compete with, why do you have merge rights on a Stellar repository?

tomquisel:
@nebolsin that's a great auth scheme! I'll update the PR to specify that approach.

tomquisel:
@istrau2 I'll beef up the `/info` endpoint to support more flexible `extra_fields`.

As far as your question, I'm taking time to improve the quality of the Stellar ecosystem and help it grow. I have merge rights because the Stellar team knows me and trusts me to use it responsibly.

tomquisel:
Ok, `/info` is now more flexible (extra form fields don't include types, but I think these are unnecessary as the data is entered and transmitted as strings), and the auth flow is now what @nebolsin describes.

We'll merge this week if there are no other suggestions.

nebolsin:
@tomquisel I read the spec one more time and it looks pretty solid, great job! 

This auth scheme is in fact not specific for anchor integration and can be used as a universal web authentication flow for Stellar wallets (for example, we use it in Mobius DApp Store to authenticate the wallet with a DApp backend). So it might be better to keep SEP-0006 focused on anchor-specific topics and extract auth flow into a separate SEP-0010 "Stellar Web Authentication".

Let me know if you like this idea, we can add SEP-0010 here or submit it in a separate PR.

tempofr:
I started implementation of the info 

1) We probably need language... by putting info it is unilingual. 
The deposit and withdraw calls also probably need ISO 639-1 codes
2) We have the fields in the non interactive/interactive forms.. repeat here? I Maybe an indicator that interactive or non interactive is supported?
3) countries - I have this in our TOML but it may make sense to have it here. For example USD could be sent to many countries

tomquisel:
@nebolsin I like the idea of separating the auth flow out into SEP-10. If you want to split it out into a separate PR, that'd be fantastic.

tomquisel:
@tempofr the `language` and `countries` ideas make sense to me. Could you send me an example of an `/info` request/response incorporating those fields?

No need to put your fields in the `/info` endpoint. Only fields that need to be passed into `/deposit` or `/withdraw` should be listed in `/info`. Your fields are covered by your interactive flow. I'll clarify that in the description.

tomquisel:
I'm just going to separate the auth flow out into SEP-10 real quick as part of this PR.

tomquisel:
I've finished another round of changes, including:

* separating web authentication out into [SEP-10](https://github.com/stellar/stellar-protocol/blob/sep0006-expansion/ecosystem/sep-0010.md)
* adding a `lang` parameter to `/info`
* adding `DELETE /customer/...`.

I'll merge shortly if there are no other suggestions.

nebolsin:
@tomquisel Thanks for moving this forward. I'll create a PR with a few fixes and suggestions for SEP-10 targeting this branch in a few hours.

paultiplady:
I have a quick question on SEP-6 / SEP-9 - we have the field:

```
bank_number | Number identifying bank (routing number in US)
```

I'm a little concerned that this field won't be workable for global payments -- for example, in many countries, there are a number of different "bank numbers"; e.g. to pay someone in India, it might be:

SWIFT (if you're sending them a wire)
[IFSC](https://en.wikipedia.org/wiki/Indian_Financial_System_Code)
INRN

Or a less esoteric example, to get to the UK it might be:

SWIFT
Sort code
IBAN

Additionally, the anchor might only be able to process a subset of the bank number types, e.g. only supporting wire to the UK, even though a local would naturally provide "sort code" in this case.

The generic international payments API that my system uses is (unfortunately) quite complex, and the required fields are dependent on what country the beneficiary is located in. (This is quite standard for international payment rails, in my experience).

Not to mention that certain countries require additional regulatory information such as patronymic name in Russia (in addition to first/last), or CPF codes in Brazil. (An even bigger can of worms).

Is this issue something that should be discussed in the scope of SEP-6 (plus extensions)? Happy to open another issue if that would be more appropriate.

tomquisel:
@paultiplady thanks for bringing this up! Can you raise it as a separate issue and tag me since this PR is merged? It's definitely relevant to both SEP-6 and SEP-9.

PR #144, opened 2018-08-03
A few fixes and improvements for SEP-0010

nebolsin:
Some updates for SEP-0010: Stellar Web Authentication (based on a work started in #136):
- [x] use a single auth endpoint (GET the challenge, POST the response)
- [x] generalise text from an `anchor` to a `web service`
- [x] a few minor updates and clarifications
- [x] add recommendations for JWT claims and signing scheme (either symmetric HMAC-based, in which case a client should consider a token opaque, or via Ed25519 which allows a client or third-party to verify the issued JWT)

/cc @tomquisel


tomquisel:
@nebolsin thanks for the changes & fixes, they all look good to me. Don't forget to update SEP-1 to include WEB_AUTH_ENDPOINT rather than WEB_AUTH_SERVER.

Let me know when your changes are done and I'll merge it in.

tomquisel:
@nebolsin looks good! Merging in.

PR #148, opened 2018-08-13
add new seps to table of contents

cecilygardner:
I added sep 9/10 bullet points to the TOC. Please merge when it's appropriate to list them.


stanford-scs:
How are these not autogenerated?  We shouldn't be manually maintaining this kind of information.

Issue #159, opened 2018-09-01
Txrep: Canonical human-readable representation of Stellar transactions


stanford-scs:
## Preamble

    SEP: Not assigned
    Title: Txrep: human-readable low-level representation of Stellar transactions
    Author: David Mazières
    Status: Draft
    Created: 2018-08-31

## Simple Summary

Txrep is a human-readable representation of Stellar transactions that functions like an assembly language for XDR.

## Abstract

This document specifies *txrep*, a human-readable format for Stellar transactions. Txrep is unambigous and machine-parsable. Binary XDR transactions can be decompiled into txrep format and recompiled to the exact same binary bytes. Txrep is designed to be parsed directly into data structures generated by XDR compilers, ideally by the very XDR code generated by those compilers.

## Motivation

Bug reports, test vectors, semantic specifications in CAP documents, and Stellar documentation all need a way to talk concretely about the contents of transactions. Without a canonical textual format, different documents will each devise their own way of describing transactions, sometimes ambiguously, making it hard to relate information from multiple sources.

Furthermore, advanced Stellar users and devlopers need a way to craft arbitrary transactions, including potentially invalid ones for test cases. This functionality is currently available from the [Stellar laboratory transaction builder](https://www.stellar.org/laboratory/#txbuilder), but that is not a good solution for high security, as it requires a browser (and in practice requires trusting an HTTPS certificate). Moreover, the transactions one builds in a web browser cannot easily be audited before signing, or described in documentation, or scripted, or placed under version control.

Finally, "dumb" ink-on-paper contracts may need to specify Stellar transactions unambiguously. For example, a legal contract promising to deliver tokens with a lock-up period in exchange for a wire transfer may want to specify the exact lock-up mechanism used through a human readable description of the Stellar transactions involved. Using txrep, this description can be unambiguous.

## Specification

A txrep file consists of a number of lines, each describing the value of a field. We describe the line format using a simple BNF-like notation in which brackets (\[...\]) indicate optional contents, pipe (|) indicates alternatives, and an asterisk (\*) indicates zero or more repetitions of the previous symbol. Literal text (e.g., `:`, `.`, `[`, `]`, `._present`, `.len`, or `_`) indicates the occurence of those specific characters in the txrep source.

Each line of txrep has the following format:

> *line* = *field* `:` SP\* *value* \[*comment*\] LF | `:` *comment* LF | LF
> 
> *comment* = zero or more characters other than LF
> 
> SP = space (ASCII 32)
[...]

PR #170, opened 2018-09-17
Resolve issue #159

stanford-scs:


ecosystem/sep-0011.md
```diff
@@ -0,0 +1,201 @@
...
+
+A transaction without timebounds would contain this line:
+
+    tx.timeBounds._present: false
+
+For fields of type pointer-to-pointer, the pseudoselector `._present` describes whether the outermost pointer is NULL or not. If `._present` is `true`, then `._inner_present` describes whether the nested pointer is NULL or not. `_inner` can be repeated as many times as needed for arbitrarily nested pointers.
+
+Variable-length arrays use the pseudo-selector `.len` with an integer value to indicate the number of elements of the variable-length array. Only indices from 0 to len-1 are meaningful. For example, a transaction with one operation might look like this:
+
+    tx.operations.len: 1
```
bartekn: This potentially opens many issues with memory allocation we observed in XDR parsers. I can write:
```
tx.operations.len: 2147483647
```
and allocate big part of computer's memory. This is especially dangerous on servers that need to parse txrep transactions.

Not sure if it's the right place but we should state that parsers **must** check the type max length before allocating the memory. Maybe "Writing parser guidelines" section?

stanford-scs:
Superseded by 171

PR #171, opened 2018-09-18
Fix issue #159

stanford-scs:
Fucking github A) won't let me update pull requests from unknown repos, and B) forced me to move my branches around so much and made it painful enough to view blobs that my previous pull requests somehow dropped a bunch of commits.  I think this one will be correct, but I'll have to inspect after creating the pull request.


stanford-scs:
Fuck github.  This isn't right either.

PR #172, opened 2018-09-18
fix #159

### LINKED ISSUE #159: Txrep: Canonical human-readable representation of Stellar transactions

stanford-scs:


stanford-scs:
Can someone (@nikhilsaraf) r+ me on this shit before I lose my fucking mind?  If it's not perfect we can always fix it after the fact, but I want to get some of my work onto an actual github branch.

PR #179, opened 2018-09-26
Added cap-0004

jonjove:
Added cap-0004, which describes improvements to rounding for cross offer.

### DIFF
--- a/core/cap-0004.md
+++ b/core/cap-0004.md
@@ -17,7 +17,7 @@ As of protocol version 9, crossing offers can lead to unbounded relative roundin
 As of protocol version 9, crossing offers can lead to unbounded relative rounding error and can even leave the book in a crossed state. We propose a new algorithm to resolve these issues. This algorithm will prevent unbounded relative rounding error by adding an error threshold, and will prevent the book becoming crossed by guaranteeing that the less valuable offer must always be removed from the book. The new algorithm also guarantees that any given offer can only suffer a single adverse rounding event by rounding in favor of the offer with more total value (which, equivalently, means the offer which remains in the book).
 
 ## Motivation
-Consider the following situation:
+In what follows, quantities expressed like "100 stroop X" mean 100 times the minimum representable quantity of the asset X. Consider the following situation:
 
 - In transaction 1, account A creates an offer selling 100 stroop X at a price of 3 Y / 2 X.
 - In transaction 2, account B creates an offer selling 10 stroop Y at a price of 1 X / 2 Y.
@@ -31,7 +31,7 @@ As of protocol version 9, 1 stroop Y is exchanged for 1 stroop X meaning that th
 
 There is still another issue with the rounding algorithm as of protocol version 9. Whenever offers cross, at least one of the offers must execute entirely in order to avoid a crossed book. There are situations, however, where the rounding algorithm as of protocol version 9 will allow the book to remain crossed. Consider the following situation:
 
-- In transaction 1, account A creates an offer selling 100 X at a price of 5 Y / 1 X.
+- In transaction 1, account A creates an offer selling 100 X at a price of 10 Y / 1 X.
 - In transaction 2, account B creates an offer selling 1 stroop Y at a price of 1 X / 10 Y.
 
 As of protocol version 9, nothing is exchanged and the book remains crossed.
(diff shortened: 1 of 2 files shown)


stanford-scs:
I find the notation "100 stroop X" confusing, because I thought stroops were only for XLM.  Can you just say 10 microX (where .1 microX is the minimum unit) or something?  Or maybe just say that for the purposes of exposition, X is the minimum unit of asset MegaX, and Y the minimum unit of asset MegaY?

> * In transaction 1, account A creates an offer selling 100 X at a price of 5 Y / 1 X.
> * In transaction 2, account B creates an offer selling 1 stroop Y at a price of 1 X / 10 Y.
> As of protocol version 9, nothing is exchanged and the book remains crossed.

What's confusing about this example is that wouldn't this still be the case if A had the same price as B (10Y/1X)?  In other words, I think the point is just that B is trying to buy something that can't be represented, but the example made me also think about the fact that the prices are different.

> In order to prevent unbounded relative error during rounding, we introduce a new error threshold. If the rounding error exceeds the error threshold, nothing is exchanged.

It's not clear what this means.  Doesn't rounding usually involve a 50% threshold?  So shouldn't the most anyone ever loses be 0.5 of the minimum unit of whatever asset they are acquiring?

Oh, I think this has to do with path payments.  So maybe explain that earlier, or have a forward reference?

> The offer which initially had less total value is removed from the book regardless of whether it has been executed entirely, so it is guaranteed that the book never remains crossed.

Shouldn't this also just be a side-effect of normal rounding?  If I'm selling 1 X at a rate of 10X/4Y, then I'd be getting 0.4 Y, which rounds to 0, so just remove the thing from the orderbook.

jedmccaleb:
We have been using "stroops" to refer to the smallest unit of any asset

jonjove:
@stanford-scs Before responding to each of your questions individually, I want to offer one overall clarification. In your questions you mention phrases like “rounding usually” and “normal rounding”, but I want to emphasize that there are many distinct algorithms for rounding. According to [the Wikipedia entry about IEEE-754](https://en.wikipedia.org/wiki/IEEE_754#Rounding_rules), that standard alone defines 5 rounding algorithms. These algorithms are not immediately applicable to the situation of crossing offers since in this setting there are two quantities (the amount of X sent and the amount of Y received) that need to be rounded jointly, rather than a single quantity which needs to be rounded. Analogous to IEEE-754 defining rounding algorithms for a single quantity, this CAP defined a rounding algorithm for a pair of quantities that has properties particularly suited to the situation of crossing offers.

### Question 1
> I find the notation "100 stroop X" confusing, because I thought stroops were only for XLM. Can you just say 10 microX (where .1 microX is the minimum unit) or something? Or maybe just say that for the purposes of exposition, X is the minimum unit of asset MegaX, and Y the minimum unit of asset MegaY?

I have added a sentence clarifying excatly what is meant by a quantity like "100 stroop X".

### Question 2
> What's confusing about this example is that wouldn't this still be the case if A had the same price as B (10Y/1X)? In other words, I think the point is just that B is trying to buy something that can't be represented, but the example made me also think about the fact that the prices are different.

You are indeed correct that this example would be unchanged if both offers had the same price. Fixed.

### Question 3
> It's not clear what this means. Doesn't rounding usually involve a 50% threshold? So shouldn't the most anyone ever loses be 0.5 of the minimum unit of whatever asset they are acquiring?

There was an omitted word in the last sentence that you quoted. All mentions to rounding error should be referring to relative rounding error in the price, so the quoted text should read "In order to prevent unbounded relative error during rounding, we introduce a new error threshold. If the relative rounding error exceeds the error threshold, nothing is exchanged."

### Question 4
> Shouldn't this also just be a side-effect of normal rounding? If I'm selling 1 X at a rate of 10X/4Y, then I'd be getting 0.4 Y, which rounds to 0, so just remove the thing from the orderbook.

I am unsure how to respond to the question of whether this is a "side-effect of normal rounding" since I am uncertain what exactly "normal rounding" would correspond to in the context of crossing offers. It is clear that not every rounding algorithm has this property because, as mentioned in the motivation section, the rounding algorithm that was in use as of protocol verison 9 did not have this property. However, the requirement that you quoted from the proposal is definitely necessary in the context of the rounding algorithm described therein. Since the rounding algorithm includes a threshold for relative rounding error, above which nothing is exchanged, it must be resolved which offer should remain in the book in this case.

PR #183, opened 2018-09-30
SEP draft for Dynamic Asset Metadata

orbitlens:
Draft for #96 (SEP for Issuing large numbers of Assets).

I'm also going to create a simple reference implementation for the described service in JS. It will work as a ready-to-use app for serving assets metadata from the static files in the TOML format (@nullstyle arguments look convincing) and as a bootstrap service for the custom implementation.

### DIFF
--- a/drafts/draft-0004.md
+++ b/drafts/draft-0004.md
@@ -1,0 +1,143 @@
+## Preamble
+
+```
+SEP: <to be assigned>
+Title: Dynamic Asset Metadata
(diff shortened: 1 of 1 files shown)


JeremyRubin:
I think that doing a signed merkle tree (depth tags and perhaps chunked leaf nodes) might be better because then you can do queries against untrusted edge servers to get proofs of the asset. 

You can also do a merkle trie so you get higher assurance that you can't include conflicting entries

Issue #133, opened 2018-06-27
Address Fee-Paying For Future-Proofing Transactions


jedmccaleb:
This will allow you to set a higher fee so the tx will still work during surge pricing also helps future proof your txs.

Current Proposals:
- [CAP-0010](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0010.md)
- [CAP-0015](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0015.md)
- [CAP-0019 & CAP-0015](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0019.md)

robdenison:
One mechanism that could enable this would be a uniform-price auction:

1) Sort transactions by priority (either using the [current algorithm](https://github.com/stellar/stellar-core/blob/d200c00e80e9584a5f7f3c673f252e6dd52eebe4/src/herder/TxSetFrame.cpp#L135), or using the algorithm suggested [here](https://github.com/stellar/stellar-protocol/issues/75#issuecomment-404969973)). 
2) Select transactions until the total number of transactions (or operations, see #75) is too high.
3) If the transaction queue is now empty (i.e., every transaction has now been selected), all transactions pay the minimum fee necessary (based on the minimum feerate * their operation count).
4) If the transaction queue is not empty, then all transactions pay the same feerate as the lowest-feerate transaction that was in fact included in the block.

That way, transactions can set a fee higher than the minimum, with the comfort of knowing that their transactions will only pay the minimum necessary to actually get included in a block.

I think the challenges described in [this post](https://ethresear.ch/t/first-and-second-price-auctions-and-improved-transaction-fee-markets/2410/1) should already mostly be avoided thanks to Stellar's fee mechanism, where fees are pooled and redistributed through inflation, rather than being given to a particular miner who is responsible for choosing transactions for that block. Validators do not have any particular incentive to manipulate transactions to force fees to be higher, and it would be costly to create fake transactions themselves to do so.

MonsieurNicolas:
> If the transaction queue is not empty, then all transactions pay the same feerate as the lowest-feerate transaction that was in fact included in the block.

I think there is an interesting edge case to discuss.

I don't know if it's worth dealing with it, but it's probably worth discussing.

What we're talking about is picking the top `n` transactions
```
tx[1..n].maxfeerate >= X
tx[n+1].maxfeerate == Y
```

[...]

Issue #75, opened 2018-01-17
Update "surge pricing" to be more fair/limit based on operations


MonsieurNicolas:
I think that the implementation to enforce a maximum number of transactions predated the addition of operations.

How this works today:
the "surge pricing" logic kicks in when a validator encounters more transaction candidates than `maxTxPerLedger` (defined in the ledger header). The first pass at filtering is done by sorting transaction candidates by fee ratio (defined as `fee/min_fee` where `min_fee` is `base_fee*nb_operations`).
The idea is that in a situation where too many transactions are pending, we prioritize the ones with higher average fees.
If fees are equal, the `sourceID` of the transactions is used as a tie breaker.

Proposal for new behavior:
1. build a transaction set that maximizes the number of operations in a ledger, up to the maximum `maxOperationsPerLedger` defined by the network
2. sort transactions using the average fee per operation as the way to compare transaction fees (done simply by comparing `left.fee * right.nb_ops` with `right.fee * left.nb_ops`)
3. break ties using a deterministic yet randomizing scheme, giving priority to different accounts for each ledger.  Using `sourceID XOR hash_ofPreviousLedger` seems to fit the bill.

In order to implement 1 and 2, the candidate transaction set is build as follows:
1. trim the transaction set such that it only includes valid transactions (including sequence number and fee)
2. perform a lexicographical sort the candidate transactions by fee (descending), xored `source ID` (ascending), sequence number (ascending)
3. iterate over that sorted list:
   1. if the transaction can fit in the candidate set (ie, `newTxSetOpCount <= maxOperationsPerLedger`), add it to the set
   2. otherwise, skip the transaction and any other transactions from the same source account

Something else to discuss here is what to do with the ledger header.
Right now the XDR is:
```
struct LedgerHeader
{
    uint32 ledgerVersion;    // the protocol version of the ledger
    Hash previousLedgerHash; // hash of the previous ledger header
    StellarValue scpValue;   // what consensus agreed to
    Hash txSetResultHash;    // the TransactionResultSet that led to this ledger
    Hash bucketListHash;     // hash of the ledger state

    uint32 ledgerSeq; // sequence number of this ledger

    int64 totalCoins; // total number of stroops in existence.
                      // 10,000,000 stroops in 1 XLM

    int64 feePool;       // fees burned since last inflation run
    uint32 inflationSeq; // inflation sequence number

    uint64 idPool; // last used global ID, used for generating objects

[...]

PR #189, opened 2018-10-08
First draft of CAP0005

MonsieurNicolas:


istrau2:
I am not sure where we are supposed to be commenting on this but I think that the fee calculation is somewhat misleading. Effectively, there is no "fee", there are only suggested fees depending on what you want to accomplish similar to this: https://web3js.readthedocs.io/en/1.0/web3-eth.html#getgasprice

Perhaps there should be a calculation where one could provide parameter on what they were trying to accomplish (i.e. speed vs affordability)? This should obviously be an endpoint in horizon as well.

MonsieurNicolas:
@istrau2 you should comment in one of the issues linked from the CAP, we're trying to avoid having feedback all over the place

istrau2:
@MonsieurNicolas I am not seeing any links. Where are these links you mention?

MonsieurNicolas:
each CAP has a "discussion" field in the header. For this one, it's:
Discussion: https://github.com/stellar/stellar-protocol/issues/75
https://github.com/stellar/stellar-protocol/issues/133

Issue #188, opened 2018-10-05
Allow sending account to add trustline on receiving account


tomquisel:
Currently, sending an airdrop or funding a new Stellar account with a non-XLM token is a messy, multistep process for both sender and receiver. Issuer and user experience can be improved by adding some mechanism for a sending account to add a trustline on a receiving account.

# Deposit & Sending Requirements
## Deposit with asset issuer (mandatory)
An issuer should be able to completely process a user’s deposit of an external asset without needing to wait on the user to take an intermediate step, such as adding a trustline.

## An illustrative SEP-6 deposit flow (crypto asset example):

1. User wants to deposit an external crypto asset with an Issuer and receive the Issuer’s anchored asset on the Stellar network.
2. User makes a SEP-6 deposit request to the Issuer via the user’s Stellar wallet, communicating their Stellar account id (A) to the Issuer.
3. Issuer asks for any information they need (KYC info, bank acct, etc…).
4. User provides needed information.
5. Issuer sends deposit address to user.
6. The user sends their external crypto asset to the deposit address to initiate the deposit.
7. When the Issuer receives the external asset, they send a Stellar payment of the same amount (minus fees) to the user’s Stellar account, completing the deposit.

## The Requirements

- When initiating a deposit, the user must not need to sign any transactions.
- The Issuer must not need to wait on any user action to complete the deposit, once the user has sent the external crypto asset to the Issuer. The single exception to this is if the user does not need any XLM to take the action and the information that the transaction has completed (on the issuer side) is present on the chain.
- The deposit must arrive at the user’s original account id A. Only at this point is the deposit complete.
- The user must not have to add extra XLM from some external source to their account to complete the deposit.
- The deposit process must complete successfully even in the case where the user spends XLM for some other purpose as soon as it hits their account.
- The deposit process must complete successfully even if the user’s account is not yet created, or if the user has no trustline for the deposit asset.

tomerweller:
With the current base reserve model, it will allow you to force fees on other accounts. This should probably come with some form of trust-line sponsorship model or a complete revamp of storage fees. 

Regarding airdrops, one possible solution is to setup a new account per drop along with the relevant trustline, and then transfer ownership using multisig. Essentially dropping an account on a signer rather than dropping an asset on an account. However, this requires more ecosystem support for multiple accounts per signer.

tempofr:
The issue is most  clients who pay with ether or BTC do not have a stellar
account.

[...]

PR #197, opened 2018-10-26
Draft of cap-0006

### LINKED ISSUE #180: Allow buy offers
Currently, stellar only allows for sell offers. However, translating buy offers to sell offers does not preserve the intent of the trader. 

Quoting @ScottHogge from the stellar's keybase team: 
> Lets say you want to buy exactly 10 Marbles for at most $1 each... if you re-frame this as an offer to sell $10 at a price of 1 Marble/$,  you could end up getting 20 Marbles at $0.5 if the orderbook contains an offer to sell Marbles at $0.5 --while the desired effect would be just 10 Marbles at $0.5.  This behavior is different than in any traditional market (stocks, options, futures, etc...) and is kinda problematic for the financial industry adopting SDEX usage when they have a duty to execute customer orders as directed.

jonjove:
#180

### DIFF
--- a/core/cap-0006.md
+++ b/core/cap-0006.md
@@ -1,0 +1,181 @@
+## Preamble
+
+```
+CAP: 0006
+Title: Add ManageBuyOffer Operation
+Author: Jonathan Jove
+Status: Draft
+Created: 2018-10-24
+Discussion: https://github.com/stellar/stellar-protocol/issues/180
+Protocol version: TBD
+```
+
+## Simple Summary
+We introduce the `ManageBuyOffer` operation with functionality similar to the
+`ManageOffer` operation except that the amount is specified in terms of the
+`buying` asset instead of the `selling` asset.
+
+## Abstract
+The `ManageOffer` operation specifies the maximum amount of the `selling` asset
+that should be sold by the offer. It is not, however, possible to express the
+maximum amount of the `buying` asset that should be bought by the offer. These
+constraints are not equivalent because at the time of offer submission it is not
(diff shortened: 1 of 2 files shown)


MonsieurNicolas:
removed the "resolves" as this is a draft

MonsieurNicolas:
merging, so that we continue discussing in #180

skirsch:
This is awesome. This is one of the two major "mis-features" of Stellar, and this addresses it. THANK YOU. The other mis-feature on my list is inability to send someone my USD directly. there should be a flag on an account to enable this.

theaeolianmachine:
@skirsch - are you discussing the need for someone to establish a trustline before sending them a token, or something else? Is there another issue in this repo already addressing what you're talking about?

skirsch:
@theaeolianmachine yes, there is another issue on the "how can I send someone my USD" (since they don't have a trustline) problem: https://github.com/stellar/stellar-protocol/issues/188

Is that what you were asking?

theaeolianmachine:
Yes, that's correct - wanted to make sure it was that issue. We're actively looking at several proposals, including SEP-0013 & CAP-0007, along with a few others that haven't been drafted yet.

skirsch:
Great. fixing the how can I send my token problem is I think one of the reasons ripple failed. it's awkward today as you know.

oriana13:
Hello @skirsch,
May I ask you for your solution regarding the trust line? Because currently I have the same problem: I want to create a function that sends an NFT, the only problem is that the recipient does not have a trust line and when I create it in my function, I do not have the secret key of the recipient who must sign ...

PR #199, opened 2018-10-26
Add draft-0005: Data Entry Namespaces

MisterTicot:
A small SEP that defines how to represent namespaces into account data entry. Please give me hints about how we should proceed to validate and merge that one as the process is not clear to me yet.


MisterTicot:
The draft is available here: https://github.com/stellar/stellar-protocol/blob/06453860c207a8b784142979b3d773cf549ec51a/drafts/draft-0005.md

drafts/draft-0005.md
```diff
@@ -0,0 +1,138 @@
...
+data entries, and provides guideline about how to parse them and to handle
+edge cases.
+
+## Motivation
+
+As a generic mechanism to store key/value pairs on the ledger, data entries
+accept a very wide range of use cases. If each actor use it following its own
+syntax, it will lead to a confusing experience.
+
+Introducing namespaces allow to hierarchize and organize data, in such a way
```
pselden: What's the benefit to being able to turn it into a JSON structure? Is there a specific example you can show as to how this improves things over using the keys directly, besides just having a tidy looking data object?

drafts/draft-0005.md
```diff
@@ -0,0 +1,138 @@
...
+accept a very wide range of use cases. If each actor use it following its own
+syntax, it will lead to a confusing experience.
+
+Introducing namespaces allow to hierarchize and organize data, in such a way 
+that sorting the key/value entries will result in a readable output even when 
+using them for various purpose on the same account.
+
+It also permit to parse the whole tree at once, and provides guidelines for a 
+generic API that would allow direct access to data entries, by converting data 
+tree into JSON for example. This could make sense for some sensitive data that
```
pselden: Can you provide an example as to what this means?

MisterTicot: I assume you asked the same question twice ? This is not clear to me which part you're pointing to.

**Avoiding name conflicts:** This is generally why we use namespaces.

**Tidy looking:** This is not only about having a tidy looking object, this is also about displaying a clear and understandable data entries tree to the user. It's more professional than:

```
anchorRedirect = true
ticot.multisig.protocol = ...
ticot.multisig.options = ...
FAVORITE_COIN = ...
my_eth_entry_point = ...
pzY9c8 = ...
SuperAnchor:BTCwallet = ...
tradeAccount = ...
tradeOpts = ...
```

**Consistency:** It is clear that if each SEP use different syntax to name data entries it's quickly going to be a mess.

**Promoting standard solutions:** Defining a few standard fields and their usage would help there (like wallet,config,profile). Could be done with dedicated SEP. It allows dev to know exactly were to look for data rather than jungling with each vendor-specific solution.

**Parsing the whole tree at once:** All data entries respecting this SEP can be parsed at once with a tiny dedicated util. This will make working with data entries dev-friendly. Example, if you want a list of user external wallets:

```js
function externalWallets (account) {
  return SDK.AccountResponse.dataTree(account).wallet
}
```

**Turning into JSON:** This is only a side effect of this SEP, so maybe I shouldn't have mentioned it but it is clearly an interesting use case. It makes easy to serve data as nicely formatted JSON directly from the blockchain, without having the client needing to use a SDK or even knowing about it. The use case here are very broad so it's mainly a matter of imagination.

**Example:**
* I want to have my federation server and assets configuration locked under my 5-signers account and needs at least 60% votes to be changes (provided you read the data from a GET request)
* I want to serve live the data of my daily audits, that have to be signed by at least 2 different independent parties.
* I want to publish the shasum of software release, signed by 3 contributors at least
* ... and so on ...

Once a trusted server (Horizon?) offers data account entries as well-formatted JSON responses, it's only a matter of signing a few transaction and your data is served.

pselden: This particular comment I was more wondering about how it improves things for the sensitive data case you mentioned.

pselden: I'll also notes that since SEPs are mostly just ecosystem enhancements and not really baked into the core that there's nothing stopping people from polluting/using inconsistent data entries. I really think this is application/use case specific -- the main benefit I can see would be "reserving" some namespaces for specific use-cases, but that would really be the responsibility of other SEPs to define those.

pselden:
So I definitely like the idea here, in that we want some way to group similar data fields, but I'm unsure whether making this a full "SEP" is worth it since then tools will have to pay the cost to implement them, especially all the requirements around what can be in a data field and what can't. Since this is not enforced at the protocol level then I feel like it's sort of a lost cause to have any sort of key enforcement.

drafts/draft-0005.md
MisterTicot: This is not meant to prevent free use of data entries, this is meant to provide guidelines for who want to use them a structured, compatible & consistent way; which is especially desirable when it comes to writing SEP:

This is not a revolutionary SEP, this is only a tiny proposal based on the fact that naming things with consistency lead to a clearer environment than multiplying syntaxes. This really is one of the basis of good programming.

johansten:
Personally, I don't think it makes much sense. When you start having the need for this, that's probably the time you should look into storing data off-chain, before it becomes to expensive. On-chain IS expensive, and shouldn't be taken lightly.

MisterTicot:
Actually the situation right now is that everybody have to code the bits to parsed data entry from their base64 representation its own way, name data entry its own way and end up making something incompatible. So the cost you're talking about is already here. This SEP allow to implement a standard tool that would work for everybody so it's removing it.

Besides I don't see why it should have to be enforced over everybody. This is only a base on which we can write clean standards.

MisterTicot:
@johansten I think there's a misunderstanding. This is not about how much data entry you use *personally*, this is about the fact that **1) a bunch of services are going to use them** and **2) any naming conflict can have serious consequences**.

The tool we use to deal with this situation is namespaces. This is why we organize files under directories, npm packages under scopes, URL under domains, phone numbers under region and so on.

Namespace is an well-know concept and its properties & advantages are easy to understand because we're relying on them all the time.

johansten:
@MisterTicot 

I'm all for namespacing. I'm just opposed to storing json blobs on chain.. It seems a lot more appropriate to point your key to an IPFS file, since the Content ID changes if the content does.

MisterTicot:
This is really not the center of this proposal.

drafts/draft-0005.md
```diff
@@ -0,0 +1,142 @@
...
+The namespace syntax had to be familiar, and terms have to be written in a way
+that is valid accross a broad range of programming languages.
+
+This protocol uses JavaScript syntax because it is the language the most 
+developers uses. It is also familiar to anybody using JSON, which is a format 
+widely used to pass data, and the one already used by horizon API.
+
+Parsing the data tree must lead to the same namespace/key/value structure 
+regardless of the programming language being used. This is the reason why term 
+syntax must be restricted, as any unusual character could lead to errors in
```
pselden: If we're thinking about this in terms of JSON, then I think that you should allow anything that can be a valid JSON key -- which is everything.

drafts/draft-0005.md
```diff
@@ -0,0 +1,142 @@
...
+developers uses. It is also familiar to anybody using JSON, which is a format 
+widely used to pass data, and the one already used by horizon API.
+
+Parsing the data tree must lead to the same namespace/key/value structure 
+regardless of the programming language being used. This is the reason why term 
+syntax must be restricted, as any unusual character could lead to errors in 
+this or that language. The lowcase restiction is meant to prevent a different 
+data tree interpretation between case-sensitive and case-insensitive languages.
+
+Forbidding the use of numbers and/or undescore is advisable if it appears that
```
pselden: Minor typo: underscore. But which programming languages have problems parsing underscores?

drafts/draft-0005.md
```diff
@@ -0,0 +1,142 @@
...
+this or that language. The lowcase restiction is meant to prevent a different 
+data tree interpretation between case-sensitive and case-insensitive languages.
+
+Forbidding the use of numbers and/or undescore is advisable if it appears that 
+they prevent the data tree to be properly parsed in any well known programming 
+language.
+
+**Identifier/namespace conflict**
+
+This conflict arise when a term is used both as an identifier and a namespace,
```
pselden: This sort of conflict is why I feel there should be less emphasis on turning it into a data structure and more just some loose guidelines about what are the namespace delimiter and what things are considered "under" the namespace. That way you can just call out some "known namespaces" and not really worry about the details like this.

drafts/draft-0005.md
MisterTicot: None that I know, but I don't know the rules of all major commonly used programming language, hence the **if**.

drafts/draft-0005.md
pselden: But then why would the name matter at all for the programming language? We're just talking about strings. When you mention "key" it implies a JSON key.

drafts/draft-0005.md
MisterTicot: It's true I could have chosen that strategy. Turning into data structure is not a must have: it is a very convenient feature for devs that comes at a no cost with this design. It asked me an additional effort to put it right but that's it.

Now I assume that by saying *this sort of confict* you're introducing the idea that there are other possible conflict that are not properly handled by this SEP, and that turning into data structure might require additional effort. Can you elaborate on that?

Because, if there's no other kind of conflict and everything works fine, I don't see any valid reason to drop this feature.

drafts/draft-0005.md
pselden: I guess i'm stepping around my main opinion here: that I don't think namespace keys should have _any_ restrictions in them, and that this sep could then boil down to a suggestion of: "use dots (.) to group hierarchically related data under a namespace".

drafts/draft-0005.md
pselden: No, I just don't think it should be part of the SEP at all since it is application/use case specific how you might want to use these namespaces/whatever data structure you decided to build out of them.

drafts/draft-0005.md
MisterTicot: > But then why would the name matter at all for the programming language?

I don't know how to put more clearly than I already did. You won't be able to parse the structure if key names are not valid in a given language.

>  When you mention "key" it implies a JSON key

This is only your interpretation. I'm using key in reference of a key/value pair which's a generic concept.

If I understand your position, you want to remove the feature of being able to parse the data tree into a structure, at the cost of having each one having to code its own solution to handle the tree; which strange because you used the argument of "cost of implementation" a couple of days ago. Do you realize there are programming languages in which it won't be a trivial task?

Another drawback I can see in this proposal is that we loose in cleanness, as we could end up with a tree like:

```
profile.name
Profile.myPubKey
PROFILE.iMaGr34Th4cKeR
```

Enforcing lowcase actually ensure an item goes under the intended namespace.

So, I'd say the cost of removing character restrictions is pretty high. It is OK to have opinions about what the SEP should be, but if there's no rational motivation and it only have drawbacks why should we follow it?

drafts/draft-0005.md
MisterTicot: The usage of a namespace might or might not be application specific. The way to write them obviously require a standard.

drafts/draft-0005.md
MisterTicot: (answer in next item)

drafts/draft-0005.md
pselden: > You won't be able to parse the structure if key names are not valid in a given language.

Parse into what? I'm totally lost here about here now since we're not talking JSON and we're not talking any specific languages so I have no context besides JSON. Any language that can use strings can build a tree out them.

As far as lowercase vs uppercase -- sure, it's cool to define whether "CONFIG" and "config" are the same namespace, but since we can't actually restrict what gets in the data entries I don't think it makes sense to restrict it here as well beyond saying "namespaces are considered case insensitive".

MisterTicot: By the way saying that I could make design choices for their "coolness" is kind of insulting.

pselden: > it is easier to reference conf.foo than conf.['foo*.!']

But you wouldn't need to do that unless you you had a namespace called 'foo*!' that you cared about accessing. 

The benefit from removing the restrictions is that it simplifies the SEP and gets at the core bits that we should really care about: that we are encouraging people to group things using dots and setting the stage for "well known" namespaces. As it stands, my opinion is that the restrictions are adding some complexity for no benefit.

MisterTicot: Ok, so the benefits are:

* **It is by far the simpler way to make it case insensitive**
* It is also the simpler way to incentive a "clean" usage of namespaces
* In some languages, the easiest / most efficient way to build the data tree may actually comes with restrictions (for example in LISP languages you would use symbols, which comes with restrictions). As we don't know about all languages that may end up interfacing with Stellar, it is easier to solve it with tight rules.
* It is easy to handle with a simple regexp

The disadvantages is:

* It takes two lines to define the REGEXP and a few other to explain the rational behind this choice.

Now the benefits of removing restrictions:

* You can use keys like `=(G_G)=`
* You can use keys like `I love spaces!`

The disadvantages:

* You break the sorted raw display because of higher case letters
* It requires more code & computation to ensure case insensitiveness; So it makes it harder to implement.
* It is actually not clear which character you can use or not, as Stellar keys comes with its own restrictions for keys, but anyway there will still be a restriction of same kind.
* The advantages are not real advantages

MisterTicot:
I tried to look at what we discussed so far at various angle and here's what I got:

**The need for a naming convention:**

Independently from the question of characters restriction, I think this SEP must provide a clear naming convention. It is what most dev groups use to obtain a cohesive result, as each individual have its own preference on that matter. The usual standards are HIGHCASE, lowcase and CamelCase. To be meaningful, this SEP must settle the debate, as it aims at defining clear guidelines for a consistent use of data entry names in SEPs.

**The need for a REGEXP:**

If we want to write a few utilities that deal with those namespaces, we'll need introduce some REGEXP to deal with edge cases such as string with double dots, dots at the beginning, dots at the end. There are also legit characters that are prone to confusion, like space or asterisks. I think this is good (but not required) to rule them out.

Defining a REGEXP that discriminate between valid and invalid entries leaves no place to personal interpretations. It makes it more likely that the standard is harmoniously implemented. That's simpler to have a concise REGEXP than one that include a ton of special characters; It also leads to clearer path. Hence, a restrictive REGEXP seems to be a good solution.

Lowcase-only and highcase-only naming conventions permit to write a REGEXP that will select only valid keys and remove any confusion about low/high case, while CameCase leaves some space to confusion on that matter.

**Having a parsable tree:**

I think this is the most debatable part of this SEP. In fact, in a lot of simple use case using a tree vs a flat structure doesn't bring much benefit. The thing is, we don't have to actually use a tree structure anyway.

Now, I can see use cases where being able to parse the data as a tree structure is convenient. For example, when you need to iterate over the members of a scope (like a 'wallet' scope); Or when you need to test for the presence or the absence of a namespace (does the account have a 'conf.multisig' namespace?)

Now, because of the two first requirements (naming convention and valid key definition), enforcing that namespaces must represent tree structure only cost the conflict-resolution line & its rationale. It is also what namespace are actually supposed to be. In any case, allowing a path to be both a key and a namespace would be confusing.

Overall this part is not as important as the two others, which are crucial to this SEP.

MisterTicot:
I wrote two possible implementations of this SEP that exposes two method:

* `accountData.read(account, [convert])` parse valid namespace entries and convert them to the desired format.
* `accountData.write(account, accountData)` generate a transaction that change **account** data entries to reflect **accountData**.

The first implementation, [flat.js](https://github.com/MisterTicot/stellar-draft-0005/blob/master/flat.js), parse account data entries as a list of keys. It takes 45 lines of code so it's nice for being compact.

The second implementation, [tree.js](https://github.com/MisterTicot/stellar-draft-0005/blob/master/tree.js), parse account data entries as a tree. It takes 70 lines of code. It is not as compact and efficient, but it is dev-friendly.

Both implementations comes with a demo that perform a few basic operations. What we can see is that the tree version makes it a breeze to manipulate the account data tree, while with the flat version everything is a bit more complicated. The reason is the flat version expose namespace as keys, not as actually namespaces.

I didn't write a demo for the case were you want to update the account data entries using a JS namespace, as it is too cumbersome to do with flat.js.

Both implementation have their benefit and are compatible: tree.js could actually be written as a layer ontop of flat.js.

If you want to run the demo (Linux command-line):

```
git clone https://github.com/MisterTicot/stellar-draft-0005
cd stellar-draft-0005
npm install
node flat.js
node tree.js
```

JeremyRubin:
I've thought about this a bit.

I don't really strongly understand the use case where you need this data stored on-chain. I think I'd like to strongly understand what the actual use cases are here so that we can better understand how people are expecting to use this feature. Overall I'd tend towards deprecating this feature if there isn't a strongly motivated use case. Users could still write their own library to use data entries, but we could no longer expose the data-entry operations outside of consensus. You're right -- that should be a separate issue if it's going to happen, but the reason it's relevant to discuss here is we may wish to not further encourage adoption of standards around the feature if it's to be phased out.



That said, as a SEP, this seems to be acceptably specified. I would think about doing some compression of some sort -- e.g., syntax like a.b.c,d,f = hello,world,example would allow you to use fewer entries.

MisterTicot:
@JeremyRubin I answered in your other post: I'd wouldn't like the debate about arbitrary data publishing to make us loose the focus here. However I can answer here about specific use cases related to this SEP:

* [Bootstrapping Multisig Coordination](https://github.com/stellar/stellar-protocol/pull/208) stores the coordinator configuration URL on-chain.
* [On-chain Signature & Transaction Sharing](https://github.com/stellar/stellar-protocol/pull/158) stores its configuration on-chain as well. (not to mention that of course it publishes the signatures & transactions on-chain)

theaeolianmachine:
Hey @MisterTicot,

To move this forward with the new process (and actually get some more momentum into this), could you do the following?

1. Rename the file for how drafts are to be named in that process, and remove the table edit? I'll assign a number when this gets merged in and take care of that for you :).
1. Take a look at the process in the above readme, and let me know if you have any questions.

Once you choose the track, it's up to you to put the time into editing the draft until you're ready for bringing it to a vote. The intent will be to merge this in, and when you make a new PR moving it to `Awaiting Decision` we'll assign a buddy to give feedback and open deliberation to the public. Any draft feedback we suggest you discuss on our public mailing list, so thanks for already doing that! Finally, do note the new process for submitting drafts for future drafts.

MisterTicot:
It definitely needs to go on the Standard track − else it is meaningless. Maybe we'll need to discuss standard namespaces and their usage − such as "config", "profile", "wallet" − see https://groups.google.com/forum/m/#!topic/stellar-dev/VhEhf_s-kBI

Other than that I think the SEP is ready.

theaeolianmachine:
Set this to SEP-0018 in #268. Would make sure to continue to get any additional community feedback by referencing the draft and making additional updates as necessary. When ready to move this further in the process, submit a PR as mentioned in the README. Thanks!

PR #201, opened 2018-11-02
[SEP-6] Clarify asset codes

vcarl:
Some assets are issued with a different asset code than what they anchor (e.g. WSD and USDT for USD), the language used left it ambiguous which code should be used.


vcarl:
As for the reasoning, @tomquisel phrased it well in our team Slack:

> Because you're telling the anchor what you have. What they give you on the other end is up to them

PR #203, opened 2018-11-07
CAP-0007 Draft: Determinisitc Accounts

JeremyRubin:


core/cap-0007.md
```diff
@@ -0,0 +1,150 @@
...
+Currently, when a contract must create an account as a part of it's settlement (such as in Starlight or in an Account Transfer),
+creating the account results in an unpredictable sequence number which prevents the presigning of transactions for it. Furthermore,
+accounts can not be created and modified at the same time preventing initializing an account on behalf of another party.
+
+
+## Specification
+
+We assume a function `PublicKey derive(PublicKey pk, const *char buf, size_t
+buflen)` as specified by [ChainKD Non Hardened Extended Public
+Key](https://chain.com/docs/1.2/protocol/specifications/chainkd#derive-non-hardened-extended-public-key) which returns
```
bartekn: There's a [paper](https://cardanolaunch.com/assets/Ed25519_BIP.pdf) describing some vulnerabilities in ChainKD. Could you check if these affect this CAP in any way? Also, I couldn't find any implementations of ChainKD online.

I'm wondering if instead of deriving the key we could just make the signer field required and create an account with the `master_key = 0`, `signer[0].key = op.signer` and `signer[0].weight = 1`. This will be much easier to use as public key of a new account can be just SHA256 of tx hash and op id so no special derivation libraries are needed in stellar-core and SDKs to create smart contracts.

---

```xdr
struct CreateDeterministicAccountOp
{
    int64 startingBalance;
    Signer signer;
};
```

bartekn: > So we discussed that before and I think it was highly desired that account id continues to be a public key.

It's not the case currently as stellar-core does not check if a new account's public key (created by `create_account`) is a valid ed25519 public key.

I'm fine with ChainKD (if it's confirmed to be safe) but if there's no real advantage of using it, I'm voting for a simple hash as it's much easier option.

danrobinson: Strong, strong vote in favor of using a hash rather than a public key.

core/cap-0007.md
```diff
@@ -0,0 +1,150 @@
...
+
+
+
+## Rationale
+
+CreateDeterministicAccount is a minimal deterministic account derivation scheme.
+
+The newly created account is deterministically derived from
+
+	- The transaction creating it, which transitively commits to:
```
danrobinson: This would mean that the transaction creating it can't send money to the newly created account, or have operations sourced from it (because there'd be a circular hashing problem). I had been worried about this, but now I realized that it should probably be fine, since you could split the creating transaction (the one from some other source account) into two steps, one of which creates the account and includes any other steps that should be committed to in the account ID, and the next of which does anything that shouldn't be committed into the account ID. You could even authorize these two transactions atomically in a single preauth. So I think what you're suggesting here does actually cover essentially all cases, including the ones you were considering adding the more complicated bitfields for! Good thinking

bartekn:
I was wondering if SQL schema changes belong here but it seems that [CAP-0003](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0003.md) includes them. The one important change is to add a key type in `accounts` table and change primary key from `(accountid)` to  `(accountid, account_type)`. This should be done to prevent creating _normal_ accounts with the same `accountid` as the deterministic account (before the transaction creating the deterministic account is included in the the ledger).

Another idea is to use [`strkey`](https://github.com/stellar/stellar-core/blob/master/src/crypto/StrKey.h#L17) encoding to make sure the first letter of a deterministic account id is different (maybe `D`?). This way no DB changes are required and it will be easy to distinguish account types. I think I like it more.

core/cap-0007.md
JeremyRubin: Correct. We could also allow operations to specify account with a trick (e.g., have a pending account ID that only hashes up to the operation in question and then finalize a different account ID).

JeremyRubin:
I think it's fine to create a normal account with the same ID as a deterministic account, so long as the prefixes differ and are signed. 

This proposal does have a different letter for the key type already I think, because we change the type tag. 

Let me know if my understanding of type tag is wrong. 


A simple way to handle SQL stuff is to just use a different table for deterministic accounts. This may be good for performance reasons because we can remove the uniqueness constraints on it as they are cryptographically guaranteed to be unique.

core/cap-0007.md
```diff
@@ -0,0 +1,150 @@
...
+The result of this operation is an account created with the following state:
+```c++
+AccountEntry
+{
+    accountID = derive(op.keyToDeriveFrom, SHA512()
+										.hash(tx.hash)
+										.hash(op.index)
+										.digest() ,64),
+    balance = op.startingBalance,
+    seqNum = 0,
```
MisterTicot: Sequence number 0 have a special signification in SEP-0007 where it means "replace me by current sequence number". I don't see any real incompatibility, but maybe this is better to use seqNum=1 anyway as sequence=0 were not supposed to be possible until here?

MonsieurNicolas: ... or not use magic numbers in the first place? I thought that idea was abandoned in favor of templates as it's brittle and a recipe for bad code in the wild? In the future we may also want to use the full range of sequence numbers.

JeremyRubin:
There are a couple issues in the proposal as is that @danrobinson and I uncovered in discussion.

We'll take it back to the whiteboard for a moment, and post an updated proposal when applicable.

Not going to close so that discussion on the general idea can continue.

core/cap-0007.md
JeremyRubin: @MisterTicot if the seq num starts at zero, then the next transaction is at seqNum 1, so there is no incompatibility.

We will never execute a txn at sequence number 0.

JeremyRubin:
I updated this CAP to address the concerns.

JeremyRubin:
Updated to include a spec for an alternative to just referencing the last created account.

Noted that merges should not look at the latest ledger as det accounts are not replayable.

PR #205, opened 2018-11-08
SEP-0013 Account Transfer Permissionless Payments Protocol (@p2p)

JeremyRubin:


theaeolianmachine:
Given the desire to go another route (be it CAP-0013 or some further iteration) rather than the proposal in SEP-0013, I'm closing this and #225 out.

PR #207, opened 2018-11-08
Update sep-0006.md

istrau2:
Adds a `/transaction` endpoint for querying/validating a specific transaction.
Adds a `kind` parameter in `/transactions`.

@tomquisel

### DIFF
--- a/ecosystem/sep-0006.md
+++ b/ecosystem/sep-0006.md
@@ -390,6 +390,7 @@ Name | Type | Description
 `account` | string | The stellar account ID involved in the transactions
 `no_older_than` | UTC ISO 8601 string | (optional) The response should contain transactions starting on or after this date & time
 `limit` | int | (optional) the response should contain at most `limit` transactions
+`kind` | string | (optional) The kind of transaction that is desired. Should be either `deposit` or `withdrawal`.
 `paging_id` | string | (optional) the response should contain transactions starting prior to this ID (exclusive)
 
 On success the endpoint should return `200 OK` HTTP status code and a JSON object with the following fields:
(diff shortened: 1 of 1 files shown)


istrau2:
@tomquisel as an aside (unrelated to this PR), is there a reason we don't use asset_issuer throughout all the endpoints to identify an asset. We have a use case for it. We are developing an anchor-as-a-service and would like the ability to offer multiple customers the same code. It will be impossible to offer that if asset_issuer isn't used to identify an asset. Of course, if an anchor is not using different issuing accounts, they can always just ignore the parameter. It would change the response of the `/info` endpoint though...

istrau2:
@tomquisel Any thoughts? We are trying to migrate to SEP0006 and running into these issues. Thanks.

tomquisel:
@istrau2 thanks for submitting the PR! It basically looks good to me. A few minor things:

* can you update the version number for SEP-6?
* can you specify which arguments to `/transaction` are optional? I assume it's any one of the three, but being explicit seems wise.
* Just a question: where does the client get the transaction id from? It seems like right now the only way to get a transaction id is from the `/transactions` endpoint. Is there any other approach you're using to communicate the transaction id to the wallet?

tomquisel:
@istrau2 as far as the `asset_issuer` aside, I like the idea of adding that argument, but I don't think the protocol will be backwards compatible if we add it. A wallet that doesn't know that it must be passed to `/info` will get the wrong result. The best approach I can think of is to specify a different `TRANSFER_SERVER` for each of your clients (since presumably each will have their own `stellar.toml`), and each client's server can live at a different path on your hosted server. Will that meet the need?

istrau2:
@tomquisel Ok, fixed the first two points.

With regard to the third point - this is mostly for another application that has a transaction (gotten by whatever means) and wants to get it by id. Couple of examples:
- We are working on an anchor as a service where the asset specific logic lives on another server (relay server). When the relay server receives a notification that it needs to execute a withdrawal, it first goes back and confirms the transaction with the transfer server. It needs to be able to fetch a transaction by id.
- Lots of times, client applications (like react front ends) have a route that looks like `/transactions/:id` where the app needs to be able to render that on first load. It would have to fetch the transaction by id to do that.

As I am writing this, I see you responded to my `asset_issuer` request. Having multiple processes for different clients isn't really feasible. I think this protocol is fairly new, is it not? Couldn't we just add in the `asset_issuer` property? Anyway, this is only not-backward compatible for client applications. The anchors that don't need it could simply ignore it. There aren't very many client applications that I am aware of that are working on SEP0006 (outside of StellarX and Stellarport).

istrau2:
@tomquisel I added it to the endpoint list.

istrau2:
Also, with regard to `asset_issuer`, it is a pretty easy piece of information to include as I would assume that any client application that has the code, has the issuer available as well.

tomquisel:
@istrau2 I'm happy with this PR, but don't have permissions to approve it. @jedmccaleb?

Ishai, can you also open a separate PR for the `asset_issuer` addition?

istrau2:
@tomquisel hmm, yeah I suppose we could use:
```
TRANSFER_SERVER=https://ourtransferserver.com/asset_issuer
```
and then just parse that in the same endpoint for all.

I am going to give that a shot before I submit a PR for a `asset_issuer` param.

PR #208, opened 2018-11-14
Bootstrapping Multisig Transaction Submission SEP

pselden:
[Link to SEP](stellar-protocol/drafts/sep_pselden_3d6e9988.md)

[Google Groups Discussion](https://groups.google.com/forum/#!topic/stellar-dev/-erucceOKbs)


drafts/draft-0005.md
```diff
@@ -0,0 +1,188 @@
...
+```
+
+## Simple Summary
+
+Provides a standard way for multisig accounts to designate where transactions should be submitted for coordination of additional signatures.
+
+## Abstract
+
+An account may designate which multisig server to use in a data entry that points to a server with a `stellar.toml` file. This file contains a `MULTISIG_ENDPOINT` field which is an API where multisig transactions should be submitted. Wallets should implement this SEP so they can support multisig accounts without requiring them to run a multisig service themselves.
+
```
MisterTicot: There's a big hole: how am I suppose to actually get pending transactions? Coordination have two directions: how do I know my co-signers did send transactions to the coordinator?

drafts/draft-0005.md
```diff
@@ -0,0 +1,188 @@
...
+Example:
+
+`POST https://stellarguard.me/api/transactions `
+
+```json
+{
+  "uri": "web+stellar:tx?xdr=AAAAAHFd0%2BHQV5u%2FY%2FfM3%2BVelUr1IWwSFL8CUDIAUudUdD4MAAAAZAADyI8AAAADAAAAAAAAAAAAAAABAAAAAAAAAAEAAAAAcV3T4dBXm79j98zf5V6VSvUhbBIUvwJQMgBS51R0PgwAAAAAAAAAAACYloAAAAAAAAAAAVR0PgwAAABAEw8ODG0iixkbAHg1aJATAnZS2531PhGauuSvFDad2WxHKzIenUNbc7K5mGiSpe5jvqe19OQCbNFuBjqN11jfBw%3D%3D"
+}
+```
+
```
MisterTicot: This need to be changed into a proper POST request such as:
{
  "xdr": ...,
  "callback": ...
}

drafts/draft-0005.md
```diff
@@ -0,0 +1,188 @@
...
+{
+  "id": "c585ceee-b1b9-4009-aa37-b8544346a036",
+  "statusHref": "https://stellarguard.me/api/transactions/c585ceee-b1b9-4009-aa37-b8544346a036",
+  "extras": {
+    "isStellarGuard": true,
+    "url": "https://stellarguard.me/transactions/c585ceee-b1b9-4009-aa37-b8544346a036"
+  }
+}
+```
+
```
MisterTicot: This seems too much complicated to me: why introducing a new kind of id when we already have txHashes ?

drafts/draft-0005.md
```diff
@@ -0,0 +1,188 @@
...
+1. Before submitting the transaction, it should contain at least one valid signature.
+1. `POST` to the `MULTISIG_ENDPOINT` endpoint with the SEP-0007 `tx` URI in the body as the parameter `uri`.
+1. The service responds with a 200 HTTP status code and a JSON object specified in the Submit Transaction Endpoint Response section.
+
+## Rationale
+
+When implementing StellarGuard for the Stellar Account Viewer, it was rightly pointed out it should avoid vendor-specific implementation. This SEP provides an alternative approach that can be used by tools that want to support multisig accounts without implementing a multisig server for each product. 
+
+One design choice was whether to skip the `stellar.toml` step and add the fully qualified multisig server endpoint to the account's data field. The `stellar.toml` approach affords greater flexibilty for the server to change its implementations. It is much easier to update the `stellar.toml` than to update the data entry of every account that uses it.
+
```
MisterTicot: I'm still wondering about that one: URL changing from things such as coordinator are not a very welcome event.

drafts/draft-0005.md
```diff
@@ -0,0 +1,188 @@
...
+Access-Control-Allow-Origin: *
+```
+
+### Multisig Transaction Flow
+
+1. Determine whether a transaction requires additional signatures in order to be valid.
+1. If the transaction requires more signatures, look up the data entry with the key `multisig.domain` on the transaction source account.
+1. Resolve the `stellar.toml` associated with the account's `multisig.domain` entry and return the `MULTISIG_ENDPOINT` field.
+1. Convert the transaction to a SEP-0007 style URI, setting any additional parameters such as `callback`.
+1. Before submitting the transaction, it should contain at least one valid signature.
```
MisterTicot: As said, this should be vendor specific

drafts/draft-0005.md
```diff
@@ -0,0 +1,188 @@
...
+
+In order to comply with browser cross-origin access policies, the service should provide wildcard CORS response HTTP header. The following HTTP header must be set for the API endpoints:
+
+```
+Access-Control-Allow-Origin: *
+```
+
+### Multisig Transaction Flow
+
+1. Determine whether a transaction requires additional signatures in order to be valid.
```
MisterTicot: Since the discussion on the list I think this should be removed. A point to solve: what must be the behavior if the transaction is valid? send it to coordinator anyway? send it to horizon? I think the coordinator must always be in charge of posting validated transactions for consistency (maybe it got its own way to deal with it).

drafts/draft-0005.md
```diff
@@ -0,0 +1,188 @@
...
+
+A library that implements this protocol has been published here: [https://github.com/stellarguard/multisig-utils](https://github.com/stellarguard/multisig-utils)
+
+A demo using that library that submits to the StellarGuard multisig endpoint: [https://stellarguard.github.io/multisig-utils/demo/](https://stellarguard.github.io/multisig-utils/demo/)
+
+## Outstanding Questions/Discussion
+
+1. Should we specify that multisig services expose and return a `statusHref` field for status updates in the response data and also dictate how that should endpoint should function? Applications may have specific use-cases or implementations that this may not work with (for instance, what if they want to use event source or websockets instead of a polling endpoint -- or if they do not want to expose a public status endpoint at all for privacy reasons). However, if we do not specify exactly how it should work then it's possible there will be multiple incompatible implementations.
+
+1. Do we need to explicitly define how the endpoints should look, such as adding `/submit` or `/status` to the MULTISIG_ENDPOINT?
```
MisterTicot: It looks like we don't.

drafts/draft-0005.md
```diff
@@ -0,0 +1,188 @@
...
+
+1. Should we use `MULTISIG_SERVER` or `MULTISIG_ENDPOINT` as the key in stellar.toml. `SERVER` is consistent with the naming of `FEDERATION_SERVER`, but `MULTISIG_ENDPOINT` more accurately describes that it is not just a hostname, it's the actual path to the endpoint.
+
+1. What should the data entry key be for the multisig domain? `multisig.domain` aligns with any particular "namespaces" implementation that may come out of [this discussion](https://github.com/stellar/stellar-protocol/pull/199). However `multisig_domain` is more consistent with the field `home_domain` used on the account.
```
MisterTicot: conf.multisig.domain is likely to be the thing.

drafts/draft-0005.md
```diff
@@ -0,0 +1,188 @@
...
+The server implementation is currently live on [https://test.stellarguard.me](https://test.stellarguard.me) and [https://stellarguard.me](https://stellarguard.me).
+
+A library that implements this protocol has been published here: [https://github.com/stellarguard/multisig-utils](https://github.com/stellarguard/multisig-utils)
+
+A demo using that library that submits to the StellarGuard multisig endpoint: [https://stellarguard.github.io/multisig-utils/demo/](https://stellarguard.github.io/multisig-utils/demo/)
+
+## Outstanding Questions/Discussion
+
+1. Should we specify that multisig services expose and return a `statusHref` field for status updates in the response data and also dictate how that should endpoint should function? Applications may have specific use-cases or implementations that this may not work with (for instance, what if they want to use event source or websockets instead of a polling endpoint -- or if they do not want to expose a public status endpoint at all for privacy reasons). However, if we do not specify exactly how it should work then it's possible there will be multiple incompatible implementations.
+
```
MisterTicot: * If we use txHashes we can remove the whole statusHref thing.
* Dictate how the endpoint should function is the job of this SEP.
* We should allow vendor-specific extension, and let them out of the realm of this SEP, or, we can make event source / websocket an *optional* extension whose behavior is specified by this SEP. In any case coordinators/utils must implement the base protocol.

drafts/draft-0005.md
```diff
@@ -0,0 +1,188 @@
...
+```json
+{
+  "id": "c585ceee-b1b9-4009-aa37-b8544346a036",
+  "status": "pending",
+  "uri": "web+stellar:tx?xdr=AAAAAHFd0%2BHQV5u%2FY%2FfM3%2BVelUr1IWwSFL8CUDIAUudUdD4MAAAAZAADyI8AAAADAAAAAAAAAAAAAAABAAAAAAAAAAEAAAAAcV3T4dBXm79j98zf5V6VSvUhbBIUvwJQMgBS51R0PgwAAAAAAAAAAACYloAAAAAAAAAAAVR0PgwAAABAEw8ODG0iixkbAHg1aJATAnZS2531PhGauuSvFDad2WxHKzIenUNbc7K5mGiSpe5jvqe19OQCbNFuBjqN11jfBw%3D%3D",
+  "signers": ["GBYV3U7B2BLZXP3D67GN7ZK6SVFPKILMCIKL6ASQGIAFFZ2UOQ7AZB5V"]
+}
+```
+
```
MisterTicot: Please for god sake just send me back the raw XDR.

drafts/draft-0005.md
pselden: Just reiterating what we talked about privately: "config.multisig.domain" (looks like we have plenty of chars leftover in they key size so no need to abbreviate it).

drafts/draft-0005.md
MisterTicot: Would be nice to be able to review the whole design, rather than just half of it.

drafts/draft-0005.md
MisterTicot: I don't think it's supposed to be this way: You, as the designer, are supposed to provide the rationale as to why you impose such a strange use of POST request into the protocol - rather than me, as the reviewer, am suppose to motive the fact that POST request should be used as intended.

> there are other benefits of using SEP-0007 URLs that will help interoperability

Such as?

drafts/draft-0005.md
MisterTicot: > You can have two different signing requests that have the same tx hash but the callback is different

That's weird to introduce complexity for such a far-fetched edge case. If providing multiple callbacks is really important (which I doubt) it would be better to allow to do so into the POST request. It's the same transaction and you probably don't want to collect signatures twices / have two entries for it in your system.

> think of an enterprise that is using Stellar that has their own internal multisig service that does not need to interoperate

Isn't interoperability the point of writing a protocol ? Why an enterprise needs to be told how to implement its own internal signature collector?

drafts/draft-0005.md
MisterTicot: Ok don't want to overthink on that one.

drafts/draft-0005.md
MisterTicot: Then nothing prevents the coordinator to use them if he consider then necessary. I would put the question the other way: what's the benefit of forbidding coordinator to accept unsigned transaction - maybe they authenticated the user by some other means. How does it matter to us?

I don't think I would actually send unsigned transaction but I certainly wouldn't forbid possibilities just because I don't see how it may be put at use in the future.

drafts/draft-0005.md
MisterTicot: Actually 've said exactly the opposite: if an account did enable multisig coordinator, **then** it likely need it, **then** its transaction are unlikely to be valid at that point. **Even if** the transaction were valid, you would send it to coordinator in case it triggers side-effect (logging transaction) or in case in have a special treatment of them (sending to a particular endpoint, email notification, ...). Either you respect user configuration, either you don't. If you do, you clearly don't need to force wallet checking for transaction validity.

MisterTicot: The point is: a signature coordinators are going to act as a submissions service.

drafts/draft-0005.md
MisterTicot: Ok we go this way.

drafts/draft-0005.md
MisterTicot: There are already people implementing that so I think it would be nicer to work on that now rather than saying them to do it again because it's not standard later. Also wallet are going to need to know how they are supposed to handle listening. I don't think it should be a mandatory thing but that's needed.

drafts/draft-0005.md
MisterTicot: Fetch the transaction & its signatures to present them to the user.

drafts/draft-0005.md
nikhilsaraf: I agree with your goal of keeeping things flexible here. The questions I’ve posed above are important to be answered before we can allow offering both signed and unsigned options in the procol, otherwise the protocol will be _broken_ because wallets will not know whether the service accepts signed or unsigned XDRs.

Anyhow, since we don’t have a use case right now then it does not make sense to bloat up the SEP, which only makes it more expensive to implement. We can always extend the SEP if we have a real use case for this later.

drafts/draft-0005.md
nikhilsaraf: When you say “wallets are going to need to know how they are supposed to handle listening”, what are you referring to? What will wallets listen for? — I fear that this conversation is going to go down the thread of multisig coordination so I will say it again, this SEP is **not a protocol for how to do multisig coordination**.

drafts/draft-0005.md
MisterTicot: I think you didn't got it: how am I getting to display pending transaction **in my wallet**? If my wallet can't get pending transactions from a multisig service how do I offer user to sign them?

drafts/draft-0005.md
MisterTicot: I didn't said it's non-standard, I said it's weird. You still said nothing about the benefits.

drafts/draft-0005.md
MisterTicot: If someone setup signature collection for its account he'll need it. I don't get how you can miss that.

My point was not about how people would use it, but about the fact that if you happen to send part of your transactions to multisig coordinator (for some reason it's not valid yet) and some other not (maybe you got the signature from somewhere else like preTxAuth) you're going to create inconsistencies into the coordinator service. Either you use it either you don't, but one time yes one time no looks like a good way to create a mess.

Now my proposal allow to make it consistent while to remove the bloat for wallet implementors to have to check for transaction validity, which is a non-trivial task really. You can keep this requirement if you want: but as I shown it's not really useful and it makes the SEP significantly harder to implement.

drafts/draft-0005.md
nikhilsaraf: I’ll say it again.. This SEP **does NOT** describe _how a user will sign **pending transactions** or **signature requests** or **multisig requests**_ or anything else of that nature. That falls into the area of **how to do multisig coordination**, which this SEP is **strictly not**.

What this SEP does describe is **how to send a signature request to an external multisig service**.

drafts/draft-0005.md
MisterTicot: It's still about having the wallet being able to expose the status of the transaction but somehow you don't want it in there so I really don't get what you're trying to do. I'm going to leave it as it is because now that whole thing is nothing but confusing to me.

drafts/draft-0005.md
nikhilsaraf: Here are some of the benefits:

1. Access to SEP-0007 libraries (both for multisig services and wallets that choose complex payment requests beyond the basic encoding that I coded up in the email thread). **This is a general benefit you get when building layers of protocols (think HTTP vs. TCP, why do you use HTTP for websites and not use TCP socket programming for everything?)**
2. Take advantage of any future improvements to SEP-0007 that expand the capabilities of defining a unique payment request.
3. SEP-0007 URLs are standardized and can be hashed to provide a **unique ID** with a guarantee that it is the full request and is not missing any information in the hash value. This can be used by the multisig service _if they want to_ and is logically a very good ID to use.
4. Security guarantees with a well-defined protocol of how to sign payment requests and also how to verify them — also easily usable from a shared open-source library.

drafts/draft-0005.md
nikhilsaraf: I have not understood your response. Can you provide context?  — it’s confusing when you use pronouns like “it” then I do not know what you are referring to

By definition, a wallet is capable of checking if a transaction is valid, since it will submit to the network when the user hits submit. So what is the **additional cost** on a wallet implementer?

drafts/draft-0005.md
MisterTicot: Well then, why is an endpoint to get tx status specified in that SEP?

I'm sorry you have to use a lot of those dark characters here but maybe the lack of clarity is not only from my side, you know.

nikhilsaraf: The [HTTP specification for a POST request](https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.5) species the following:
> If a resource has been created on the origin server, the response SHOULD be 201 (Created) and contain an entity which describes the status of the request and refers to the new resource, and a Location header (see section 14.30).

To comply with the HTTP POST request semantics we need to provide a link to fetch the _status_ of this newly created resource and also therefore need to support a **status** endpoint to allow _retrieval of this newly created resource_.

drafts/draft-0005.md
nikhilsaraf: The wallet can expose the status of the transaction [to the user] by displaying the _signers_ from the response.

I don’t see a benefit of exposing **signatures** in addition to the **signers**. These are two different pieces of information and **signers** is more valuable to expose.

drafts/draft-0005.md
MisterTicot: Your answers:

1. Wallets should not send unsigned transaction to coordinator.
2. Example use case: a smart-contract coordinator that accept external inputs from a (likely authenticated) non-signing tier; Or more generally signature coordinator that can be configured to accept external tx request for authenticated tiers.

drafts/draft-0005.md
MisterTicot: I think most wallets are actually not able to check if a transaction is valid; I think most wallet doesn't even handle multi-signature at all. They just have 1-signers account, they sign and they send to Horizon knowing it is valid.

If you have the coordinator handle multisignature, maybe those wallet can join for free; If you don't they will not join because it's actually not so easy checking for transaction validity. And that is the additional cost I'm talking about. And for what? Being able to introduce a non-consistent behavior.

drafts/draft-0005.md
nikhilsaraf: > Why is it a problem than signatures can be retrieved?

My concern is security.
You can get signatures for one transaction XDR with a fixed set of parameters (today we only have a callback but in the future we could have other fields too). Today you only sign full transactions but in the future we could have partial signing of transactions.
Forcing a service to expose _sensitive_ information such as signatures in the protocol leaves open an attack vector where these signatures can be retrieved and applied to a request with different parameters, which is why I'm hesitant.

I understand your use case for transparency between these services, however, would you want your coordinator to transparently pass data between co-signers even if that data was _sensitive_ such as private keys or passwords (or signatures in this case, which is also _sensitive_)?

To state my claim here more explicitly... I'm saying that **signatures are sensitive information** and should be treated more securely.

--- 
> If the service retains information it's not coordination anymore.

Correct, this is **NOT** coordination, as I've described a number of times earlier. Coordination is one use case and a coordination protocol can expose signatures if it deems that to be an important part of the coordination, that is **out of scope for this SEP**.

--- 

> I'm not willing to tell user a tx is signed without having the app check the signature

The protocol in the SEP is defining that the multisig service should only display a `signer` in the list if it has the `signature` for it. So if the multisig service is following the protocol then wallets can trust that they have the relevant signature. At the end the wallet is only going to work with a multisig service that the **user** has chosen via the data field so the user will trust the result of the `signers` list from the multisig service. This should be a good enough reason to trust the result of the signers list from the multisig service without verification of the actual signature, no?

drafts/draft-0005.md
nikhilsaraf: Ok, we can update the first step in the workflow from `Determine whether a transaction requires additional signatures in order to be valid.` to `Determine whether the transaction has been signed by at least 1 signer`.

This will remove the requirement from wallets to perform checks on whether the transaction needs more signatures.

drafts/draft-0005.md
nikhilsaraf: 1. Ok, so we are in agreement that all XDRs that are sent to the multisig service via this protocol should be **signed**?
2. This is inconsistent with ^^your point one above^^. If you have a use-case for sending XDRs that are not signed to a multisig service then why do you say that `wallets should not send unsigned transactions to coordinator` in your point above?

In your example use case, what is an example of an authenticated tier? How will it be authenticated? Please provide more details on this.

drafts/draft-0005.md
MisterTicot: Look Nikhil you keep saying this is not the coordination SEP but status fetching is clearly a coordination component and must be handled as such, because obviously coordination will depends on this component. If it doesn't, it means you're going to introduce another endpoint that mostly does the same thing than this one and that wouldn't be so great.

Now you got a good point saying that prior to publication signatures can be sensitive, but that's not right to put it on the same level than password or private keys. On the ledger signatures are a public data - password and private keys are not. Please don't add to my confusion with this kind of comparisons.

In a trustless design a signature that is sent to the network must be considered compromised - that is public. Now that's worth thinking twice about this but I'd say that the normal coordinator behavior should be to expose signatures (actually on-chain signature collection is plainly public and trustless and in itself that's not an issue).

There may be situations were signatures must kept secret before publication but I think that would be the symptom of a missing or a wrongly designed feature. If you're planning on offer support for that kind of thing it means you're stepping out of the trustless environment to go back to the trust-a-third party one. This is not the way.

> At the end the wallet is only going to work with a multisig service that the user has chosen via the data field so the user will trust the result of the signers list from the multisig service. This should be a good enough reason to trust the result of the signers list from the multisig service without verification of the actual signature, no?

Again mate we're in the crypto space and we're playing everyday with concepts such as "redundancy", "distributed", "trustless". However there's a strong tendency to go back to the old centralized thinking mode. Would you say that because I trust an Horizon node I don't need a decentralized ledger? No. A proper signature coordinator for a trustless ledger must be trustless as well, right? Else the whole thing is not trustless anymore. I think that's how that game is meant to be played.

drafts/draft-0005.md
MisterTicot: Ok. Well there's still something wrong about this design - maybe you're not willing to admit it. That would be solved by the alternative SEP-0007 design I sent you without loosing any of the benefits you mentioned.

drafts/draft-0005.md
pselden: That requirement is already listed as:
> Before submitting the transaction, it should contain at least one valid signature.

It becomes a bit awkward if I move it to the front because it means I need two "early exits". If you can think of a good way to word that I'll move it to the front, but it really doesn't which order it is in as long as all the steps are taken.

pselden:
Changes based on discussion:

    1. `multisig.domain` data entry changed to `config.multisig.domain`
    2. Removed the requirement to validate whether the transaction is totally valid before submission.
    3. Renamed the proposal to "Bootstrapping Multisig Transaction Submission"
    4. Included "xdr" in the status response.

Also removed the discussions section as those details have been resolved now.

Let me know what you think. Regarding point 4, now that I've included that it seems there's less need for this and now it's more of a convenience, since it can be inferred from the XDR instead. Maybe it could be removed?

drafts/draft-0005.md
MisterTicot: I think this is not resolved yet. Being able to register several time the same transaction comes with a lot of trickyness.

The main one is that it would make coordination complicated in case each co-signer received the tx request independently: because instead of coordinating the signatures under the same entry (by txHash) it will create as much entries as signatures are sent (by id).

It raises a lot of question: must the signatures shared between different entries with the same txHash? What is the cost of implementing the required checks? Is it OK to use signature of someone who signed tx with callback1 for validating tx with callback2? Can't I abuse this system by sending the same tx request again and again & change only a few characters?

Being able to have the same tx registered several times will also complicate everybody's work; because you'll have to handle that special case at every level (coordinator, wallets, utils).

Some weeks ago I pointed to @pselden  that if I sent the same tx request twice at StellarGuard it would create two different entries for that transaction. Then if I validate one the other is still there. It shows that this kind of stuff is not so easy to deal with: there's a lot of logic to get right so the user gets what he'd expect.

In short: this is a lot of troubles for a feature that will probably not really get used (multiple callbacks).

Now if you go with identification by txHash it's all solved from start. So I maintain this is the way to go.

drafts/draft-0005.md
MisterTicot: I think we can resolve this point by keeping it this way for now, because you guys are right that in general it's the safest thing to do. If coordinators have to deviate from this rule to implement new features they will simply do so - no harm here.

drafts/draft-0005.md
MisterTicot: Resolved

drafts/draft-0005.md
MisterTicot: Resolved

drafts/draft-0005.md
MisterTicot: Resolved

drafts/draft-0005.md
MisterTicot: Resolved

drafts/draft-0005.md
MisterTicot: Resolved by last update

drafts/draft-0005.md
MisterTicot: I'm not sure if it is resolved or if you want to discuss this some more @nikhilsaraf

drafts/draft-0005.md
MisterTicot: Indeed my point was not to discuss StellarGuard design but to exemplify what will happen if coordinator use arbitrary IDs. StellarGuard accounts only have one external co-signer, now what do you think will happen if 10 co-signers are posting signatures for a given transaction?

You're saying you don't want to enforce coordinators to use txHash: why wouldn't you want to do that if it actually solves a whole group of issues and makes simpler?

Some of the issues that come to mind:

* Duplicates entries for the same transaction
* Impossible to probe endpoint for a given transaction status prior to know its ID
* Force wallet to track additional data (more data to encrypt/store/submit)
* Make changing coordinator a complicated process (it's going to be a complete mess wallet-side)

Advantages of using txHash:

* Predictability of entry ID: once a wallet receive an external transaction request it can immediately probe its status
* Changing coordinator doesn't affect TxPool state
* Signatures are properly aggregated under the same entry

drafts/draft-0005.md
pselden: Yeah agreed this part _is_ coordination specific, but only in the "let's do the minimum work to make things functional and then we can explore additions from there based on real use cases".

drafts/draft-0005.md
pselden: I think we're at an impasse here. I don't know how comfortable I feel making this dependent on an overhaul of SEP-0007

drafts/draft-0005.md
pselden: These are good arguments. Especially compelling is the fact that that you won't have to store any other information about the transaction besides the hash.

Some potential downsides to weigh:
- How do you handle submitting the same resubmitting the transaction with other parameters (callback?).
- How do you handle resubmission of transactions in the even the first one failed and or was rejected by signers?

These could be left up to the coordinator without much impact though (they could decide whether they accept it, overrwrite it, allow a duplicate after the first one is rejected).

That said, @nikhilsaraf I'm convinced that there's enough benefits to make this change. Your thoughts?

drafts/draft-0005.md
nikhilsaraf: I don't necessarily agree with the XDR being returned here but we can consider this resolved for now.

drafts/draft-0005.md
nikhilsaraf: we can revisit this later if needed.

drafts/draft-0005.md
nikhilsaraf: if we do a rework of SEP-0007 at any point in time then the expectation is that this SEP would also use that new protocol, whatever it ends up being. agreeing with @pselden here that this is not a reason to block this SEP. marking this thread as resolved.

drafts/draft-0005.md
nikhilsaraf: the points brought up for using tx hash for **coordination** makes sense, however I want to stress again that multisig coordination is a _specfic case of multisig accounts_ and this bootstrapping SEP is trying to accommodate all cases.

When the multisig coordination SEP comes out, it will be a good avenue to discuss whether coordinators should use tx hashes or not. This is the reason why this SEP supports a **string** ID type which is more general and flexible.

As for storing the ID, that's a convenience that the wallet can (and probably should) provide to it's users. Are they not already storing some information related to the user's account? If they don't want to store anything then they can always additionally allow the user to enter an ID that is retrieved from the vendor's website.

I think we've already gotten way into the weeds here with this one, a string ID is very flexible and if that doesn't work out we can always amend this SEP accordingly.

@pselden added my comments here from our discussion. My understanding is that you support using string IDs as well, please feel free to unresolve this thread and add a comment if I misunderstood that from our discussion.

nikhilsaraf:
@pselden I think this SEP is good to be merged, can you make the following updates to this PR to get it ready to merge?

- [ ] move the document to `ecosystem/sep-0013.md`
- [ ] assign a SEP number of 13 in the document
- [ ] update `ecosystem/readme.md` to include a link to this new document in `ecosystem/sep-0013.md`

drafts/draft-0005.md
pselden: Yes, I support string IDs. The tipping point for me was your point that this doesn't _have_ to be use for coordination at all, and reiterating that this is not about the coordinator.

pselden:
Updated with SEP number and new location.

drafts/draft-0005.md
MisterTicot: I sincerely don't understand your point and I don't think this is resolved. Getting into the details of how this is going to work is required to take the good decisions. The problems I pointed to have not been addressed so far - thus it is not correct to set this point as resolved.

MisterTicot: I sincerely don't understand your point and I don't think this is resolved. Getting into the details of how this is going to work is required to take the good decisions. The problems I pointed to have not been addressed so far - thus it is not correct to set this point as resolved.

MisterTicot:
Feel like a bit of a rush to me. I think all points should be properly resolved before merging. I'd warn against any temptation to skip issues & force merge. Things have to be solved anyway and now is the best time.

In particular my point on the mailing list about remaining cinflict with oc-multisig haven't been answered yet. Oc-multisig will probably not become the most use solution, but there's not doubt it's going to have it's userbase. As a protocol, this SEP must take it in account.

pselden:
Replied in list. Data is clearly a more convenient place to store and access this configuration, so unless there is a very strong feeling that data entries _will_ be removed then I don't think like there's a compelling reason to change it.

JeremyRubin:
There's an active discussion in #221 which is worth weighing in on IMO (bias -- as I'm one of the main authors there) discussing the relative merits of on vs off chain data.

MisterTicot:
Just an update an that one:

We're going to use oc-multisig to build a decentralized anchor for Stellar community photons. 

We didn't solved all the compatibility problem between this SEP & oc-multisig. As this SEP stalled, I have no choice but going ahead with a non standard solution that would introduce inconsistencies & confusion in the future.

The problematic point is that we failed to agree on sharing the same field for coordinator adress (config.multisig.coordinator)

Basically there are three solutions :

* Both this SEP & oc-multisig SEP use config.multisig.coordinator & the worst case scenario is a fail when this field is not of the good type for one of the SEP
* This SEP uses config.multisig.domain but mention that implementations have to check for config.multisig.coordinator if oc-multisig is enabled (lead to inconsistencies if implementator skip this step).
* This SEP continues ignoring oc-multisig configuration, in which cases we'll end up with double multisig configuration in accounts & that'll be a total mess before long.

Of course the first solution is the better one as it's the only one that guarantees a consistent behavior in all cases.

pselden:
https://github.com/stellar/stellar-protocol/pull/208#issuecomment-458407866 - again I'll mention that I feel strongly that we do NOT overload fields like that. The pseudo duck-typing of a field (does it look like a uri, does it look like an address, does it look like ___) is just really strange and makes things inflexible in the future.

Regarding OC-Multisig -- if that was indeed its own accepted SEP then of course we'd say something here about checking it, and maybe just roll it all into the same thing and have them build on each other. But as far as I can tell right now you're asking to add support in this SEP for a product that isn't a standard.

And as far as stalling goes -- @nikhilsaraf can you comment on if there's any timeline or if you want to get more feedback or what the next steps might be?

pselden:
https://github.com/stellar/stellar-protocol/pull/208#issuecomment-458394111 - the idea is that it breaks a user free from having to choose a wallet OR a multisig coordinator (or a wallet with its own coordinator). With enough support for this, in the future you could pick your wallet, you could pick your own coordinator, and they don't necessarily have to be related. Basically it opens the door for standalone coordinators (enterprise?) and lets more people use multisig.

MisterTicot:
Yeah it's nothing ducktaped but simply using object type as a switch. It's perfectly robust, I'm sorry if you don't see it. Now If you don't like it no problem but do you have a better solution? Because else, standard or non-standard, we're going to have to deal with serious coordination failure very soon.

pselden:
Not ducktape, duck type (https://en.wikipedia.org/wiki/Duck_typing)

The solution I see would be to check both fields or have a separate "type" field if you want to be explicit. Otherwise your problem with supporting one but not the other gets WORSE because instead of ignoring it it just crashes or errors out.

MisterTicot:
I think you hadn't conduct the analysis correctly.

Scenario 1: only one protocol supported, shared field:
Field value of wrong type, coordination fails; But it is detects that a coordinator has been set up.
A well coded UX would show a type error.

Scenario 2: only one protocol supported, two fields:
Doesn't detect that a coordinator is already set, automatically setup a new one or ask the user to do so. The likely consequence is that each co-signer wallet will try to coordinate on its on protocol without never issuing an error, but a lot of consequences are possible. If wallet can detect a problem, it won't be able to solve it anyway and fixing the situation will require manual intervention. That'll be a total mess. The keyword here is in fact **unpredictible behavior**.

The case you're advocating for is clearly worse, and this SEP would actually miss its point of reaching interoperability in multisig coordination.

pselden:
Although I think it muddies the purpose of the field a little bit, in the end of the day it's a small part of this here, so if `config.multisig.coordinator` makes things just work better for you then I'll change it to that. In the end it won't change how this SEP functions besides the name of the data field.

MisterTicot:
That's great because it allows me to go ahead pushing innovations (decentralized anchors) without introducing any major drawback.

johansten:
> [#208 (comment)](https://github.com/stellar/stellar-protocol/pull/208#issuecomment-458394111) - the idea is that it breaks a user free from having to choose a wallet OR a multisig coordinator (or a wallet with its own coordinator). With enough support for this, in the future you could pick your wallet, you could pick your own coordinator, and they don't necessarily have to be related. Basically it opens the door for standalone coordinators (enterprise?) and lets more people use multisig.

And how do you get rid of the coordinator siloing? The coordinators don't talk do each other..
All you're doing is negotiating *which* single-point-of-failure to use, as if that improves anything.

pselden:
@MisterTicot I've updated the SEP with the data field name change.

MisterTicot:
Great, I'll put myself at work to update on-chain coordination SEP as soon as possible.

A few words about the current state of this SEP − all points have been successfully resolved but 2:

1. Use a proper POST request to post the tx+signatures. We stayed stuck on that one even while all the arguments favoring the SEP-0007 case have been defeated. My point of view is that this part of the design now resides out of the realm of intellectual honesty. We agreed to move the debate to SEP-0007 design itself, taking the risk promote a protocol that'll have to be modified after being partly adopted − or to keep it flawed forever.

2. Register transactions by txHash. Unfortunately this was not possible to rationally confront pros & cons on that one, as the whole question have been pushed away as "being part of signature coordination". The fact is that this SEP is going to be part of signature coordination; and while I asked more precisions about how the whole thing is going to work, I got none. Being someone who actually implemented a functional signature coordination system, I got the impression that you guys actually don't have a clue about how it is going to work. Else, you would have been able to defend your design choice instead of pretending it's out of subject (when it's not). Here again, we keep a flaw for later (or forever).

P.S.: Nothing personal, protocol quality is my focus. :kissing_heart:

pselden:
Feel free to make a PR against my branch or a separate proposal (I don't mind if you take most of this if you want to).

@nikhilsaraf I believe the Satoshipay team may have been working on a SEP for the actual coordination aspect of it, is that correct?

nikhilsaraf:
@pseldon yes, I’ve been working with the Satoshipay team on how coordination can work. Once that’s been well tested we will propose that as a SEP.

sui77:
I just implemented a sample [SEP-0008](https://github.com/stellar/stellar-protocol/blob/4da1e71cda4fb1a287f2f49263ac7a340e04c6df/ecosystem/sep-0008.md) wallet/server (https://sep8.sui.li/) . SEP-0008 is about automatically 2-of-2 co-signing using an approval_server that is defined in an anchors stellar.toml. It has some flaws and while I tried to figure out where to address them, I came across this proposal to notice that it is related.

- SEP-0008 is a good solution when it comes to automated signing. The cosigning endpoint has 5 possible responses and gives you a good chance to get a synchronous signed response right away without polling: 
     - success (returns signed tx), 
     - pending (tx needs to be re-submitted after given timeout), 
     - revised (reject, but proposes changes to the original tx), 
     - action_required (refers to a website), 
     - rejected.

- SEP-0008 is flawed in that point that it locates the information about the cosigning service in the anchor/issuer=>stellar.toml=>[[CURRENCIES]] section, making it impossible to use different assets on the same account.

- This draft has a much better approach to define the location of cosigning endpoints, but it also lacks a bit flexibility since it will only work with one cosigning coordinator per account.

(
- This draft has also some other flaws where I would prefer a SEP-0008 like endpoint: 
  - MULTISIG_ENDPOINT.extras response is too vague to be implemented in a standardized way.
  - The status endpoint does not give me a clue on why it's pending or how it intends to get the missing signatures.
  - status = failed should definitely come along with a decriptive error message.
  - Who is actually submitting the transaction and when?
)
  

So here is my combined proposal:

- Optionally allow multiple data entries, to be precise one per signer:

    - multisigdomain.SigHint1 = example.com
    - multisigdomain.SigHint2 = sui.li

- Flow
   - Determine all cosigning services
   - Walk through corresponding cosigning endpoints as described in SEP-0008 until you reach the required threshold or determine you wont ever reach it.
   - Submit the transaction to the network yourself

theaeolianmachine:
Hey @pselden, @nikhilsaraf,

We recently made some changes to clarify the SEP process and make it more straightforward for contribution. Could you do the following things so we can push this forward?

2. Rename the file for how drafts are to be named in that process, and remove the table edit? I'll assign a number when this gets merged in and take care of that for you :).
3. Take a look at the process in the above readme, and let me know if you have any questions.

Once you choose the track, it's up to you to put the time into editing the draft until you're ready for bringing it to a vote. The intent will be to merge this in, and when you make a new PR moving it to Awaiting Decision we'll assign a buddy to give feedback and open deliberation to the public. Any draft feedback we suggest you discuss on our public mailing list, so thanks for already doing that! Finally, do note the new process for submitting drafts for future drafts.

pselden:
Renamed file, updated preamble, removed accidental deletion of readme.

theaeolianmachine:
Set this to SEP-0019 in #274. Would make sure to continue to get any additional community feedback by referencing the draft and making additional updates as necessary. When ready to move this further in the process, submit a PR as mentioned in the README. Thanks!

PR #210, opened 2018-11-14
Add CAP-0009 Linear/Exterior Immutable Accounts

JeremyRubin:


tomerweller:
`AUTH_PARTIAL_LINEAR` is a great addition to Stellar's compliance story as it can facilitate a sign-to-receive scheme, essentially enforcing [SEP0003](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0003.md). Merchants have asked for something like this. Though, can someone think of a better name? 

While I understand the need for something like `AUTH_EXTERIOR_IMMUTABLE` and `AUTH_EXTERIOR_IMMUTABLE_ACTION`, this sounds a bit cumbersome. I vaguely remember @jonjove commenting that if removing trustlines is the main use case, it can be achieved by adding an exterior immutable flag to trustlines rather than to the account. @jonjove?

PR #214, opened 2018-11-15
Update sep-0006.md

istrau2:
Modifies `how` parameter in `deposit` response to be an object or string. The motivation here is to allow anchors to return a string response that can easily be formatted by client apps (e.g. "depositing more than 10 BTC will require manual confirmation that may take up to 24 hours" or similar).


istrau2:
@tomquisel

tomquisel:
@istrau2 [Here's](https://github.com/stellar/stellar-protocol/pull/215) my suggestion for this.

istrau2:
@tomquisel Ok, thanks for the link.

PR #218, opened 2018-11-27
IPFS Support SEP Draft

sbsends:


sbsends:
@stanford-scs, can you weigh in? Happy to make changes based on community feedback. 

Re: 100 links per operation.
This could be achieved by combining multiple files in IPFS and then referencing the head. 

I like the idea of a single DATA op that has specific account targets. This would effectively combine the two parts of the proposal since you could convey communication information **and** IPFS hashes.

theaeolianmachine:
Hi @sbsends — as part of our new process can you please do the following?

* Pick which track you'd like to propose this idea to: Standard or Informational. Info about both can be found here: https://github.com/stellar/stellar-protocol/blob/master/ecosystem/README.md
* Take a look at the process in the above? Once you choose the track, it's up to you to put the time into editing the draft until you're ready for bringing it to a vote. The intent will be to merge this in, and when you make a new PR moving it to `Awaiting Decision` we'll assign a buddy to give feedback. Any draft feedback we suggest you discuss on our [public mailing list](https://groups.google.com/forum/#!forum/stellar-dev). Also, do note the new process for submitting drafts for future drafts. I'll rename and move this file, along with giving it a number when this is merged in.

theaeolianmachine:
Went ahead and committed this as a draft.

PR #224, opened 2018-12-06
SEP DRAFT-0004 updated accordingly to comments from Tom Quisel and public discussion

orbitlens:
* Simplified API - single URL endpoint for all operations.
* Only JSON output format.
* Support for the multiple issuer accounts.
* Added link to the reference service implementation.

### DIFF
--- a/drafts/draft-0004.md
+++ b/drafts/draft-0004.md
@@ -6,21 +6,34 @@ Title: Dynamic Asset Metadata
 Author: OrbitLens <orbit.lens@gmail.com>, Paul Tiplady <paul@qwil.com>
 Status: Draft
 Created: 2018-09-30
-Updated: 2018-09-30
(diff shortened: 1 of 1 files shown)


theaeolianmachine:
Hey @orbitlens I'd love to put this on track for the SEP process updates. Could you do the following?

* Add a track (Informational or Standard) that you wish this to be on, and consider the pros and cons of each for what you're trying to do. Remember you can always make something Informational to get implementers to work with your idea before moving it to a Standard.
* The file has been moved to sep-0014.md - can you apply your diff to that file? It should be unchanged otherwise.

orbitlens:
@theaeolianmachine Done. New pull request is here: #261
So far I lean towards the Standard track for this SEP as it may be referenced from a few other proposals I have in mind.

PR #226, opened 2018-12-13
Corrects verification of challenge response

istrau2:
The spec is fairly primitive and will break down for more sophisticated setups (like I am sure many anchors have).

@tomquisel

### DIFF
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -108,9 +108,10 @@ To validate the challenge transaction the following steps are performed by the s
 * verify that transaction has time bounds set, and that current time is between the minimum and maximum bounds;
 * verify that transaction contains a single [Manage Data](https://www.stellar.org/developers/guides/concepts/list-of-operations.html#manage-data) operation and it's source account is not null;
 * verify that transaction envelope has a correct signature by server's signing key;
-* verify that transaction envelope has signatures from valid signers of the operation's source account. If the account does not exist on the public Stellar network, then a signature by the operation source must be attached to the transactions. Otherwise, if the account does exist on the public Stellar network, signatures must be present on the transaction with summed weight greater than the medium threshold on the operation's source account;
+* verify that transaction envelope has signatures corresponding to whichever roles are to be included in the token (see the [roles section](#claims) for more info on roles).
 * use operations's source account to determine the authenticating client and perform any additional service-specific validations.
 
+#### Claims
 Upon successful validation service responds with a session JWT, containing the following claims:
 
 * `iss` (the principal that issued a token, [RFC7519, Section 4.1.1](https://tools.ietf.org/html/rfc7519#section-4.1.1)) — the URI of the web service (`https://example.com`)
@@ -118,6 +119,16 @@ Upon successful validation service responds with a session JWT, containing the f
 * `iat` (the time at which the JWT was issued [RFC7519, Section 4.1.6](https://tools.ietf.org/html/rfc7519#section-4.1.6)) — current timestamp (`1530644093`)
 * `exp` (the expiration time on or after which the JWT must not be accepted for processing, [RFC7519, Section 4.1.4](https://tools.ietf.org/html/rfc7519#section-4.1.4)) — a server can pick its own expiration period for the token, however 24 hours is recommended (`1530730493`)
 * `jti` (the unique identifier for the JWT, [RFC7519, Section 4.1.7](https://tools.ietf.org/html/rfc7519#section-4.1.7)) — hex-encoded hash of the challenge transaction (`f0e754c6ab988eb5fb2811c2cacc4d3d2aa4334763b8df7285ef341921e2530a`)
+
+#### Optional Claims
+The JWT may also contain the following optional claims:
+
+* `roles`
+The JWT may optionally include a `roles` (array) claim. In the case that there is no `roles` claim, consumers should assume the roles are equivalent to `['Master']` (see below). `roles` are permission claims related to the **operation's** source account. The available roles are:
+  1. `Master` - The JWT requestor signed with the account's master key (this says nothing about the weight of the master key relative to any of the account's thresholds or whether or not the account even exists on the Stellar network at all).
+  2. `Low` - The JWT requestor signed with signatures summing to equal or greater than the account's low threshold.
+  3. `Medium` - The JWT requestor signed with signatures summing to equal or greater than the account's medium threshold.
+  4. `High` - The JWT requestor signed with signatures summing to equal or greater than the account's high threshold.
 
 #### Request
 
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -108,7 +108,7 @@ To validate the challenge transaction the following steps are performed by the s
 * verify that transaction has time bounds set, and that current time is between the minimum and maximum bounds;
 * verify that transaction contains a single [Manage Data](https://www.stellar.org/developers/guides/concepts/list-of-operations.html#manage-data) operation and it's source account is not null;
 * verify that transaction envelope has a correct signature by server's signing key;
-* verify that transaction envelope has signatures from valid signers of the operation's source account with summed weight greater than the medium threshold on the operation's source account;
+* verify that transaction envelope has signatures from valid signers of the operation's source account. If the account does not exist on the public Stellar network, then a signature by the operation source must be attached to the transactions. Otherwise, if the account does exist on the public Stellar network, signatures must be present on the transaction with summed weight greater than the medium threshold on the operation's source account;
 * use operations's source account to determine the authenticating client and perform any additional service-specific validations.
 
 Upon successful validation service responds with a session JWT, containing the following claims:
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -107,7 +107,7 @@ To validate the challenge transaction the following steps are performed by the s
 * verify that transaction source account is equal to the server's signing key;
 * verify that transaction has time bounds set, and that current time is between the minimum and maximum bounds;
 * verify that transaction contains a single [Manage Data](https://www.stellar.org/developers/guides/concepts/list-of-operations.html#manage-data) operation and it's source account is not null;
-* verify that transaction envelope has signatures from valid signers of the transaction's source account with summed weight greater than the medium threshold on the transaction's source account;
+* verify that transaction envelope has a correct signature by server's signing key;
 * verify that transaction envelope has signatures from valid signers of the operation's source account with summed weight greater than the medium threshold on the operation's source account;
 * use operations's source account to determine the authenticating client and perform any additional service-specific validations.
 
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -107,8 +107,8 @@ To validate the challenge transaction the following steps are performed by the s
 * verify that transaction source account is equal to the server's signing key;
 * verify that transaction has time bounds set, and that current time is between the minimum and maximum bounds;
 * verify that transaction contains a single [Manage Data](https://www.stellar.org/developers/guides/concepts/list-of-operations.html#manage-data) operation and it's source account is not null;
-* verify that transaction envelope has signatures with summed weight greater than the medium threshold on the transaction's source account;
-* verify that transaction envelope has signatures with summed weight greater than the medium threshold on the operation's source account;
+* verify that transaction envelope has signatures from valid signers of the transaction's source account with summed weight greater than the medium threshold on the transaction's source account;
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0010.md
```diff
@@ -107,8 +107,8 @@ To validate the challenge transaction the following steps are performed by the s
 * verify that transaction source account is equal to the server's signing key;
 * verify that transaction has time bounds set, and that current time is between the minimum and maximum bounds;
 * verify that transaction contains a single [Manage Data](https://www.stellar.org/developers/guides/concepts/list-of-operations.html#manage-data) operation and it's source account is not null;
-* verify that transaction envelope has a correct signature by server's signing key;
-* verify that transaction envelope has a correct signature by the operation's source account;
+* verify that transaction envelope has signatures from valid signers of the transaction's source account with summed weight greater than the medium threshold on the transaction's source account;
+* verify that transaction envelope has signatures from valid signers of the operation's source account with summed weight greater than the medium threshold on the operation's source account;
```
nebolsin: What is the reasoning behind threshold level selection here? I believe that depending on the use case it might be low, medium or high, which leads me to think that it might be better to move such check on a higher level. Authentication process goal is to verify that certain request comes from a user who has a private key of a certain public key, and threshold check looks more like authorization (verifying that the identified user has privileges to perform certain action on the account), which is outside of the scope for this SEP.

ecosystem/sep-0010.md
```diff
@@ -107,8 +107,8 @@ To validate the challenge transaction the following steps are performed by the s
 * verify that transaction source account is equal to the server's signing key;
 * verify that transaction has time bounds set, and that current time is between the minimum and maximum bounds;
 * verify that transaction contains a single [Manage Data](https://www.stellar.org/developers/guides/concepts/list-of-operations.html#manage-data) operation and it's source account is not null;
-* verify that transaction envelope has a correct signature by server's signing key;
-* verify that transaction envelope has a correct signature by the operation's source account;
+* verify that transaction envelope has signatures from valid signers of the transaction's source account with summed weight greater than the medium threshold on the transaction's source account;
```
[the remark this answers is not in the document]
istrau2: I see. I read the SEP as saying that the anchor needed to sign with a key for the issuing account. I am ok with the signing account instead though.

ecosystem/sep-0010.md
istrau2: Well, the issue is that this SEP is not really possible to implement (from my POV) without some sort of specification on which sigs are needed by the client. I do agree that if we are being complete, we should specify exactly which threshold is needed for each operation. However, I think it was mostly for time's sake the medium threshold was chosen.

istrau2: In other words, there are many users that have multisig accounts of various forms. How does this SEP deal with them? Do we just produce a token (and allow a user to change KYC docs) for a 1/3 multisig signature? That doesn't seem right...

ecosystem/sep-0010.md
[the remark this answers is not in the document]
nebolsin: I think that spec explicitly says about `SIGNING_KEY` usage [here](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0010.md#challenge). It makes sense because this spec targets any web service which needs to authenticate a user via Stellar wallet, not specifically anchors. So the whole concept of "issuing account" is kind of meaningless here.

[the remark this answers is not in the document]
istrau2: Yeah, I fixed it. Some issuers might also have their issuing accounts locked.

ecosystem/sep-0010.md
nebolsin: The way I see it is that this SEP operates primarily with keypairs, and in general doesn't require an associated Stellar account to exist. For example, it can be used in a centralized multi-sig coordinator (simple web service where you authenticate with a particular keypair and receive a  list of transactions pending signature by your keypair), or there might be a case where corresponding account doesn't exist yet (like the service allows withdrawals to non-funded account via `create_account` operation, but wants to ensure that funds wont be locked out because of a typo, or stollen if some malware tampered with destination address).

I think one possible improvement is to support auth with several keypairs at once (like you pass a comma-separated list of public keys when requesting the challenge and then on the auth stage service can check that the tx is signed by all those keys). But since this SEP is based on `wallet <-> web-service` interaction over relatively short period of time, I don't quite see a real use-case for multi-sig on the auth transaction, because the wallet would have to collect all required signatures a) very quickly (recommended expiration for the challenge tx is 5 minutes); and b) co-signers would have to sign very generic transaction "allow A to perform operations with our shared account X on service S", with no granularity over operations and basically no way to revoke such permission until auth session expires.

> However, I think it was mostly for time's sake the medium threshold was chosen.

My point was that since this SEP defines a generic authentication (not _authorization_) protocol, it doesn't seem to be a proper place to define specific thresholds at all. Authorization logic is generally app-specific but can be defined in higher-level SEPs when necessary.

istrau2: I've updated the text to include the case of the account not existing on pubnet (in which case, the client just needs to send the signature by the master key).

To your point, I understand that this is just an authentication SEP. However, the problem is that right now it is specified as the authentication standard for SEP-6 and SEP-12. Currently, the way this SEP is written, the overall system is broken (because clients are getting authenticated for accounts where they shouldn't be).

I agree with you that an authorization SEP is the right way to do this (where you have application specific tokens and different permission levels etc.) I don't have the time to solve this elegantly in a separate SEP (thus the medium threshold specification). If you do, please go ahead.

I am just a SEP-6 and SEP-12 implementer and simply cannot implement as the whole thing currently exists.

ecosystem/sep-0010.md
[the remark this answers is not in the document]
tomquisel: Great point!

ecosystem/sep-0010.md
tomquisel: @nebolsin @istrau2 This is a great discussion.

@nebolsin you make a great point about this protocol needing to function with key pairs that don't have associated Stellar accounts. I think we should be sure to preserve that and state it as a goal of the protocol in the Abstract section.

I think the use case of multi-sig accounts is important, and will become more common. I'm thinking of users of services like StellarGuard or users with SEP-8 accounts that include one or more issuers as signers. I think the best approach is for this protocol to specify a sensible default way to handle created Stellar accounts (using medium threshold), and make it clear that different applications can choose different thresholds or ways of handling multi-sig accounts. That maximizes the simplicity of building on SEP-10, while also allowing flexibility in certain cases where medium threshold isn't appropriate.

I'm not too worried about the recommended time it takes to sign a transaction appropriately. Implementers can bump it up if it's causing an issue in their case, and automatic signing like SEP-8 will be more and more common.

Thoughts?

nebolsin: @tomquisel I totally agree that multi-sig use case is very important, I'm just not convinced that it should be included into this SEP as a requirement. And as an optional feature it's already allowed by the last item in verification rules so there's no conflict between this SEP and multi-sig: 

> * use operations's source account to determine the authenticating client and perform *any additional service-specific validations*

We can definitely add a note for implementors describing the security considerations for multi-sig case and some hints about recommended setup. 

But let me stop right here for now, because I just realized that even with proposed changes this SEP will not work for StellarGuard case, because as far as I know it receives partially-signed tx, adds a final missing signature after 2FA check and submits the transaction to the network. And this SEP intentionally makes challenge transaction invalid so it will fail if submitted. And for now I don't see a clear solution for this scenario, need to think more about it. Any ideas?

nebolsin: @tomquisel Another consideration against mixing authentication and authorization here is that JWT token obtained via this SEP is a session token and it will remain valid for some prolonged period (currently the spec advises to use 24 hours). But during that time account multi-sig configuration on Stellar could change, for example a person who requested a JWT token could be excluded from co-signers or his weight is adjusted. If we perform authorization as part of authentication, the JWT will encode account multi-sig configuration at issue time and it will remain static throughout token lifetime, which have pretty bad security implications.

I think that's a pretty strong argument to keep authentication focused on identity (so that JWT token only claims ability to sign transactions on behalf of `sub` keypair), while authorization, including multisig checks, should be enforced on a service level for every particular action. 

For example, in case of SEP-12 `/customer` endpoint handler should first extract authenticated keypair from JWT and verify that it is authorized to perform the requested action on the account. To support multi-sig setups it should either accept multiple JWT tokens, or we can extend this SEP to allow single JWT to claim control of multiple keypairs. I'll try to draft a PR for this latter approach.

istrau2:
@nebolsin 
I have to say that I disagree with your last post, authorization **should** be included in the JWT, whether or not it is included in the SEP is a preference thing. Including basic authorization related claims (such as groups, roles and permissions) is a standard, widely adopted practice. This is because centralizing authorization logic (at least the generic, non-application specific kind of logic) is a really good idea, so that one can have a focused authentication/authorization service that is responsible for this kind of thing. For example, think of what google, facebook etc. do when you request an OAuth2 JWT from them, they prompt the user to select which permissions to allow to the app.

More to the point, for our (i.e. Stellarport's) specific implementation of this SEP, we **need** to do authentication and authorization in one place. Our KYC server is not capable of doing the basic authorization (i.e. determining which thresholds are met by certain signatures) because it is not connected to a Stellar node.

In terms of some sort of permissions changing on the underlying account, this is a general problem with JWTs, not really specific to our use case. Any time a service issues a JWT, the underlying permissions on the account might change. That change will not be reflected in the JWT. The way to solve this is by limiting the JWT to a short enough period of validity. In fact, I am not sure why this spec advises a validity period, I would think that is very application specific.

Perhaps the solution here is for the JWT to include an optional `roles` claim which can include `Master`, `Low`, `Medium` and `High` roles. `Master` means that the token guarantees that this user has signed with the master key. The other three correspond to the various Stellar account thresholds. In the case that there is no `roles` claim, the consuming application should assume the `Master` claim, this makes the change in the SEP backwards compatible.

What do you think?

ecosystem/sep-0010.md
istrau2: @nebolsin
I have to say that I disagree with your last post, authorization should be included in the JWT, whether or not it is included in the SEP is a preference thing. Including basic authorization related claims (such as groups, roles and permissions) is a standard, widely adopted practice. This is because centralizing authorization logic (at least the generic, non-application specific kind of logic) is a really good idea, so that one can have a focused authentication/authorization service that is responsible for this kind of thing. For example, think of what google, facebook etc. do when you request an OAuth2 JWT from them, they prompt the user to select which permissions to allow to the app.

More to the point, for our (i.e. Stellarport's) specific implementation of this SEP, we need to do authentication and authorization in one place. Our KYC server is not capable of doing the basic authorization (i.e. determining which thresholds are met by certain signatures) because it is not connected to a Stellar node.

In terms of some sort of permissions changing on the underlying account, this is a general problem with JWTs, not really specific to our use case. Any time a service issues a JWT, the underlying permissions on the account might change. That change will not be reflected in the JWT. The way to solve this is by limiting the JWT to a short enough period of validity. In fact, I am not sure why this spec advises a validity period, I would think that is very application specific.

Perhaps the solution here is for the JWT to include an optional roles claim which can include Master, Low, Medium and High roles. Master means that the token guarantees that this user has signed with the master key. The other three correspond to the various Stellar account thresholds. In the case that there is no roles claim, the consuming application should assume the Master claim, this makes the change in the SEP backwards compatible.

What do you think?

istrau2: I have modified the PR with the optional roles claim instead.

theaeolianmachine:
Latest update on this is to take a second look everything that's been said here; next steps are on @tomquisel (don't merge in yet).

istrau2:
@tomquisel @theaeolianmachine @nebolsin 

We are using this in production and it is working great. Specifically, our SEP10 server produces tokens with a roles array containing one or more of the following options:
`['Master', 'Low', 'Medium', 'High']`
then other servers can allow actions depending on these roles. Some examples:

1. Our KYC server only allows uploading KYC docs for jwts with a `High` role.
2. Our chat server and federation servers only allow chatting or updating a federation record for jwts with a `Medium` role.
3. Our 2FA server only signs transactions submitted that have been signed by the master. This example isn't using the JWT directly (because it has actual transactions it can look at) but illustrates how sometimes, looking at the `Master` role might be necessary.

theaeolianmachine:
Hey @tomquisel, a few things so that we know next steps:

* @tomquisel are you leaning more one way or the other on the approaches in #226 vs #240?
* Are there any other folks on the platform team that you'd like to hand this off to for taking a look at?

tomquisel:
Closing this out. I think it's correct that we don't want to mix authentication and authorization in the same SEP. Let's continue discussion in #331 .

PR #227, opened 2018-12-16
Update sep-0012.md

istrau2:
Adds the option to add a `KYC_SERVER` to send SEP-12 client requests to so that anchor services can split apart apart KYC and TRANSFER functions between web services if desired.


tomquisel:
@istrau2  This makes sense, although I worry about confusion with the `COMPLIANCE_SERVER`. Do you have a need to host your `/customer` endpoint somewhere different from your transfer endpoints?

istrau2:
@tomquisel What is the `COMPLIANCE_SERVER`?

Yes, we need to host our `/customer` endpoint on a different server from the transfer endpoints.

istrau2:
@tomquisel Ok, done.

PR #228, opened 2018-12-16
Update sep-0006.md

istrau2:
Upon doing the work on integrating SEP12, I have noticed what seems to me, an inconsistency within this SEP.  

The SEP states that upon using the `/withdrawal` endpoint, a transfer server has the right to return a demand for KYC. However, how is the the anchor going to know if KYC is needed or not if they do not have a Stellar account to associate the request with?

Throughout SEP6, SEP10 and SEP12, KYC is associated with a Stellar account. It is not possible by using the SEPs directly to associate KYC info with an outside destination because there is currently no standard way of proving a user's rights to that withdrawal destination.

Instead, we are associating KYC with a Stellar account (like Coinbase associates KYC info with a Coinbase account). If so, the Stellar account needs to come in with the request to withdraw.

I hope that is clear...


istrau2:
@tomquisel

istrau2:
Not necessary.

PR #229, opened 2018-12-17
Draft of SEP-13a

tomquisel:
This draft explains how CAP-13 can be used to provide a convenient and easy-to-implement deposit experience for new Stellar network users.

You can see the full SEP [here](https://github.com/tomquisel/stellar-protocol/blob/tq/sep-13a-draft/ecosystem/sep-0013a.md).

### DIFF
--- a/ecosystem/sep-0013a.md
+++ b/ecosystem/sep-0013a.md
@@ -32,52 +32,64 @@ This SEP is an alternative to SEP-13, and has some extra desirable properties:
 
 ## Specification
 
-The general setup is the case of an issuer (Isabel) processing a deposit of an external asset (FunBucks) for a specific user (Ulric). Ulric has already initiated a deposit with Isabel, and specified the Stellar account ID (U) where he wants the deposit to arrive.
+The general setup is the case of an Issuer processing a deposit of an external asset (FunBucks) for a specific User. The User has already initiated a deposit with the Issuer, and specified the Stellar account ID (U) where they want the deposit to arrive.
 
 We consider a few cases:
 
 1. U is not yet created
 1. U is created, but has no FunBucks trustline
-1. U is created, has a FunBucks trustline, but has insufficient XLM to create an offer (in this case Ulric's account is wedged even if the FunBucks deposit is received)
+1. U is created, has a FunBucks trustline, but has insufficient XLM to create an offer (in this case the User's account is wedged even if the FunBucks deposit is received)
 1. U is created, has a FunBucks trustline, and has plenty of XLM
 
 ### Case 1: Account Needed
 
-In this case, Isabel creates U and deducts the cost from the value of the deposit.
+In this case, the Issuer creates U and deducts the cost from the value of the deposit.
 
-1. Ulric deposits 10 FunBucks with Isabel, external to the Stellar network. Isabel must now transfer 10 FunBucks tokens to U on Stellar.
-1. Isabel determines the amount of XLM that will be required to create a functional account for Ulric. The breakdown is:
+1. The User deposits 10 FunBucks with the Issuer, external to the Stellar network. the Issuer must now transfer 10 FunBucks tokens to U on Stellar.
+1. The Issuer determines the amount of XLM that will be required to create a functional account for the User. The breakdown is:
     - 2 * `base reserve` (for the new account)
     - 1 `base reserve` (for the trustline)
-    - 1 `base reserve` (so Ulric will be able to create an offer to sell more FunBucks once the deposit completes)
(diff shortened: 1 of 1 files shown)


tomquisel:
@JeremyRubin I'd love your thoughts.

JeremyRubin:
Thanks for preparing this document Tom!

The core issue with your proposed mechanism comes down to race conditions. I think it's simplest to think about the protocol as follows:

```
while user not paid:
  get account state for user's account;
  determine which case pattern matches the account state;
  generate and sign transactions for the case;
  broadcast the correct transaction;
```

Your proposal is crafted to handle one sort of bad interleaving, where the account is created ahead of time.

However, accounts can also be merged too. E.g., consider this loop:

```


while user not paid:
  get account state for user's account;
  merge account;
  determine which case pattern matches the account state;
  generate and sign transactions for the case;
  broadcast the correct transaction;
```

JeremyRubin:
Thanks for preparing this document Tom! It is a nice complement to side by side with SEP-13 to see the relative merit of each approach.

The core issue with your proposed mechanism comes down to race conditions. I think it's simplest to think about the protocol as follows:

```
while user not paid:
  get account state for user's account;
  determine which case pattern matches the account state;
  generate and sign transactions for the case;
  broadcast the correct transaction;
```

Your proposal is crafted to handle one sort of bad interleaving, where the account is created ahead of time.

However, accounts can also be merged too. E.g., consider this loop:

```
while user not paid:
  create account;
  get account state for user's account;
  merge account;
  determine which case pattern matches the account state;
  generate and sign transactions for the case;
  broadcast the correct transaction;
```

This loop never terminates! This could, if not appropriately handled, cost the issuer a lot of money.

How you specified it so far it's unclear if the issuer should retry many times or just one time and what should happen after the second failure.

Just handling the error demonstrated above isn't sufficient as there is a broad class of related issues that could cause problems. E.g., trustline is added then removed, concurrent issuers paying at the same time, etc. It's generally a hard problem to solve and as an issuer you are legally responsible (imo) for handling these cases correctly. Perhaps something like proper use of exterior immutable accounts would help.


There are a few other categories of problem with this approach that I see around the handling of fees but I think those are addressable with my feebalance proposal so I'll hold back on those for now.




Lastly, to draw the contrast between the two approaches (13 and 13A), 13 is largely immune to the sort of problem here because the actions are deterministic.

tomquisel:
@JeremyRubin Thanks for the feedback!

This is a good point about race conditions, and those while loops potentially not terminating.

The intent of this SEP is that in the general case, if there's an error with the funding transaction, the issuer should just fail and return funds to the user. No retries. So that's how the account merged race condition would be handled, as well as the case of a trustline being removed. If a user does something like merging an account or removing a trustline at the same time they have a pending deposit, it's pretty unclear what result the user wants. It may even be they are making an effort to cancel the deposit. As a result, failing the deposit and refunding seems like the best approach.

The only case where an issuer should retry is this case: 

> In case 1, a third party creates U before Isabel

And in that case, the issuer retries only once before giving up.

Let me know if that helps address those concerns.

tomquisel:
@s-a-y @satbam what are your thoughts on this SEP?

JeremyRubin:
@tomquisel it would be good to see the proposal amended with the exact logic you want in such failure cases.

To me at least it seems a complicated/worrisome from a compliance PoV to implement a wallet against such a spec. I think it's very good for issuers to be able to immediately clear their liability for a user deposit to the network.

The complete protocol spec would also help me get a sense for the magnitude of the griefing caused by this vector and what measures an issuer would need to take to prevent such users from using too much fees.

tomquisel:
@JeremyRubin Thanks for that suggestion! I've added a precise error handling flow. Let me know if you think it's onerous and I can simplify it more.

I also think issuers will be able to quickly clear their liability with their flow. They try the transaction at most twice (no reason this should take more than a minute), and will then either be able to issue a refund or the transaction will have succeeded and the deposit will be completed.

MonsieurNicolas:
Trying to understand the process a bit better when it comes to SEPs (even after reviewing #247 I don't see a mention of PRs being rejected) @theaeolianmachine : should this be closed as well? I see that you closed #225 and #205 (that depend on CAP7/CAP12), and I don't see a real difference with this one that depends on CAP13 (that is nowhere close to "final").
None of those solutions are ready, and I don't know if the other approaches are "broken beyond repair", more that the first round of feedback on them is "too complicated".

theaeolianmachine:
Yeah I hadn't gotten to this one yet — I guess in keeping with the spirit of things I should just merge Jeremy's stuff in, even if it lingers forever.

theaeolianmachine:
Went ahead and merged this as SEP-0017.

PR #231, opened 2018-12-19
Update sep-0012.md

istrau2:
Updates multiple SEPs to use the standard Authorization header for the submission of tokens obtained via SEP-10.


tomquisel:
@jedmccaleb

accordeiro:
Hi @istrau2 – sorry for the delay on getting back to you. I’m currently focusing on addressing the issues & PRs related to SEP6/9/10/12, so you should have faster response times from now on :)

I’ve discussed this issue with @tomquisel and, since some applications can’t rely on headers (see https://github.com/stellar/stellar-protocol/issues/311#issuecomment-496617356), it would make more sense to require servers to support both `Authorization` header and `jwt=` GET param as ways to provide the JWT, but to recommend clients to use the former.

Does that make sense to you? If so, could you update the PR to reflect that? 

Please let me know if I can help in any way.

[Also worth noting that there's a broader discussion about auth on #311]

istrau2:
@accordeiro I think there is some miscommunication here which is leading to confusion.

This PR updates SEP6 and SEP12 with a recommendation on how clients submit tokens acquired via SEP10 to those particular **servers**. Passing the jwt in this fashion is standard. Please see here: https://tools.ietf.org/html/rfc6750#section-2.1

The comment from #311 is referring to how clients should pass tokens to other **client applications** (i.e. doing a browser redirect). Of course, in this case, adding a header would not be possible.

That said, I have no problem accepting the jwt as a query parameter as well. I just wanted to clarify that this PR actually does not cover the case in #311 as they are in fact two separate issues.

Let me know if you still want me to update (in which case I will go ahead and do that).

accordeiro:
@istrau2 yes, we're on the same page about those being separate issues, but thought it was worth referencing the discussion – sorry for the miscommunication.

> Let me know if you still want me to update (in which case I will go ahead and do that).

Yes, please do! :)

istrau2:
@accordeiro done.

ecosystem/sep-0012.md
```diff
@@ -7,7 +7,7 @@ Author: Interstellar
 Status: Accepted
 Created: 2018-09-11
 Updated: 2018-09-11
-Version 1.0.0
+Version 2.0.0
```
accordeiro: Since we're keeping the changes backwards-compatible, do we need to update the major version? Or should we use `1.1.0` instead?

ecosystem/sep-0012.md
```diff
@@ -52,7 +64,6 @@ The fields below should be placed in the request body using the `multipart/form-
 Name | Type | Description
 -----|------|------------
 `account` | `G...` string | The Stellar account ID to upload KYC data for
-`jwt` | string | The JWT previously sent by the anchor via the [SEP-10](sep-0010.md) authentication flow
```
accordeiro: Do we need to delete this line, now that we're keeping the `jwt` param support?

accordeiro:
@istrau2 – please let me know if you're available to address these suggestions, otherwise I'd be happy to finish them up.

ecosystem/sep-0012.md
istrau2: Yep, your'e right. Done.

ecosystem/sep-0012.md
istrau2: Well, I thought about putting it back in but I thought since we are recommending against it, even putting it in with an (optional) might make people think that it is simply optional where really it is not recommended.

PR #233, opened 2018-12-20
Update sep-0012.md

istrau2:
Adds requirement for uploading file type fields after text fields as required by some web servers.


istrau2:
@tomquisel

PR #240, opened 2019-01-08
Support multisig in SEP-10, SEP-12 authentication

tomquisel:
An alternative to #226 for supporting multisig in SEP-10. @nebolsin @istrau2 let me know your thoughts.

It goes the route of supporting multiple signers on JWTs, and leaving authorization to SEPs that build on it.

One possible addition would be to have the ability to encode a permissions list into the transaction so multiple signers could see exactly which permissions are being requested. That adds complexity, though.

### DIFF
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -7,12 +7,12 @@ Author: Sergey Nebolsin <sergey@mobius.network>, Tom Quisel <tom.quisel@gmail.co
 Status: Accepted
 Created: 2018-07-31
 Updated: 2018-08-14
-Version 1.0.1
+Version 1.1.1
 ```
 
 ## Simple Summary
 
-This SEP defines the standard way for clients such as wallets or exchanges to create authenticated web sessions on behalf of a user who holds a Stellar account. A wallet may want to authenticate with any web service which requires a Stellar account ownership verification, for example, to upload KYC information to an anchor in an authenticated way as described in [SEP-6](sep-0006.md).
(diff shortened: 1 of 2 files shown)


istrau2:
Hi Tom,

To put it bluntly, I don't like this proposal (please don't take this the wrong way). This for several reasons:

1. I am not seeing what we gain by having the `signers` array. It seems that by fixing the signers, all we are doing is restricting the flexibility the user has with respect to which signers actually do the signing. For example, in Stellarport, we just present the transaction to the user and let them sign it as they will (even externally) before submitting it back to the JWT server. Ultimately, the service producing the JWT needs to look at an array of signatures and validate them against the actual signers of the `account`.
2. This really doesn't solve the  issue of a consuming service. If I have a  service that is consuming this JWT (that is not connected to Stellar), how am I to run any authorization logic? The whole point of a JWT (in  my mind) is that it has some basic authorization claims in it (like roles and groups) so that other services which cannot compute these basic claims can still do authorization using these basic building blocks which were produced by the authorization server (thus my suggestion in #226 to include the roles array).

nebolsin:
Hey guys,

Sorry for being silent for a long time, it’s been a holiday week here in Russia and I got caught up on other activities. I’ll provide feedback here and on #226 tomorrow.

nebolsin:
@istrau2 Authorization is a complicated matter which will require a lot of effort to research and specify. It's very interesting topic and hopefully we'll get to that discussion, but it's outside the scope of this SEP. This one is basically about "Login via Stellar wallet" feature for web sites or web services. It's intended to be as generic as possible and only define proof-of-possession auth protocol using Ed25519 keys for Stellar ecosystem (something like [bit.id](https://github.com/bitid/bitid), [Trezor Login / SLIP-13](https://github.com/satoshilabs/slips/blob/master/slip-0013.md), etc). Everything else is up to web service itself or higher level SEPs.

tomquisel:
Closing this out due to issues.

PR #241, opened 2019-01-08
Add fee endpoint for complex fee structures

tomquisel:
See the full new [SEP](https://github.com/tomquisel/stellar-protocol/blob/add-fee-endpoint/ecosystem/sep-0006.md) and the new [fee section](https://github.com/tomquisel/stellar-protocol/blob/add-fee-endpoint/ecosystem/sep-0006.md#fee).

### DIFF
--- a/ecosystem/sep-0006.md
+++ b/ecosystem/sep-0006.md
@@ -6,8 +6,8 @@ Title: Anchor/Client interoperability
 Author: stellar.org
 Status: Accepted
 Created: 2017-10-30
-Updated: 2018-11-07
-Version 2.3.0
+Updated: 2019-01-08
+Version 2.4.0
 ```
 
 ## Simple Summary
@@ -38,6 +38,7 @@ To support this protocol an anchor acts as a server and implements the specified
 * [`GET /deposit`](#deposit): required
 * [`GET /withdraw`](#withdraw): required
 * [`GET /info`](#info): required
+* [`GET /fee`](#fee): optional (only needed for complex fee structures)
 * [`GET /transactions`](#transaction-history): optional, but strongly recommended
 * [`GET /transaction`](#single-historical-transaction): optional, but strongly recommended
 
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0006.md
```diff
@@ -387,6 +396,8 @@ For each withdrawal asset, it contains:
 * The fee structure
 * `min_amount`: Optional minimum amount. No limit if not specified.
 * `max_amount`: Optional maximum amount. No limit if not specified.
+* `fee_fixed`: Optional fixed (flat) fee for withdraw. In units of the deposited asset. Leave blank if there is no fee or the fee schedule is complex.
+* `fee_percent`: Optional percentage fee for withdraw. In percentage points. Leave blank if there is no fee or the fee schedule is complex.
```
vcarl: I think it would be better to have blank only indicate a complex fee schedule, with `0` needed here to indicate no fee. Wallets wouldn't want to request `/fee` unnecessarily (esp if it's not implemented), allowing blank to represent no fee would force a more complex check for whether to request `/fee` or not.

ecosystem/sep-0006.md
```diff
@@ -399,7 +410,56 @@ The `fields` object contains a key for each field name and an object with the fo
...
+
+```
+GET https://api.example.com/fee?operation=withdraw&asset_code=ETH&amount=0.5
+```
+
+On success the endpoint should return `200 OK` HTTP status code and a JSON object with the following fields:
+
+Name | Type | Description
+-----|------|------------
+`fee` | float | The total fee (in units of the asset involved) that would be charged to deposit/withdraw the specified `amount` of `asset_code`
```
vcarl: Perhaps `fee`, with optional `fee_percent`, and `fee_fixed` fields so wallets could break it down for users? Just spitballing

ecosystem/sep-0006.md
tomquisel: @vcarl that's a great point. My thought was that wallets would use this stanza to decide if a complex fee structure is in play:

```
"fee": {
    "enabled": true
}
```

I think we need to keep the missing field implies zero semantics for backwards compatibility.

Issue #146, opened 2018-08-08
Cosigned assets: NopOp and COAUTHORIZED_FLAG


tomerweller:
(Posting on behalf of David Mazières)

# Motivation

A number of Stellar asset issuers need greater control over assets
than simply authorizing particular accounts to hold the asset.  As a
result, we are seeing people build solutions in which the asset issuer
is actually a co-signer on asset holders' accounts.  This solution has
several drawbacks.  In particular, the asset owner cannot unilaterally
withdraw XLM or other assets from his or her account, and it is very
hard to trade one such restricted asset for another (as both issuers
would need to be cosigners on the account, giving one issuer
permission to authorize transactions on the other's asset).

This proposal is a first cut at attempting to address the problem
through minimal stellar-core changes.

# Co-signer authorized trust lines

To accommodate asset issuers with restrictions, we propose adding
another mode for trustlines of `AUTH_REQUIRED` assets, called
"cosigner authorized," that lies somewhere between being not
authorized and fully authorized.  The idea is to allow issuers to
require asset holders to request co-signing of transactions involving
the asset without being cosigners on asset holders' accounts.

```c
enum TrustLineFlags
{
    // issuer has authorized account to perform transactions with its credit
    AUTHORIZED_FLAG = 1,
    // issuer allows transactions that are co-signed by issuer low threshold
    COAUTHORIZED_FLAG = 2
};
```

There are now three modes for the trustline of an `AUTH_REQUIRED`
asset, based on `TrustLineEntry::flags`:

* If `(flags & 3) == 0`, the source account is not authorized.  Any
  `PAYMENT`, `PATH_PAYMENT`, `MANAGE_OFFER`, or
  `CREATE_PASSIVE_OFFER`, with the source account causes the whole
  transaction to fail, with only two exceptions:

    * A `PATH_PAYMENT` is allowed to contain the asset in `path`, so
      long as it does not apear in `sendAsset` or `destAsset`, and

    * A `MANAGE_OFFER` is allowed if it sets `amount` to 0 (to delete
      the offer).

  In addition, any operation with a different (even authorized) source
  account will fail if it attempts to send the asset to an account
  with `(flags & 3) == 0`.  Finally, existing orders on the order book
  belonging to the account with `(flags & 3) == 0` are immediately
  invalid and cannot be filled.  This behavior should be close or
  identical to the status quo.

* If `(flags & 3) == COAUTHORIZED_FLAG` for an account, then the
[...]

PR #244, opened 2019-01-24
add cap0016: cosigned assets

tomerweller:
Migrating from https://github.com/stellar/stellar-protocol/issues/146
closes https://github.com/stellar/stellar-protocol/issues/146

### DIFF
--- a/core/cap-0016.md
+++ b/core/cap-0016.md
@@ -1,0 +1,189 @@
+## Preamble
+
+```
+CAP: 0016
+Title: Cosigned assets: NopOp and COAUTHORIZED_FLAG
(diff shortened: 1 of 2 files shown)


MonsieurNicolas:
I don't think you want to close 146 as it's the issue used to track the conversation related to this CAP. If you agree, update your comment to avoid closing the issue @tomerweller

theaeolianmachine:
It feels clunky though to keep the thread going in the issue though, no? Seems like we'd rather push any future conversation to the mailing list — we can certainly announce it's in draft and copy over the existing conversation.

PR #246, opened 2019-01-30
Add optional `amount` argument to SEP-0006

vcarl:
I realized that anchors have no insight into how much users are attempting to withdraw or deposit, leaving validation of min/max entirely up to the wallet initiating the transfer. This seems less than ideal, the anchor should have an opportunity to enforce their limits as well.


theaeolianmachine:
@tomquisel can you take a quick look — I think this change is very minor and should be easy to merge in.

tomquisel:
I'm not sure I'm in favor of this. There are already two mechanisms for the issuer to communicate min and max amounts to the wallet (`/info`, and the `/deposit` responses). I think the SEP just gets more confusing by adding this explicitly. We can't control how much of the asset the user eventually sends to the issuer in a deposit or a withdrawal, so having the `/deposit` endpoint reject a request with the wrong amount sounds nice, but the user can still just go ahead and send the wrong amount anyway. Better for the wallet to just read the min and max amounts, and enforce client side, I think, rather than enforce in two places (neither of which is totally effective). Closing for now, but feel free to reopen if you disagree.

PR #247, opened 2019-01-31
Finalize CAP/SEP Process

theaeolianmachine:
Added additional clarity to the CAP process, and added the new SEP process.


MonsieurNicolas:
If we're going to use a folder like "draft" to track state, we should be consistent and use folders for all other possible "states" that exist. I see that abandoned proposals go into some "archive" state for example, and we have "final" that is quite important.

That said, I think the reason we need a draft folder with the current process is because of the first PR: yet, the first draft should be a first draft of the final document (relative high bar), and as such should have a number.

In Ethereum they just assign the number before merging, by using for example the pull request ID as ID generator, so the burden on editors is low.

README.md
```diff
@@ -1,58 +1,52 @@
...
-
-### SEP proposals
-
-1. Write your SEP using `sep-template.md` (`DRAFT`).
-2. Place it in the /drafts directory.
-2. Create a PR in this repository.
-3. SEP number assigned (`ACCEPTED`) or SEP rejected (`REJECTED`).
-4. Discussion and changes.
-5. SEP merged (`FINAL`) or SEP rejected (`REJECTED`).
+See [CONTRIBUTING](CONTRIBUTING.md) for more information.
```
tomquisel: How about: `See [CONTRIBUTING](CONTRIBUTING.md) to learn how to contribute.` just to be more clear.

core/readme.md
```diff
@@ -1,86 +1,186 @@
...
+* **Pending** — A CAP that is ready for final deliberation by the CAP Core Team. After a maximum of
+  three meetings, a vote will take place that will set the CAP's intended FCP disposition
+  (**FCP: Acceptance/Rejection**) or go back into a **Draft** state.
+* **FCP: [Acceptance/Rejection]** — A CAP that has entered a Final Comment Period (FCP) with an
+  intended disposition. After one week has passed, during which any new concerns should be
+  addressed, the CAP will head towards its intended disposition [**Acceptance/Rejection**] or go
+  back into a Draft state.
+* **Accepted** — A CAP that has been formally accepted and is ready for implementation. It is
+  expected to be included in a future version of the protocol.
+* **Finalized** — A CAP that has been implemented in Stellar Core in the version specified.
```
tomquisel: What happens to rejected CAPs? `Archived`?

core/readme.md
```diff
@@ -1,86 +1,186 @@
...
+  * When possible, solutions should be considered outside of core protocol changes such as via
+    [SEPs (Stellar Ecosystem Proposals)](../ecosystem/readme.md) to minimize complexity in the
+    Stellar protocol.
+  * When possible, proposals should minimize the impact of changes to the smallest surface area and
+    shallowest depth (i.e. sticking to the higher levels of the software) of the protocol
+    architecture possible to make changes predictable and easier to test and reason about. Changes
+    should be surgical, and minimal invasive. As a result, changes that affect lower levels of the
+    implementation have a higher bar for acceptance.
+  * In order from the highest level to the lowest level systems, the systems are:
+    * Historical / Ledger XDR
```
tomquisel: This list looks reversed?

core/readme.md
```diff
@@ -1,86 +1,186 @@
...
+* Consider contacting experts in a particular area for feedback while you're hashing out the
+  details.
+
+### Creating a CAP Draft
+Draft a formal proposal using the [CAP Template](../cap-template.md), and submit a PR to this
+repository. You should make sure to adhere to the following:
+
+* Make sure to place the draft in the `core/drafts` folder.
+* Use the following format for the filename of your draft
+  `cap_{email_username}_{hyphen-separated-cap-title}.md`.
```
tomquisel: Maybe github username instead of email? Email is a little longer / more awkward and may feel exposing to submitters.

ecosystem/readme.md
```diff
@@ -1,14 +1,118 @@
...
+  ecosystem, but has not been formally standardized by SDF, and is not endorsed by SDF for
+  adoption. Typically a SEP can start as `Informational` to gain traction within the ecosystem
+  before moving to the `Standards` track.
+* **Standard** — A SEP on the `Standards` track is one that aims for formal standardization and
+  endorsement by SDF for adoption. Typically a Standard has a higher bar towards acceptance, and it
+  requires approval by 2 SDF members of the SEP Team.
+
+## SEP Status Terms
+* **Archived** - A SEP that did not head towards a disposition due to a lack of consensus _and_
+  support. Generally open to revival with additional edits.
```
tomquisel: Is this also where rejected SEPs go?

ecosystem/readme.md
```diff
@@ -1,14 +1,118 @@
...
+  formal draft!
+- Consider contacting experts in a particular area for feedback while you're hashing out the
+  details.
+
+### Creating a SEP Draft
+Draft a formal proposal using the [SEP Template](../sep-template.md), and submit a PR to this
+repository. You should make sure to adhere to the following:
+
+* Use the following format for the filename of your draft
+  `cap_{email_username}_{hyphen-separated-cap-title}.md`.
```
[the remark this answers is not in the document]
tomquisel: I think it makes sense, it's only in the `ecosystem/draft` folder. It's good to be explicit about what the file is.

sep-template.md
```diff
@@ -1,35 +1,42 @@
...
 ## Abstract
 A short (~200 word) description of the technical issue being addressed.
 
-## Motivation
-Should clearly explain why the existing protocol specification is inadequate to address the problem that the SEP solves. SEP submissions without sufficient motivation may be rejected outright.
-
 ## Specification
 The technical specification should describe the syntax and semantics of any new feature.
 
 ## Rationale
```
tomquisel: Having both Motivation and Rational is confusing for me. Maybe make it more clear how they're different?

ecosystem/readme.md
```diff
@@ -1,14 +1,118 @@
...
+  before moving to the `Standards` track.
+* **Standard** — A SEP on the `Standards` track is one that aims for formal standardization and
+  endorsement by SDF for adoption. Typically a Standard has a higher bar towards acceptance, and it
+  requires approval by 2 SDF members of the SEP Team.
+
+## SEP Status Terms
+* **Archived** - A SEP that did not head towards a disposition due to a lack of consensus _and_
+  support. Generally open to revival with additional edits.
+* **Draft** - A SEP that is currently open to discussion and feedback.
+* **FCP** - A SEP that has entered the Final Comment Period (FCP), and has one week remaining
```
vcarl: My experiences implementing SEP-6 says that there needs to be more feedback after a final draft has been reached. We made several significant changes to the spec (data shape, /fee endpoint) due to issues discovered during implementation.

It might be worth gating "ready for wide adoption" behind a limited number of prototype implementations to try to discover needs that weren't anticipated. A hands-on implementation with a key partner and members of the SEP team seems like it would give a lot of opportunity to optimize and clarify the spec text.

ecosystem/readme.md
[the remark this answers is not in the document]
theaeolianmachine: I went ahead and changed this to be `cap_githubusername_sha256ofproposal` for when someone sends the PR. The person who's merging it in will actually rename the file to the next number, and update the number in the proposal.

core/readme.md
theaeolianmachine: Yeah I actually just changed it to `Rejected` or Perma-`Draft`. Basically if something doesn't ever get a decision or loses momentum will just stay as a draft.

core/readme.md
theaeolianmachine: Fixed.

core/readme.md
theaeolianmachine: Yup, makes sense — see update for using the sha256 of the file to make it reasonably unique when folks are submitting a PR.

ecosystem/readme.md
theaeolianmachine: Yeah decided to just make it all one flat folder where the metadata on the file says the final word. That way we can use say a static site or the like to show off what's in what stage, such as http://eips.ethereum.org/all

sep-template.md
theaeolianmachine: I changed Rational to `Design Rationale` as Motivation is supposed to be "What's the problem that needs to be addressed" where as Design Rationale is supposed to be "Why this design over others to solve the problem". Let me know if you still think it's confusing, open to suggestions.

ecosystem/readme.md
theaeolianmachine: @tomquisel as well.

One thing I did was add an "Active" category where a SEP can continue to be updated, but is generally worth putting out into the world. The TOML file for example would always be "Active". Things should naturally graduate towards "Final" when they can, and I think honestly we can beta test Draft proposals out? Happy for additional feedback though, I do like this idea.

theaeolianmachine:
@MonsieurNicolas @tomquisel,

I went ahead and decided to flatten out the structure a la the EIP directory, and use the metadata at the top of the proposal to determine it's fate. It's less management on the maintainers and submitters, and while it'll be a little clunky for now, we only have a limited number of both. Ideally as we get more drafts we'll have some form of static site or HTML generator that uses that metadata.

I'll update all of the existing metadata to a format that Gatsby can use sometime in the future once this merges.

Issue #221, opened 2018-12-02
Deprecate Data Entries


JeremyRubin:
See https://github.com/stellar/stellar-protocol/pull/199 for some discussion.

The data entries API doesn't really make sense for most use cases -- in most cases; the person is better served doing the thing off-chain. 

The only documented use case I could find with a cursory search is https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/stellar-dev/mzIP9OuegyI/bnPP0LVcAAAJ. 

Pinning Stellar.toml is a nice idea, but there are so many practical issues with the approach, something much simpler makes more sense -- issuers provide time-windowed certificates which guarantee a period where the toml is up to date and if you issue a contract with a mintime/maxtime matching it you should expect (but not be guaranteed) for it to work.

# Compatible Removal
I propose that we soft-deprecate the data entries by removing data entries from the public facing APIs while keeping them in consensus should someone be relying on the functionality currently. We should also strongly communicate the deprecation, and collect community feedback for a period of one year.

After the one year feedback period, a policy layer (read -- the software that nominators use to select transactions, but is not enforced by validators otherwise) should block transactions with data entry modifications.

Then, after an additional period, we can explore significantly increasing the reserve for Data Entries -- this enables contracts relying on Data Entries to still be processed if the person sends additional fund to the account to cover the unplanned excess.

I'm not sure I'd ever advocate full removal of data entries, given that a user may already have a compatibility issue, but perhaps with a long enough horizon for community input we can fully remove the feature.

# Repairable Removal
Alternatively, I propose that we, in the next update, fully remove data entries (transactions with that operation are invalid). If there is strong motivation for the feature in the future, I propose that we could then add back an identical feature such that those transactions would again become valid, or a new type of data entry could be added to better match the needs of users. This is semantically somewhat equivalent to dramatically increasing the minfee for a data entry operation (from the perspective of an engineer implementing a system to handle data entries).

MisterTicot:
I have a lot of data about data entries and their usefulness; I'll try to go to
the point.


**Why would you store data on a decentralized Ledger?**

The same reason that you're posting any operation: you want authenticated
information to be publicly available in a resilient manner.

Decentralized database are actually going to solve a lot of security issues that
were a pain to deal with so far.


**What are the possible use-cases?**

Simple ones:

* Storing your public key
* Storing your wallets address
* Storing dApps configuration
* Storing hash of software/files/source-code

Advanced one (dApp based on a side-chain):

[...]

PR #257, opened 2019-02-17
SEP-0008 update/replace: Generalize "Regulated Asset" to "Automated Co-Signing"

sui77:
@tomerweller 

I've just implemented a sample SEP-0008 wallet + approval server ( https://github.com/sui77/stellar-sep8-example ) and came to the conclusion that in one point SEP-0008 is conceptually flawed. It should be updated/replaced to reflect a more generic "Automated Co-Signing" instead of "Regulated Asset" procedure.

## What's the issue?

A stellar multisig account that has one anchor as co-signer requires approvals for any transaction (other anchors asset payments, XLM native payments, non-payment operations) but the required metadata is located in the [[CURRENCIES]] section of the stellar.toml at the home domain of one trustline. Any non primary anchor asset related transaction would require to look up every trustlines stellar.toml and walk through all found approval services in a trial and error fashion. 

One could argue, that in most cases this specific setup will have a single anchor cosigner only, accounts will be managed by that anchor, the first found approval_server will be the right one almost every time and the anchor is not interested in XML transactions. Nevertheless it feels wrong and bugs me.

If you think about it, co-signing isn't even directly related to an asset but obviously related to the signers. And it's not specifically about asset regulation by an anchor but more generally about account management by a 3rd party. I could think of these other use cases:

- I could give my kids a co-signed account and my approval server would restrict their daily pocket money budgets.
- Web wallets could manage accounts without being able to access user funds but being able to help securing non tech savvy users with additional 2fa or transaction limits. Even on standardized 3rd party wallets.
- Multisig services could gather all signatures of a m-of-n account in a standardized way (ok, that might need additional protocol definition, but on a good base)

## What to change? Options to be discussed.

The one thing to change is the location of metadata.

### Option 1:
- Set up user account.home_domain
- Look up users account.home_domain/stellar.toml instead of issuers account.
- Locate approval_server in a new generic [OUR_SIGNERS] section instead of the [[CURRENCIES]] section.

Simple and my favorite one. I've updated the document to reflect this option.

### Option 2:
- Set up every signers account.home_domain
- Look up every signers account.home_domain => stellar.toml 
- Locate approval_server in a new generic [OUR_SIGNERS] section instead of the [[CURRENCIES]] section.

Would be more flexible, but requires every signer to have an account, more lookups and adds too much complexity for my taste.

### Option 3:
- Use data entries to refer to each signers home_domain

Found this idea when I stumbled over the [Multisig Coordination proposal](https://github.com/stellar/stellar-protocol/blob/4da1e71cda4fb1a287f2f49263ac7a340e04c6df/drafts/draft-0005.md) ). 
Thought it's cool at first sight but I don't like the idea of refering to a stellar.toml different than the accounts home_domain. (Btw: Still I don't support the request for [deprecating data entries](https://github.com/stellar/stellar-protocol/issues/221) )

## Backwards Compatibility
This would be a major change that would not be compatible with existing implementations. But since SEP-0008 is pretty new I think nobody has implemented it yet anyways (other than maybe using it as a guideline for specific closed solutions). I've crawled all accessable stellar.tomls and could only find two approval_server entries on testnet (one is me and the other links to a disfunctional endpoint) and none on mainnet.

### DIFF
--- a/ecosystem/sep-0008.md
+++ b/ecosystem/sep-0008.md
@@ -41,15 +41,15 @@ Implementing automated co-signing consists of these parts:
     5. *Rejected?* Wallet should display given error message to the user.
 
 ## Stellar.toml
-Co-signers will advertise the location of their Approval Service through their stellar.toml file. This is done in the [OUR_SIGNERS] section.
-[OUR_SIGNERS] is an array of strings that contain the signers public key and the corresponding Approval Service URL separated with a space.
+Co-signers will advertise the location of their Approval Service through their stellar.toml file. This is done in the` OUR_SIGNERS` array of strings that contain the signers public key and the corresponding Approval Service URL separated with a space.
 The keyword ALL instead of a public key may be used to refer to an Approval Service for any signer if not specified otherwise.
 
 ### Example
 ```toml
-[OUR_SIGNERS]
-    "ALL https://goat.io/tx_approve"
+OUR_SIGNERS = [
+    "ALL https://goat.io/tx_approve",
     "GB75YSNJQ6NTS2OIMWEXPRAW2EEKMCBO6WS5EPDCM2TM3CIDVFTY6TBE https://sep8.sui.li/tx_approve"
+]
 ````
 
 ## Approval Server
--- a/ecosystem/sep-0008.md
+++ b/ecosystem/sep-0008.md
@@ -2,7 +2,7 @@
 
 ```
 SEP: 0008
-Title: Regulated Assets
+Title: Automated Co-Signing
 Author: Interstellar.com
 Status: Accepted
 Created: 2018-08-22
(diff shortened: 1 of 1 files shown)


tomerweller:
Thanks @sui77! Sorry for the late reply, things have been busy. I will definitely dive deeper into your implementation as soon as I can. Is it open source? 

I agree with most of your statements. However, it seems that what you're outlining is an approach for general multi-sig orchestration and I would consider discussing it as a potential replacement  or enhancement on the [multisig sep](https://github.com/stellar/stellar-protocol/blob/4da1e71cda4fb1a287f2f49263ac7a340e04c6df/drafts/draft-0005.md) you linked. 

The current implementation of regulated assets as an "issuer as signer on account" is less than ideal and is temporary. The protocol should have an idiomatic way of giving an issuer alleviated signing permissions *without* being a signer on every holder's account. [CAP16](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0016.md), or likelier [CAP18](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0018.md) should provide that protocol level support and SEP8 in it's current form should be able to support these unmodified.

sui77:
Just pushed it into a repo https://github.com/sui77/stellar-sep8-example 

Sure, it is a general multisig thing but it comes with just a relatively tiny change to SEP-8 and IMHO has a  much better flow than the multisig SEP which just defines submitting tx and polling status.

Thanks for refering to the CAPs, going to observe them but I guess they will be rolled out too late for me so I have to start with something SEP-8 like and migrate later.

sui77:
After actually reading CAP18 I think I got it now. I allways thought about setting up a user account with the anchor as co-signer but it also works by just toggeling the trustlines without affecting other assets on the same account at all, right? And it should already work fine this way without CAP18, which just adds flexibility for offers (I guess unauthorized trustlines can already just hold an asset... nevermind will try and figure that out)? The only downside would be adding 2 or 4 extra trustline toggeling operations for a "simple" payment. 

Okay, from that point of view it makes perfectly sense the way it is and we can close this PR.

tomerweller:
Thanks @sui77

PR #259, opened 2019-02-22
Update sep-0006.md

istrau2:
Standardizes passing JWT through to the interactive customer info required url.

### DIFF
--- a/ecosystem/sep-0006.md
+++ b/ecosystem/sep-0006.md
@@ -233,7 +233,7 @@ The response body should be a JSON object with the following fields:
 Name | Type | Description
 -----|------|------------
 `type` | string | Always set to `interactive_customer_info_needed`
-`url` | string | URL hosted by the anchor. The wallet should show this URL to the user either as a popup or an iframe.
+`url` | string | URL hosted by the anchor. The wallet should show this URL to the user either as a popup or an iframe. IMPORTANT: If the client application obtained a JWT via [SEP0010](sep-0010.md), the client app should add it as an `Authorization: Bearer <jwt>` header when requesting from this `url`.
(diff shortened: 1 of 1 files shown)


istrau2:
@tomquisel

Ryanmtate:
Thanks @istrau2;

Yes-- having the JWT standardized in the url would be great.

This may be able to be pre-populated by the anchor server via the `/deposit` response, as long as the request was authenticated with the token... probably a good idea for the client to parse the query string to ensure that no fields are missing when opening the new window.

```json
{
  "type": "interactive_customer_info_needed",
  "url" : "https://api.example.com/kycflow?account=GACW7NONV43MZIFHCOKCQJAKSJSISSICFVUJ2C6EZIW5773OU3HD64VI&token=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJHQTZVSVhYUEVXWUZJTE5VSVdBQzM3WTRRUEVaTVFWREpIREtWV0ZaSjJLQ1dVQklVNUlYWk5EQSIsImp0aSI6IjE0NGQzNjdiY2IwZTcyY2FiZmRiZGU2MGVhZTBhZDczM2NjNjVkMmE2NTg3MDgzZGFiM2Q2MTZmODg1MTkwMjQiLCJpc3MiOiJodHRwczovL2ZsYXBweS1iaXJkLWRhcHAuZmlyZWJhc2VhcHAuY29tLyIsImlhdCI6MTUzNDI1Nzk5NCwiZXhwIjoxNTM0MzQ0Mzk0fQ.8nbB83Z6vGBgC1X9r3N6oQCFTBzDiITAfCJasRft0z0",
  "interactive_deposit": true
}
```

istrau2:
@Ryanmtate Oh, you are right about that. Well, in that case, I suppose there may not be a need to standardize passing the JWT. Since the anchor is generating the url and has already has access to the JWT in the `/deposit` request, the anchor can just add it to the url in the response if it would like.

PR #262, opened 2019-03-12
Add refunded transaction status to SEP-0006

istrau2:
Adds a refunded transaction status.


istrau2:
@tomquisel

tomquisel:
I like the idea of this field, although it could lead to confusion. A deposit / withdrawal could have hit one of the other issues (like `no_market`), and then been refunded. We'd want to communicate to the user both the underlying reason for the refund `no_market` and the fact that it has been refunded. Maybe `refunded` should be a separate field?

accordeiro:
@istrau2 any thoughts on @tomquisel's suggestion?

istrau2:
@tomquisel @accordeiro Sounds good to me. I am ok with either approach really.

theaeolianmachine:
@istrau2 can you make the change so we can merge this in?

istrau2:
@theaeolianmachine sorry, done.

theaeolianmachine:
Thank you!

PR #269, opened 2019-03-15
CAP proposal: INITENTRY and METAENTRY.

graydon:
This is an updated and expanded version of the INITENTRY (and METAENTRY) CAP resulting from the past month of prototyping work over on https://github.com/stellar/stellar-core/pull/1950

I believe it's addressed all the feedback I've got so far from discussion in the community and with other core devs, so I'm proposing this draft for consideration.

### DIFF
--- a/core/cap_graydon_3b96ae30.md
+++ b/core/cap_graydon_3b96ae30.md
@@ -1,0 +1,223 @@
+## Preamble
+
+```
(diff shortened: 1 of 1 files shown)


theaeolianmachine:
Changed this to CAP-0020 in #275.

PR #304, opened 2019-05-08
toml content-type recommendation

tomerweller:


ecosystem/sep-0001.md
```diff
@@ -30,6 +30,12 @@ You must enable CORS on the `stellar.toml` so people can access this file from o
 Access-Control-Allow-Origin: *
 ```
 
+It is also recommedned to set a `text/plain` content type so that browsers render the contents, rather than prompting for a download.
```
ire-and-curses: typo: recommedned -> recommended

PR #306, opened 2019-05-15
Mark `dest` as optional in SEP-0006 withdraw

jimdanz:
If an anchor is requiring interactive customer information, it may be possible to omit `dest` in the initial withdraw API call.  The reason being that the anchor can collect the dest information interactively while they collect the customer information.

I would propose that anchors that want to declare that dest is required should do so via the Info endpoint.

### DIFF
--- a/ecosystem/sep-0006.md
+++ b/ecosystem/sep-0006.md
@@ -6,8 +6,8 @@ Title: Anchor/Client interoperability
 Author: SDF
 Status: Active
 Created: 2017-10-30
-Updated: 2019-01-08
-Version 2.4.0
+Updated: 2019-06-21
+Version 2.5.0
 ```
 
 ## Simple Summary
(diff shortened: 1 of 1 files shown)


accordeiro:
Hi @jimdanz – your reasoning makes sense, as long as that information is gathered interactively. 

Could you also update the [Info Example](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0006.md#response-2) to illustrate how an anchor server could inform that `dest` is optional or required?

theaeolianmachine:
Hey @jimdanz, did you see @accordeiro's comment? Anything you need from us for you to add the example? Would love to get this merged in.

accordeiro:
@tomquisel & @theaeolianmachine I've just pushed an example for the `/info` endpoint. I think it makes sense to add the `optional` flag here because on upstream/master this asset's withdrawal is `authentication_required`, as in:

https://github.com/stellar/stellar-protocol/blob/edcbbd825188c83c47c3c7272f54fe7b3d1eaf00/ecosystem/sep-0006.md#L406-L409

Let me know your thoughts.

theaeolianmachine:
Merging this in given @tomquisel's approval

Issue #307, opened 2019-05-20
Proposal: remove [QUORUM_SET] from stellar.toml


sui77:
I'd like to propose to remove [QUORUM_SET] from the stellar.toml specification in [SEP-1](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0001.md).

Reasons:
- It bloats the file without serving a useful purpose.
- Even if, it can not reflect multiple validator nodes with different qset configs under the same domain.
- Lot of people seem to just copy&paste the `"$self", "$lab1", "$nelisky1","$jianing","$eno","$donovan"` example.
- It's prone to be outdated when validator qsets are changed but stellar.toml is not updated (as currently seen on https://www.stellar.org/.well-known/stellar.toml ).
- All information that it should serve is already reliably accessable via SCP (or stellarbeat) and can be simply related to SEP-20 self-verified validators

tomerweller:
Agreed. @rice2000 is working on a SEP1 cleanup PR. This will definitely be part of it.

theaeolianmachine:
Marking #308 to close this out once merged, feel free to take a look with any feedback @sui77.

sui77:
Thx, nothing to do here any more :)

PR #308, opened 2019-05-22
Change how sep1 presents validator information

rice2000:
This proposal streamlines the stellar.toml specification in the wake of SEP20, essentially implementing a suggestion @MonsieurNicolas made in the discussion of that SEP.

Rather than expressing attributes of validators independently in the Account Information global fields, organizations would create a list of validators (really an array of tables) similar to the Currency Documentation section.

Each validator an organization runs would get its own entry, and relevant info would be presented per validator.  This change will make it easier to add new properties to validators (if we want to), and to write tools like automated configuration generation.

This proposal also gets rid of fields that no longer match stellar core configurations (such as DESIRED_BASE_FEE and DESIRED_MAX_TX_PER_LEDGER), as well as quorum information that is more reliably discovered via SCP or via a quorum explorer like stellarbeat.  There’s a tendency for that info to be out of date since organizations often update their configuration without amending their stellar.toml.

Since this is a breaking change, this proposal also adds version information as a global field to help crawlers know what fields to expect in a stellar.toml.

If adopted, this proposal would also require amending SEP20, which currently assumes nodes are declared in the Account Information global fields.

Specifically, we’d remove the following:
NODE_NAMES
OUR_VALIDATORS
ASSET_VALIDATOR
DESIRED_BASE_FEE
DESIRED_MAX_TX_PER_LEDGER
KNOWN_PEERS
HISTORY
[QUORUM_SET]

And add:
VERSION
[[VALIDATORS]]
NODE_PUBLIC_KEY
NODE_NAME
HISTORY

In addition this PR fixes https://github.com/stellar/stellar-protocol/issues/307.

### DIFF
--- a/ecosystem/sep-0001.md
+++ b/ecosystem/sep-0001.md
@@ -150,7 +150,7 @@ These fields go in the `stellar.toml` `[[VALIDATORS]]` list, one set of fields f
 
 Field | Requirements | Description
 ------|--------------|------------
-ALIAS | string | A name for display in stellar-core configs that conforms to `^[a-z0-9-]+$`
+ALIAS | string | A name for display in stellar-core configs that conforms to `^[a-z0-9-]{2,16}$`
 DISPLAY_NAME | string | A human-readable name for display in quorum explorers and other interfaces
 PUBLIC_KEY | `G...` string | The Stellar account associated with the node
 HOST | string | The IP:port or domain:port peers can use to connect to the node
--- a/ecosystem/sep-0001.md
+++ b/ecosystem/sep-0001.md
@@ -150,7 +150,7 @@ These fields go in the `stellar.toml` `[[VALIDATORS]]` list, one set of fields f
 
 Field | Requirements | Description
 ------|--------------|------------
-ALIAS | string | A name for the node (containing no spaces) for display in stellar-core configs
+ALIAS | string | A name for display in stellar-core configs that conforms to `^[a-z0-9-]+$`
 DISPLAY_NAME | string | A human-readable name for display in quorum explorers and other interfaces
 PUBLIC_KEY | `G...` string | The Stellar account associated with the node
 HOST | string | The IP:port or domain:port peers can use to connect to the node
@@ -247,23 +247,23 @@ image="https://pbs.twimg.com/profile_images/666921221410439168/iriHah4f.jpg"
 fixed_number=10000
 
 [[VALIDATORS]]
-ID="domain.au"
+ALIAS="domain-au"
 DISPLAY_NAME="Domain Australia"
-ADDRESS="core-au.domain.com:11625"
+HOST="core-au.domain.com:11625"
 PUBLIC_KEY="GD5DJQDDBKGAYNEAXU562HYGOOSYAEOO6AS53PZXBOZGCP5M2OPGMZV3"
 HISTORY="http://history.domain.com/prd/core-live/core_live_001/"
 
 [[VALIDATORS]]
-ID="domain.sg"
+ALIAS="domain-sg"
 DISPLAY_NAME="Domain Singapore"
-ADDRESS="core-sg.domain.com:11625"
+HOST="core-sg.domain.com:11625"
 PUBLIC_KEY="GAENZLGHJGJRCMX5VCHOLHQXU3EMCU5XWDNU4BGGJFNLI2EL354IVBK7"
 HISTORY="http://history.domain.com/prd/core-live/core_live_002/"
 
 [[VALIDATORS]]
-ID="domain.us"
+ALIAS="domain-us"
 DISPLAY_NAME="Domain United States"
-ADDRESS="core-us.domain.com:11625"
+HOST="core-us.domain.com:11625"
 PUBLIC_KEY="GAOO3LWBC4XF6VWRP5ESJ6IBHAISVJMSBTALHOQM2EZG7Q477UWA6L7U"
 HISTORY="http://history.domain.com/prd/core-live/core_live_003/"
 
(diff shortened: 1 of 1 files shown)


MonsieurNicolas:
For consistency, can we drop the `NODE_` prefixes for the fields under `VALIDATORS`:
```toml
[[VALIDATORS]]
PUBLIC_KEY="(G..."
NAME="string"
HISTORY="(optional)string"
```

Also, I think we should make `NAME` a friendly name (alphanum, no spaces allowed) and maybe complement it with an optional `DESCRIPTION` (if we want something else more free form)?
I think we need also `PEER` (to allow people to easily discover those validators).

tomquisel:
This is a great change!

I think @stanford-scs felt `ASSET_VALIDATOR` was important?

tomerweller:
1. This is really good. 
2. Following up on @MonsieurNicolas's proposal. I think it might be good to have both a display name and an identifier, but I wouldn't go as far as a description (we probably want to keep that in the entity level, not specific validators.

So something like:
```toml
[[VALIDATORS]]
ID="sdf.au"
DISPLAY_NAME="SDF Australia" 
PUBLIC_KEY="GCGB2S2KGYARPVIA37HYZXVRM2YZUEXA6S33ZU5BUDC6THSB62LZSTYH"
HISTORY="http://history.stellar.org/prd/core-live/core_live_001/"
```
3. I'd define history as "URI" and not a "URL" (which has a web context), to accommodate things like ipfs.

MonsieurNicolas:
We also need to add `ADDRESS` to the mix and deprecate `KNOWN_PEERS` (unless we want to allow people to advertise other peers that are not validators, but I don't think there is a use case for this)

tomerweller:
I think it's valuable to add a public horizon url, if there is one. But not sure how to structure this. Some options: 
1. One top level `HORIZON_URL` for the entity. 
2. A top level `HORIZON_URLS` arrays.
3. `HORIZON_URL` per horizon. 

Leaning towards 1, as the best scenario is that horizon has a load balancer that will direct you to a regional horizon/core instance. 

@MonsieurNicolas what do you think?

tomerweller:
@rice2000, I think we need to make it clear that the TOML file describes an entity, more than it describes a domain. (we don't want entities having multiple TOML files if, for example, their validators run on multiple domains)

rice2000:
@tomerweller That makes sense to me re: clarifying entity, not domain.
@MonsieurNicolas question about `PEER`: by that, I think you mean the port for inbound connection.  Is that correct?  Is that usually called a peer in this context?  Or should the field be `PEER_PORT`?

Tons of great suggestions here.

MonsieurNicolas:
@rice2000 yeah I mean replace `PEER` (in my first response) by `ADDRESS` (which deprecates the `KNOWN_PEERS` list)

`HORIZON_URL` @tomerweller yeah I don't think we should map multiple horizon clusters to stellar.toml : it also doesn't need to be an array, people that have global deployments use GSLB (so use a global address like "horizon.stellar.org" to really resolve to things like "horizon.us.stellar.org") so your option 1 seems right.

entities vs domains: I am not sure I understand the statement "validators run on different domains", in other SEPs, there is an assumption (that seems fine to me) that `HOME_DOMAIN` equates to the entity.
The fact that a validator may use a different domain (its `ADDRESS`) is irrelevant: people can use IP addresses in there if they want (thanks to SEP20, you can always reverse lookup the home domain from a validator's public key).
I think it's probably the same with the `HORIZON_URL` (but for this to work, we need to be able to ask Horizon for its home domain and it doesn't seem to be possible right now the '/' endpoint doesn't tell me anything about that).

rice2000:
Okay: I updated to include the suggestions above.

Specifically:
- Added `HORIZON_URL` to the Account Information global fields
- Amended validator fields.  Now they are ID, DISPLAY_NAME, PUBLIC_KEY, ADDRESS, and HISTORY
- Updated the example stellar.toml accordingly.

I think that's everything, but @theaeolianmachine @MonsieurNicolas @tomerweller: please take a look when you get a chance.

theaeolianmachine:
Looks great to me — will wait up to 2 days (June 5th) to hear back from the other reviewers otherwise will merge this in.

theaeolianmachine:
This PR fixes https://github.com/stellar/stellar-protocol/pull/308.

sui77:
> ID, DISPLAY_NAME, PUBLIC_KEY, ADDRESS, and HISTORY

- What's the difference between ID and PUBLIC_KEY?
- Naming: I'd prefer `HOST` instead of `ADDRESS` which might be confused with "Stellar Address"

marcinx:
+1 on Suat's ADDRESS -> HOST suggestion. 

What's the purpose of the ID field? If it's a naming suggestion for TOML aliases would it be more straightforward to call it ALIAS?

sui77:
Agree, ALIAS sounds more meaningful.

Keep in mind [SEP-20 Validator self verification](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0020.md) also needs an update as soon as this one is finalized.

rice2000:
@sui77 @marcinx I agree with changing ADDRESS to HOST
re: ID: @tomerweller I think adding that field was your suggestion. Can you clarify the purpose for me?  I think calling it ALIAS makes sense, but would love your input before making that change.

tomerweller:
@rice2000 `ALIAS` sounds good

rice2000:
Okay: I made the suggested changes.

tomerweller:
Regarding ALIAS, I suggest we use an explicit regex for things to look uniform. 
How about `^[a-z0-9-]$`? (I think that means lower case letters + digits + the hypen sign, but someone should double check)

rice2000:
@tomerweller I used the amended regex for `ALIAS`: `^[a-z0-9-]+$`

marcinx:
I think the "updated date" in the header needs to be bumped up to reflect the latest changes. It still shows: `Updated: 2018-09-11`

PR #313, opened 2019-05-27
update SEP-0011 to reflect new canonical asset format.

stanford-scs:


bartekn:
The part of the doc changed here was referenced in https://github.com/stellar/go/pull/1754 because it's using asset format.

I don't think below is true:
> Codes in `Asset` and `AllowTrustOp` should escape `\`, `:` and any byte outside the range 0x21-0x7e, but no other bytes.  Trailing `\x00` should not be shown except as needed to show an asset of type `ASSET_TYPE_CREDIT_ALPHANUM12` fewer than 5 non-zero bytes.


in the context of stellar-core code. The relevant code is [here](https://github.com/stellar/stellar-core/blob/07af9db4b8f3a22f1427761c2a70dee103bfafa7/src/util/types.cpp#L97) and it looks like asset code must pass [`isalnum`](https://en.cppreference.com/w/cpp/string/byte/isalnum) function so only alphanumeric characters can form a valid asset code.

MonsieurNicolas:
@bartekn is correct: I don't know where this comes from. Both the code and [documentation](https://www.stellar.org/developers/guides/concepts/assets.html#alphanumeric-4-character-maximum) are basically describing `alphanum` (with zero bytes only allowed as padding characters at the end of the asset code)

PR #314, opened 2019-05-27
Update SEP-6 to clarify the trust line creation flow on /deposit

### LINKED ISSUE #309: SEP-6 clarify behavior when user has no trustline or asset is AUTH_REQUIRED
[This section of SEP-6](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0006.md#1-success-no-additional-information-needed) describes what to do on deposit when the user has no account, but it does not explicitly say what the wallet and issuers should do if the user has no trustline or if the asset is `AUTH_REQUIRED`.

Here's my first instinct:

* no trustline
    * issuer detects if account has enough XLM to create a trustline. If not, the issuer sends the needed amount of XLM to the account to create a trustline
    * the wallet knows that a trustline is needed since the user initiated a deposit, and it adds the trustline ASAP. Possibly it has to wait for the XLM from the issuer to arrive before adding it.
    * issuer listens for user to add trustline, and then sends the remainder of the deposit (after paying for the trustline XLM) to the account
* `AUTH_REQUIRED` asset with no trustline:
    * same as above, except the issuer also approves the newly detected trustline just before sending the asset to the user.

Also, if the account has not uploaded KYC info necessary to approve a trustline for the account, the issuer should obviously return one of the `customer_info_needed` responses to deposit requests until the KYC conditions are met.

accordeiro:
This PR fixes #309. 

The goal is to detail the steps that should be taken during the deposit flow, when the given Stellar account doesn't have a trust line for the asset that is being deposited.

### DIFF
--- a/ecosystem/sep-0006.md
+++ b/ecosystem/sep-0006.md
@@ -149,7 +149,7 @@ The deposit flow can only be fulfilled if the Stellar `Account` has established 
 2. If `Account` doesn't have enough funds, `Wallet` starts listening for transactions to the given `Account`, waiting for it to have enough funds for a trust line.
 3. When asked for a deposit, `Anchor` detects if `Account` has enough XLM to create a trust line. If it doesn't, `Anchor` sends the needed amount of XLM to the `Account` for creating a trust line. `Issuer` then starts listening for trust line creations for that `Account`.
 4. `Wallet` detects the arrival of funds in the `Account`, and establishes a trust line.
-5. `Anchor` detects the trust line creation in the `Account`. If the asset is `AUTH_REQUIRED`, `Anchor` either approves the new trust line, or returns any of the `customer_info_needed` responses until KYC conditions are met.
+5. `Anchor` detects the trust line creation in the `Account`. If the asset is `AUTH_REQUIRED`, `Anchor` approves the new trust line.
 6. `Anchor` proceeds with the deposit flow.
 
 ## Withdraw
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0006.md
```diff
@@ -129,14 +132,25 @@ Examples:
...
+###### Stellar account doesn't trust asset
+
+The deposit flow can only be fulfilled if the Stellar `Account` has established a trust line for the given asset. To ensure this is accomplished, when initiating the deposit flow, `Wallet` should check if the `Account` has a trust line for the given asset. If it doesn't:
+
+1. `Wallet` checks if `Account` has enough funds to create a trust line. If it does, skip to step `4`.
+2. If `Account` doesn't have enough funds, `Wallet` starts listening for transactions to the given `Account`, waiting for it to have enough funds for a trust line.
+3. When asked for a deposit, `Anchor` detects if `Account` has enough XLM to create a trust line. If it doesn't, `Anchor` sends the needed amount of XLM to the `Account` for creating a trust line. `Issuer` then starts listening for trust line creations for that `Account`.
+4. `Wallet` detects the arrival of funds in the `Account`, and establishes a trust line.
+5. `Anchor` detects the trust line creation in the `Account`. If the asset is `AUTH_REQUIRED`, `Anchor` either approves the new trust line, or returns any of the `customer_info_needed` responses until KYC conditions are met.
```
tomquisel: I believe both of these special cases only come up after the issuer has received the deposit. The issuer needs to have KYCed the user before giving deposit instructions, so at the stage that the issuer is handling the "stellar account doesn't trust asset case", the user will already have satisfied KYC requirements. As a result, the issuer should be able to immediately approve the new trustline and does not need to give any further `customer_info_needed` responses.

accordeiro: Yes, that makes sense – I've pushed some changes to reflect this behavior.

PR #322, opened 2019-06-04
SEP-6: allow anchors to indicate auth required

tomquisel:
Provide a mechanism in  SEP-6 that anchors can use to indicate that their deposit, withdrawal, and transaction endpoints require [SEP-10](sep-0010.md) authentication.

### DIFF
--- a/ecosystem/sep-0006.md
+++ b/ecosystem/sep-0006.md
@@ -398,7 +398,7 @@ The response should be a JSON object like:
     },
     "ETH": {
       "enabled": true,
-      "authorization_required": false,
+      "authentication_required": false,
       "fee_fixed": 0.002,
       "fee_percent": 0
     }
@@ -439,7 +439,8 @@ The response should be a JSON object like:
     "authentication_required": true
   },
   "transaction": {
-    "enabled": false
+    "enabled": false,
+    "authentication_required": true
   }
 }
 ```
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0006.md
```diff
@@ -380,13 +398,15 @@ The response should be a JSON object like:
     },
     "ETH": {
       "enabled": true,
+      "authorization_required": false,
```
accordeiro: I think you meant `authentication_required`, right?

ecosystem/sep-0006.md
```diff
@@ -415,7 +435,8 @@ The response should be a JSON object like:
     "enabled": false
   },
   "transactions": {
-    "enabled": true
+    "enabled": true,
```
accordeiro: Should we add `"enabled": true` to the `"transaction"` info as well? Is there any case where `/transactions` requires authentication, but `/transaction` doesn't?

ecosystem/sep-0006.md
```diff
@@ -455,7 +476,7 @@ The `fields` object contains a key for each field name and an object with the fo
 
 The wallet should display a form to the user to fill out any fields with unknown values as part of the deposit/withdrawal flow. Each field should be an text `input`, unless `choices` is specified, in which case a dropdown should be used.
 
-An anchor should also indicate in the `/info` response if they support the following optional endpoints:
+An anchor should also indicate in the `/info` response if they support the following optional endpoints and if they require authentication:
```
accordeiro: Now that there is more than one field that should be provided for each endpoint, should we list them here to keep consistency? Something in the lines of:

An anchor should also indicate in the `/info` response if they support the `fee`, `transactions` and `transaction` endpoints, by providing the following fields for each:

* `enabled`: `true` if the endpoint is available.
*  `authentication_required`:`true` if client must be [authenticated](#authentication) before accessing the endpoint.

accordeiro:
I've updated this PR to address the suggestions. @tomquisel please let me know your thoughts :)

ecosystem/sep-0006.md
debnil: I don't think there's a way to populate the `/transaction` request without a prior call to `/transactions`, because the first request requires at least one of a specific transaction's three identifiers. I'm not sure what this implies for authentication - but it's extremely unlikely, absent a very large DoS, that requesting `/transaction` without the prior authenticated call to `/transactions` would succeed.

accordeiro: Yeah, I agree that it's unlikely.

In any case, I think it's still worth it to make the `/transaction` and `/transactions` calls auth criteria consistent, since if an anchor doesn't want to disclose information to non-authenticated sources about the whole transaction list, it probably wouldn't want do it for a specific transaction either (regardless of how hard it is to guess their identifier). Does that make sense?

debnil: Yup, it does make sense! Agree that it's worth adding.

PR #323, opened 2019-06-07
add some clarifications to CAP-0015 regarding handling of duplicate bumpFee transactions

MonsieurNicolas:


MonsieurNicolas:
@orbitlens @stanford-scs

stanford-scs:
> If `innerTx` was not present, its associated effective fee is 0.

I don't understand what this means.  A fee bump transaction should be invalid if `innerTx` is not present.

> (as to ensure that sequence numbers are respected).

so as to ensure...

> which will contain a `feeCharged` of  `0` if it was not already included

Again, shouldn't the FeeBump be invalid if the `innerTx` is invalid (including because `innerTx` already executed so the sequence number is stale)?

MonsieurNicolas:
Thanks @stanford-scs 

Would `If innerTx was not already present in the transaction set, its associated effective fee is 0.` be clearer?

MonsieurNicolas:
@theaeolianmachine : can be merged now

PR #335, opened 2019-06-26
Move SEP-0020 to Active.

theaeolianmachine:
SEP-0020 has been gaining traction without comment for some time, so
we're going to move this past FCP to Active.


theaeolianmachine:
Cc @johansten @rice2000, my plan is to merge this in tomorrow unless I hear anything from either of you saying otherwise.

PR #342, opened 2019-07-05
Improve & clarify interactive SEP-6 flow

tomquisel:
This is a breaking protocol change (the new version is `3.0.0`), but it's a change in parts of the protocol that very few implementors have covered so far, so I think the negative impact will be small.

The breaking changes:
* If a server uses `postMessage` to tell the wallet about status changes in an interactive flow, change the format of the message to contain a transaction object.
* Specify that wallets should poll the `/transaction` endpoint to see the status of an interactive deposit / withdrawal, instead of polling the `/deposit` or `/withdrawal` endpoints.

This change handles the case where anchors use the interactive flow to collect per-request info (like SMS verification, or which connected bank account the user wants to use for this request).

Other changes:

* `interactive_customer_info_needed` response
    * Add an `id` response field for the `interactive_customer_info_needed` response. This allows wallets to check the status of a particular ongoing interactive deposit or withdrawal request via the `/transaction` endpoint.
    * Remove unnecessary `interactive_deposit` flag.
* Clarify that the `callback` URL passed into an interactive flow popup's URL may also respond in failure cases. This allows wallets to more gracefully handle failed interactive requests.
* Add the optional `jwt` parameter to the interactive flow popup's URL. This allows for users to go through an anchor's interactive flow without having to re-authenticate.
* Add fields to the `/transaction` response that the wallet needs to handle ongoing transactions
* Add advice on how a wallet should use the responses it receives from the server
* Clarify that `customer_info_status` is used to see status of uploaded customer info (happens one time per customer generally), as in the non-interactive flow, and that the transaction object is used to see the status of a transaction, both historical and ongoing. The transaction object is used when the user must submit interactive information for every transaction.


tomerweller:
Thanks @tomquisel. It's a messy job but someone has to do it :)

I'm not a huge fan of adding the `callback` argument for the interactive flow url. Do we really need it? Sounds like an overhead both for the anchor and for the wallet to implement. While in fact just polling the `/transaction/` endpoint should suffice. 

In fact, if by convention the wallet queries that as soon as the webview closes, there shouldn't be any lag. No?

tomquisel:
@tomerweller I made several more changes, see the Additional changes section. I decided to keep the `callback` because I think it's good to allow implementors to avoid polling if they want for responsiveness and better performance. I don't think that closing the window is a good enough trigger, because the user may just leave the window open on success and background it. So you'd get some lag.

ecosystem/sep-0006.md
```diff
@@ -276,32 +276,82 @@ Example:
...
+
+The response body must be a JSON object with the following fields:
 
 Name | Type | Description
 -----|------|------------
 `type` | string | Always set to `interactive_customer_info_needed`
 `url` | string | URL hosted by the anchor. The wallet should show this URL to the user either as a popup or an iframe.
 `interactive_deposit` | boolean | (optional) Flag indicating that depositing is also handled in the anchor's interactive customer info flow. The wallet need not make additional requests to `/deposit` to complete the deposit. Defaults to `false`. Only relevant for responses to `/deposit` requests.
+`id` | string | (optional) The anchor's internal ID for this deposit / withdrawal request. Can be passed to the [`/transaction`](#single-historical-transaction) endpoint to check status of the request.
```
accordeiro: I love this `id` field – it's definitely something that will enable us to track transactions more easily once they have moved from the `/deposit` or `/withdraw` flows.

My question is: should we be more explicit about when this `id` needs to be generated by the anchor? Is it only when the interactive KYC is completed? Or should an `id` be generated every time the endpoint is hit? 

Here's a suggestion, which might make the wallet-anchor integration even simpler:

1. We generate an `id` for each `/deposit` or `/withdraw` request, regardless of it being interactive or not
2. On future calls to `/deposit` or `/withdraw`, the wallet also provides an `id` query param – this will remove the need for the Wallet to always replicate all request params (+ newly required data) until it gets a success response, since `/deposit` is not stateless anymore
3. As soon as a 200 is received, the status of that deposit or withdrawal can be checked via `/transaction` by passing along that `id`

I think my reasoning is that we're basically doing a REST API around the `/transactions` resource, but using `GET /withdraw` or `GET /deposit` instead of `POST /transactions`, so it might make sense for them to share the same id.

ecosystem/sep-0006.md
```diff
@@ -276,32 +276,82 @@ Example:
...
+
+To do this, the wallet needs to either poll the `/transaction` endpoint with the `id` provided in the `interactive_customer_info_needed` response from the anchor until the necessary information is available, or register a callback with the anchor as described above. Either way, the anchor's response will contain the transaction fields described in the [/transactions](#transaction-history) endpoint.
+
+The wallet must use the response fields in the following way to complete the withdrawal:
+
+- `status`: `incomplete` means the user is either still going through the interactive flow with the issuer, or has abandoned the transaction part way through. The wallet should simply keep polling or waiting in this case. The wallet should provide the user with an `x` or other way to abort the transaction and return to their wallet in this case.
```
accordeiro: Should we provide a way for the Wallet to inform the Anchor that the transaction has been aborted by the user? Or should it be the Anchor's responsibility to define the criteria for timing out a deposit / withdrawal?

ecosystem/sep-0006.md
```diff
@@ -276,32 +276,82 @@ Example:
 
 Response code: `403 Forbidden`
 
-An anchor that requires the user to fill out information on a webpage hosted by the anchor should use this response. A wallet that receives this response should open a popup browser window to the specified URL.
+An anchor that requires the user to fill out information on a webpage hosted by the anchor should use this response. This can happen in situations where the anchor needs KYC information about a user, or when the anchor needs the user to perform a custom step for each transaction like entering an SMS code to confirm a withdrawal or selecting a bank account. A wallet that receives this response should open a popup browser window to the specified URL. Next, the wallet must do one of two things:
```
accordeiro: Should we mention somewhere here that these popups will likely run on mobile apps, so they need to be functional on major mobile browsers?

ecosystem/sep-0006.md
tomquisel: Thanks Alex, this makes sense. It's definitely a bummer that we need to be as backwards-compatible as possible, there are some strange choices in the existing SEP.

Here's what I'm thinking we can do for your suggestion:

1. Add an optional `id` parameter to `/deposit` and `/withdraw` success responses. These are useful for calling `GET /transaction`. They don't help with repeated calls to `/deposit` or `/withdraw`, because clients won't make any more of those requests once they get a success response.
1. After receiving a `interactive_customer_info_needed` response, the `interactive_deposit: false` case is the only one where the client needs to make another request to `/deposit`. We can specify that if the server has included an `id`, then the client doesn't need to re-include any parameters on its second call to `/deposit`, it just needs to supply the `id`.
1. Add an optional `id` parameter to the  `non_interactive_customer_info_needed` response, and specify that the client doesn't need to re-include any parameters if it supplies the `id` to future requests to `/deposit` or `/withdraw`.

Backwards compatibility analysis:

- old client and new server: may not work, client will need to be updated to check for `id` if server depends on statefulness
- new client and old server: will work since new clients check for `id`

However, I'm not too worried about this, because I don't think any existing flows are using the repeated requests to `/deposit` or `/withdraw` yet.

What do you think?

ecosystem/sep-0006.md
tomquisel: Good question. If the wallet closes the anchor popup when the `x` is clicked, then there's no need. The case that concerns me is where the user clicks the `x` in the wallet and then finishes the anchor flow. They'd end up back at the wallet trying to complete the transaction, but the wallet would not have the right interface up. That can't happen with the wallet closing the popup, though. So I think that's the best approach.

ecosystem/sep-0006.md
tomquisel: Great call!

ecosystem/sep-0006.md
accordeiro: This approach makes sense to me! Thanks :)

ecosystem/sep-0006.md
accordeiro: Agreed!

PR #345, opened 2019-07-09
Clarify choice of passphrase in SEP-10

### LINKED ISSUE #339: SEP-10 should define a network passphrase used when signing challenge transaction
Transaction signatures depend on the network passphrase. Challenge endpoint (and SEP-10 itself) do not define what network should be used when signing so server and client can use different networks. SEP-10 should define which network should be used for signing (probably pubnet).

tomquisel:
* Clarify choice of passphrase for signatures in SEP-10
* Also add an optional response parameter to the challenge endpoint: `network_passphrase` to make debugging signature issues easier.
* Replace outdated reference to SEP-6 with reference to SEP-12

Addresses problem mentioned in https://github.com/stellar/go/pull/1468
Closes #339 

@nebolsin I'd love your feedback.

### DIFF
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -6,13 +6,13 @@ Title: Stellar Web Authentication
 Author: Sergey Nebolsin <sergey@mobius.network>, Tom Quisel <tom.quisel@gmail.com>
 Status: Active
 Created: 2018-07-31
-Updated: 2018-08-14
-Version 1.0.1
+Updated: 2019-07-09
+Version 1.1.0
 ```
 
 ## Simple Summary
 
-This SEP defines the standard way for clients such as wallets or exchanges to create authenticated web sessions on behalf of a user who holds a Stellar account. A wallet may want to authenticate with any web service which requires a Stellar account ownership verification, for example, to upload KYC information to an anchor in an authenticated way as described in [SEP-6](sep-0006.md).
+This SEP defines the standard way for clients such as wallets or exchanges to create authenticated web sessions on behalf of a user who holds a Stellar account. A wallet may want to authenticate with any web service which requires a Stellar account ownership verification, for example, to upload KYC information to an anchor in an authenticated way as described in [SEP-12](sep-0012.md).
 
 ## Abstract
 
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0010.md
```diff
@@ -65,20 +65,23 @@ GET https://auth.example.com/?account=GCIBUCGPOHWMMMFPFTDWBSVHQRT4DIBJ7AD6BZJYDI
...
-* signature by the web service signing account
+* `transaction`: an XDR-encoded Stellar transaction with the following:
+  * source account set to server's signing account
+  * invalid sequence number (set to 0) so the transaction cannot be run on the Stellar network
+  * time bounds: `{min: now(), max: now() + 300 }` (we recommend expiration of 5 minutes to give user time to sign transaction)
+  * operations: `manage_data(source: client_account, key: '<anchor name> auth', value: random_nonce())`
+    * The value of key is not important, but can be the name of the anchor followed by `auth`. It can be at most 64 bytes.
+    * The value must be a 64 byte long base64 encoded cryptographic-quality random string
+  * signature by the web service signing account
+* `network_passphrase`: (optional but recommended) Stellar network passphrase used by the server. This allows the client to verify that it's using the correct passphrase when signing.
```
bartekn: Looks good to me. I'm a little bit worried that `network_passphrase` is optional. It requires code that handles this situation and if developer forgets about it it may result in null pointer exceptions or using empty passphrase.

nebolsin:
I want to point out that server’s network passphrase is implicitly bound to the challenge transaction as part of server’s signature, and the client is required to “verify that transaction envelope has a correct signature by server's signing key“, which I think provides a reliable way to protect from network mismatch issues and halt the flow before it gets to client-side signing.

It might be convenient to extend the response with `network_passphrase` (or should it actually be specified in stellar.toml? looks like currently SEP-1 doesn’t define this field) to simplify debugging or improve error reporting, but I think that the server’s signature check should still act as the ultimate verification.

ecosystem/sep-0010.md
tomquisel: Agreed. I think this is a necessary evil to make the change backwards compatible.

bartekn:
> I want to point out that server’s network passphrase is implicitly bound to the challenge transaction as part of server’s signature, and the client is required to “verify that transaction envelope has a correct signature by server's signing key“, which I think provides a reliable way to protect from network mismatch issues and halt the flow before it gets to client-side signing.

I think the problem we're trying to solve is having to guess the passphrase if the first signature is incorrect. I agree that in most cases it will be pubnet or testnet passphrase but it's not defined anywhere so in reality the passphrase can be anything. Also, in https://github.com/stellar/stellar-protocol/issues/339#issuecomment-509631764 I pointed out that we had very similar problem in the past (two components were talking to each other but using different passphrases) and it was really difficult to debug.

PR #346, opened 2019-07-14
Add Documentation for signing a SEP-10 JWT

drdgvhbh:


tomquisel:
Thanks for the contribution! It's such a good point that this is unspecified right now.

Specifying `SIGNING_KEY` sounds great. Do you think we need to specify HS256? The JWT self-documents what algorithm is used, so it seems viable to leave it up to the server. @nebolsin any opinion?

drdgvhbh:
Maybe just document that you can use any algorithm instead because I didn't even know that

tomquisel:
Actually, thinking about this more, it's very important that the initial transaction that the server sends to the client is signed by the server's `SIGNING_KEY`, but I don't think we care how the JWT is signed. All that matters is that the server is able to verify its own signatures internally. I think I'm just going to close the PR. Feel free to reopen if you think I'm missing something.

PR #356, opened 2019-07-25
Describe recommended SEP-6 implementations

tomquisel:
Describe recommended wallet and anchor implementations to demystify all the possible ways of implementing SEP-6. See new [recommendations section](https://github.com/stellar/stellar-protocol/blob/tomquisel-patch-1/ecosystem/sep-0006.md#recommendations).

### DIFF
--- a/ecosystem/sep-0006.md
+++ b/ecosystem/sep-0006.md
@@ -75,7 +75,7 @@ SEP-6 lays out many options for how deposit and withdrawal can work. These are r
 
 ### Basic Wallet Implementation
 
-* Identify anchors you want to support manually in advance
+* Identify anchors you want to support manually, and test them with your wallet to be sure they work before whitelisting them. We encourage you to support as many anchors as possible.
 * For each anchor, use information from its [`stellar.toml`](sep-0001.md) file and its [`/info`](#info) endpoint to display useful information to the user about the asset they've picked.
 * Provide a UI that allows users to pick an asset, anchor, and amount to use for deposit or withdraw. The UI should display the asset's fee structure (if possible) as well as information such as the address of the anchor and description of the asset from the `stellar.toml` file.
 * **Use the `/info` endpoint**
@@ -111,7 +111,7 @@ SEP-6 lays out many options for how deposit and withdrawal can work. These are r
     * We recommend you use the [interactive flow](#3-customer-information-needed-interactive). It's flexible and simple to integrate.
     * To start an interactive flow, provide the [Customer info needed response](#3-customer-information-needed-interactive).
     * Include the `id` field in your response so the wallet can check up on the status of the transaction if it wants.
-    * We expect you'll want to use [authentication](#authentication) in the interactive flow, so be sure to accept the [`jwt` parameter](#adding-parameters-to-the-url) so the user does not have to re-login. Support the `callback` parameter as well.
+    * We recommend you use [SEP-10 authentication](#authentication) in the interactive flow and do not separately prompt for password to achieve a good user experience (although asking for MFA when confirming a transaction or requiring email confirmation is reasonable). Be sure to accept the [`jwt` parameter](#adding-parameters-to-the-url) so the user does not have to re-login. Support the `callback` parameter as well.
 * **Interactive deposit**
     * Your interactive deposit popup will do everything needed to initiate the deposit without the user needing to interact with the wallet further. It should either directly initiate the deposit, or handle displaying information (such as reference number or bank account number) that the user will need to complete their deposit.
 * **Interactive withdrawal**
--- a/ecosystem/sep-0006.md
+++ b/ecosystem/sep-0006.md
@@ -67,6 +67,50 @@ Valid CORS headers are necessary to allow web clients from other sites to use th
 Access-Control-Allow-Origin: *
 ```
 
+## Recommendations
+
+SEP-6 lays out many options for how deposit and withdrawal can work. These are recommendations for getting a wallet or anchor implementation working with minimal effort while providing a great user experience.
+
+### Basic Wallet Implementation
+
+* Identify assets you want to support manually in advance
+* For each asset, use information from its `stellar.toml` and its `/info` endpoint to display useful information to the user like transaction fee structure, types of deposits and withdrawal supported, and details about the anchor like street address.
+* Use `/info` to determine:
+    * anchor endpoints that require authentication, and perform [authentication](#authentication) via SEP-10 before hitting those endpoints
+    * Fee structure: if `fee_fixed` and `fee_percent` are provided, show this to the user early in the process so they're fully informed.
+    * If the `/fee` endpoint is enabled, use it for computing fees when you need to show them to the user.
+* While `/info` allows an anchor to communicate non-standard fields that are needed for `/deposit` or `/withdraw`, it's easier for a basic wallet implemention to hard-code extra fields that are needed on a per-anchor basis, and ensure those fields are passed in properly.
+* As a conservative measure, pass in any optional fields (including amount, which you should ask the user for) for `/deposit` or `/withdraw` that the wallet has on-hand, such as `email_address` and `account`.
+* For `/deposit`
+    * If the issuer's `/deposit` endpoint immediately returns success:
+        * display the deposit info that came back from the endpoint to the user, including fee and you're done! The user will execute the deposit exernally using the instructions if they want to.
+    * If the issuer responds with an interactive flow, handle it as [described in detail](#3-customer-information-needed-interactive). We expect this is the most common flow, and it should work smoothly.
+    * For now, do not handle the non-interactive or status responses to the `/deposit` request.
+    * Handle the [special cases](#special-cases), they're relatively common.
+* For `/withdraw`
+    * If the issuer's `/withdraw` endpoint immediately returns success:
+        * Provide an interface to the user that allows them to pick a withdrawal amount and view the total fee. The user can then confirm and your wallet will initiate the withdrawal by sending the Stellar payment to the address provided by the issuer.
+    * If the anchor responds with an interactive flow, handle it as [described in detail](#3-customer-information-needed-interactive).
+    * For now, do not handle the non-interactive or status responses to the `/withdraw` request.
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0006.md
```diff
@@ -67,6 +67,50 @@ Valid CORS headers are necessary to allow web clients from other sites to use th
...
+* For `/withdraw`
+    * If the issuer's `/withdraw` endpoint immediately returns success:
+        * Provide an interface to the user that allows them to pick a withdrawal amount and view the total fee. The user can then confirm and your wallet will initiate the withdrawal by sending the Stellar payment to the address provided by the issuer.
+    * If the anchor responds with an interactive flow, handle it as [described in detail](#3-customer-information-needed-interactive).
+    * For now, do not handle the non-interactive or status responses to the `/withdraw` request.
+* It provides a better experience for users to show deposits or withdrawal they've completed in the past via the `/transactions` endpoint, but it's not strictly necessary.
+
+### Basic Anchor Implementation
+
+* Provide a full-featured implementation of `/info`. The more info you provide there, the better a wallet can relay deposits and withdrawals from the user to you.
```
accordeiro: I think implementing the `"fields"` part of both `"deposit"` and `"withdraw"` aren't really useful if we're favoring the interactive flow.

What do you think about recommending anchors to focus on the other parts of the `/info` endpoint (enabled, fees, amounts), rather than asking them to provide as much info as possible?

ecosystem/sep-0006.md
```diff
@@ -67,6 +67,50 @@ Valid CORS headers are necessary to allow web clients from other sites to use th
...
 ```
 
+## Recommendations
+
+SEP-6 lays out many options for how deposit and withdrawal can work. These are recommendations for getting a wallet or anchor implementation working with minimal effort while providing a great user experience.
+
+### Basic Wallet Implementation
+
+* Identify assets you want to support manually in advance
+* For each asset, use information from its `stellar.toml` and its `/info` endpoint to display useful information to the user like transaction fee structure, types of deposits and withdrawal supported, and details about the anchor like street address.
```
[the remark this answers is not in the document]
rice2000: It also might prevent confusion to explicitly state that the `/info` endpoint is covered in this SEP, whereas the `stellar.toml` is covered in SEP-1.  Something like:

"For each asset, use the `/info` endpoint to discover and display useful information to the user like fee structure and types of deposits and withdrawals support, and check the issuer's `stellar.toml` (as specified in SEP-001) to discover details about the issuer such as business name and street address."

ecosystem/sep-0006.md
```diff
@@ -67,6 +67,50 @@ Valid CORS headers are necessary to allow web clients from other sites to use th
...
+* Use `/info` to determine:
+    * anchor endpoints that require authentication, and perform [authentication](#authentication) via SEP-10 before hitting those endpoints
+    * Fee structure: if `fee_fixed` and `fee_percent` are provided, show this to the user early in the process so they're fully informed.
+    * If the `/fee` endpoint is enabled, use it for computing fees when you need to show them to the user.
+* While `/info` allows an anchor to communicate non-standard fields that are needed for `/deposit` or `/withdraw`, it's easier for a basic wallet implemention to hard-code extra fields that are needed on a per-anchor basis, and ensure those fields are passed in properly.
+* As a conservative measure, pass in any optional fields (including amount, which you should ask the user for) for `/deposit` or `/withdraw` that the wallet has on-hand, such as `email_address` and `account`.
+* For `/deposit`
```
rice2000: Small visual suggestion: I think this would be easier to read and process if the three sections -- "use info to determine," "For deposit," and "For withdrawal" -- were more clearly delineated.  Maybe bold those?  Or make them some kind of header?

ecosystem/sep-0006.md
```diff
@@ -67,6 +67,50 @@ Valid CORS headers are necessary to allow web clients from other sites to use th
...
+    * If the issuer's `/withdraw` endpoint immediately returns success:
+        * Provide an interface to the user that allows them to pick a withdrawal amount and view the total fee. The user can then confirm and your wallet will initiate the withdrawal by sending the Stellar payment to the address provided by the issuer.
+    * If the anchor responds with an interactive flow, handle it as [described in detail](#3-customer-information-needed-interactive).
+    * For now, do not handle the non-interactive or status responses to the `/withdraw` request.
+* It provides a better experience for users to show deposits or withdrawal they've completed in the past via the `/transactions` endpoint, but it's not strictly necessary.
+
+### Basic Anchor Implementation
+
+* Provide a full-featured implementation of `/info`. The more info you provide there, the better a wallet can relay deposits and withdrawals from the user to you.
+* Decide which endpoints, if any, need to be authenticated, and declare that properly in the `/info` endpoint.
```
rice2000: I think it would be helpful to mention + link to SEP-10 here (basically parallel the wallet recommendation section)

ecosystem/sep-0006.md
```diff
@@ -67,6 +67,50 @@ Valid CORS headers are necessary to allow web clients from other sites to use th
...
+* Pick your approach to [fees](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0006.md#fee). We strongly recommend using `/info` to express fees if possible as it provides a better user experience (the user can see the fee structure in the wallet early in the process).
+* If you need no KYC info, explicit user signup, or other interaction from users to show deposit info or process a withdrawal, you can have your `/deposit` and `/withdraw` endpoints immediately return success with all info that the user will need to complete the deposit or withdrawal.
```
rice2000: "If you don't need..." seems a little clearer to me that "need no."

ecosystem/sep-0006.md
```diff
@@ -67,6 +67,50 @@ Valid CORS headers are necessary to allow web clients from other sites to use th
...
+* Otherwise, we recommend you provide an interactive deposit or withdrawal flow.
+* For both interactive deposit and withdrawal:
+    * Use the [Customer info needed response](#3-customer-information-needed-interactive) and include the `id` field so the wallet can check up on the status of the transaction (using the `id`) if they want.
+    * You should almost certainly use [authentication](#authentication) in the interactive flow, so be sure to accept the [`jwt` parameter](#adding-parameters-to-the-url) so the user does not have to re-login. Support the `callback` parameter as well.
+* Interactive deposit
```
rice2000: Same comment as above: I think this will be easier to read if "For both interactive deposit and withdrawal," "Interactive deposit," and "Interactive withdrawal" function as delineators.

tomquisel:
@accordeiro that's a great point! I'll simplify the anchor section to just describe the interactive flow.

What do you think about applying the same simplification to the wallet section? I'm torn on it, because a number of existing anchors (like apay and naobtc) use the "immediate success" flow, and I'd like for wallets to be able to interact with them as a baseline. But it is more complicated to have the two branches...

ecosystem/sep-0006.md
```diff
@@ -67,6 +67,50 @@ Valid CORS headers are necessary to allow web clients from other sites to use th
 Access-Control-Allow-Origin: *
 ```
 
+## Recommendations
+
+SEP-6 lays out many options for how deposit and withdrawal can work. These are recommendations for getting a wallet or anchor implementation working with minimal effort while providing a great user experience.
+
+### Basic Wallet Implementation
+
+* Identify assets you want to support manually in advance
```
tomerweller: This is interesting. I think that if a wallet chooses to whitelist assets, they should do that in the balance/trustline level. If I have a trustline to asset X, I expect the wallet to present the deposit/withdraw flow that X issuer provides.

Some wallets might choose to abstract away the existence of trustlines or support just one currency, that changes things a bit but I don't think we should optimize for this use case. 

For simplicity, I would remove this point and add a qualifier in the next point.

ecosystem/sep-0006.md
```diff
@@ -67,6 +67,50 @@ Valid CORS headers are necessary to allow web clients from other sites to use th
...
+* Pick your approach to [fees](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0006.md#fee). We strongly recommend using `/info` to express fees if possible as it provides a better user experience (the user can see the fee structure in the wallet early in the process).
+* If you need no KYC info, explicit user signup, or other interaction from users to show deposit info or process a withdrawal, you can have your `/deposit` and `/withdraw` endpoints immediately return success with all info that the user will need to complete the deposit or withdrawal.
+* Otherwise, we recommend you provide an interactive deposit or withdrawal flow.
+* For both interactive deposit and withdrawal:
+    * Use the [Customer info needed response](#3-customer-information-needed-interactive) and include the `id` field so the wallet can check up on the status of the transaction (using the `id`) if they want.
+    * You should almost certainly use [authentication](#authentication) in the interactive flow, so be sure to accept the [`jwt` parameter](#adding-parameters-to-the-url) so the user does not have to re-login. Support the `callback` parameter as well.
```
tomerweller: Should we explicitly ask anchors to *not* use additional authentication (user, pwd) on their side?

accordeiro:
That's a good question. My initial thought was what you described: there would be no downsides on asking anchors to only implement the interactive flow, but asking wallets to do the same would make them incompatible with a few anchors.

I also don't love the two branches, but if we want to support these "immediate success" flow anchors as a baseline, it seems like we should keep both approaches as a recommendation for wallets – at least for now. 

Maybe we could get in touch with said anchors and start surfacing a migration to the interactive flow, and plan to update the wallet recommendation to solely interactive flow when we have sufficient adoption?

ecosystem/sep-0006.md
tomerweller: Following the previous point I would change "For each asset" to "For each asset the account has a trustline to"

ecosystem/sep-0006.md
tomquisel: This is more about only whitelisting integrations that the wallet has tested and confirmed work. I'll make it more clear.

ecosystem/sep-0006.md
tomquisel: They might want to deposit something with no trustline (yet), though.

tomquisel:
@rice2000 do you think the formatting is better now? Or should I use headers instead?

rice2000:
> @rice2000 do you think the formatting is better now? Or should I use headers instead?

I think this works great!  Aces!

PR #358, opened 2019-07-31
SEP6: make withdraw type optional

tomerweller:
Interactive withdraws can facilitate multiple types so no need for the wallet to know in advance.


tomquisel:
This is tricky. The field is only optional if the anchor decided to give an interactive experience that lets the user choose type. So maybe it's better to leave it required, and then leave it up to the anchor to ignore it if it wants.

tomerweller:
Thanks @tomquisel for clarifying!

PR #375, opened 2019-08-13
Simplify sep6 authentication

msfeldstein:
Currently there's two ways to send the JWT token to SEP6 endpoints, headers or query params. I propose that we simplify this to always use queryParams, since every client can send them this way.  My worry is that a server might develop using one or the other authentication strategies and forget to do the other one, keeping only one way will help ensure better support through the ecosystem.

@tomquisel @tomerweller

### DIFF
--- a/ecosystem/sep-0006.md
+++ b/ecosystem/sep-0006.md
@@ -49,12 +49,7 @@ As stated, Anchors may support [SEP-10](sep-0010.md) web authentication to enabl
 1. anchors must set the `authentication_required` field in their [info endpoint](#info)
 1. clients must submit the JWT previously obtained from the anchor via the [SEP-10](sep-0010.md) authentication flow to all API endpoints.
 
-The JWT should be included in all requests as request header:
-```
-Authorization: Bearer <JWT>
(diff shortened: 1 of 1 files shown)


tomquisel:
I really like the goal of consistency. That said, using the `Authorization` header is [idiomatic](https://jwt.io/introduction/) (search for `Bearer`), and so I think we should use the header everywhere we can.

The `jwt` param of course is a hack for the interactive situation where the wallet opens an iframe / popup and is unable to set headers. This [stack overflow] (https://stackoverflow.com/questions/32722952/is-it-safe-to-put-a-jwt-into-the-url-as-a-query-parameter-of-a-get-request) discusses putting JWTs in URLs, and it seems like a practice we should minimize.

It's also a breaking change, and several have already implemented with the header, so it's probably something we'd want to put in SEP-6-next in any case.

msfeldstein:
This makes sense.  Perhaps we should clarify that the url param is only for
that single case?

tomquisel:
That sounds great! Feel free to PR that 😄

PR #378, opened 2019-08-22
[SEP6] Expect amount from interactive webapp

msfeldstein:
In the interactive flow, we should expect the amount to come from the interactive webapp, not the wallet.  This will be more clear if we can separate interactive and non-interactive into their own seps, but for now, give instruction on how to pass the amount back.


ecosystem/sep-0006.md
```diff
@@ -399,6 +399,7 @@ The wallet must use the response fields in the following way to complete the wit
 - `withdraw_anchor_account`: send the withdrawal payment to this Stellar account 
 - `withdraw_memo`: (if specified) use this memo in the payment transaction to the anchor
 - `withdraw_memo_type`: use this as the memo type
+- `amount_in`: the amount expected in the stellar payment
```
accordeiro: Small fix: `stellar` -> `Stellar`

PR #382, opened 2019-08-28
[SEP6] Show example postMessage syntax

msfeldstein:
We've had lots of questions on how exactly the postMessage is supposed to work, so show some example code


tomquisel:
Nice!

PR #383, opened 2019-08-29
SEP-12 is deprecated for a SEP-6 implementation.

tempofr:
* Anchors may support [SEP-12](sep-0012.md) customer info transfer if they need to handle KYC-ing customers


msfeldstein:
I agree with the intention here but want to figure out the right way to do this.  While we do recommend to use the interactive flow, some anchors/wallets still use the non-interactive flow so i don't want to break anyone who is still using non-interactive.

That said I don't know if anyone actually uses SEP-12 currently so maybe it is best to just remove vs add a deprecation warning.

@accordeiro what do you think on this? I definitely don't want people to think they should implement sep12.

msfeldstein:
It sounds like no one uses this, merging.

PR #384, opened 2019-08-29
Add deprecated status

msfeldstein:
SEP12 has been deprecated inside the spec but is currently shown as "Active" in the list.  We should make it clear that we clean out old seps.

### DIFF
--- a/ecosystem/README.md
+++ b/ecosystem/README.md
@@ -29,6 +29,7 @@
   implemented.
 * **Superseded: [New Final SEP]** - A SEP that which was previously final but has been superseded
   by a new, final SEP. Both SEPs should reference each other.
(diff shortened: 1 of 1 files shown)


accordeiro:
I think this makes sense to me. One thing worth investigating before we merge: on SEP-12, we mention

> Therefore, SEP-12 is deprecated for anyone using it **as part of a SEP-6 implementation**.

Does that mean that there are other use cases in which we still recommend SEP-12? If not, should we remove that ambiguity from the SEP-12 too?

msfeldstein:
It's a technically functional sep which so far no one has needed to use.  I'd prefer to deprecate or move back to draft until there is acutally a need for it.

That said, this doesn't actually do anything to sep12, it just creates a deprecated status

accordeiro:
You're right, I got confused by the mention of SEP12 on the description.

msfeldstein:
I actually have other changes to make so i might as well do this all in one go and get broader community feedback

PR #385, opened 2019-08-29
Simplify/Clarify the SEP process

msfeldstein:
- Remove the Information vs Standard track distinction.  It seems confusing and I'm not sure what the value prop of having the complexity is.
- Create a `Deprecated` status, and explain how a SEP would get put into such status.
- Modify the `Final` status to be one in which there are live implementations being used.  Previously, `Final` and `Active` were too similar, and often misused.  Requiring a Final sep to have multiple implementations will give us much more confidence to mark something as final since we probably got most of the edge cases handled.
- Replace Johnny with Michael as a SEP team member

### DIFF
--- a/ecosystem/README.md
+++ b/ecosystem/README.md
@@ -1,28 +1,17 @@
 # Stellar Ecosystem Proposals (SEPs)
 
-## SEP Tracks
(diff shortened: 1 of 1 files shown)


msfeldstein:
Separating into multiple changes

PR #387, opened 2019-08-29
Remove Informational vs Standard track distinction

msfeldstein:
Remove the Information vs Standard track distinction. It seems confusing and I'm not sure what the value prop of having the complexity is.

### DIFF
--- a/ecosystem/README.md
+++ b/ecosystem/README.md
@@ -9,14 +9,12 @@
   which any new concerns should be addressed, the SEP's status will become `Active` or `Final`,
   depending on the content of the SEP.
 * **Active** - An actively maintained SEP that is intended for immediate adoption by the entire
-  ecosystem by its author(s), and if it is a Standard, by SDF as well. Additional updates may be
-  made without changing the SEP number.
+  ecosystem. Additional updates may be made without changing the SEP number.
 * **Final** - A finalized SEP that is intended for immediate adoption by the entire
-  ecosystem by its author(s), and if it is a Standard, by SDF as well. A final SEP should only be
-  updated to correct errata.
+  ecosystem. A final SEP should only be updated to correct errata.
 
 ### Additional Statuses
-* **Rejected** - A Standards SEP that has been formally rejected by the SEP Team, and will not be
+* **Rejected** - A SEP that has been formally rejected by the SEP Team, and will not be
   implemented.
 * **Superseded: [New Final SEP]** - A SEP that which was previously final but has been superseded
   by a new, final SEP. Both SEPs should reference each other.
--- a/ecosystem/README.md
+++ b/ecosystem/README.md
@@ -40,7 +40,7 @@
 ### Draft Proposals
 
 | Number | Title | Author | Status |
-| --- | --- | --- | --- | --- |
+| --- | --- | --- | --- |
 | [SEP-0013](sep-0013.md) | DEPOSIT_SERVER proposal | @no, @ant, @manran, @pacngfar | Draft |
 | [SEP-0014](sep-0014.md) | Dynamic Asset Metadata | OrbitLens, Paul Tiplady | Draft |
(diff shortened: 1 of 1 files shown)


ecosystem/README.md
```diff
@@ -1,14 +1,5 @@
...
 
-## SEP Tracks
-* **Informational** — A SEP on the `Informational` track is one that is open to adoption by the
-  ecosystem, but has not been formally standardized by SDF, and is not endorsed by SDF for
-  adoption. Typically a SEP can start as `Informational` to gain traction within the ecosystem
-  before moving to the `Standards` track.
-* **Standard** — A SEP on the `Standards` track is one that aims for formal standardization and
-  endorsement by SDF for adoption. Typically a SEP Standard has a higher bar towards acceptance,
-  and it requires approval by 2 SDF members of the SEP Team.
-
```
leighmcculloch: The word `Standard` is mentioned in the SEP Status Terms section. I can't comment on the lines directly, but its line 12, line 15, line 19.

msfeldstein: updated

MonsieurNicolas:
As this distinction was made to reduce the friction for non SDF proposals, I'd ask @rice2000 what he thinks about it

tomerweller:
I agree that this might be confusing but would not remove it without addressing the issue it tries to solve. There needs to be a way for people in the ecosystem to inform the community of a convention they are using, without going through the hassle of the standardization process. This is a similar model to [EIPs](https://eips.ethereum.org/).

ecosystem/README.md
```diff
@@ -18,49 +9,47 @@
...
+| [SEP-0008](sep-0008.md) | Regulated Assets | Interstellar | Final |
+| [SEP-0009](sep-0009.md) | Standard KYC / AML Fields | SDF | Active |
+| [SEP-0010](sep-0010.md) | Stellar Web Authentication | Sergey Nebolsin, Tom Quisel | Active |
+| [SEP-0011](sep-0011.md) | Txrep: Human-Readable Low-Level Representation of Stellar Transactions | David Mazières | Active |
+| [SEP-0012](sep-0012.md) | Anchor/Client Customer Info Transfer | Interstellar | Active |
+| [SEP-0020](sep-0020.md) | Self-verification of validator nodes | Johan Stén | Active |
 
 ### Draft Proposals
 
 | Number | Title | Author | Track | Status |
```
rice2000: Delete | Track | column heading so table renders properly

rice2000:
Currently, the claim is that: "The bar for accepting an `Informational` SEP is much lower than one for a `Standard`..."  However, the only real difference I see is that the `Standard` track requires 3 approvals, whereas an `Informational` requires 2.  That doesn't seem like a meaningful difference to me, and it doesn't really seem to reduce friction.

Reducing complexity _does_ reduce friction, so I'm all for getting rid of the tracks.

msfeldstein:
Yeah i get the point of the informational track, and in this case i feel it would be better to have something just totally separated.  If it's purely informational we don't need a real review or to go through the stages, we would just want to ensure the proposals are written well and are at least somewhat useful to have in this repo.

Another solution might be to have informational seps be another status vs a separate track. These could go `Draft -> Awaiting Decision -> Informational` instead of going to active or final.

tomquisel:
Is there an example of an informational SEP? Or one being concocted somewhere?

msfeldstein:
There are a few draft proposals here: https://github.com/stellar/stellar-protocol/tree/master/ecosystem, and one that looked like sep6 that we ended up rejecting since sep6 subsumed it.

msfeldstein:
https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0022.md
and 
https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0021.md

leighmcculloch:
This PR looks like it has been superseded by #555. I'm going to close it because of that. If I'm making a mistake by closing this, sorry!, please reopen.

PR #388, opened 2019-08-29
Add Deprecated status to SEP Process

msfeldstein:
Create a deprecated status and explain how to move a SEP into it.  Its important that we remove any cruft from our SEPs to ensure people have a clear understanding of the state of the ecosystem, and to prevent people from implementing SEPs that won't be compatible with anyone else.

### DIFF
--- a/ecosystem/README.md
+++ b/ecosystem/README.md
@@ -10,6 +10,7 @@
   and it requires approval by 2 SDF members of the SEP Team.
 
 ## SEP Status Terms
+* **Deprecated** - A SEP that was previously on an active track but has been deprecated and is no longer suggested for use. There may be legacy usage of a deprecated SEP.
 * **Draft** - A SEP that is currently open for consideration and actively being discussed.
 * **Awaiting Decision** — A mature and ready SEP that is ready for approval by the SEP
   Team. If enough the approval requirements are met by SEP team members, the SEP will move towards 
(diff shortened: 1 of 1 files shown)


ecosystem/README.md
```diff
@@ -10,6 +10,7 @@
   and it requires approval by 2 SDF members of the SEP Team.
 
 ## SEP Status Terms
+* **Deprecated** - A SEP that was previously on an active track but has been deprecated and is no longer suggested for use. There may be legacy usage of a deprecated SEP.
```
leighmcculloch: I think we should move this into the `Additional Statuses` section. The statuses in the `SEP Status Terms` section are for proposals that are accepted or on track for being accepted. The statuses in the `Additional Statuses` appear to be for proposals that are no longer in use, which a deprecated proposal is more related.

ecosystem/README.md
```diff
@@ -146,6 +147,10 @@ From there, the following process will happen:
   * If the SEP requires active maintenance, such as having an open schema, it should be marked at
     `Active`.
   * Otherwise, the SEP should be marked as `Final`.
+  
+### Regression
+* It is possible for a SEP to move from `Active` to `Draft` or `Deprecated` if it is never adopted, or is abandoned by the community.
```
leighmcculloch: A `Draft` proposal should probably move to `Rejected` rather than `Deprecated`. Otherwise there's not much distinction between the two?

leighmcculloch: Oh, I misread the sentence. It's saying that `Active` can move to `Draft`/`Deprecated`, not that `Active`/`Draft` can move to `Deprecated`. Sorry!

msfeldstein: Yeah i'm not so sold on this sentence. I want to be able to move back to `draft` because there are some seps that seemingly would be useful in the future so i don't want to put them in a rejected status

leighmcculloch: Maybe break it up into two statements.
```suggestion
* It is possible for a SEP to move from `Active` to `Draft` if it is not adopted and it needs further work.
* It is possible for a SEP to move from `Active` to `Deprecated` if it is never adopted and abandoned.
```

ecosystem/README.md
rice2000: I agree: this should move to `Additional Statuses`.  The current `Status Terms` list is in order of evolution from Draft through Final.

PR #389, opened 2019-08-29
Update the definitions of Active and Final status

msfeldstein:
- Modify the Final status to be one in which there are live implementations being used. Previously, Final and Active were too similar, and often misused. 
- Requires a Final SEP to have multiple implementations will give us much more confidence to mark something as final since we probably got most of the edge cases handled.
- Strongly suggests an Active SEP needs to have a reference implementation

### DIFF
--- a/ecosystem/README.md
+++ b/ecosystem/README.md
@@ -18,8 +18,7 @@
   which any new concerns should be addressed, the SEP's status will become `Active`.
 * **Active** - An actively maintained SEP that is intended for immediate adoption by the entire
   ecosystem. Additional updates may be made without changing the SEP number.
-* **Final** - A finalized SEP that is being used in live products and will not be changed aside from
-  minor errata.
+* **Final** - A finalized SEP will not be changed aside from minor errata. For a proposal to be a candidate to be made Final it must be being used in live products.
 
 ### Additional Statuses
 * **Rejected** - A Standards SEP that has been formally rejected by the SEP Team, and will not be
--- a/ecosystem/README.md
+++ b/ecosystem/README.md
@@ -144,7 +144,8 @@ From there, the following process will happen:
 * Minor changes may be made as more implementations are brought online highlighting any edge cases.
 
 ### Active -> Final
-* Once there are live implementations being used in production, the SEP team can move the SEP to `Final` status.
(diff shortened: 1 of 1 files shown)


ecosystem/README.md
```diff
@@ -15,14 +15,11 @@
...
+  which any new concerns should be addressed, the SEP's status will become `Active`.
 * **Active** - An actively maintained SEP that is intended for immediate adoption by the entire
-  ecosystem by its author(s), and if it is a Standard, by SDF as well. Additional updates may be
-  made without changing the SEP number.
-* **Final** - A finalized SEP that is intended for immediate adoption by the entire
-  ecosystem by its author(s), and if it is a Standard, by SDF as well. A final SEP should only be
-  updated to correct errata.
+  ecosystem. Additional updates may be made without changing the SEP number.
+* **Final** - A finalized SEP that is being used in live products and will not be changed aside from
+  minor errata.
```
leighmcculloch: This statement makes it sound like the primary reason for a proposal to be a candidate for being 'Final' is that it has been adopted in live products. I think we could have 'Active' proposals that have a few live implementations that aren't ready for being finalized.

I think we should reverse the sentence and make the focus that the primary attribute of a 'Final' proposal is that it will not change.

>A finalized SEP will not be changed aside from minor errata. For a proposal to be a candidate to be made `Final` it must be being used in live products.

ecosystem/README.md
```diff
@@ -141,11 +138,16 @@ From there, the following process will happen:
...
-    `Active`.
-  * Otherwise, the SEP should be marked as `Final`.
+### FCP -> Active
+* If no major concerns are brought up, the SEP is marked as `Active` by your SEP buddy.
+* Ideally there will be a reference implementation exhibiting the behavior and value of the SEP before moving to active state.
+* Active SEPs should be brought into production by ecosystem members.
+* Minor changes may be made as more implementations are brought online highlighting any edge cases.
+
+### Active -> Final
+* Once there are live implementations being used in production, the SEP team can move the SEP to `Final` status.
```
leighmcculloch: Rather than live implementations being the trigger, can we make it a requirement? I think the trigger should be that a decision has been made that the proposal is complete, proven, and won't be extended, and one of the most important attributes to consider when making that decision is if it's being used in live products.

Issue #330, opened 2019-06-20
SEP-0001: limit stellar.toml size to 100kb


tomerweller:

ire-and-curses:
[Context](https://github.com/stellar/go/issues/1429). We want to standardise this number for all SDKs so that interoperability is assured. Using the federation limit seems to make sense.

PR #390, opened 2019-08-30
[SEP1] Standardize max filesize

msfeldstein:
Need to standardize a max file size for stellar.toml so parsers can ensure they can handle any toml.  Context in https://github.com/stellar/stellar-protocol/issues/330

### DIFF
--- a/ecosystem/sep-0001.md
+++ b/ecosystem/sep-0001.md
@@ -38,6 +38,8 @@ It is also recommended to set a `text/plain` content type so that browsers rende
 content-type: text/plain
 ```
 
+`stellar.toml` can have a maximum file size of 100kB.
+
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0001.md
```diff
@@ -38,6 +38,8 @@ It is also recommended to set a `text/plain` content type so that browsers rende
 content-type: text/plain
 ```
 
+`stellar.toml` can have a maximum file size of 100kB.
```
leighmcculloch:
```suggestion
`stellar.toml` can have a maximum file size of 100KB.
```

I think uppercase `K` is more commonly used in computer data measurements.

According to [Wikipedia](https://en.wikipedia.org/wiki/Kilobyte) we should write this as `100KB` rather than `100kB` because lowercase `k` is from SI units and is `1000` and uppercase `K` is more common used for binary units and is from IEC binary units, `1024`. See 

The value in the code is `100 * 1024`:

https://github.com/stellar/go/blob/fd815d6/clients/stellartoml/main.go#L5-L6

msfeldstein: thanks i was wondering about the kCase

PR #403, opened 2019-09-16
SEP6 clarify error messaging

### LINKED ISSUE #402: SEP0006: Make it clear that `error` responses should be strings
/deposit and /withdraw endpoints sometimes return errors. The examples indicate that they should return a property `error` that's a string, but that's not made clear in the spec's language.

So my suggestion is:

- Canonize the `error` response property, required for errors, and require that it be a string. It should be human-readable and user-friendly.
- Add an optional `error_metadata` property that must be an object, and can contain more information about the error.

msfeldstein:
For /withdraw and /deposit, we need to indicate that the error message is a string to be displayed to users, and also have a language option for localization.

Fixes #402


msfeldstein:
What do you expect error_metadata to be used for?

morleyzhi:
@msfeldstein That's so the anchor can provide more structured data about the error, like a list of invalid inputs, for example.

msfeldstein:
We can do a follow up for that if we need, im not sure how it would work since we would need to either define a set of available values, or clients and anchors would need to work together to come to consensus among themselves about what fields are and mean.

PR #404, opened 2019-09-16
Improve withdrawal flow on wallets that enable cross-border payments

### LINKED ISSUE #395: Withdraw flow issues on wallets that enable cross-border payments
This issue describes a use case for Wallets that support cross-border payments, and a few issues that have been encountered while trying to lay out the correct withdrawal flow.

### Use case
- User holds a balance in their Stellar account of some USD token, let's call it `USDZ`
- User wants to withdraw that balance in their native currency, say Brazilian Real (BRL). The anchor that issues BRL tokens calls those tokens `BRLZ`

### User / Wallet / Anchor interaction flow
In order to accomplish that use case, we could use the following flow:
1. User indicates to Wallet that it wants to withdraw a part of its `USDZ` balance in their native currency
2. Wallet shows how many `USDZ` it would cost for the user to withdraw a certain amount of BRL (the Wallet can use the [Stellar Ticker API](https://ticker.stellar.org/markets.json) to calculate the conversion rate). In this example, let's say at the given time the Wallet informs the user that they would have to use `100 USDZ` of their balance to withdraw `400 BRL`.
3. With the given to-withdraw amount in hands (`400 BRL`), the Wallet hits the `BRLZ` anchor's `/withdraw` endpoint, passing `amount=400` as a parameter.
4. As we're currently recommending to anchors, the response for the `/withdraw` request would be a `403 (Forbidden)` with a URL pointing to the interactive flow
5. Wallet opens the interactive URL, and the User continues with the KYC / withdrawal process through the `BRLZ` Anchor's UI, informing their bank account info, etc.
6. Once the process is completed, the Anchor provides a Stellar address for the Wallet to make the `BRLZ` transfer (via `postMessage`) before it fulfills the external deposit
7. The Wallet trades (maybe via a path payment to itself) `100 USDZ` to `BRLZ`, then sends the newly acquired `BRLZ` to the given Stellar address, and polls the `/transaction` to get status updates about the withdrawal, until the transaction's status is marked as `complete`.

### Problems
There are a few issues with this approach.

#### P1: Double input of amount

The user has to input a `BRL` amount early in the process, in step 2. However, as we're suggesting anchors to collect most info through the interactive flow, that user will likely be presented with an "amount" input again in the interactive flow. Even if the amount is pre-populated, giving the user the ability to change that amount might lead to an inconsistent state in the later steps. 

##### Suggested solution
One solution for that is to provide some kind of `amount_readonly=true` parameter to the interactive popup URL, and either skip the amount input or display it as read-only during the interactive withdrawal flow.

#### P2: Exchange rate fluctuation

On step `2`, the Wallet uses an exchange rate (`ER_1`) to calculate how many `USDZ` are needed to withdraw `400 BRL`. However, this exchange rate might have changed to `ER_2` when the actual exchange happens, which is only on step `7`. Wallets could ask the user if they would like to update the `USDZ` amount to reflect the exchange rate, but we should have a solution in case the user doesn't want to do that (or can't, in case `100 USDZ` is their full balance and `ER_2` is more expensive than `ER_1`).

##### Suggested solution
In order to provide a smooth user experience in this scenario, we could recommend/require anchors to be flexible to accept `398 BRLZ` or `402 BRLZ`, as long as the correct `memo` is provided in the transaction.

### Questions
- Do the suggested solutions make sense?
- Is there any alternative proposal for addressing this use case?

@morleyzhi @brunomuler please let me know if this flow correctly describes what we've discussed last Friday
@tomquisel @msfeldstein @tomerweller looking forward to getting your thoughts on this :)

accordeiro:
This PR fixes #395. It aims to:

- Emphasize the importance of accepting an amount value as a parameter in both `/withdraw` and the interactive flow URL
- Point out that anchors need to be flexible about the informed amount vs. what is actually transferred to their Stellar accounts (a 10% margin might be a good start)
- Instruct anchors to watch for other types of `payment` operations besides `payment`

I'm still a bit unsure whether I've made changes in the right places, since we now have the Basic Implementation recommendations and the actual specs. 

@msfeldstein @tomerweller @tomquisel please let me know your thoughts.


ecosystem/sep-0006.md
```diff
@@ -420,7 +424,7 @@ The wallet must use the response fields in the following way to complete the wit
 - `withdraw_anchor_account`: send the withdrawal payment to this Stellar account. 
 - `withdraw_memo`: (if specified) use this memo in the payment transaction to the anchor.
 - `withdraw_memo_type`: use this as the memo type.
-- `amount_in`: the amount expected in the stellar payment.
+- `amount_in`: the amount expected in the Stellar payment (with an acceptable error margin of 10%)
```
msfeldstein: The amount_in should never fluctuate right?  Only the amount that actually ends up sent? The error margin here is confusing.

accordeiro: Yeah, I agree it's confusing. I wanted to make it clear for wallets that there is an acceptable error margin, but this is probably not the spot. I'll push a change to address this shortly.

ecosystem/sep-0006.md
```diff
@@ -115,12 +116,16 @@ SEP-6 lays out many options for how deposit and withdrawal can work. These are r
...
 * **Interactive deposit**
     * Your interactive deposit popup will do everything needed to initiate the deposit without the user needing to interact with the wallet further. It should either directly initiate the deposit, or handle displaying information (such as reference number or bank account number) that the user will need to complete their deposit.
 * **Interactive withdrawal**
     * Your withdrawal flow will have to pass control back to the user's wallet, so it can initiate the withdrawal with a Stellar payment to your withdrawal address. You'll need to communicate the withdrawal address, amount and status to the wallet using the [`callback` parameter](#adding-parameters-to-the-url), and also by making it available on your `/transaction` endpoint. See [details](#guidance-for-wallets-completing-an-interactive-withdrawal).
+    * If an `amount` parameter is provided to the interactive URL, make sure to use that value to pre-populate the amount on the interactive form. This greatly improves the user experience, especially in wallets that enable cross-border payments.
```
msfeldstein: Remove the reference to cross-border payments here

ecosystem/sep-0006.md
```diff
@@ -115,12 +116,16 @@ SEP-6 lays out many options for how deposit and withdrawal can work. These are r
...
+    * In order to fulfill a withdrawal, a wallet must make a payment to the Stellar address that the anchor provides. It is the anchor's job to watch for Stellar payments to the given address and make the external transaction as soon as they're detected. It is recommended for anchors to listen to at least `payment` and `path_payment` [operations](https://www.stellar.org/developers/horizon/reference/endpoints/payments-all.html), although supporting the remaining ones is important for covering all use cases.
```
msfeldstein: > It is recommended for anchors to listen to at least `payment` and `path_payment`

Should we mention our SDKs do this for our /payment streaming endpoints already?

ecosystem/sep-0006.md
```diff
@@ -115,12 +116,16 @@ SEP-6 lays out many options for how deposit and withdrawal can work. These are r
...
+    * When the wallet is exchanging relatively unstable currencies, there might be slight fluctuations on value between the informed withdrawal `amount` and the actual transferred amount. It is recommended for anchors to accept an amount fluctuation of up to ±10%, and  adjust the amount to be transferred (and fees) to reflect the actual value received.
```
msfeldstein: I'd like to make this section a little clearer that the unstable currencies isn't something the anchor needs to worry about, but they should expect that the amount requested to withdraw might not be the exact amount they receive, and they should be flexible.

msfeldstein: I'd prefer to say "Some wallets might..." to be clear that this isn't something that always happens, and just a feature some wallets add

ecosystem/sep-0006.md
```diff
@@ -102,6 +102,7 @@ SEP-6 lays out many options for how deposit and withdrawal can work. These are r
         * Provide an interface to the user that allows them to view the withdrawal details including the computed fee. The user can confirm and then the wallet will initiate the withdrawal by sending the Stellar payment to the address provided by the issuer.
     * If the anchor responds with an interactive flow, handle it as [described in detail](#3-customer-information-needed-interactive).
     * For now, do not handle the non-interactive or status responses to the `/withdraw` request.
+    * For wallets that enable cross-border payments: exchange rate fluctuations might require withdrawal values to slightly vary from the originally provided `amount`. Anchors are instructed to accept a variation of ±10% between the informed `amount` and the actual value sent to the anchor's Stellar account. The withdrawn amount will be adjusted accordingly.
```
msfeldstein: make the same type of "Some wallets" change here too and we should be good to go

PR #410, opened 2019-09-19
Update sep-0006.md

istrau2:
Adds pending_kyc transaction status

Sometimes a user might initiate a transaction but still needs to complete kyc before the anchor can complete the deposit/withdrawal

### DIFF
--- a/ecosystem/sep-0006.md
+++ b/ecosystem/sep-0006.md
@@ -732,6 +732,7 @@ Name | Type | Description
 * `pending_anchor` -- deposit/withdrawal is being processed internally by anchor.
 * `pending_stellar` -- deposit/withdrawal operation has been submitted to Stellar network, but is not yet confirmed.
 * `pending_trust` -- the user must add a trust-line for the asset for the deposit to complete.
+* `pending_kyc` -- the user must complete kyc before the deposit / withdrawal can complete.
 * `pending_user` -- the user must take additional action before the deposit / withdrawal can complete.
 * `pending_user_transfer_start` -- the user has not yet initiated their transfer to the anchor. This is the necessary first step in any deposit or withdrawal flow.
 * `incomplete` -- there is not yet enough information for this transaction to be initiated. Perhaps the user has not yet entered necessary info in an interactive flow.


msfeldstein:
I was considering this, but would actually rather simplify the status codes to make consistency more likely. Would `incomplete` handle this use case?

```
`incomplete` -- there is not yet enough information for this transaction to be initiated. Perhaps the user has not yet entered necessary info in an interactive flow.
```

istrau2:
@msfeldstein idk, I kinda think more statuses are better. So long as they stay distinct and focused. A status can be very helpful to a user, clever app developer or even administrator trying to diagnose an issue. 

For example, how do I as an app developer trying to integrate with SEP6 handle an incomplete status?  Handling pending_kyc status is quite clear, redirect the user to the interactive_url. Same thing with a user or administrator, the status is clear and helpful.

I took a look at your SEP24 proposal. In general, I am all for simplifying SEP6 but not at the expense of the user. I'd prefer a long complicated SEP that results in a top-notch user experience than a much simpler SEP that results in a user experience that is even 1% less intuitive.

msfeldstein:
I  think you handle `incomplete` the same way as `pending_kyc`, just open  the interactive url.  That's the reason i am 100% geared towards the  interactive flow, it requires so much less knowledge between the wallets  and anchors.  If we add more statuses, and a wallet doesn't know how to handle it it will result in a broken experience.  @tomquisel @accordeiro what do you think about the decision between a larger number of more specific status codes vs a single `incomplete` status code that just launches the interactive URL.

Yeah I think the 24 proposal will  result in a better user experience (that's all i really care about and why I did it).  Complex SEPs end up with incompatible incomplete implementations, which is what we were seeing.  Entities didn't quite understand whether things were for interactive, or non-interactve, and what was actually required for each flow.

Please jump in on the discussion in 24, no one has commented yet and I want to get people raising the issues of what we'd be losing with the simplification.

msfeldstein:
Regarding UX, I think the biggest thing all these seps have been missing is UX flows, and I think we saw the result of that with the constant flux of SEP6.  I'm hoping to have a solid UX flow for SEP24 illustrating how things work from a user perspective as well as what status codes are available at what point, etc

msfeldstein:
Also I'm not against adding this status code but i want to make sure it actually adds value over just saying the transaction can't be completed yet for reasons which include KYC.  What would the wallet do differently with a pending_kyc status vs an incomplete status?

istrau2:
@msfeldstein I suppose I am not understanding what `incomplete` means. To me, `incomplete` means the deposit or withdrawal is incomplete for whatever reason (i.e. it is a sort of catchall for incomplete transactions). By `incomplete` do you mean `interactive_info_needed`?

msfeldstein:
That's how i read it.  I'm feeling like this is just a documentation issue on the incomplete status.  Incomplete is a pretty bad name, everything besides `complete` is obviously incomplete.

istrau2:
@msfeldstein With regard to SEP24, I will jump in soon (please ping me if I don't, I am @istrau2 on keybase). Its just that I am going through a particularly stressful period right now with Stellarport and totally engrossed in that atm (the stress comes from KYC related issues thus the original PR).

istrau2:
No longer need this. @msfeldstein had a good suggestion on how to avoid this scenario.

PR #417, opened 2019-10-03
[SEP1] Add sep10 to the list of SIGNING_KEY usages

msfeldstein:
@tomerweller is there any reason these should be separate fields?


tomerweller:
> is there any reason these should be separate fields?

Maybe if an entity implements both SEP3 and SEP10, and wants to maintain separate environments without sharing private keys between them. Given that SEP3 is practically DOA and I don't think anyone implements both, I don't see any issue right now.

> :shipit:

PR #419, opened 2019-10-04
Deprecate SEP3

msfeldstein:
Seems to be deprecated in the ecosystem at large and we recommend sep6/12 instead according to https://www.stellar.org/developers/guides/compliance-protocol.html.  Moving this to deprecated status to avoid confusion

### DIFF
--- a/ecosystem/sep-0003.md
+++ b/ecosystem/sep-0003.md
@@ -4,9 +4,9 @@
 SEP: 0003
 Title: Compliance protocol
 Author: stellar.org
-Status: Active
+Status: Deprecated
 Created: 2017-10-30
-Updated: 2018-07-15
+Updated: 2019-10-04
 Version 1.0.1
 ```
 


leighmcculloch:
Does this mean the [compliance] and [bridge] servers are deprecated as well? The linked document definitely favors SEP-6/12 over SEP-3, but it also acknowledges that most cases will be better served by them, not all. What's the use case for SEP-3 that we might be missing here?

[compliance]: https://github.com/stellar/go/tree/master/services/compliance
[bridge]: https://github.com/stellar/go/tree/master/services/bridge

ire-and-curses:
I'd also appreciate clarification on the intended status of compliance and bridge. We occasionally get people asking whether to use them. @tomquisel

msfeldstein:
That's why I want to deprecate anything that isn't actively being used,
even if it might be in the future. No reason to have people start off
confused and have to maintain baggage, if someone finds a need for it we
can reinstate it and probably do it with some changes they need.

tomquisel:
I'm happy with the deprecation, it sounds like a good move, and we can always bring it back to life. To be fair, I think the reason external groups get confused and try to use SEP-3 is not because of the SEPs, but rather a result of out-of-date documentation, guides, or presentations we're giving.

leighmcculloch:
If we're going to deprecate it, should we do the same with `compliance` and `bridge`? Or keep maintaining them on the basis that someone might pick it up even though it's no longer SEParized?

tomquisel:
I'd say your call. I think it depends on whether some partners we don't know about are using them, and what the maintenance cost is.

msfeldstein:
@rice2000 is looking into more concrete answers on whether anyone is using it, but unless he comes back with something i think its safe to say we should deprecate them as well.

And i agree we need to clean up the documentation/guides, i can make a PR for that as well

msfeldstein:
I'm not sure why there's 2 PRs for this...

msfeldstein:
Closed in favor of https://github.com/stellar/stellar-protocol/pull/420

PR #420, opened 2019-10-04
Deprecate SEP3

msfeldstein:
according to [our docs](https://www.stellar.org/developers/guides/compliance-protocol.html) we don't recommend sep3 usage anymore and i don't know anyone who's actually implemented it.  Deprecate it now to avoid confusion in future.

### DIFF
--- a/ecosystem/README.md
+++ b/ecosystem/README.md
@@ -33,7 +33,6 @@
 | --- | --- | --- | --- | --- |
 | [SEP-0001](sep-0001.md) | stellar.toml specification | SDF | Standard | Active |
 | [SEP-0002](sep-0002.md) | Federation Protocol | SDF | Standard | Final |
-| [SEP-0003](sep-0003.md) | Compliance Protocol | SDF | Standard | Active |
 | [SEP-0004](sep-0004.md) | Tx Status Endpoint | SDF | Standard | Final |
 | [SEP-0005](sep-0005.md) | Key Derivation Methods for Stellar Accounts | SDF | Standard | Final |
 | [SEP-0006](sep-0006.md) | Anchor/Client Interoperability | SDF | Standard | Active |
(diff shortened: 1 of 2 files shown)


leighmcculloch:
Copying over my commented from https://github.com/stellar/stellar-protocol/pull/419#issuecomment-538558468:

Does this mean the [compliance] and [bridge] servers are deprecated as well? The linked document definitely favors SEP-6/12 over SEP-3, but it also acknowledges that most cases will be better served by them, not all. What's the use case for SEP-3 that we might be missing here?

[compliance]: https://github.com/stellar/go/tree/master/services/compliance
[bridge]: https://github.com/stellar/go/tree/master/services/bridge

msfeldstein:
Unclear why there's 2 PRs for this but some more conversation in https://github.com/stellar/stellar-protocol/pull/419

msfeldstein:
@rice2000 @tomquisel is the reason that this was never adopted by anyone because no one had a business need for it, or because it wasn't flexible enough to serve most needs (a la non-interactive sep6)?

tomquisel:
I think it was lack of business need, but I don't know all the history. @jedmccaleb do you know?

tomquisel:
I was chatting with Lisa at the offsite, and she brought up how Tempo and Cowrie want to interact. Basically, they want to exchange client KYC info server-to-server, just like SEP-3 and the compliance server provide. @tomerweller and @msfeldstein maybe it's worth syncing to see if SEP-3 is the right answer and if it's worth un-deprecating SEP-3 for it? Sorry to get that thought in after the cutoff 😄

msfeldstein:
Can you tag someone from those orgs to take a look and see if they want to use this?  If this would solve their problems i'm happy to undeprecate

rice2000:
@tempofr Thoughts?

tomerweller:
I think that by creatively squinting at sep24 you can mimic sep3. Essentially, you need to have the sender initiate a withdraw on behalf of the recipient before they start the deposit on their end. Only after getting the thumbs up to withdraw, they deposit and complete the withdraw using a path payment.

Obviously, wallets that support this flow would need to join the creative squinting. I still think it's better to have one deposit/withdraw standard. 

@msfeldstein thoughts?

msfeldstein:
i would be ok with adding some color to SEP24 to make this flow more
obvious so we don't have to squint

tomquisel:
I'm onboard for the SEP-24 B2B approach! It'd be amazing if we could use the same tooling (polaris and demo wallet) to support / test both cases.

PR #425, opened 2019-10-08
Add email as an example of sep9

msfeldstein:
The old sep6 had some stuff about passing email address from a wallet to an anchor.  This special case has been removed in favor of just supporting any SEP9 fields that the wallet may have.  It is useful to have the string `email address` there so that people can ctrl+f for `email` to find instructions on how to do this.

### DIFF
--- a/ecosystem/sep-0024.md
+++ b/ecosystem/sep-0024.md
@@ -86,7 +86,7 @@ SEP-24 lays out many options for how deposit and withdrawal can work. These are 
 * **Make a request to `/deposit` or `/withdraw`.**
     * This will respond with the interactive url needed to proceed with KYC and deposit/withdraw details.
 * **For `/deposit` and `/withdraw`**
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0024.md
```diff
@@ -86,7 +86,7 @@ SEP-24 lays out many options for how deposit and withdrawal can work. These are
 * **Make a request to `/deposit` or `/withdraw`.**
     * This will respond with the interactive url needed to proceed with KYC and deposit/withdraw details.
 * **For `/deposit` and `/withdraw`**
-    * Optionally attach any fields from [SEP-9](sep-0009.md) as query parameters in the interactive URL, in order to let the anchor pre-fill them in the interactive flow UI.  This is optional, but can create a much nicer user experience.
+    * Optionally attach any fields from [SEP-9](sep-0009.md) as query parameters in the interactive URL, in order to let the anchor pre-fill them in the interactive flow UI.  This is optional, but can create a much nicer user experience.  Email address and name are good examples of fields to help pre-fill the anchors website.
```
morleyzhi: I think this should be `email_address` instead of "Email address", ditto for `name`, to hammer the point home that these are specifically-named fields.

msfeldstein: Good call

PR #426, opened 2019-10-10
Add 'native' as an endorsed string of the native asset

leighmcculloch:
### What
Add `"native"` as an endorsed string representation of the native asset.

### Why
The standard defines `"NATIVE"` to be the string representation of the native asset. In practice we use the lowercase string `"native"` in Horizon and in Ticker. While the specification does mention that any string up to 12 characters to be a valid reference to the native asset I think it is worth naming `"native"` explicitly since we have been using it in our services.

cc @stanford-scs


leighmcculloch:
@stanford-scs Moving our conversation from https://github.com/leighmcculloch/stellar-protocol/pull/1#issuecomment-540312037 to here:

> I don't feel strongly about 'native' vs 'NATIVE'. But in real tools I think it is more useful to put the currency symbol, like `XLM`. That's why it accepts any string without colon-issuer as the native asset. In `stc`, the name of the native asset is a parameter in a configuration file, and it defaults to `XLM` for the public network and `TestXLM` for the test network.

Got it, I think that makes a lot of sense for rendering to string, but that flexibility comes at a cost of robust error protecting parsing back into machine format. e.g. If someone naively encodes the asset `BTC` they are unlikely to expect it to be encoded to the native asset. It'd be desirable if the parser was allowed to raise an error.

If the known valid use cases are `XLM`, `NATIVE`, and `native`, could we make only those asset names valid as referring to the native asset?

leighmcculloch:
Closing this change because @stanford-scs has opened a similar change #429 which I think better addresses the goal of this PR.

PR #427, opened 2019-10-10
Make optional fields required

msfeldstein:
Making too many fields optional creates a large array of places where wallets and anchors can get out of sync.  There's no reason for most of these fields to be optional so lets just require them.

### DIFF
--- a/ecosystem/sep-0024.md
+++ b/ecosystem/sep-0024.md
@@ -548,7 +548,7 @@ Name | Type | Description
 `stellar_transaction_id` | string | transaction_id on Stellar network of the transfer that either completed the deposit or started the withdrawal.
 `external_transaction_id` | string | (optional) ID of transaction on external network that either started the deposit or completed the withdrawal.
 `message` | string | (optional) Human readable explanation of transaction status, if needed.
-`refunded` | boolean | True if the transaction was refunded.
+`refunded` | boolean | True if the transaction was refunded, false otherwise.
 
 
 ### Fields for deposit transactions
--- a/ecosystem/sep-0024.md
+++ b/ecosystem/sep-0024.md
@@ -565,7 +565,7 @@ Name | Type | Description
 Name | Type | Description
 -----|------|------------
 `withdraw_anchor_account` | string | If this is a withdrawal, this is the anchor's Stellar account that the user transferred (or will transfer) their issued asset to.
-`withdraw_memo` | string | Memo used when the user transferred to `withdraw_anchor_account`.  Return null if the withdraw is not ready to receive payment, for example if KYC is not completed.
+`withdraw_memo` | string | Memo used when the user transferred to `withdraw_anchor_account`.  Assigned null if the withdraw is not ready to receive payment, for example if KYC is not completed.
 `withdraw_memo_type` | string | Memo type for `withdraw_memo`.
 `from` | string | Stellar address the assets were withdrawn from
 `to` | string | Sent to address (perhaps BTC, IBAN, or bank account in the case of a withdrawal, Stellar address in the case of a deposit).
--- a/ecosystem/sep-0024.md
+++ b/ecosystem/sep-0024.md
@@ -544,7 +544,7 @@ Name | Type | Description
 `amount_out` | string | Amount sent by anchor to user at end of transaction as a string with up to 7 decimals. Excludes amount converted to XLM to fund account and any external fees.
 `amount_fee` | string | Amount of fee charged by anchor.
 `started_at` | UTC ISO 8601 string | Start date and time of transaction.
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0024.md
```diff
@@ -540,15 +540,15 @@ Name | Type | Description
...
-`amount_fee` | string | (optional) Amount of fee charged by anchor.
-`started_at` | UTC ISO 8601 string | (optional) Start date and time of transaction.
-`completed_at` | UTC ISO 8601 string | (optional) Completion date and time of transaction.
-`stellar_transaction_id` | string | (optional) transaction_id on Stellar network of the transfer that either completed the deposit or started the withdrawal.
+`amount_in` | string | Amount received by anchor at start of transaction as a string with up to 7 decimals. Excludes any fees charged before the anchor received the funds.
+`amount_out` | string | Amount sent by anchor to user at end of transaction as a string with up to 7 decimals. Excludes amount converted to XLM to fund account and any external fees.
+`amount_fee` | string | Amount of fee charged by anchor.
+`started_at` | UTC ISO 8601 string | Start date and time of transaction.
+`completed_at` | UTC ISO 8601 string | Completion date and time of transaction.  Return null for in-progress transactions.
```
leighmcculloch: The `Return` feels a little odd here, like it's documentation for a function call. We're describing a piece of data and can just say the value it has in the edge case, without eluding to the operations the data is used in.
```suggestion
`completed_at` | UTC ISO 8601 string | Completion date and time of transaction.  Assigned null for in-progress transactions.
```

ecosystem/sep-0024.md
```diff
@@ -557,18 +557,18 @@ Name | Type | Description
...
 
 Name | Type | Description
 -----|------|------------
-`withdraw_anchor_account` | string | (optional) If this is a withdrawal, this is the anchor's Stellar account that the user transferred (or will transfer) their issued asset to.
-`withdraw_memo` | string | (optional) Memo used when the user transferred to `withdraw_anchor_account`.
-`withdraw_memo_type` | string | (optional) Memo type for `withdraw_memo`.
-`from` | string | (optional) Stellar address the assets were withdrawn from
-`to` | string | (optional) Sent to address (perhaps BTC, IBAN, or bank account in the case of a withdrawal, Stellar address in the case of a deposit).
+`withdraw_anchor_account` | string | If this is a withdrawal, this is the anchor's Stellar account that the user transferred (or will transfer) their issued asset to.
+`withdraw_memo` | string | Memo used when the user transferred to `withdraw_anchor_account`.  Return null if the withdraw is not ready to receive payment, for example if KYC is not completed.
```
leighmcculloch:
```suggestion
`withdraw_memo` | string | Memo used when the user transferred to `withdraw_anchor_account`.  Assigned null if the withdraw is not ready to receive payment, for example if KYC is not completed.
```

ecosystem/sep-0024.md
```diff
@@ -540,15 +540,15 @@ Name | Type | Description
...
+`amount_out` | string | Amount sent by anchor to user at end of transaction as a string with up to 7 decimals. Excludes amount converted to XLM to fund account and any external fees.
+`amount_fee` | string | Amount of fee charged by anchor.
+`started_at` | UTC ISO 8601 string | Start date and time of transaction.
+`completed_at` | UTC ISO 8601 string | Completion date and time of transaction.  Return null for in-progress transactions.
+`stellar_transaction_id` | string | transaction_id on Stellar network of the transfer that either completed the deposit or started the withdrawal.
 `external_transaction_id` | string | (optional) ID of transaction on external network that either started the deposit or completed the withdrawal.
 `message` | string | (optional) Human readable explanation of transaction status, if needed.
-`refunded` | boolean | (optional) Should be true if the transaction was refunded. Not including this field means the transaction was not refunded.
+`refunded` | boolean | True if the transaction was refunded.
```
leighmcculloch: I think it is worth calling out that this should or shouldn't be set to a value in the non-refunded case. Should it be omitted? Should it be false? Can it be either? If we make it always required and false it will be probably easier to validate implementations.
```suggestion
`refunded` | boolean | True if the transaction was refunded, false otherwise.
```

ecosystem/sep-0024.md
leighmcculloch: Could also drop the word `Assigned` and make it `Null for ...`.

msfeldstein: Thanks yeah, poor wording

ecosystem/sep-0024.md
msfeldstein: Yeah the goal was to always have a value there, will clarify

morleyzhi:
I think `id` from /deposit and /withdraw should be non-optional. And we should clarify that `id` should be equal to the `id` returned by /transaction and /transactions.

PR #428, opened 2019-10-10
Add recommendation for encoding phone numbers in federated addresses

leighmcculloch:
### What
Add recommendation for encoding phone numbers in federated addresses and tidy up the section that defines the requirements of usernames and domains.

### Why
We provide an example of how email addresses can be used in federated addresses which is helpful as it points out to services whose users usernames are their email address that the service can simply use that existing username as the username of the federated address without requiring their users to choose a new username.

Phone numbers are increasingly becoming the user identifiers of many services and it is possible a service will use a phone number as the username.

We should encourage a common uniform approach to encoding phone numbers into usernames using the recommended approaches from industry so that it becomes straight forward to query services for phone numbers without needing to know that some use country codes, others don't, some use separates, etc.

### DIFF
--- a/ecosystem/sep-0002.md
+++ b/ecosystem/sep-0002.md
@@ -33,7 +33,7 @@ For example: `maria@gmail.com*stellar.org`.
 
 #### Phone Numbers
 
-Phone numbers may be used as the username. When encoding a phone number as the username the ITU-T recommendations E.123 and E.164 should be followed. Specifically, the E.123 international notation with a leading `+` as demonstrated in E.123 Section 2.5, with spaces be ommitted, to ensure phone numbers are consistently encoded and do not collide when mixing international phone numbers.
+Phone numbers may be used as the username. When encoding a phone number as the username the ITU-T recommendations E.123 and E.164 should be followed. Specifically, the E.123 international notation with a leading `+` as demonstrated in E.123 Section 2.5 to ensure phone numbers are consistently encoded and do not collide when mixing international phone numbers. Spaces should be omitted.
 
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0002.md
```diff
@@ -20,9 +20,25 @@ For example: `bob*stellar.org`:
...
+
+Email addresses may be used as the username as the `@` symbol is allowed in the username. This allows for using email addresses as the username of an address when the users username is their email address.
+
+For example: `maria@gmail.com*stellar.org`.
+
+#### Phone Numbers
+
+Phone numbers may be used as the username. When encoding a phone number as the username the ITU-T recommendations E.123 and E.164 and should be followed and that spaces be ommitted to ensure phone numbers are consistently encoded and do not collide when mixing international phone numbers.
+
+For example: `+14155550100`.
```
leighmcculloch: The example phone number is part of the reserved set of numbers for fictional use in North America.

leighmcculloch:
Should this maybe be a separate SEP that supplements recommendations for how to encode phone numbers into federated addresses?

ecosystem/sep-0002.md
msfeldstein: should this example have the `*stellar.org` suffix?

msfeldstein:
I don't think it needs to be a separate sep, it doesn't change the SEP it just clarifies usage of it.

ecosystem/sep-0002.md
leighmcculloch: Thanks, fixed in ee77b3a.

PR #431, opened 2019-10-11
add confirms to info

antb123:
This is useful for traders and for people using anchors for payments

### DIFF
--- a/ecosystem/sep-0024.md
+++ b/ecosystem/sep-0024.md
@@ -443,6 +443,7 @@ The JSON object contains an entry for each asset that the anchor supports for SE
 * `enabled`: `true` if SEP-6 deposit for this asset is supported
 * `min_amount`: Optional minimum amount. No limit if not specified.
 * `max_amount`: Optional maximum amount. No limit if not specified.
+* `confirms`: Optional number of confirms before the crypto transation will be released on stellar.
 * `fee_fixed`: Optional fixed (flat) fee for deposit. In units of the deposited asset. Leave blank if there is no fee or the fee schedule is complex.
 * `fee_percent`: Optional percentage fee for deposit. In percentage points. Leave blank if there is no fee or the fee schedule is complex.
 


msfeldstein:
This seems really specific for bitcoin/eth.  I'm not against adding it, but i wonder if there's a way to do it in a more general sense.  I don't want to have super specific fields for all different types of transaction if possible

leighmcculloch:
The concept of a specific number of confirms is unique to blockchain transactions, but the concept of time before funds are available are common to all payment networks. We could generalize the field to refer to the time until the funds will be available, where the unit of time is customizable and can be seconds, minutes, or blockchain confirmations.

```markdown
* `eta`: Optional. Estimated time until the funds will be sent to the address.
* `eta_units`: Optional. Required if `eta` is specified. Units that `eta` is expressed in. Valid values:
   * `seconds`
   * `confirms`
```

msfeldstein:
I'd be cool with the generalization leigh suggests

I also wonder if this belongs in /info or if the `status_eta` field in `/transaction` response would work (adding `eta_units`).

msfeldstein:
I don't really know what having this field in /info would enable as far as usability

leighmcculloch:
@msfeldstein I think adding it to `/info` would serve to prevent it from being a surprise if the time period is a long period of time. Given that people generally expect blockchain powered payments to be faster than real-world, it's good for an anchor to be upfront if their USD to token deposit flow that uses ACH is going to be 5-7 days turn around.

msfeldstein:
Sgtm, one the pr is updated I'll merge

ecosystem/sep-0024.md
```diff
@@ -443,6 +443,7 @@ The JSON object contains an entry for each asset that the anchor supports for SE
 * `enabled`: `true` if SEP-6 deposit for this asset is supported
 * `min_amount`: Optional minimum amount. No limit if not specified.
 * `max_amount`: Optional maximum amount. No limit if not specified.
+* `confirms`: Optional number of confirms before the crypto transation will be released on stellar.
```
kiafer35: qp7yukqy94a70xgrwrqy3n9aymmkrmv59ukdqzaktt

kiafer35: kiafer

ecosystem/sep-0024.md
```diff
@@ -443,6 +443,7 @@ The JSON object contains an entry for each asset that the anchor supports for SE
 * `enabled`: `true` if SEP-6 deposit for this asset is supported
 * `min_amount`: Optional minimum amount. No limit if not specified.
 * `max_amount`: Optional maximum amount. No limit if not specified.
```
kiafer35: qp7yukqy94a70xgrwrqy3n9aymmkrmv59ukdqzaktt

msfeldstein:
@antb123 if you still want this please update to the suggestion leigh mentioned and reopen, closing for now since its inactive

PR #433, opened 2019-10-11
[SEP-24] Stop sending PII in URLs

msfeldstein:
The process of sending PII  as query parameters in a URL are a security risk.  We should be sending them in a secure way to the anchor.

More info on PII Leaking: https://www.owasp.org/index.php/Information_exposure_through_query_strings_in_url

### DIFF
--- a/ecosystem/sep-0024.md
+++ b/ecosystem/sep-0024.md
@@ -83,6 +83,10 @@ SEP-24 lays out many options for how deposit and withdrawal can work. These are 
     * If the `/fee` endpoint is enabled, use it for computing fees when you need to show them to the user.
 * **Authentication**
    * Perform [authentication](#authentication) via SEP-10 before hitting those endpoints
+* **Requested Fields**
+   * If the anchor requests personal information in `/info`, and your app has it or wishes to collect it, send the information to the anchor via [SEP-12](sep-0012.md).
+   * Before sending personal information to an anchor, the user should confirm that they wish to share that information with that anchor, and be given the opportunity to decline, and this step will be skipped.
+   * This step is optional and anchors should be able to continue without any of this information, but it does create a better user experience.
 * **Make a request to `/deposit` or `/withdraw`.**
     * This will respond with the interactive url needed to proceed with KYC and deposit/withdraw details.
 * **For `/deposit`**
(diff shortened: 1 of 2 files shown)


ecosystem/sep-0024.md
```diff
@@ -83,6 +83,10 @@ SEP-24 lays out many options for how deposit and withdrawal can work. These are
     * If the `/fee` endpoint is enabled, use it for computing fees when you need to show them to the user.
 * **Authentication**
    * Perform [authentication](#authentication) via SEP-10 before hitting those endpoints
+* **Requested Fields**
+   * If the anchor requests personal information in `/info`, and your app has it or wishes to collect it, send the information to the anchor via [SEP-12](sep-0012.md).
+   * Before sending personal information to an anchor, the user should confirm that they wish to share that information with that anchor, and be given the opportunity to decline, and this step will be skipped.
```
leighmcculloch: The statement `, and this step will be skipped` is confusing on first read. I don't think that part of the sentence offers much given the next dot point discusses that it can continue without it. Can we drop these few words?

ecosystem/sep-0024.md
```diff
@@ -83,6 +83,10 @@ SEP-24 lays out many options for how deposit and withdrawal can work. These are
...
+   * This step is optional and anchors should be able to continue without any of this information, but it does create a better user experience.
```
leighmcculloch: Is the reason it is optional here because the information could be requested during the interactive flow? If so, can we make a comment here about that, I didn't realize it initially.

ecosystem/sep-0024.md
```diff
@@ -452,7 +459,7 @@ The JSON object contains an entry for each asset that the anchor supports for SE
...
-* `fields`: [SEP-9](sep-0009.md) fields for personal information the anchor requests from the wallet, to be provided to the `/deposit` endpoint.  These fields are all optional, so an anchor should be able to proceed even if the wallet provides nothing, usually by requesting the info during the KYC interactive process.  Each field can optionally have a `description` property which can be used by the wallet to explain to the user why the anchor is requesting it, and give the opportunity to prevent it from being sent.
+* `fields`: Personal information the anchor requests from the wallet.  Allowed fields are defined in [SEP-9](sep-0009.md) and should be submitted to the anchor via [SEP-12](sep-0012.md) before requesting the `/deposit` endpoint.  These fields are all optional, so an anchor should be able to proceed even if the wallet provides nothing, usually by requesting the info during the KYC interactive process.  Each field can optionally have a `description` property which can be used by the wallet to explain to the user why the anchor is requesting it, and give the opportunity to prevent it from being sent.
```
leighmcculloch: Suggested edit:

>so an anchor should be able to proceed even if the wallet provides nothing, usually by requesting the info during the KYC interactive process

=>

>so an anchor should be able to proceed even if the wallet provides nothing. Any information not provided via SEP-12 the anchor can request the user to enter manually during the interactive process

ecosystem/sep-0024.md
```diff
@@ -403,7 +407,12 @@ The response should be a JSON object like:
       "fee_fixed": 5,
       "fee_percent": 1,
       "min_amount": 0.1,
-      "max_amount": 1000
+      "max_amount": 1000,
+      "fields": {
+        "email_address": {
+          "description": "Your email address will be used for for transaction status updates."
```
leighmcculloch:
```suggestion
          "description": "Email address for transaction status updates."
```
Start with the name of the field, so if it is displayed alongside a field it describes the field starting with its name. Also briefer will make it easier for including in more apps.

ecosystem/sep-0024.md
```diff
@@ -83,11 +83,12 @@ SEP-24 lays out many options for how deposit and withdrawal can work. These are
     * If the `/fee` endpoint is enabled, use it for computing fees when you need to show them to the user.
 * **Authentication**
    * Perform [authentication](#authentication) via SEP-10 before hitting those endpoints
+* **Requested Fields**
+   * If the anchor requests personal information in `/info`, and your app has it or wishes to collect it, send the information to the anchor via [SEP-12](sep-0012.md).
```
leighmcculloch: SEP-12 is currently deprecated. We need to un-deprecate it if we're going to recommend its use.

ecosystem/sep-0024.md
bancha3352: sep-0024.md

msfeldstein:
Looking into a cleaner approach in https://github.com/stellar/stellar-protocol/pull/438

msfeldstein:
Closing in favor of https://github.com/stellar/stellar-protocol/pull/438

PR #439, opened 2019-10-24
[sep24] Add fee_minimum, clarify fees

### LINKED ISSUE #437: SEP-0024 - Specify how to interpret both fee_percent & fee_fixed set
The spec is not clear about how to interpret the presence of both `fee_fixed` and `fee_percent` in the response of the `/info` endpoint.

I always assumed (and we implemented it like this in the Solar wallet at the time) that it means you have to pay the fixed fee *plus* the percentual fee.

Now we realized that anchorusd.com say in their docs that they charge "2% of transaction size, with a $5 minimum fee" on withdrawal, while returning a `"USD": { "enabled": true, "fee_fixed": 5, "fee_percent": 2, ... }`.

msfeldstein:
It was unclear if fee_fixed was additive with fee_percent, or if it was a minimum.  This adds `fee_minimum,` and clarifies that fixed+percent are additive.

Fixes #437

### DIFF
--- a/ecosystem/sep-0024.md
+++ b/ecosystem/sep-0024.md
@@ -81,7 +81,7 @@ SEP-24 lays out many options for how deposit and withdrawal can work. These are 
 * For each anchor, use information from its [`stellar.toml`](sep-0001.md) file and its [`/info`](#info) endpoint to display useful information to the user about the asset they've picked.
 * Provide a UI that allows users to pick an asset, anchor, and amount to use for deposit or withdraw. The UI should display the asset's fee structure (if possible) as well as information such as the address of the anchor and description of the asset from the `stellar.toml` file.
 * **Use the `/info` endpoint**
-    * Fetch the asset's deposit & withdawal fee structure: if `fee_fixed` and `fee_percent` are provided, show this to the user early in the process so they're fully informed.
+    * Fetch the asset's deposit & withdawal fee structure: if `fee_fixed`, `fee_percent`, or `fee_minimum` are provided, show this to the user early in the process so they're fully informed.
     * If the `/fee` endpoint is enabled, use it for computing fees when you need to show them to the user.
 * **Authentication**
    * Perform [authentication](#authentication) via SEP-10 before hitting those endpoints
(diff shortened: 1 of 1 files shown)


istrau2:
@msfeldstein so, if I am withdrawing 1 ABC with a fee_percent of 5 and fee_fixed of 0.01, does that mean:
`fee = 0.01 + 1 * 0.05 = 0.06`
or
`fee = 0.01 + (1 - 0.01) * 0.05 = 0.0595`

I always assumed the first...

msfeldstein:
i assume the first, though that is an interesting accounting question.  I don't think i've seen the latter used, where you remove the fixed_fee from the amount you're paying the percent on

msfeldstein:
i can clarify that

istrau2:
@msfeldstein Yeah, Idk. In principal, its just kind of funny to charge the fee_percent on the fee_fixed as well, kind like you are paying a fee on your fee... but I think that's how fee structures typically work, its definitely easier that way.

leighmcculloch:
I would expect the first where the transaction amount is included in the percent fee, with the fixed fee added on after. In mixed fixed and percent fee schedules I've seen in the wild the amount wasn't reduced by any part of the fee, and the fixed fee wasn't added before the percent.

i.e. `(amount * percent fee) + fixed fee = total fee`

msfeldstein:
Updated to clarify that formula.  I've also never seen it any other way but better to be clear.

ecosystem/sep-0024.md
```diff
@@ -445,16 +446,20 @@ The JSON object contains an entry for each asset that the anchor supports for SE
...
 * `enabled`: `true` if SEP-6 withdrawal for this asset is supported
 * `min_amount`: Optional minimum amount. No limit if not specified.
 * `max_amount`: Optional maximum amount. No limit if not specified.
-* `fee_fixed`: Optional fixed (flat) fee for withdraw. In units of the deposited asset. Leave blank if there is no fee or the fee schedule is complex.
-* `fee_percent`: Optional percentage fee for withdraw. In percentage points. Leave blank if there is no fee or the fee schedule is complex.
+* `fee_fixed`: Optional fixed (base) fee for withdraw. In units of the withdrawn asset.  This is in addition to any `fee_percent`.
+* `fee_percent`: Optional percentage fee for withdraw in percentage points. This is in addition to any `fee_fixed`.
+* `fee_minimum`: Optional minimum fee in units of the withdrawn asset.
+
+If fee_fixed or fee_percent is provided, fees are calculated as `(amount * percent fee) + fixed fee = total fee`.  If the fee structure doesn't fit this model, leave them blank and provide the [`/fee`](#fee) endpoint instead.
```
leighmcculloch: Suggested rewording so that all mentions of percent and fixed match the names already used. Adds backticks around named fees also like other mentions of them.
```suggestion
If `fee_fixed` or `fee_percent` are provided, the total fee is calculated as `(amount * fee_percent) + fee_fixed = fee_total`.  If the fee structure doesn't fit this model, leave them blank and provide the [`/fee`](#fee) endpoint instead.
```

leighmcculloch: What does `leave them blank` mean? Should they be `null`, or should the fields not be provided, or does it mean blank strings (`""`)?

msfeldstein: Updated to omit

PR #441, opened 2019-10-29
SEP-0007: add support for mutable URI requests

nikhilsaraf:
SEP-0007 should support mutable requests using a new `replace` param for the `tx` operation type. This will follow the semantics of the [Txrep (SEP-0011)](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0011.md) specification to easily reference and identify parts of the `xdr` that needs to be replaced.


nikhilsaraf:
merging this in for now @patrickxb and @pselden 
can make further changes if there are suggestions.

pselden:
The "hint" part does introduce some possibility of it being used for scams/phishing - EX: "You won 10000! Sign this transaction to claim it!" but i'm not sure how to avoid that since otherwise how would it be used.

Some questions that it may help to clarify in the specification itself:

1. There's no guidance on what to use for the parts to be replaced in the initial transaction -- I assume that means it can be anything?
2. Are hints required for all parts (looks like it, since that's really what "defines" what variables need to be replaced.
3. Is there a maximum length of hint?
4. Also when parsing it seems like the simplest way to parse this is with a bunch of string splits on the separator characters. Unfortunately when you do it that way, you cannot have any of the separator characters in your hint text or it will fail in different ways. Here's a codepen that shows this: https://codepen.io/pselden4/pen/dyyJyxQ?editors=0012. So we probably need to explicitly disallow those control characters in the hint text. And if we do that, we should probably pick some less common control characters, ESPECIALLY the comma, since that's very likely to be used. There's some consideration to use something like JSON to directly encode it in the URL, but then obviously it greatly increases the URL length.


Some other thoughts:

It seems difficult to design an interface that can present this intelligently without having the user just blindly enter replacements for X and Y. Having it so generic presents some challenges and it'll be interesting to see how wallets will handle it.

nikhilsaraf:
@pseldon that is great feedback. Yes, we should find better delimiters. Will comment in more depth on this next week.

nikhilsaraf:
@pselden (fyi @andywer @pitsevich @patrickxb)

I'll introduce a new PR with the changes once we finalize based on this discussion, see responses below

----

> The "hint" part does introduce some possibility of it being used for scams/phishing - EX: "You won 10000! Sign this transaction to claim it!" but i'm not sure how to avoid that since otherwise how would it be used.

Yes this is a risk as it's another avenue for scammers to send messages to users. The wallet should provide enough information so any such scam messages are made clear to the user, for example, the wallet can specifically warn the user about these messages and to beware of scam messages every time it takes user input. Until we fid a better solution, I think this should be open-ended to easily instruct the user. 

> There's no guidance on what to use for the parts to be replaced in the initial transaction -- I assume that means it can be anything?

Yes, this can be anything.

> Are hints required for all parts (looks like it, since that's really what "defines" what variables need to be replaced.

Yes

> Is there a maximum length of hint?

No. I'm considering a limit of 1024 characters for each hint, which is very generous but still provides some level of protection against unexpected length-based attacks. thoughts?

> Also when parsing it seems like the simplest way to parse this is with a bunch of string splits on the separator characters. Unfortunately when you do it that way, you cannot have any of the separator characters in your hint text or it will fail in different ways. Here's a codepen that shows this: https://codepen.io/pselden4/pen/dyyJyxQ?editors=0012. So we probably need to explicitly disallow those control characters in the hint text. And if we do that, we should probably pick some less common control characters, ESPECIALLY the comma, since that's very likely to be used. There's some consideration to use something like JSON to directly encode it in the URL, but then obviously it greatly increases the URL length.

You make a very good point. We need a new delimiter. With the current spec there are three disallowed chars: `:`, `,`, `;`. We can reduce this to one character by repeating it for different levels of splitting. Let's use the colon (`:`) since there's little reason to use that in the hint.
The `,` is replaced by `::` and the `;` is replaced by `:::`
So the following example can be transformed like so:
(original example)
```
sourceAccount:X,operations[0].sourceAccount:Y,operations[1].destination:Y;X:account from where you want to pay fees,Y:account that needs the trustline and which will receive the new tokens
```

(updated example)
```
sourceAccount:X::operations[0].sourceAccount:Y::operations[1].destination:Y:::X:account from where you want to pay fees::Y:account that needs the trustline and which will receive the new tokens
```

With this approach, you can first split strings by `:::` and then for each part you can split it by `::` and then finally by `:`.

> Some other thoughts:
> 
> It seems difficult to design an interface that can present this intelligently without having the user just blindly enter replacements for X and Y. Having it so generic presents some challenges and it'll be interesting to see how wallets will handle it.

Wallets should do validity checks on the user input for example, entering an invalid account in a `sourceAccount` field should fail with UI feedback, entering text for an amount field should similarly present UI feedback, entering a very large amount value should present an additional confirmation dialog to the user, and before signing the transaction the values entered manually should be displayed to the user while being clearly highlighted as user entered input along with a confirmation dialog.

pselden:
One other question came up as I was implementing this: is it a requirement that seqNum be present if sourceAccount is present?

nikhilsaraf:
@pselden Two points on this:

a) wallets should implicitly set the sequence number to the latest value (i.e. replace it) if the source account needs to be replaced.
b) if the source account is set then the `seqNum` should only be replaced if it is specified as a `replace` param. i.e. seqNum=0 is valid moving forward.

Note that my suggestion in (b) is different from the currently published version of the SEP and I will need to update the SEP to reflect these two points. The current version of the SEP says:

>  If the source account is set and the sequence number is 0 then the URI handler should replace the sequence number before signing.

However, I intend to update the SEP to reflect my statements above, unless someone raises a concern about this for any reason.
Allowing a seqNum=0 value to be a valid SEP-0007 request will give us the ability (in the future) to repurpose signed transactions that are invalid on the network (because seqNum=0) such as authenticating a user's account to a callback, or any other similar usage.

pselden:
By the way I published version 1.1.0 of https://github.com/stellarguard/stellar-uri with the replace functionality (it uses the published version, not the version suggested in the comments since it's not canon yet).

You can check out [a demo of it here](https://stellarguard.github.io/stellar-uri/demo/) by clicking on the "Sample Tx" URI button and entering some replacements, then clicking "Perform Replacements".

Please do let me know if there are any updates.

PR #444, opened 2019-10-31
Remove confusing sha stuff for sep submission

msfeldstein:
Currently we require some confusing SHA hash generation to submit a SEP proposal which seems overly confusing, so I replace it with just a short title of whats going on.

The important thing is just to have a unique name here, it will be changed to a number as soon as the draft is accepted (not finalized, just accepted as a draft).  I don't want to have barriers to entry to creating proposals


msfeldstein:
yeah i think its good to have some clear instruction as to what it should be, but maybe even taking out the username is good.  just `sep_{title}.md`: `sep_newdeposit.md`

PR #448, opened 2019-10-31
SEP-2: Add updated timestamp and version

leighmcculloch:
### What
Add `Updated` timestamp to the last date SEP-2 was updated, and a `Version` field, setting it to `1.1.0` assuming it has been updated once since its initial release.

### Why
In #428 I added instructions to the SEP for how to use a phone number as a username in a federated address. I missed adding an `Updated` timestamp when I made that change, and I've discovered reviewing other PRs that we do that normally.

I chose version `1.1.0` because there was no version field previously, but it is in `Final` status, so I assume if it hadn't had any major updates prior to now it was version `1.0.0`.

### DIFF
--- a/ecosystem/sep-0002.md
+++ b/ecosystem/sep-0002.md
@@ -6,6 +6,8 @@ Title: Federation protocol
 Author: stellar.org
 Status: Final
 Created: 2017-10-30
+Updated: 2019-10-10
+Version 1.1.0
 ```
 
 ## Simple Summary


satoshi0x:
I have a minor objection to the version number. I think it matters if we're going to be perfect with formatting we're consistent with how things are audited.

> I believe the correct version if we count including the first version of SEP-0002 would be the draft for the DEPOSIT_SERVER proposal which the author indicated was like but different from the Federation servers on the date of his draft. The draft was the first commit here and dated 2018- 08-21, but subsequent commits basically deleted all of the code written here by the first commit and changed everything to "Federation protocol" and reverted most if not all the important parts of the draft. He also backdated his edits to 2017-10-30 without and reference as to where those were, but that would be we either went back a version or we are now on the third. The status he put for his Federation protocol was "Accepted"... if you're calling yours final (and sorry I'm late but this is a buggy buggy S-Ecosystem Protocol that most web apps etc alike aren't consistent enough with) - then are you not version 1.0.0 OR are you in fact version 1.2.0? I guess that's up to consensus but I don't see anyone auditing these like me.

I left that as a comment in (#448). Aside from that my problem is the "Jed McCaleb" account that edited the first Draft entirely deleting sections for the "DEPOSIT_SERVER" proposal made a subsequent amount of edits relating to the Federation protocol he replaced all of the Draft proposed code with using a different "jedmccaleb" account (one that you can see on his account page on GitHub). The first one is not clickable, no avatar to ID, I could careless if he is the same person I have to question it just like I would when you see someone using the handle "satoshi" or "Satoshi Nakamoto".... yes?

So for version numbers. Lets just say we count all three, or even just the first two but yours as the latest, but first final so 1.0.0. The draft should be 0.x.x in my opinion and the second changes that were "Approved and Reformatted" could be 0.x+1.1 (as in a minor change).  Even though we have no clue as to who both accounts are - if they're the same or not. If it was a different person making the change under the Jed McCaleb and jedmccaleb account I would vote 0.x+2.0 for the last change before the final version you published. So something like 0.3.0 before you and yours being 1.0.0.

PR #454, opened 2019-11-12
[SEP26] Non-interactive updated portion of SEP6

msfeldstein:
Similar to breaking out the interactive portions of sep6 we also need to create something for the non-interactive portions.  This takes just the non-interactive portions of sep6, updates it with some of the improvements we made in sep24, and creates a standalone non-interactive transfer server proposal.

Changes from sep6:
Removed all references to interactive flow
Changed `/deposit` and `/withdraw` URL endpoints so any legacy implementations can run side by side
Changed `/deposit` and `/withdraw` from GET to POST so we can securely pass parameters
Require SEP10

Changes from SEP6->26 here: https://github.com/stellar/stellar-protocol/commit/fdd4aea85c09751d369bfe8ffdafaea6a4ddfa5d


msfeldstein:
no longer necessary as sep6 has been updated to deprecate the interactive portion

PR #455, opened 2019-11-12
[SEP24] Remove the 403 forbidden stuff from the interactive endpoint

msfeldstein:
One of the vestigial remains of the evolution of SEP6 is that the call to return the interactive URL should return as a `403 forbidden`.  This made sense when sep6 could be interactive or non-interactive and there were multiple ways to get through the flow, 403 was the closest thing to "More user info needed".  Given that this is now the one and only happy path we should probably cut the cord on this 403 response, and just show it as the correct response.

It would be good to land this at the same time as https://github.com/stellar/stellar-protocol/pull/438 since its already a breaking change regarding these endpoints, we can do it all at once.

### DIFF
--- a/ecosystem/sep-0024.md
+++ b/ecosystem/sep-0024.md
@@ -253,9 +253,9 @@ Example:
 
 ## Deposit and Withdraw shared responses
 
-### 3. Interactive customer information needed
+### 2. Interactive customer information needed
 
-Response code: `403 Forbidden`
+Response code: `200 OK`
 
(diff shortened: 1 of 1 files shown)


istrau2:
To be clear, this would change on the new SEP24 endpoint. However on the proposed SEP26 endpoint, it would have to remain as 403, correct?

msfeldstein:
I will update that one too, unless you think there is a reason to keep it.

msfeldstein:
Sorry that is incorrect.  The endpoints in sep26 that are 403 make sense as 403, so i will leave them.

istrau2:
Yeah, agreed, on SEP26 I think it should remain as 403.

istrau2:
hmm what about SEP24 withdawal? Doesn't that need a 403 response (because 200 means anchor has everything it needs)?

msfeldstein:
I feel like sep24 withdraw should always go through the interactive flow if for nothing else than to show the confirmation of exactly whats going to happen (exact fees, exact withdraw account details, etc).  Those things are always so anchor specific im not sure its a good UX to just press withdraw then confirm

istrau2:
+1 get rid of 403 entirely in the interactive workflow.

PR #459, opened 2019-11-18
[SEP24] Don't recommend adding jwt to interactive url

### LINKED ISSUE #457: SEP-24 passing the JWT into the interactive url
This issue relates to the following part of SEP-24:

> Since the user has authenticated with SEP-0010, the wallet should pass it into the interactive URL... If the JWT is valid: The anchor must honor the user's authenticated session. The anchor should respond with a 302 redirect response that sets a cookie authenticating the user and redirects the user to a different URL that does not contain the jwt parameter. This is to avoid the JWT appearing in the user's browser history. After the redirect, the authenticated interactive flow must continue as usual.

The way I read this, the SEP is telling wallets that they should append the JWT as a `&jwt=<JWT>` query parameter on the interactive url that is returned by the anchor. Then, the anchor should parse the JWT and redirect 302 to prevent the JWT being included in the browser history.

I understand the motivation here, however I disagree with the approach:

- If the anchor wants the JWT included in the url, it should already have it in there (remember, the JWT was already passed to the anchor in the `/deposit` or `/withdraw` endpoint in which the anchor is forming the interactive url). The wallet shouldn't need to append the JWT to the url.

- Ultimately, it is the anchor's responsibility to implement security around this workflow. The anchor is forming the redirect url as well as producing the page that is redirected to. I think this SEP can suggest a way for anchors to deal with the security of sending the JWT (it was helpful to me), but probably not mandate a certain approach. I can think of many ways of dealing with the issue of the JWT appearing in browser history including using a 302 redirect on the regular sep10 JWT as suggested, producing a special one-time use (or very short-lived) JWT for this redirect (so that it is harmless if it gets recorded in history) as well as having the redirected page simply replace the browser history. Perhaps there is an anchor who doesn't even need a JWT at all in the redirect (I believe there are anchors that don't use SEP10 at all currently).

Instead of mandating that the wallet pass the JWT in the redirect, I think it is better to suggest to anchors a proper way of going about security but ultimately, leaving the implementation up to the anchor.

msfeldstein:
Fixes #457

Having wallets add JWTs to interactive URLs seems like a fishy practice.  If the anchor expects a JWT in the interactive url it can put it there itself.  We should leave the specific security of interactive webapps to the anchors discretion.  Instead we add a small hint as to how to keep continuity from authenticated API calls to fresh interactive flow requests.


yuriescl:
I agree, the Anchor should deal with the security of the URL.
The JWT generated in SEP10 is supposed to be used between the Wallet and the Anchor, the user shouldn't have access to it since the JWT is signed by who owns the account private key. It should be up to the Wallet to decide granting the user access to that private key.
My suggestion is, as an Anchor, when the first `POST /transactions/deposit/interactive` is called from the Wallet, an internal unique ID (`deposit_id`) for that deposit is generated and saved in database. Then, a new JWT token (`one_time_jwt`) is generated, with expiration time of 10 minutes and only allowed to be used once, and its `jti` field contains that `deposit_id`, so it allows retrieving the deposit info:

```python
from datetime import timezone, timedelta
from datetime import datetime as dt

deposit_id = new_deposit_uuid() # unique ID for the deposit
current_time = dt.now(timezone.utc).timestamp()
jwt_dict = {
    "iat": current_time,
    "exp": current_time + timedelta(minutes=10).total_seconds(),
    "jti": deposit_id,
}
one_time_jwt = jwt.encode(jwt_dict, settings.SERVER_JWT_SECRET, algorithm='RS256').decode("utf-8")

return JsonResponse({
  "type": "interactive_customer_info_needed",
  "url" : "https://website.anchor.example.com/deposit-form?token=" + one_time_jwt,
  "id": deposit_id
})
```

This allows confirming the token authenticity and retrieving all the deposit info based on the `deposit_id` in the `jti` field:

```python
jwt_dict = jwt.decode(token, settings.SERVER_JWT_PUBKEY, algorithm='RS256')
current_time = dt.now(timezone.utc).timestamp()
if current_time < jwt_dict["iat"] or current_time > jwt_dict["exp"]:
    raise Exception("invalid token")

# ... check if token was already used, etc ...
```

JakeUrban:
I like the approach @yuriescl outlined. 

However I want to point out that since the new JWT has an expiration, we need to provide wallets the ability to retrieve a new JWT if their original one expires before it is able to start the interactive flow.

I'll file a new PR for SEP-24 that adds a `GET /transactions/<transaction>/interactive` endpoint, which generates a fresh one-time JWT for the wallet to use in the event the one-time JWT the wallet received from `POST /transactions/<transaction>/interactive` expires.

yuriescl:
@JakeUrban That new JWT I suggested would be used in the interactive URL:  
```
{
  "type": "interactive_customer_info_needed",
  "url" : "https://api.example.com/kycflow?token=<new_jwt>",
  "id": <transaction_id>
}
```
That URL will be accessed by the user. It can be used any number of times (never mind the "one-time token" idea). But in an usual scenario, it will be only used once by the popup window. Once the user finishes the KYC interactive flow, `new_jwt` will be invalidated.
The Wallet doesn't need to know `new_jwt` because it can use the existing SEP10 JWT for querying transactions by their ids.

PR #460, opened 2019-11-19
SEP-0026 proposal for Non-interactive Anchor/Wallet Asset Transfer

efritze:


efritze:
As an overview, here's how SEP26 differs from SEP6.

- Got rid of interactive workflow. Already handled by [sep24](sep-0024.md) 
- Got rid of /info endpoint. Asset metadata now found in stellar.toml file.
- Got rid of /fee endpoint. Fee metadata now found in stellar.toml file.
- Now supports multiple deposit options.
- Proposal for i18n of asset metadata
- Anchored assets can now specify if they support funding of unfunded wallets.

msfeldstein:
Looking forward to reading this more in depth tomorrow.  At first glance I'm curious about the authentication protocols, what other authentication protocols do you forsee? And do you think its safe to have non authenticated assets?  Why would you want that?

msfeldstein:
I'm also worried about losing machine-readable fee information, without that the wallets don't have the ability to calculate the correct amount to send

msfeldstein:
I also have [another sep26 proposal](https://github.com/stellar/stellar-protocol/pull/454) out there but if this works better for people im more than happy to abandon mine.

msfeldstein:
One thing im definitely a fan of is the `transfer_protocols = ["sep26"]` addition.  The /info endpoint for sep24 and sep26 is going to clash and it may be tough to support both protocols for one asset

efritze:
> At first glance I'm curious about the authentication protocols, what other authentication protocols do you forsee?

One of our goals with this SEP was to make it future proof.

 Though SEP10 is the only standardized authentication protocol we have right now, this might change in the future. 
 
 We already have the case where an anchored asset may not need any form of authentication. At least not for all transfer endpoints.

There could very well be another standardized authentication protocol for anchors that also require the wallets/clients to be authorized in order to access the anchor's transfer services. i.e. both the account owner and the wallet/client itself need to be authenticated.

There are also many other common authentication protocols that might become SEP standards in the future e.g. proof of location, selfie, email, 2FA, OTP and others.

The beauty about using authentication protocols is I could very well create a non-standard authentication protocol for my own assets without waiting for my custom authentication protocol to  become a standard. Wallets are free to implement my custom authentication protocol, or not. The market will decide.

gdini2003:
For the withdraw request, this proposal is missing dest_extra request parameter. The dest_extra request parameter is essential for fiat anchors that process bank withdrawals. Bank withdrawals require bank account number and bank name at a minimum.
I believe dest_extra request parameter should be included.
dest_extra | string | (optional) Extra information to specify withdrawal location. For crypto it may be a memo in addition to the dest address. It can also be a routing number for a bank, a BIC, or the name of a partner handling the withdrawal.
-- | -- | --

gdini2003:
> I'm also worried about losing machine-readable fee information, without that the wallets don't have the ability to calculate the correct amount to send

I concur

efritze:
> And do you think its safe to have non authenticated assets? Why would you want that?

We definitely recommend authentication for all fiat assets. However there are cases where authenticated transfer endpoints might not be needed at all, and cases where authenticated transfer endpoints might actually be detrimental.

I could very well be an NGO offering cows for impoverished communities using the Stellar blockchain. Or I could very well be a government bureau using the Stellar blockchain to distribute fuel(petrol) bonds to various other public offices. In both cases, transparency might be one of my main goals, and thus I might not even want authentication for any of my asset transfer endpoints. 

Providing authenticated endpoints for other public crypto endpoints does not necessarily enhance privacy. In the case of an Ethereum anchor for example, asset deposits and withdrawals will appear on two public blockchains. The public Stellar blockchain, and the public Ethereum blockchain. It might be trivial to link both just by looking at transactions on both public chains. An anchor may decide it's not worth the hassle to implement an authentication protocol for these sort of assets.

Certain anchored assets might have transfer endpoints that are more sensitive than others. Some fiat anchors might not require authentication to view deposit instructions, but they will most likely need authentication for withdrawals and transaction history.  

There's finally the issue of user experience. Some wallets require you to sign every single transaction. This is especially true for hardware wallets. If your asset doesn't really benefit that much from authenticated transfer endpoints, you may wish to improve user experience by getting rid of SEP10 or any other authentication protocol.

efritze:
> I'm also worried about losing machine-readable fee information, without that the wallets don't have the ability to calculate the correct amount to send

You won't lose any machine-readable fee information in this new SEP26 proposal, and wallets will have the ability to read the exact calculated fees that will be charged without any ambiguity.

There are crazy fee structures out there. Some are fixed, some depend on time of the day, some depend on the weather, some depend on withdrawal options, some depend on location, some depend on amount, some depend on the account owner, some depend on seasons, some depend on priority, some depend on a combination of any of the above, etc.

We are of the opinion that the only way to accurately determine fees is when you are about to perform an actual deposit or withdrawal i.e. when querying the */deposit* and/or */withdraw* endpoints. Both endpoints will now return a field called *fee*, which already contains the **calculated fee** for that particular deposit or withdrawal using whatever crazy fee structure or business logic the anchor implemented. The /deposit and /withdraw endpoints will also have access to any custom fields required to accurately determine the required fee. E.g. a different fee depending on the priority specified for BTC withdrawals. With a calculated fee being returned, we don't need the previous fee_fixed, and fee_percent fields that were returned in SEP6. No ambiguity either. There's only one fee returned: the calculated fee. We also don't need a separate /fees endpoint.

Now, end users may still want to know the fee structure of an anchored asset before deciding to use the anchored asset. That's why we define the fee structure in the stellar.toml file. It is free form (human-readable), and not necessarily machine-readable because it only serves as a guide to the end user to make an informed decision. 

Case study: apay.io . Apay.io used to only offer fixed fees before. Now their model has changed into a combination of percentage fees plus fixed fees. Precisely, their new fee structure for withdrawing BTC is  0.1% + 0.0001 BTC (at the time of this writing). With this new SEP, apay will simply write this same information as their fee structure in their stellar.toml file. Wallets will display this new human-readable fee structure to the end user, giving the end user more data to make an informed decision to use  apay.io or not. 

There are usually two steps involved whenever a user wishes to withdraw: Data entry, and confirmation. In the first screen, the user will see apay.io's BTC withdrawal information has not changed (still 0.1% + 0.0001 BTC).  When the end user enters all withdrawal details including amount, the next screen will show the actual fee that will be charged, after apay has done all its internal fee calculations. Wallets can take into consideration  this  non-ambiguous machine-readable calculated fee, and the end user can decide to confirm the withdrawal or not.

Another case study: XAF anchor which we run. Our withdrawal fees depend on whether you are withdrawing using mobile money, or whether you are withdrawing to a bank account.  In the case of mobile money, withdrawal fees depend on the amount being withdrawn since the mobile operator charges different fees based on the amount being sent. The withdrawal fees also depend on the mobile operator you are withdrawing to. In our Stellar.toml file we will specify a url displaying all these intricate details regarding our fee structure, and the end users can decide if it is worthwhile for them, or not. 


When an end user decides to withdraw XAF, the wallet will show the withdrawal screen where the user can select either a bank transfer withdrawal, or mobile money withdrawal. If the user selects mobile money withdrawal, the wallet will once more show the end user the withdrawal fee structure (which could be a link to an external url ). In the first screen, the user will enter relevant withdrawal details. In the confirmation screen, the calculated fee with all our business logic already applied  will be displayed to the end user. Wallets can take into consideration this  non-ambiguous machine-readable calculated fee, and the end user can decide to confirm the withdrawal or not based on the calculated fee displayed.

efritze:
> For the withdraw request, this proposal is missing dest_extra request parameter. The dest_extra request parameter is essential for fiat anchors that process bank withdrawals. Bank withdrawals require bank account number and bank name at a minimum.

Nothing stops you from specifying all these extra fields in this new SEP26 proposal. In this new SEP26 proposal, you specify extra deposit and withdrawal parameters in the stellar.toml file. 

You can actually give those fields more meaningful names like bank_name, instead of the generic dest_extra name.

Please take a look at the sample stellar.toml file  we did for FIATUSD in the SEP26 proposal.

msfeldstein:
ah yeah, im definitely down with your new proposal for fees then: Human readable up front, and exact calculation at withdraw time.  Wallets might want something stronger up front but i don't think its feasible to really do outside the context of an actual transaction

msfeldstein:
What was the main motivation for moving to a toml solution instead of an /info endpoint?

ecosystem/sep-0026.md
```diff
@@ -0,0 +1,931 @@
...
+
+The deposit endpoint allows a wallet to get deposit information from an anchor, so a user has all the information needed to initiate a deposit. It also lets the anchor specify additional information (if desired) that the user must submit non-interactively via a popup browser window or via [SEP-12](sep-0012.md) to be able to deposit.
+
+If the given account does not exist, or if the account doesn't have a trust line for that specific asset, see the [Special Cases](#special-cases) section below.
+
+
+### Request
+
+```
+POST TRANSFER_SERVER/deposit
```
msfeldstein: This endpoint is list as `GET` on line 407

s-a-y:
in general I'm not a big fan of moving to stellar.toml, because there is currently a 100kb limit, which is very easy to reach when you put multiple assets there.
Otherwise I'm still reading

efritze:
> What was the main motivation for moving to a toml solution instead of an /info endpoint?

One less endpoint to consult. 

Currently on our Interstellar wallet, we need to query two https endpoints whenever a user selects an asset to interact with: the `stellar.toml` endpoint, and the `/info` endpoint. Note that we have to wait for the stellar.toml to load in order to know the right transfer server to use to query the /info endpoint. Note also that the /info endpoint was marked as optional in the previous SEP6 spec, so the /info endpoint might not even exist! That's a wasted http call. We felt we could fix all the above mentioned points with one https call to stellar.toml, and get all the info we need. It should load faster too. (1)

Secondly, TOML is a much better configuration file format than JSON ([opinion1](https://www.lucidchart.com/techblog/2018/07/16/why-json-isnt-a-good-configuration-language/), [opinion2](https://everythingsysadmin.com/2015/06/toml-vs-json.html)). The /info endpoint uses the less human-readable JSON for what is essentially configuration information.

(1) Wallets could cache the two endpoints. I rather cache one endpoint than two.

efritze:
> in general I'm not a big fan of moving to stellar.toml, because there is currently a 100kb limit, which is very easy to reach when you put multiple assets there.

From what we've understood, the 100kb was an arbitrary limit. It was increased from 5kb to 100kb. It can be increased from 100kb to 200kb.

That said, we need not make our new stellar.toml proposal more verbose than necessary. One of the areas we've improved upon to reduce stellar.toml's  filesize in our sep26 proposal is i18n.

Our initial approach to i18n was to place all language localizations in the same stellar.toml file. e.g. title = "Name", title_fr = "Nom", title_es = "Nombre". This could easily inflate the size of stellar.toml, especially for anchors that wish to support as many languages as possible. Not only did this approach increase the size of the stellar.toml configuration file, it also made stellar.toml a little bit more *noisy*.

Our new proposal in this SEP is to append `?lang=languageCode` to the stellar.toml endpoint. 

For example, https://homedomain.com/.well-known/stellar.toml?lang=es should load a stellar.toml configuration with Spanish translations, if the anchor supports Spanish localization. If the anchor doesn't yet support Spanish localization, the stellar.toml configuration should load with the anchor's default language.

msfeldstein:
Making stellar.toml localizable is a pretty big change, and would make people need to support localization both at the transfer server level and now at the stellar.toml level, which is far more complicated than the current static file approach.  I'm not certain i see the value in 1 less network request, especially since the existing stellar.toml is static and can be easily cached.

I also don't really agree on the TOML > JSON point you make here.  The /info endpoint is a configuration file (it's not a file at all).  It's machine generated and machine consumed.  Most of those opinions are about authoring TOML vs JSON, and i agree with those opinions, but you're not authoring anything here.  You can still use TOML as a config file that your server can read from if you want.

msfeldstein:
There are some changes here that i think we likely want to adopt whichever approach we take

- Per Asset Transfer Server URL.  This has been broached before and is an important change
- Future proofing by changing static fields into arrays.  `authentication_protocol`, `transfer_protocol` (existing `/info`  just says if something is available for deposit, not whether its available for sep24 vs sep26)
- Fund Wallet features
- Possibly move towards the 'human readable fee structure with exact fee at deposit time' approach, though i know wallet's like having a machine readable fee structure ahead of time
Is there anything else in here that you cannot do with the `/info` endpoint approach?

efritze:
Discussions regarding the best approach to localize stellar.toml have been moved to https://github.com/stellar/stellar-protocol/issues/464

efritze:
SEP26 no longer uses stellar.toml for extra asset information.

We refactored sep26 to use (SEP27 #474) for custom asset information and localization. 

We also added an interactive workflow for one-time information needed by an anchor .

marcinx:
Hey! Came across this recently. Looks promising!

Two feedback items: 

* We should probably break up the "fee" attribute into fee_fixed and fee_proposal. Currently it is hardcoded as an arbitrary string, e.g. "0.1% + 1 USD", which makes it basically impossible to automatically process and calculate amounts programmatically. I haven't come across any anchors that have a need for non fixed+percentage type of fees but for those we could probably re-introduce the /fee endpoint if absolutely needed. I know this has been discussed above in the comments but I still don't understand how this would work in a fully automated way when arbitrary strings are permitted as `fee`. It is not sufficient to display the fees after /withdrawal or /deposit was invoked. Some services need to calculate and display these beforehand.

* The KYC portion of this draft basically breaks the "non-interactive" promise of this SEP as it calls for graphical user interfaces and interactive web views. Is there a reason to not use [SEP-12](sep-0012.md) along with [SEP-9](sep-0009.md) alongside this SEP? With those, the whole process of depositing/withdrawing and transfering KYC/AML information between a service and anchor could be implemented programmatically in its entirety. Anchors who absolutely need to collect KYC/AML themselves via web views could still do this via [SEP-0024](sep-0024.md). This would enable this draft as a stand-alone programmatic/non-interactive interface that can peacefully co-exist with SEP-24, which would provide a graphical UI/interactive interface for anchor/service interaction.

Thoughts? @msfeldstein @efritze @gdini2003

PR #466, opened 2019-11-22
SEP-0010: allow JWT iss field to be server public key

leighmcculloch:
### What
Allow the JWT generated by the server in SEP-10 to contain an issuer (field: `iss`) that is the public key of the server used in the challenge process, along with the URL.

Example JWT:
```json
{ 
  "iss": "GAHSZCNU4MTXFWND3P2CJIOQRN5PXG5XPAWS3MTLH67NGDDK577UTZKG",
  "sub": "GDOAAGSRBZREJ7UEUUHLJLRH6BNIQU6X763QDF65HSTL2HZH7VVUJXNN",
  "iat": 1574382332,
  "exp": 1574385932,
  "jti": "9ec36c2f0b4e0569e89776cb05c546719dc791fc6d4ac614131d26fc0fe3b124"
}
```

### Why
Non-URL strings are allowed by [rfc7519] with the only requirement being that they do not contain a semi-colon character (`:`). This constraint by SEP-10 to limit the field to a URI seems unnecessary. In most cases the servers that will be validating the JWT are going to be also in the control by the same entity that issued the JWT so it is mostly an implementation detail what is placed in this field.

Not all servers are aware of their absolute URI. Many servers are deployed behind load balancers or frontend web servers and they may not otherwise have a need to know their absolute URI, so this unnecessary requirement will be inconvenient without any gain.

We use public keys to identify servers in other places like the TOML and it would be consistent if we allowed and encourage the issuer to identify itself with its public key everywhere. We should at least expand the SEP to allow for that consistency.

[RFC7519]: https://tools.ietf.org/html/rfc7519#section-4.1.1

### CC
@tomquisel @nebolsin – Original authors.

### DIFF
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -6,8 +6,8 @@ Title: Stellar Web Authentication
 Author: Sergey Nebolsin <sergey@mobius.network>, Tom Quisel <tom.quisel@gmail.com>
 Status: Active
 Created: 2018-07-31
-Updated: 2019-07-09
-Version 1.1.0
+Updated: 2019-11-21
+Version 1.2.0
 ```
 
 ## Simple Summary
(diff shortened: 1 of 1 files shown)


leighmcculloch:
I accidentally opened this PR while I was still only part way through getting my thoughts down in the PR description. I've updated it to completely capture the reasoning behind this change.

tomquisel:
I think this is a good point.

I don't have super strong feelings, but I'd rather leave the `iss` field alone for these reasons though:

* Using a URL is standard, so it's more what devs expect
* By specifying a particular domain or subdomain, it allows the issuer to namespace the JWT in a clear way (maybe they want to allow access to their `anchor.domain.com` site but not `domain.com`)
* By having two possibilities, it's a little harder to reason about compatibility

I'd suggest we add another optional claim field, maybe called `server_account` or `iss_account`.

leighmcculloch:
>Not all servers are aware of their absolute URI.

I realize now that when I typed this I was thinking of the URI as a URL, but it's not. An implementer who wishes to adhere to the SEP but for who it is inconvenient to include the issuing servers URL can define a URI to represent their key on a domain they own, e.g.: `https://jwt.example.com/GAHSZCNU4MTXFWND3P2CJIOQRN5PXG5XPAWS3MTLH67NGDDK577UTZKG`.

---

@tomquisel Responding to your comments:

>Using a URL is standard, so it's more what devs expect

I haven't managed to find evidence of this. Even Google do not use real URLs as their `iss` when issuing JWTs on Firebase. They use a URI to indicate a resource in the form `https://securetoken.google.com/<project-name>`, but the URI isn't a URL to anything meaningful and it isn't the URL of the issuing server.

>By specifying a particular domain or subdomain, it allows the issuer to namespace the JWT in a clear way (maybe they want to allow access to their anchor.domain.com site but not domain.com)

An issuer should be using the [aud] field to scope the use of a JWT rather than the `iss` field.

[aud]: https://tools.ietf.org/html/rfc7519#section-4.1.3

>By having two possibilities, it's a little harder to reason about compatibility

Agreed, I don't think it's worth changing this given it would create a choice someone would need to make, and the result of the choice is relatively unimportant, especially given they can define any URI without it needing to be a URL.

tomquisel:
Great points about `aud` and using a custom URI. Maybe it's worth giving some helpful examples of that approach in the SEP-10 spec?

leighmcculloch:
Followed up here: #468

PR #469, opened 2019-11-23
SEP-0010: Remove the jti as a required token field

leighmcculloch:
### What
Remove the `jti` as a required field on the JWT generated as part of SEP-10.

### Why
The `jti` is described in [RFC7519 Section 4.1.7] as being a unique identifier for a specific token that can be used to prevent a token from being replayed. In some situations it is desirable to only allow a token to be used one-time, even if it hasn't expired. While that might be a use-case of an anchor or other web service, it doesn't seem to be a core requirement of SEP-10. In fact SEP-10 uses a JWT as a way of getting a reusable token from a challenge-response that is not easily reused, so repeatable use of the JWT is anticipated.

Motivation for removing this now is that we are exploring other SEPs that build on SEP-10 and involve augmenting an existing JWT and issuing a new JWT that includes additional claims. Augmenting the JWT requires changing the `jti` since they're intended to be unique and the requirement of a `jti` in this SEP is inconvenient for extending and building on it. Our SEPs will be easier to compose if they hyper-focus on solving a single problem. This change reduces the requirements by removing a component that is not core to the problem solved by this SEP.

An implementer may be using the `jti` to log and improve traceability of where a token was generated and then later used. Implementing systems can still do this if they'd like either with the `jti` field or by specifying a custom claim to indicate transaction hash or any other identifier they have, but it is outside the scope of this SEP to make that a requirement in every implementation.

[RFC7519 Section 4.1.7]: https://tools.ietf.org/html/rfc7519#section-4.1.7

### DIFF
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -6,8 +6,8 @@ Title: Stellar Web Authentication
 Author: Sergey Nebolsin <sergey@mobius.network>, Tom Quisel <tom.quisel@gmail.com>
 Status: Active
 Created: 2018-07-31
-Updated: 2019-07-09
-Version 1.1.0
+Updated: 2019-12-02
+Version 1.2.0
 ```
 
 ## Simple Summary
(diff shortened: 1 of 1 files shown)


leighmcculloch:
@msfeldstein Just wondering why you merged my PR? I was planning on closing it if #489 is accepted, because while I think we can remove it, we no longer have a need to do so because I'm now targeting #489 instead of #472. #472 was driving this change because we were going to be stacking JWTs and the transaction hash as the `jti` would not be unique across them.

Could we let authors merge their PRs?

tomquisel:
Agreed, I think "author merges PR" is a good policy (at least for SDF authors).

leighmcculloch:
I was on the fence on if we should still merge this, but now it's happened I think let's run with it. I think the reasoning in the PR description stand regardless of #472 and it simplifies the SEP. 🎉 🌮 🌯

PR #471, opened 2019-11-23
SEP-0010: Make clear about what it proves and what it doesn't prove

leighmcculloch:
### What
Reduce the scope of SEP-10's promises about what it provides to proving possession of a Stellar secret key, and not possession of a Stellar account.

### Why
SEP-10 claims to provide the mechanics to help a server get proof that a client holds a Stellar account, but if a server was to implement only the functionality described in this document it would only be proving that the client is in possession of the secret key, and not any Stellar account.

Proving possession of a secret key doesn't prove any level of control of an account because an account can alter its signers. The master key for an account may have its weight reduced to zero which removes the master key as a signer, and so possession of a master key is not proof of control of an account.

It's important that if implementers of SEP-10 need to prove control of a Stellar account that they take an extra step that is not currently discussed in the document. That extra step is to verify that the key is a signer on the account and that the weight of that signature is sufficient for whatever operations and tasks the implementer is needing to perform.

This change modifies the SEP to acknowledge its limitations and not to try and solve the problem of how to authentication a Stellar account other than to point out that a server receiving the JWT would need to check with the network to confirm if the authenticated Stellar key has control over an account.

This change is important for supporting applications that do not use the master key as a signer, that rotate the signing key for a single signature account, and for accounts that have multiple signers. Implementers of these applications may read SEP-10 thinking it solves the problem of proving possession of an account when it only proves possession of a key.

### Notes

This change does not address how to use SEP-10 with multiple signers. I have another change #473 that is a work-in-progress to address that.

This change does not address how to use SEP-10 in a standard way to generate a token that could be used as proof of control of a Stellar account. I have another change #472 that is a work-in-progress to address that.

### CC

@tomquisel @nebolsin – Original authors of SEP-10.

### DIFF
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -35,7 +35,7 @@ The flow achieves several things:
 * The client is able to prove their identity using a Ledger or other hardware wallet as well as by having direct access to the secret key
 * The server can chose its own timeout for the user's session
 
-Note: The result of the challenge-response and the JWT generated when implementing this SEP will confirm that an authenticating user has possession of a Stellar secret key but does not prove signing capabilities of any Stellar account. A server receiving the JWT that needs proof of control of a Stellar account should verify if the Stellar key authenticated in the JWT is a signer of the account and has a weight that is suitable for the applications operations and purpose. An application wanting to prove that the signer has complete authority of the account may want to verify that the weight of the signer is equal or greater than all thresholds on the account.
+The result of the challenge-response and the JWT generated when implementing this SEP will confirm that an authenticating user has possession of a Stellar secret key but does not prove signing capabilities of any Stellar account. A server receiving the JWT that needs proof of control of a Stellar account should verify if the Stellar key authenticated in the JWT is a signer of the account and has a weight that is suitable for the applications operations and purpose. An application wanting to prove that the signer has complete authority of the account may want to verify that the weight of the signer is equal or greater than all thresholds on the account.
 
 ## Authentication Endpoint
 
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -69,7 +69,7 @@ Request Parameters:
 
 Name      | Type          | Description
 ----------|---------------|------------
-`account` | `G...` string | Deprecated. Use `key`.
+`account` | `G...` string | Deprecated. Use `key`. Applications should accept and send both parameters until v2.
 `key`     | `G...` string | A Stellar public key that the wallet wishes to authenticate with the server.
 
 Example:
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0010.md
```diff
@@ -22,19 +22,21 @@ The authentication flow is as follows:
...
-* The server can verify that the client holds the secret key to their account
+* The client can verify that the server holds the secret key to a particular public key
+* The server can verify that the client holds the secret key to a particular public key
 * The client is able to prove their identity using a Ledger or other hardware wallet as well as by having direct access to the secret key
 * The server can chose its own timeout for the user's session
 
+The result of the challenge-response and the JWT generated when implementing this SEP will confirm that an authenticating user has possession of a Stellar secret key but does not prove signing capabilities of any Stellar account. A server receiving the JWT that needs proof of control of a Stellar account should verify if the Stellar key authenticated in the JWT is a signer of the account and has a weight that is suitable for the applications operations and purpose. An application wanting to prove that the signer has complete authority of the account may want to verify that the weight of the signer is equal or greater than all thresholds on the account.
```
tomquisel: This is good, maybe rephrase a little more simply?

"A server receiving the JWT that needs proof of control of a Stellar account should verify if the Stellar key authenticated in the JWT is a signer of the account and has a weight that is suitable for the applications operations and purpose."
->
In the case of a server that needs proof of control of a Stellar account, the server should verify that the Stellar key authenticated in the JWT has sufficient signing weight to meet the [threshold](https://www.stellar.org/developers/guides/concepts/multi-sig.html#thresholds) suitable for the application's purposes.

ecosystem/sep-0010.md
```diff
@@ -188,6 +191,12 @@ Every other HTTP status code will be considered an error. For example:
 }
 ```
 
+## Validating control of a Stellar account
+
+Servers that receive a JWT that proves the holder has control of a secret key must take additional steps to verify that the holder of that secret key has control of a Stellar account. Even if the secret key is the master key for an account it may not have control of an account. For operations relating to a Stellar account a server on receipt of a JWT should verify in addition to the JWT's general validity that:
+
+- the public key in the `sub` has signing authority of the Stellar account at a sufficient threshold.
```
tomquisel: Maybe replace authority with `weight` to match the [multi-sig docs](https://www.stellar.org/developers/guides/concepts/multi-sig.html)?

msfeldstein:
this is good, we need to figure out our versioning scheme sooner rather than later, @rice2000 is this something we'd want to talk about at the SEP working group?

rice2000:
@msfeldstein Yes! Versioning is exactly the kind of thing we should figure out in that working group.

leighmcculloch:

PR #472, opened 2019-11-23
SEP-00XX: Add new SEP for proving authorized control of an account

leighmcculloch:
### What
Add a new SEP that provides a process to take a SEP-10 JWT and augment it with claims that capture at a point in time the control the signers in the SEP-10 JWT have over a Stellar account.

### Why
SEP-10 provides proof that a client has possession of the secret keys for one or more signers, but not any proof that the client has partial or full control of a Stellar account. Control of a Stellar account can only be determined by evaluating whether the keys in the SEP-10 JWT are signers, and comparing the weights of the those keys against the thresholds of the account. This process is not included in SEP-10 and implementers of SEP-10 need to perform this process on every request using a SEP-10 JWT.

This new protocol allows a server to take a SEP-10 JWT and determine what access it is authorized on an account and to capture that authorization in a new JWT that can be given a smaller timeout than the SEP-10 JWT. The difference in timeout for each JWT is important because if a client has access to a secret key it is likely they'll have access for sometime, but the signers on an account can change at anytime and a server may wish to more frequently check that the signers of an account are still valid. The SEP-10 JWT essentially becomes a refresh token of the JWT returned by this new protocol.

Assumes #471 is accepted.

### DIFF
--- a/ecosystem/sep-00xx.md
+++ b/ecosystem/sep-00xx.md
@@ -1,0 +1,131 @@
+## Preamble
+
+```
+SEP: 00xx
+Title: Stellar Web Authorization
+Author: Leigh McCulloch <@leighmccullcoh>
+Status: Draft
+Created: 2019-11-22
+Version 1.0.0
+```
+
+## Simple Summary
+
+This SEP defines the standard way for clients such as wallets or exchanges to expand authenticated web sessions created with SEP-10 with additional claims about control over Stellar accounts to facilitate authorization of a specific Stellar account. SEP-10 provides a JWT that proves possession of a Stellar key. This SEP provides additional claims that can be added to the JWT that prove a weight of control of a Stellar account, and a process that a client and server can follow to augment a SEP-10 JWT with those claims.
+
+## Abstract
+
+This protocol is the definition of additional claims an implementer of SEP-10 can include in the JWT that capture at a point in time the weight of control a Stellar key, or multiple keys, has for a given account.
+
+The authorization flow is as follows:
+
+1. The client obtains a [JWT](jwt.io) using SEP-10 proving possession of one or more Stellar keys.
+1. The client calls the server with the SEP-10 JWT requesting authorization of a Stellar account.
+1. The server verifies that the keys in the SEP-10 JWT are signers of the Stellar account.
+1. The server responds with the JWT augmented with additional claims representing the account and weighted control of the keys.
+1. Any future calls to the server can be authenticated by including the JWT as a parameter and a level of control of the account can be authorized by the additional claims.
+
+The flow achieves several things:
+
+* The server can verify that the client holds the secret key(s) that can sign for an account without checking the network on every request containing a SEP-10 JWT
+* The server can chose a timeout for the user's proof that they are a signer of a Stellar account independent of the timeout that they possess their secret keys
+* Since the signers of a Stellar account can change at anytime the server can choose a small timeout of JWTs of this SEP, while keeping a longer timeout of JWTs of SEP-10. A longer lasting SEP-10 JWT can be used to check and refresh proof of control of an account without the user needing to resign transactions.
+
+## Authentication Endpoint
+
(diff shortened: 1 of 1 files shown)


ecosystem/sep-00xx.md
```diff
@@ -0,0 +1,131 @@
+## Preamble
+
+```
+SEP: 00xx
+Title: Stellar Web Authorization
```
msfeldstein: A slightly more descriptive title might be useful, like `Account Control Proofs` or something.

ecosystem/sep-00xx.md
```diff
@@ -0,0 +1,131 @@
...
+
+* The server can verify that the client holds the secret key(s) that can sign for an account without checking the network on every request containing a SEP-10 JWT
+* The server can chose a timeout for the user's proof that they are a signer of a Stellar account independent of the timeout that they possess their secret keys
+* Since the signers of a Stellar account can change at anytime the server can choose a small timeout of JWTs of this SEP, while keeping a longer timeout of JWTs of SEP-10. A longer lasting SEP-10 JWT can be used to check and refresh proof of control of an account without the user needing to resign transactions.
+
+## Authentication Endpoint
+
+A web service indicates that it supports account authorization via this protocol by specifying `WEB_AUTHZ_ENDPOINT` in their [`stellar.toml`](sep-0001.md) file. This is how a wallet knows where to find the authorization server. A web server is required to implement the following behavior for the web authorization endpoint:
+
+* [`POST <WEB_AUTH_ENDPOINT>`](#authorize): exchange a SEP-10 key JWT for an account JWT
```
msfeldstein: s/WEB_AUTH_ENDPOINT/WEB_AUTHZ_ENDPOINT

ecosystem/sep-00xx.md
```diff
@@ -0,0 +1,131 @@
...
+
+To validate the request the server should complete the following steps. If any of the listed steps fail, then the authorization request must be rejected with HTTP Status Code 401. 
+* Verify the SEP-10 JWT is issued from a trusted issuer, is signed by that issuer with an expected signing method, and has not expired. The server should follow all JWT best practices to ensure the JWT can be trusted and that any other application specific claims are present and valid.
+* Verify the Stellar account is an active Stellar account.
+* Verify all the Stellar keys in the SEP-10 JWT `sub` field are signers of the Stellar account.
+
+Upon successful validation service responds with a session JWT, containing the following claims:
+
+* `iss` (the principal that issued a token, [RFC7519, Section 4.1.1](https://tools.ietf.org/html/rfc7519#section-4.1.1)) — a [Uniform Resource Identifier (URI)] for the issuer (`https://example.com` or `https://example.com/G...`)
+* `sub` (the principal that is the subject of the JWT, [RFC7519, Section 4.1.2](https://tools.ietf.org/html/rfc7519#section-4.1.2)) — the public keys of the authenticating client copied from the SEP-10 JWT
```
msfeldstein: Now that `sub` can contain more than 1 key,  we should clarify how they are separated inside the claim (comma vs space)

ecosystem/sep-00xx.md
msfeldstein: also why AUTHZ?

leighmcculloch: This new endpoint will be doing an authorization check. Authorization is sometimes abbreviated to `authz`. Authentication is sometimes abbreviated to `authn` when not wanting the abbreviation to be confused with authorization.

ecosystem/sep-00xx.md
leighmcculloch: That's clarified in #473.

leighmcculloch:
I'm exploring another route separate to this PR that builds this into SEP-10 in a single request.

The only major advantage we get by taking the route discussed in this SEP is it's possible to make the JWTs for account verification have a much smaller timeout than the JWTs that support access to a key. This is actually an important distinction because possession of a key might be considered something that's likely to be more permanent than control over an account given that account control can be changed at anytime. But if that distinction isn't important in practice, then this additional complexity of having two JWTs will be  wasted.

leighmcculloch:

PR #473, opened 2019-11-23
SEP-0010: Add ability to authenticate multiple keys at once

leighmcculloch:
### What
Expand SEP-10 to support a client proving possession of multiple keys in a single challenge-response.

### Why
I'm working on SEP-10 to solve issues with it not providing proof that a user has control of an account, and @tomquisel mentioned that another issue we've had raised in the past is that the SEP also doesn't support proving multiple signers. I've put this draft together to capture how I think we could expand SEP-10 to support multiple signatures but I'm not in a rush to merge this because it's mostly tangential to what I'm focused on at the moment. I mostly want to make sure how I'm writing up #471 and #472 would allow for multiple signatures, and I think they do.

This change is dependent on #471.

### DIFF
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -12,7 +12,7 @@ Version 1.1.1
 
 ## Simple Summary
 
-This SEP defines the standard way for clients such as wallets or exchanges to create authenticated web sessions on behalf of a user who has a Stellar secret key, who therefore may have some control of a Stellar account with that key as a signer. A wallet may want to authenticate with any web service which requires a Stellar account ownership verification, for example, to upload KYC information to an anchor in an authenticated way as described in [SEP-12](sep-0012.md).
+This SEP defines the standard way for clients such as wallets or exchanges to create authenticated web sessions on behalf of a user who has a Stellar secret key, who therefore may have some control of a Stellar account with that key as a signer. A wallet may want to authenticate with any web service which requires a Stellar account ownership verification, for example, to upload KYC information to an anchor in an authenticated way as described in [SEP-12](sep-0012.md). A client may prove possession of multiple keys in a single flow.
 
 ## Abstract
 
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0010.md
```diff
@@ -69,14 +69,20 @@ Request Parameters:
 Name      | Type          | Description
 ----------|---------------|------------
 `account` | `G...` string | Deprecated. Use `key`. Applications should accept and send both parameters until v2.
-`key`     | `G...` string | A Stellar public key that the wallet wishes to authenticate with the server.
+`key`     | `G...` string | A Stellar public key that the wallet wishes to authenticate with the server. May be repeated for multiple keys.
```
msfeldstein: Why repeating parameters vs comma separating a single one?

ecosystem/sep-0010.md
```diff
@@ -125,14 +132,14 @@ To validate the challenge transaction the following steps are performed by the s
 * verify that transaction has time bounds set, and that current time is between the minimum and maximum bounds;
 * verify that transaction contains a single [Manage Data](https://www.stellar.org/developers/guides/concepts/list-of-operations.html#manage-data) operation and it's source account is not null;
 * verify that transaction envelope has a correct signature by server's signing key;
-* verify that transaction envelope has a correct signature by the operation's source account;
+* verify that transaction envelope has a correct signatures by the operations' source account;
```
msfeldstein: "operations' source account**s**" plural

msfeldstein: also remove the 'a' between 'has a correct signatures'

tomerweller:
I think that supporting multisig authentication in a Stellar idiomatic way is not necessarily the same thing as authenticating a batch of keys. (though the latter is probably a super set of the former).

Specifically, in Stellar, multisig is about accounts and cumulative signature weight. 

So the current protocol sounds like: "Hey, I'm in control of all of these keys, let me prove it". 
A "Stellary" protocol might sound like "Hey, I'm in control of this account with medium weight, let me prove it".

So my question is: will we ever want a proof of multiple keys at once outside the context of a Stellar account multisig setup?

leighmcculloch:
@tomerweller I think that's a really good point. This change is just a building block combined with #472 to prove access to an account, it's all for that endgame.

leighmcculloch:

PR #482, opened 2019-12-03
Update CAP-0015 again

jonjove:


core/cap-0015.md
```diff
@@ -67,8 +64,7 @@ enum EnvelopeType
     ENVELOPE_TYPE_TX = 2,
     ENVELOPE_TYPE_AUTH = 3,
     ENVELOPE_TYPE_SCPVALUE = 4,
-    ENVELOPE_TYPE_TX_V1 = 5,
-    ENVELOPE_TYPE_TX_FEE_BUMP = 6
+    ENVELOPE_TYPE_TX_FEE_BUMP = 5
 };
 
 struct TransactionV1Envelope
```
leighmcculloch: Do we still need the `TransactionV1Envelope` if the `TX_V1` envelope type is being removed above? I also see it being used further down in the FeeBumpTransaction too. Does the TransactionV0  and TransactionV0Envelope structs get renamed to not include the versions?

jonjove: We do still need `TransactionV1Envelope` but the name might not be ideal anymore. This proposal makes `TransactionEnvelope` a more generic construct (which I think should retain the same name). The idea here is that `TransactionV0Envelope` can be used to process the existing `TransactionEnvelope` as of protocol version 12, while `TransactionV1Envelope` allows us to recover the extensibility of the existing `Transaction` by reattaching the discriminant to the `sourceAccount`.

The removal of `ENVELOPE_TYPE_TX_V1` is mostly an aesthetic change, allowing us to leverage the existing `ENVELOPE_TYPE_TX` instead.

PR #487, opened 2019-12-04
SEP-28: Add SEP documenting XDR Base64 Encoding

leighmcculloch:
### What
Document how XDR messages are encoded in Base64.

### Why
It appears some folks are removing padding characters from base64 encoded XDR messages and then discovering their messages are not consumable by some other systems. This won't necessarily prevent people from making a mistake, but hopefully give us something to point to when they do and prevent us from going on a repeat search if there are similar issues in the future.

This was discussed in [#1986](https://github.com/stellar/go/issues/1986).


leighmcculloch:
@tomerweller @stanford-scs I think documenting how XDR messages are encoded into base64 can stand alone from the requirements we place on building valid XDR messages.

However, I'm not sure this really belongs as a SEP. Stellar-core accepts base64 encoded XDR messages, so a definition of what is a valid XDR message is less of an ecosystem thing, and more of network thing. The reason this showed up in the ecosystem because in the ecosystem people were optimizing their base64 encoded messages in ways that made them incompatible with all SDKs. Does this belong as a SEP or in some other documentation?

leighmcculloch:
TODO:
- [ ] Add test cases demonstrating:
   - [ ] Variations that are invalid:
       - [ ] No padding on a message size not divisible by 3
       - [ ] Wrong padding char
       - [ ] Wrong alphabet (e.g. URL base64 encoding alphabet)
       - [ ] Line feeds
       - [ ] Line feeds
   - [ ] Variations that are valid:
       - [ ] No padding on a message size divisible by 3
       - [ ] Padding on a message size not divisible by 3
- [ ] Add example for JavaScript

leighmcculloch:
@dm Mentioned about test cases in SEPs, so I added some for this demonstrating valid and invalid encodings of a simple XDR message containing a vector/variable-length-array. The test cases were generated using https://play.golang.org/p/sNjE5d5mD8O.

leighmcculloch:
@stanford-scs Suggested adding copy-pasteable code for the test case byte arrays that are expected. Added in c5eae9e. Generated using https://play.golang.org/p/GXMNEUsBdrU.

stanford-scs:
r+ c5eae9e3f12eca563a8ef30ee48735ca5be0cd19

leighmcculloch:
@rice2000 This new SEP has approvals from relevant folks. I'd like to propose we number this `SEP-28` and give it status `Final` because it documents many implementations already existing (SDKs, Core, Horizon, etc) in the ecosystem.

rice2000:
> @rice2000 @msfeldstein This new SEP has approvals from relevant folks. I'd like to propose we number this `SEP-28` and give it status `Final` because it documents many implementations already existing (SDKs, Core, Horizon, etc) in the ecosystem.

This plan makes sense to me.  I say go for it.

Issue #311, opened 2019-05-25
Authorization Method not specified in SEP-0010


marcinx:
Currently SEP-0010 does not specify how to use the JWT token to authorize subsequent requests, it simply says: "Any future calls to the server can be authenticated by including the JWT as a parameter" in the abstract.

I think there should be a specification of how to use the token when communicating with protected endpoints. As such, I would suggest a header, e.g. `Authorization: Bearer <JWTTOKEN>`, this is what e.g. Stellarport already uses . What would be the formally correct way to suggest such change to the SEP? @tomquisel @nebolsin @istrau2

nebolsin:
@marcinx There're a lot of possible applications of SEP-10 and they could have different authentication/authorization requirements and capabilities (for example, system like Mobius Dapp Store acquires the JWT and then redirects user to the dapp URI with `jwt=` param, I'm not sure it would be possible to set `Authorization` header for traditional browser redirect). 

The idea was to keep SEP-10 as generic as possible and only specify the `client  <-> web_auth_endpoint` interaction necessary to acquire the session token, leaving the `client <-> protected_endpoint` communications outside the scope. How to pass the token to the protected service could be specified by a higher-level SEP (for example, [SEP-12](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0012.md#customer-put) expects it in `jwt` parameter) or it could be an implementation detail of the particular service.

I want to also mention that SEP-10 in this revision has some serious limitations:
- auth with multi-sig account is not supported (see #226 and #240 for details);
- it lacks permission granularity, which is particularly important for "real" multi-sig with a few independent parties as co-signers;
- the original design goal was to support the existing Stellar hardware wallets, so the whole flow is implemented on top of usual Stellar txs and operations, which makes it is not flexible/extendable, required some arbitrary hacks  (like making transactions intentionally invalid) and still needs additional support from the wallets to get the acceptable UX.

I thought about all this for a long time, and I currently think that we need to completely redesign web (off-chain) authentication/authorization using Stellar account, and probably build it on top of established frameworks like OAuth2/OpenID Connect.

tomquisel:
@nebolsin I'm very interested in your thoughts on how to redesign SEP-10. If we're going to do it, we should do it sooner rather than later. How would you support hardware wallets and OAuth2?

marcinx:
@nebolsin Would it make sense then to specify the authentication header in SEP-0006 instead? 

[...]

PR #489, opened 2019-12-05
SEP-10: Add multi-sig capabilities

leighmcculloch:
### What
Add requirements to SEP-10 that a server validate that the client key that has signed the challenge transaction is actually a signer of the account. Add details about how multiple signatures on a challenge transaction are a possibility.

### Why
SEP-10 has the goal to provide the mechanics to help a server get proof that a client holds a Stellar account, but if a server was to implement only the functionality described in this document it would only be proving that the client is in possession of the secret key of the account's address, the master key, and not necessarily the Stellar account. The checks required to go beyond that have until now been left up to the implementer to define.

Proving possession of a secret key doesn't prove any threshold of control of an account because an account can alter its signers. The master key for an account may have its weight reduced to zero which removes the master key as a signer, and so possession of a master key is not proof of control of an account.

It's important that implementers of SEP-10 take an extra step that is not currently discussed in the document. That extra step is to verify that the key is a signer on the account and that the weight of that signature is sufficient from the perspective of the service/anchor for whatever interactions taking place.

It's also possible for accounts to have multiple signers where no single signer will have sufficient weight to meet any threshold. In this case it's important for SEP-10 to allow a combination of signatures to prove control of the account.

This change is important for supporting applications that do not use the master key as a signer, that rotate the signing key for a single signature account, and for accounts that have multiple signers.

### SDK Changes

This change is backwards compatible in terms of how it works, but it will essentially make the verification functions in the SDKs deprecated since those functions assume the signer of the challenge transaction is always the master key.

### CC

@tomquisel @nebolsin – Original authors of SEP-10.
@tomerweller @msfeldstein

### DIFF
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -153,15 +153,6 @@ Upon successful validation service responds with a session JWT, containing the f
 * `iat` (the time at which the JWT was issued [RFC7519, Section 4.1.6](https://tools.ietf.org/html/rfc7519#section-4.1.6)) — current timestamp (`1530644093`)
 * `exp` (the expiration time on or after which the JWT must not be accepted for processing, [RFC7519, Section 4.1.4](https://tools.ietf.org/html/rfc7519#section-4.1.4)) — a server can pick its own expiration period for the token (`1530730493`)
 
-The JWT may contain additional claims to provide information about the state of the account, threshold of control, or signers.
-* `acx` — boolean indicating if the account exists
-* `thr`
-  * `low` — low threshold of the account
-  * `med` — medium threshold of the account
-  * `high` — high threshold of the account
-* `sig` — array of public keys of the signers of the challenge transaction
-* `wht` — weight of the signer(s) that signed the challenge transaction
-
 The JWT may contain other claims specific to their application, see [RFC7519].
 
 [Uniform Resource Identifier (URI)]: https://en.wikipedia.org/wiki/Uniform_Resource_Identifier
@@ -219,14 +210,6 @@ Every other HTTP status code will be considered an error. For example:
 ## JWT Expiration
 
 Servers should select an expiration time for the JWT that is appropriate for the assumptions and risk of the interactions a user can perform with it. A user may be in control of an account at the time the JWT is issued but they may lose control of the account through a change in signers. Expiration times that are too long increase the risk that control on the account has changed. Expiration times that are too short increase the number of times authentication must reoccur, and a user using a hardware signing device or who must complete a complex signing process could have a poor user experience.
-
-## JWT Validation
-
-Services that consume the JWT for authentication and authorization need to validate the JWT. General [JWT best practices](#jwt-best-practices) should be followed for validating the claims within the header and payload.
-
-Services that use the JWT as proof the user meets a threshold of control of a Stellar account should check that the threshold (`thr`) and weight (`wht`) claims are present and that they meet the threshold(s) required by the service. Services may choose to check that the signers in the JWT and their weights still meet requirements the service has by checking the state of the account on the Stellar network. If an account does not exist a server should check that the signer (`sig`) is the account.
-
-Services may optionally use the list of signers (`sig`) as authorization for providing specific signers with access to specific use cases.
 
 ## A convention for signatures
 
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -233,7 +233,7 @@ A server may choose to issue JWTs for less than all thresholds and based on any 
 
 ### Verifying Accounts that Do Not Exist
 
-A server that needs to support validating accounts that do not exist can require a signature of the master key of the account address.
+A server that needs to support validating accounts that do not exist can require a signature of the master key of the account address for accounts that do not exist.
 
 ## JWT Expiration
 
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -153,7 +153,7 @@ Upon successful validation service responds with a session JWT, containing the f
 * `iat` (the time at which the JWT was issued [RFC7519, Section 4.1.6](https://tools.ietf.org/html/rfc7519#section-4.1.6)) — current timestamp (`1530644093`)
 * `exp` (the expiration time on or after which the JWT must not be accepted for processing, [RFC7519, Section 4.1.4](https://tools.ietf.org/html/rfc7519#section-4.1.4)) — a server can pick its own expiration period for the token (`1530730493`)
 
-The JWT may contain other claims specific to their application, see [RFC7519].
+The JWT may contain other claims specific to your application, see [RFC7519].
 
 [Uniform Resource Identifier (URI)]: https://en.wikipedia.org/wiki/Uniform_Resource_Identifier
 [RFC7519]: https://tools.ietf.org/html/rfc7519
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -130,7 +130,7 @@ To validate the challenge transaction the following steps are performed by the s
 * verify that transaction envelope has a correct signature by server's signing key;
 * verify that client signatures on the transactions are signers of the operations source account;
 * verify that client signatures are correct;
-* verify that client signatures provide weight\* that meets the required threshold(s);
+* verify that client signatures provide weight that meets the required threshold(s);
 * use operations's source account to determine the authenticating client and perform any additional service-specific validations.
 
 When determining the weight of the signers the server's signer should be excluded, even if it is a signer on the account. In most cases the server should verify that the signers meet all thresholds of the account by checking that the sum of the weights is equal or greater to the low, mid and high thresholds, and fail to issue a JWT if all thresholds are not met. Some servers may have use cases to issue JWTs for less than all thresholds.
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -133,7 +133,7 @@ To validate the challenge transaction the following steps are performed by the s
 * verify that client signatures provide weight that meets the required threshold(s);
 * use operations's source account to determine the authenticating client and perform any additional service-specific validations.
 
-In most cases the server should verify that the weight of the client signatures is meet all thresholds of the account by checking that the sum of the weights is equal or greater to all the thresholds: low, mid and high. The server should only issue a JWT if all thresholds are met.
+In most cases the server should verify that the weight of the client signatures meets all thresholds of the account by checking that the sum of the weights is equal or greater to all the thresholds: low, mid and high. The server should only issue a JWT if all thresholds are met.
 
 Some servers may have use cases to issue JWTs for less than all thresholds.
 
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -22,11 +22,11 @@ The authentication flow is as follows:
 
 1. The client obtains a unique [`challenge`](#challenge), which is represented as specially formed Stellar transaction
 1. The client verifies that the transaction has an invalid sequence number 0.  This is extremely important to ensure the transaction isn't malicious.
-1. The client signs the transaction using keys that control their Stellar account
+1. The client signs the transaction using the secret key(s) that control the user's Stellar account
 1. The client submits the signed challenge back to the server using [`token`](#token) endpoint
-1. The server looks up the signers for the account
+1. The server gets the signers of the user's account
 1. The server verifies the signatures
-1. The server compares the weight provided by the signatures with the thresholds of the account against a minimum level of control required
+1. The server compares the weight provided by the signers with the thresholds of the account against a minimum level of control required
 1. If the signatures provide the minimum level of control, the server responds with a [JWT](https://jwt.io) that represents the user's session and includes the level of control the user has over the account
 1. Any future calls to the server can be authenticated by including the JWT as a parameter
 
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -206,7 +206,7 @@ Every other HTTP status code will be considered an error. For example:
 
 Services that use the JWT as proof the user meets a threshold of control of a Stellar account need to validate the JWT. General [JWT best practices](#jwt-best-practices) should be followed for validating the claims within the header and payload. If the JWT contains threshold (`thr`) and weight (`wht`) claims the service should check that the threshold required by that service is met.
 
-Older version of SEP-10 did not include the `thr` and `wht` and if the claims are not present the service can assume all thresholds were met.
+Previous version of SEP-10 did not include the `thr` and `wht` and if the claims are not present the service can assume all thresholds were met.
 
 ## A convention for signatures
 
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -22,7 +22,7 @@ The authentication flow is as follows:
 
 1. The client obtains a unique [`challenge`](#challenge), which is represented as specially formed Stellar transaction
 1. The client verifies that the transaction has an invalid sequence number 0.  This is extremely important to ensure the transaction isn't malicious.
-1. The client signs the transaction using the secret key(s) that are signers of the user's Stellar account
+1. The client signs the transaction using the secret key(s) of signers of the user's Stellar account
 1. The client submits the signed challenge back to the server using [`token`](#token) endpoint
 1. The server gets the signers of the user's account
 1. The server verifies the signatures
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -229,7 +229,7 @@ A server that needs to verify a user has complete authority of an account should
 
 ### Verifying Being a Signer
 
-A server may choose to issue JWTs for less than all thresholds and based on any other application specific logic. It's important to keep in mind that a Stellar account may have third-parties who are signers and authenticating users with less than any threshold may allow a third-party to authenticate as a user.
+A server may choose to issue JWTs for less than all thresholds and based on any other application specific logic. It's important to keep in mind that a Stellar account may have third-parties who are signers. Authenticating users with less than any threshold may allow a third-party to authenticate as a user.
 
 ### Verifying Accounts that Do Not Exist
 
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -207,7 +207,7 @@ Every other HTTP status code will be considered an error. For example:
 
 ## JWT Expiration
 
-Servers should select an expiration time for the JWT that is appropriate for the assumptions and risk of the interactions a user can perform with it. A user may be in control of an account at the time the JWT is issued but they may lose control of the account through changes in signers. Expiration times that are too long increase the risk that control on the account has changed. Expiration times that are too short increase the number of times authentication must reoccur, and for a user using a hardware signing device or who need to complete an inconvenient process to sign the challenge transaction could have a poor user experience.
+Servers should select an expiration time for the JWT that is appropriate for the assumptions and risk of the interactions a user can perform with it. A user may be in control of an account at the time the JWT is issued but they may lose control of the account through a change in signers. Expiration times that are too long increase the risk that control on the account has changed. Expiration times that are too short increase the number of times authentication must reoccur, and for a user using a hardware signing device or who need to complete an inconvenient process to sign the challenge transaction could have a poor user experience.
 
 ## JWT Validation
 
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -86,7 +86,7 @@ On success the endpoint must return `200 OK` HTTP status code and a JSON object 
 * `transaction`: an XDR-encoded Stellar transaction with the following:
   * source account set to server's signing account
   * invalid sequence number (set to 0) so the transaction cannot be run on the Stellar network
-  * time bounds: `{min: now(), max: now() + 300 }` (we recommend expiration of 5 minutes to give user time to sign transaction)
+  * time bounds: `{min: now(), max: now() + 900 }` (we recommend expiration of 15 minutes to give user time to sign transaction)
   * operations: `manage_data(source: client_account, key: '<anchor name> auth', value: random_nonce())`
     * The value of key is not important, but can be the name of the anchor followed by `auth`. It can be at most 64 bytes.
     * The value must be 64 bytes long. It contains a 48 byte cryptographic-quality random string encoded using base64 (for a total of 64 bytes after encoding).
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -26,8 +26,8 @@ The authentication flow is as follows:
 1. The client submits the signed challenge back to the server using [`token`](#token) endpoint
 1. The server gets the signers of the user's account
 1. The server verifies the signatures
-1. The server compares the weight provided by the signers with the thresholds of the account against a minimum level of control required
-1. If the signatures provide the minimum level of control, the server responds with a [JWT](https://jwt.io) that represents the user's session and includes the level of control the user has over the account
+1. The server verifies the weight provided by the signers meet the required threshold(s)
+1. The server responds with a [JWT](https://jwt.io) that represents the user's session and includes the weight and thresholds
 1. Any future calls to the server can be authenticated by including the JWT as a parameter
 
 The flow achieves several things:
@@ -37,8 +37,8 @@ The flow achieves several things:
 * The server can verify that the client holds the secret key(s) with control of an account
 * The client is able to prove their identity using a Ledger or other hardware wallet as well as by having direct access to the secret key(s)
 * The server can choose its own timeout for the user's session
-* The server can choose a minimum level of control a user needs of the account
-* Other servers can choose specific levels of control a user needs of the account for different operations
+* The server can choose the required threshold(s) a user needs on the account
+* Other servers can choose the required threshold(s) a user needs on the account for specific functionality
 
 ## Authentication Endpoint
 
@@ -130,10 +130,10 @@ To validate the challenge transaction the following steps are performed by the s
 * verify that transaction envelope has a correct signature by server's signing key;
 * verify that client signatures on the transactions are signers of the operations source account;
 * verify that client signatures are correct;
-* verify that client signatures provide a weight\* respective to their thresholds required to prove a level of control the service needs;
+* verify that client signatures provide weight\* that meets the required threshold(s);
 * use operations's source account to determine the authenticating client and perform any additional service-specific validations.
 
-When determining the weights of the signers the servers signer should be excluded, even if it is a signer on the account. In most cases the server should verify that the signers provide full control of the account by checking that the sum of the weights is equal or greater than all thresholds, low, mid and high.
+When determining the weight of the signers the server's signer should be excluded, even if it is a signer on the account. In most cases the server should verify that the signers meet all thresholds of the account by checking that the sum of the weights is equal or greater the low, mid and high thresholds, and fail to issue a JWT if all thresholds are not met. Some servers may have use cases to issue JWTs for less than all thresholds.
 
 Upon successful validation service responds with a session JWT, containing the following claims:
 
@@ -142,9 +142,6 @@ Upon successful validation service responds with a session JWT, containing the f
 * `iat` (the time at which the JWT was issued [RFC7519, Section 4.1.6](https://tools.ietf.org/html/rfc7519#section-4.1.6)) — current timestamp (`1530644093`)
 * `exp` (the expiration time on or after which the JWT must not be accepted for processing, [RFC7519, Section 4.1.4](https://tools.ietf.org/html/rfc7519#section-4.1.4)) — a server can pick its own expiration period for the token, however 24 hours is recommended (`1530730493`)
 * `jti` (the unique identifier for the JWT, [RFC7519, Section 4.1.7](https://tools.ietf.org/html/rfc7519#section-4.1.7)) — hex-encoded hash of the challenge transaction (`f0e754c6ab988eb5fb2811c2cacc4d3d2aa4334763b8df7285ef341921e2530a`)
-
-For servers that issue JWTs for signers that have less than full control, the JWT should contain additional claims that capture the level of control the user has over the account.
-
 * `thr`
   * `low` — low threshold of the account
   * `med` — medium threshold of the account
@@ -207,11 +204,9 @@ Every other HTTP status code will be considered an error. For example:
 
 ## Validating the JWT
 
-Services that use the JWT as proof of control of a Stellar account need to validate the JWT. General [JWT best practices](#jwt-best-practices) should be followed for validating the claims within the header and payload. In additional services should validate that the weight (`wht`), in respect to the account's thresholds (`thr`), provides a sufficient level of control for the operations the service is allowing the user to perform.
+Services that use the JWT as proof the user meets a threshold of control of a Stellar account need to validate the JWT. General [JWT best practices](#jwt-best-practices) should be followed for validating the claims within the header and payload. If the JWT contains threshold (`thr`) and weight (`wht`) claims the service should check that the threshold required by that service is met.
 
-If the SEP-10 server only issues JWTs if the signing weight is above all account thresholds then no additional validation will be necessary.
-
-If the SEP-10 server issues JWTs with varying levels of control over the account then services consuming the JWT may need to check that the value in `wht` is equal or greater than a value in `thr.low`, `thr.med`, and/or `thr.high`.
+Older version of SEP-10 did not include the `thr` and `wht` and if the claims are not present the service can assume all thresholds were met.
 
 ## A convention for signatures
 
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -26,18 +26,18 @@ The authentication flow is as follows:
 1. The client submits the signed challenge back to the server using [`token`](#token) endpoint
 1. The server gets the signers of the user's account
 1. The server verifies the signatures
-1. The server verifies the weight provided by the signers meets the required threshold(s)
+1. The server verifies the weight provided by the signers meets the required threshold(s), if any
 1. The server responds with a [JWT](https://jwt.io) that represents the user's session
 1. Any future calls to the server can be authenticated by including the JWT as a parameter
 
 The flow achieves several things:
 
 * Both client and server part can be implemented using well-established Stellar libraries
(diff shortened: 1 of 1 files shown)


leighmcculloch:
Note: I didn't add the list of signers to the challenge transaction, or to the initial request. The reason for this is because it's unnecessary and it doesn't reduce the number of requests we have to make to external systems to validate the signers are on the account. At best it would give the implementor the option of moving that check to the first request rather than the second.

leighmcculloch:
Another reason for why is to put the signers in the initial request would assume the caller knows who they are in the first request. It's feasible a wallet won't know what the signers will be until they present the challenge transaction to the user for signing.

ecosystem/sep-0010.md
```diff
@@ -22,18 +22,23 @@ The authentication flow is as follows:
...
-* The client can verify that the server holds the secret key to a particular account
-* The server can verify that the client holds the secret key to their account
-* The client is able to prove their identity using a Ledger or other hardware wallet as well as by having direct access to the secret key
-* The server can chose its own timeout for the user's session
+* The client can verify that the server holds the secret key to a particular public key
+* The server can verify that the client holds the secret key(s) of signers of the user's Stellar account
+* The client is able to prove their identity using a Ledger or other hardware wallet as well as by having direct access to the secret key(s)
+* The server can choose its own timeout for the user's session
+* The server can choose the required threshold(s) a user needs on the account
+* Other servers can choose the required threshold(s) a user needs on the account for specific functionality
```
tomquisel: It may be a little unclear what "Other servers" refers to. I think it's other servers (probably controlled by the same entity as the main server) that also use the JWT to decide whether the user is authorized for their specific functionality.

ecosystem/sep-0010.md
```diff
@@ -120,19 +125,28 @@ To validate the challenge transaction the following steps are performed by the s
...
+* verify that transaction contains a single [Manage Data](https://www.stellar.org/developers/guides/concepts/list-of-operations.html#manage-data) operation and its source account is not null;
 * verify that transaction sequenceNumber is equal to zero;
+* verify that transaction envelope has a correct signature by server's signing key;
+* verify that client signatures on the transactions are signers of the operations source account;
+* verify that client signatures are correct;
+* verify that client signatures provide weight that meets the required threshold(s);
 * use operations's source account to determine the authenticating client and perform any additional service-specific validations.
 
+When determining the weight of the signers the server's signer should be excluded, even if it is a signer on the account. In most cases the server should verify that the signers meet all thresholds of the account by checking that the sum of the weights is equal or greater to the low, mid and high thresholds, and fail to issue a JWT if all thresholds are not met. Some servers may have use cases to issue JWTs for less than all thresholds.
```
tomquisel: Maybe make it clear that the server is verifying that the client signature(s) meet all thresholds. Also, do you need to check all thresholds? That is, is it possible for for high >= medium >= low to be not true?

Also, maybe change "fail to issue a JWT if all thresholds are not met" to "should only issue a JWT if all thresholds are met" (minor nit, avoiding negatives)

ecosystem/sep-0010.md
```diff
@@ -120,19 +125,28 @@ To validate the challenge transaction the following steps are performed by the s
...
 * `sub` (the principal that is the subject of the JWT, [RFC7519, Section 4.1.2](https://tools.ietf.org/html/rfc7519#section-4.1.2)) — the public key of the authenticating Stellar account (`G...`)
 * `iat` (the time at which the JWT was issued [RFC7519, Section 4.1.6](https://tools.ietf.org/html/rfc7519#section-4.1.6)) — current timestamp (`1530644093`)
 * `exp` (the expiration time on or after which the JWT must not be accepted for processing, [RFC7519, Section 4.1.4](https://tools.ietf.org/html/rfc7519#section-4.1.4)) — a server can pick its own expiration period for the token, however 24 hours is recommended (`1530730493`)
 * `jti` (the unique identifier for the JWT, [RFC7519, Section 4.1.7](https://tools.ietf.org/html/rfc7519#section-4.1.7)) — hex-encoded hash of the challenge transaction (`f0e754c6ab988eb5fb2811c2cacc4d3d2aa4334763b8df7285ef341921e2530a`)
+* `thr`
+  * `low` — low threshold of the account
+  * `med` — medium threshold of the account
+  * `high` — high threshold of the account
+* `wht` — weight of the signers who signed the challenge transaction, excluding the weight of the server signer if it is also a signer of the account
```
tomquisel: Maybe just phrase it as "weight of the client signer(s)? That seems more clear than talking about excluding the server signature.

leighmcculloch: My concern here is an edge case where SEP-10 is being used by a service that offers signing. It's feasible the servers signing key will be a signer. It's also feasible someone will implement this calculation by iterating over the signers of the account and adding their weights if the key is included as a signer of the challenge transactions. It's subtle that in this situation the server is helping the user authenticate themselves. Maybe I could take this statement out of this section though given that I mention it in the paragraph added above.

leighmcculloch: e1e07e7

ecosystem/sep-0010.md
leighmcculloch: >do you need to check all thresholds? That is, is it possible for for high >= medium >= low to be not true?

It's definitely possible to configure a Stellar account with low, medium and high not in the order you described. I created an example here that does it and it works: https://repl.it/repls/UsableBurlyDesigns.

I've improved the language as you suggested in f00d22a.

ecosystem/sep-0010.md
leighmcculloch: I've attempted to provide more detail to who those servers are in 4f06a28 without spending too many words on it.

ecosystem/sep-0010.md
```diff
@@ -188,6 +206,12 @@ Every other HTTP status code will be considered an error. For example:
 }
 ```
 
+## Validating the JWT
+
+Services that use the JWT as proof the user meets a threshold of control of a Stellar account need to validate the JWT. General [JWT best practices](#jwt-best-practices) should be followed for validating the claims within the header and payload. If the JWT contains threshold (`thr`) and weight (`wht`) claims the service should check that the threshold required by that service is met.
+
+Some versions of SEP-10 did not include the `thr` and `wht` and if the claims are not present the service can assume all thresholds were met.
```
msfeldstein: why do we want to assume all thresholds were met, vs no thresholds were met?

leighmcculloch: Because that's what the SEP provides today. An implementer who is validating that the user has complete control of the account using the SEP as it currently exists is issuing JWTs without these claims, but the JWT is still claim to an account.

For an implementer who is not validating that the user has complete control of the account though this could be an issue. So maybe I should remove this statement all together and not recommend what an implementer should do. An implementer will usually be in control in the producer and consumer of a JWT, so I don't think we need to worry about cross-service compatibility.

I've removed the statement in e83f645.

ecosystem/sep-0010.md
```diff
@@ -187,6 +206,20 @@ Every other HTTP status code will be considered an error. For example:
 }
 ```
 
+## JWT Expiration
+
+Servers should select an expiration time for the JWT that is appropriate for the assumptions and risk of the interactions a user can perform with it. A user may be in control of an account at the time the JWT is issued but they may lose control of the account through changes in signers.
```
leighmcculloch: @msfeldstein @tomquisel I added a statement about expiration and removed the 24 hour recommendation we had before. I don't think this is something we should define for services. We might want to define it for specific types of services, like an anchor, but not sure we should do that generically. What do you think?

ecosystem/sep-0010.md
```diff
@@ -121,17 +126,31 @@ To validate the challenge transaction the following steps are performed by the s
...
 * `sub` (the principal that is the subject of the JWT, [RFC7519, Section 4.1.2](https://tools.ietf.org/html/rfc7519#section-4.1.2)) — the public key of the authenticating Stellar account (`G...`)
 * `iat` (the time at which the JWT was issued [RFC7519, Section 4.1.6](https://tools.ietf.org/html/rfc7519#section-4.1.6)) — current timestamp (`1530644093`)
-* `exp` (the expiration time on or after which the JWT must not be accepted for processing, [RFC7519, Section 4.1.4](https://tools.ietf.org/html/rfc7519#section-4.1.4)) — a server can pick its own expiration period for the token, however 24 hours is recommended (`1530730493`)
+* `exp` (the expiration time on or after which the JWT must not be accepted for processing, [RFC7519, Section 4.1.4](https://tools.ietf.org/html/rfc7519#section-4.1.4)) — a server can pick its own expiration period for the token (`1530730493`)
+* `jti` (the unique identifier for the JWT, [RFC7519, Section 4.1.7](https://tools.ietf.org/html/rfc7519#section-4.1.7)) — hex-encoded hash of the challenge transaction (`f0e754c6ab988eb5fb2811c2cacc4d3d2aa4334763b8df7285ef341921e2530a`)
```
msfeldstein: did you mean to put the JTI back?  I merged the JTI removal PR so make sure you want this here

ecosystem/sep-0010.md
msfeldstein: yeah i think this is right.  I don't think we can codify something that's this subjective.

ecosystem/sep-0010.md
leighmcculloch: Whoops! Nice catch. That's a merge conflict mistake on my part. Fixed in 7775d48. Thank you!

ecosystem/sep-0010.md
tomquisel: This makes sense to me. I'd also mention that a user may have to go through an inconvenient process to get a JWT (say use a Ledger Nano or use StellarGuard), so too short an expiration time could lead to poor UX.

leighmcculloch: @tomquisel I've added words about that in e1800e4c3d6b46e006e88922abfd961e73c73bb2.

nebolsin:
Personally, I'm not a fan of mixing authentication with authorization for SEP-10, because it reduces the flexibility and narrows down the SEP-10 scope to the particular use cases. There've been multiple previous discussions on this topic #226, #240, #311 where I explained my concerns about this approach.

In particular, I disagree that SEP-10 should be about solving the issue of account authentication, not key authentication. There're perfectly valid use cases for authentication with account which doesn't (yet) exist on the network (ex. exchange requires authentication with withdrawal address to protect its users from scam or to comply with regulations).

I also think that suggested threshold-based scheme is a bit limited and lacks a few important properties to be a safe authorization scheme, namely permission granularity (authorize smb. to provide KYC info for account, but don't allow other actions) and intended audience specification (authorize smb. to service A, but not service B). It's a sort of an invitation for off-chain services to build the security model after Stellar's on-chain low/med/high access levels, which I wouldn't consider a good practice.

leighmcculloch:
>I'm not a fan of mixing authentication with authorization for SEP-10

That was my thinking with #471, #472, #473 that keeps the authentication and authorization independent. However authentication and authorization are already bundled with Stellar accounts. On the network you can't prove authentication without also proving some level of control, or at least that you are a signer. This SEP already claims to perform authorization because it claims to prove that a user controls an account. My concern of going down the road of splitting out authorization separately is that the complexity is significant and would be backwards incompatible with the claims SEP-10 makes today.

>valid use cases for authentication with account which doesn't (yet) exist

That is a really good point, I hadn't considered this case. [SEP-24] supports that scenario and relies on SEP-10 and these changes would prevent an anchor from authenticating a key that does not yet have an account, so that is a blocker to this change as it stands.

I think this change could be modified to support this scenario by checking that the challenge transaction is signed by the master key and not providing threshold and weight claims in the case that the account does not exist. We could add an additional claim indicating if the account exists or not. We could also include the list of signers in the JWT as @tomquisel did in #240 to make it clear that it was the master key in this case that had been used but I was hesitant to include them because it massively increases the size of the JWT. I could see by including them it would provide greater flexibility to downstream systems layering authorization decisions on-top of signer and account pairs, rather than weights and thresholds.

Maybe the right approach is to provide all these additional claims with recommendation that implementers only include the claims they will use in their specific application?

In regards to roles and permission granularity I don't think SEP-10 can describe roles beyond the thresholds or signers defined on a Stellar account. If the SEP is to be generic it can only capture claims about a key or account as it exists or does-not-yet-exist on the network. An application can absolutely include its own claims like `role` or another SEP like SEP-24 can require the use of additional claims relating to KYC, and the SEP already contains a statement about this.

[SEP-24]: https://github.com/stellar/stellar-protocol/blob/353329faf6ace26b4db6d58c16bed58f23c34dc8/ecosystem/sep-0024.md

leighmcculloch:
@nebolsin I've pushed an amendment in be72fcc that addresses authenticating with a Stellar account that does not yet exist, and tried to do a better job of highlighting areas where an implementer may want to inject application specific logic to address the cases you mentioned.

I think the most important part of SEP-10 is its claim that the user controls the account, and the interface between the client and the SEP-10 server. The exact contents of the JWT is less critical because the consumers of the JWT are likely to be hosted by the same people and an implementer could use an application specific JWT structure and still be interoperable with wallets. An implementer can add roles and other claims to the JWT that specifically address their scenario.

Because of this I added a statement that an implementer can omit any claim from the JWT. I think it's still important we demonstrate how to use thresholds and weights as claims so that an implementer wanting to prove control of an account, which is the core purpose of the SEP, is setup to succeed at that.

ecosystem/sep-0010.md
```diff
@@ -128,31 +129,37 @@ To validate the challenge transaction the following steps are performed by the s
...
 
 * `iss` (the principal that issued a token, [RFC7519, Section 4.1.1](https://tools.ietf.org/html/rfc7519#section-4.1.1)) — a [Uniform Resource Identifier (URI)] for the issuer (`https://example.com` or `https://example.com/G...`)
 * `sub` (the principal that is the subject of the JWT, [RFC7519, Section 4.1.2](https://tools.ietf.org/html/rfc7519#section-4.1.2)) — the public key of the authenticating Stellar account (`G...`)
 * `iat` (the time at which the JWT was issued [RFC7519, Section 4.1.6](https://tools.ietf.org/html/rfc7519#section-4.1.6)) — current timestamp (`1530644093`)
 * `exp` (the expiration time on or after which the JWT must not be accepted for processing, [RFC7519, Section 4.1.4](https://tools.ietf.org/html/rfc7519#section-4.1.4)) — a server can pick its own expiration period for the token (`1530730493`)
-* `jti` (the unique identifier for the JWT, [RFC7519, Section 4.1.7](https://tools.ietf.org/html/rfc7519#section-4.1.7)) — hex-encoded hash of the challenge transaction (`f0e754c6ab988eb5fb2811c2cacc4d3d2aa4334763b8df7285ef341921e2530a`)
+* `exs` — boolen indicating if the account exists
```
tomquisel: boolen -> boolean

ecosystem/sep-0010.md
```diff
@@ -128,31 +129,37 @@ To validate the challenge transaction the following steps are performed by the s
...
+* `sig` — array of public keys of the signers of the challenge transaction
 * `thr`
   * `low` — low threshold of the account
   * `med` — medium threshold of the account
   * `high` — high threshold of the account
 * `wht` — weight of the signer(s) that signed the challenge transaction
 
-The JWT may contain other claims specific to your application, see [RFC7519].
+The JWT may contain other claims specific to your application, see [RFC7519], or omit claims that will not be used in your scenarios.
```
tomquisel: I might make two sections, one with the minimal set of claims for the JWT to function and one for suggested claims that may be useful in difference scenarios. At the end of the day the server is almost certainly making a JWT for its own use, so I wouldn't want implementers to get bogged down trying to implement many of these extra fields unless they've explicitly decided they need them.

leighmcculloch: Thanks, that's a better way to present this. 685ffb9

ecosystem/sep-0010.md
leighmcculloch: 9a96883

ecosystem/sep-0010.md
```diff
@@ -37,8 +37,9 @@ The flow achieves several things:
 * The server can verify that the client holds the secret key(s) of signers of the user's Stellar account
 * The client is able to prove their identity using a Ledger or other hardware wallet as well as by having direct access to the secret key(s)
 * The server can choose its own timeout for the user's session
-* The server can choose the required threshold(s) a user needs on the account
-* Other servers that consume the JWT can choose the required threshold(s) a user needs on the account for specific functionality
+* The server can choose the required threshold(s), if any, a user needs on the account
```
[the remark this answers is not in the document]
leighmcculloch: b9da632

istrau2:
Big fan of these changes. We are already using a very similar approach in Stellarport generated tokens (i.e. adding the threshold into the token). This approach has all the info we are including and more.

nebolsin:
> The exact contents of the JWT is less critical because the consumers of the JWT are likely to be hosted by the same people and an implementer could use an application specific JWT structure and still be interoperable with wallets.

Totally agree, in fact the original idea was to keep SEP-10 as generic as possible and make the auth token opaque for the wallet, or maybe mention JWT setup as a recommendation/default approach. For me the scope of this SEP was to define proof-of-possession auth protocol using Ed25519 keys on top of standard Stellar tools/libs. And the primary use-case in mind was kind of "Login with Stellar wallet" functionality (similar to Trezor Connect / BitID), where user is about to navigate from the to the traditional webapp and needs to identify himself with Ed25519 key.

>  I don't think SEP-10 can describe roles beyond the thresholds or signers defined on a Stellar account. <...> An implementer can add roles and other claims to the JWT that specifically address their scenario.

I think you misinterpreted my comment regarding lack of permission granularity / audience restrictions. My point was that threshold-based auth scheme should not be enforced by SEP-10 or even recommended as a default authorization system, because it lacks those important security-related features. 

With suggested changes SEP-10 server turns into an authorization server, which can issue access tokens for multiple resource servers with different authorization requirements. That indicates that client at the very least has to communicate the intended audience (service) and required threshold to the AS. This info also has to be included in the challenge tx, so that co-signers know what they're authorizing by signing it. 

Overall it seems like a pretty fundamental change for the SEP with accepted status.

leighmcculloch:
@nebolsin I see what you mean about these changes specifying how a service should perform authorization by the recommendations of using thresholds and weights, but then only goes so far, leaving many questions unanswered.

@tomquisel @nebolsin @msfeldstein @tomerweller I've reduced this proposal to improving how a server proves a user holds a Stellar account so that a server implementing the SEP will do the appropriate verifications to check the user holds a meaningful key for the account.

I think this change is aligned with the existing purpose of the SEP: to prove a user holds an account, not a key, and continues to support accounts that do not yet exist.

Could you all take another look?

ecosystem/sep-0010.md
```diff
@@ -22,18 +22,23 @@ The authentication flow is as follows:
 
 1. The client obtains a unique [`challenge`](#challenge), which is represented as specially formed Stellar transaction
 1. The client verifies that the transaction has an invalid sequence number 0.  This is extremely important to ensure the transaction isn't malicious.
-1. The client signs the transaction using the secret key for the user's Stellar account
+1. The client signs the transaction using the secret key(s) of signers for the user's Stellar account
 1. The client submits the signed challenge back to the server using [`token`](#token) endpoint
+1. The server gets the signers of the user's account
+1. The server verifies the signatures
+1. The server verifies the weight provided by the signers meets the required threshold(s), if any
```
istrau2: What is meaning of `required thresholds` in this context?

ecosystem/sep-0010.md
```diff
@@ -22,18 +22,23 @@ The authentication flow is as follows:
...
 The flow achieves several things:
 
 * Both client and server part can be implemented using well-established Stellar libraries
 * The client can verify that the server holds the secret key to a particular account
-* The server can verify that the client holds the secret key to their account
-* The client is able to prove their identity using a Ledger or other hardware wallet as well as by having direct access to the secret key
+* The server can verify that the client holds the secret key(s) to signers of their account
+* The client is able to prove their identity using a Ledger or other hardware wallet as well as by having direct access to the secret key(s)
 * The server can choose its own timeout for the user's session
+* The server can choose required threshold(s) that the user needs on the account, if any
```
istrau2: Why would an authentication server choose a threshold? It doesn't know what the token will be used for. In my view of things, these servers produce tokens that other servers use for authorization (based on their own rules).

It seems like this SEP is intending for every application server to be producing its own tokens based on its own authorization rule set. I don't agree with that approach, it seems like a single authentication server producing tokens (with claims) for all application servers to use, is the correct approach.

ecosystem/sep-0010.md
leighmcculloch: Any threshold the implementer wants the client to meet on the account. In many cases and for simplicity an implementer should probably just confirm that the signatures meet all thresholds (low, med, high) on an account to ensure the user completely holds an account in the same way they do if they have the default master key set as a signer.

An implementer can change this however they need for their specific application, requiring any lower threshold or no threshold. See this section:
https://github.com/stellar/stellar-protocol/blob/d7f11aba047a59989d0323981ec760360748b20a/ecosystem/sep-0010.md#L141-L145

ecosystem/sep-0010.md
leighmcculloch: This SEP is leaving that in the implementers control. A server can implement a verification of required thresholds to meet its definition of a user "holding an account", and a server can add any additional claims to the JWT if it wishes to provide authentication for multiple other services where varying levels of threshold are required.

leighmcculloch:
Here is an implementations of the changes to the Go SDK that modify the existing helper functions to provide the basic functionality for verifying a challenge transaction that has multiple signatures for an account: stellar/go#2071

Here is an example implementation of a SEP-10 server that requires a client to have complete control (meet all thresholds) of a Stellar account: stellar/go#2072

leighmcculloch:
**Updates:**

@tomerweller @tomquisel @msfeldstein I've added some additional recommendations relating to SEP-24. The additional text discusses the impact that accepting too little threshold can have, and explicitly states `medium` threshold should be used by anchors, and `high` threshold should be used in some situations.

I've removed statements recommending a signer meet all thresholds since any user meeting `high` threshold is able to change the thresholds and give themselves `low` and `med` capabilities even if they don't have those thresholds.

@nebolsin This backtracks a bit to take a clearer position on the use of weights and thresholds as a mechanism for determining whether the user truly holds an account. It became clear this is still important because many accounts have signers that are third party services and a third party service should not be allowed to authenticate as a user with some use cases, e.g. anchors (SEP-24).

---

Things I want to call out that this proposal does not attempt to solve:

1. This proposal does not address the possibility of clients and servers communicating about how much access is requested or preferred and I think we should continue to keep that out-of-scope until a specific use case arises. I also think servers can address that use case by displaying friendly messages to a user, and indicating signing needs to take place a second time and requesting the additional signers be involved. However, I think this should be addressed independently of this proposal.

2. This proposal does not provide a negotiation framework for multi-signer accounts where there are multiple keys spread across multiple devices. Multi-signer accounts which require signatures from other applications or hardware devices will need those applications or hardware devices to sign transactions for them and get those transactions back to the client to deliver to the server. It's very possible we could define a smooth negotiation framework such that a server could negotiate with several devices independently to authenticate a user, but since there is a wide variety in devices online/offline and existing APIs I think we need to spend more time researching this. We can open the floor for proposals for how to address this separately to SEP-10.  
   
   **Anchors may need to continue to support master key only signatures for sometime until this issue is addressed.** I could modify the proposal to continue to allow for master key regardless of its threshold proves control which keeps this backwards compatible for now.

---

**All, please take another read and help validate does this resolve the problems of:**
- Supporting single-signer accounts where the master key is no longer a signer
- Supporting multi-signer accounts where no signer has any authority on their own (single device)
- Can we move forward and keep (2) above out-of-scope, or is that a deal-breaker?

ecosystem/sep-0010.md
```diff
@@ -187,6 +201,44 @@ Every other HTTP status code will be considered an error. For example:
...
 
+## Verification
+
+The verification process confirms that a user holds an account. Depending on your application this may mean complete signing authority, some threshold of control, or being a signer of the account.
+
+An account's master key may meet not meet any threshold of control or could have had its weight reduced to zero. Most applications should not assume possession of the master key is possession of an account.
+
+An account's signers may include third-party services providing services to the user. Authenticating users with less than any threshold may allow a third-party to authenticate as a user.
+
+An account's signers may include the server's signing key if the server is a signer for the account. When determining the weight of the client sigantures the server's signature should be explicitly excluded. A server should not assist a client in authentication.
```
tomquisel: Do you think this should be explicitly stated? I can imagine there might be cases where the server does want to include its own signature.

leighmcculloch: In authentication the server should never include its own signature otherwise it will be helping a user authenticate who cannot authenticate on its own. The reason for this is that a server knows nothing about a user and so the user is on its own to prove it holds an account when communicating with the server. This would be probably exploitable if it did.

ecosystem/sep-0010.md
```diff
@@ -122,16 +127,25 @@ To validate the challenge transaction the following steps are performed by the s
 * verify that transaction has time bounds set, and that current time is between the minimum and maximum bounds;
 * verify that transaction contains a single [Manage Data](https://www.stellar.org/developers/guides/concepts/list-of-operations.html#manage-data) operation and its source account is not null;
 * verify that transaction envelope has a correct signature by server's signing key;
-* verify that transaction envelope has a correct signature by the operation's source account;
+* if the operations source account exists:
```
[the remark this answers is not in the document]
leighmcculloch: a5f989a

ecosystem/sep-0010.md
```diff
@@ -122,16 +127,25 @@ To validate the challenge transaction the following steps are performed by the s
...
+  * verify that client signatures count is one or more;
+  * verify that client signatures on the transaction are signers of the operations source account;
+  * verify that client signatures are correct;
+  * verify that client signatures provide weight that meets the required threshold(s), if any;
+* if the operations source account does not exist:
```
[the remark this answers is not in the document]
leighmcculloch: a5f989a

tomquisel:
One question is how we handle SEP-10 for users who are currently using services like StellarGuard. 

StellarGuard submits transactions directly to the network rather than passing signed transactions back to the wallet, so currently there's no way for StellarGuard to add its signature to a SEP-10 transaction. This just happens to be OK in the current SEP-10, because with StellarGuard the user continues to hold the master key, and all that the current SEP-10 requires to authenticate is a signature from the master key. So even though the master key only has half weight, it's enough to authenticate.

While this is convenient for users of StellarGuard, I don't think it's a good call going forward. It shouldn't be too difficult to fix StellarGuard to work with the new SEP-10 by modifying it to pass signed transactions back to the client instead of sending them directly to the network.

@pselden does this seem ok to you?

pselden:
It will take me a while to read through the proposal and comments here to see if it will works out, but StellarGuard already supports passing the signed transaction back to the client through a callback parameter. It would be up to the clients to use that and implement the receiver of the callback.

However, when I last did a survey (admittedly it’s been about 6 months), over half the users said they used the Stellar Account Viewer, which probably would not support this.

As long as that sort of usage is still supported it’s all good.

leighmcculloch:
>over half the users said they used the Stellar Account Viewer
The Stellar Account Viewer doesn't provide SEP-10 support so it sounds like many of the users will be unaffected by this change.

@pselden Thanks for pointing us to the callback, I hadn't seen that feature. I gave it a go at [test.stellarguard.me](https://test.stellarguard.me) and it worked great.

These are things I observed in relation to using it for SEP-10 challenge transactions:
- It looks like it'd work great for any wallet that has a backend it can use to accept the callback. ✅ 
- As far as I can tell when using a callback the transaction was not submitted to the network. That's great because the challenge transaction itself once fully signed is like a password and submitting it to anyone other than the wallet would be undesirable. ✅
- For wallets without a backend the callback won't be set. In this case the user would need to copy the resulting XDR out of the StellarGuard web UI, but because the callback isn't set the transaction will probably be submitted to the network which wouldn't be great because the signed transaction is essentially a time-limited password for authentication and shouldn't be broadcasted anywhere. Maybe StellarGuard could detect the zero sequence number which all challenge transactions have and automatically not submit the transaction since it knows the transaction would be rejected? ❓
- If the JSON response the wallet receives when submitting the transaction to StellarGuard included a URL the wallet could use to poll for the signed transaction a backend-less wallet could poll for signed XDR. 💡

Overall this doesn't seem like a blocker. Wallets/clients will need implement the callback if they are not using it today and for a period of time some wallets may become incompatible with an anchor for a subset of users until they add it.

@pselden:
- Are you able to share any stats on what % of integrated wallets already use a callback?
- For wallets that are unable to use a callback because they are backend-less, what do you think the idea 💡above?

leighmcculloch:
Another service with a similar API and behavior as StellarGuard is Lobstr Vault ([API details here](https://github.com/Lobstrco/Vault-Android#developers-integrate-lobstr-vault-with-your-service)). The use of a [callback](https://github.com/stellar/stellar-protocol/pull/489#issuecomment-567762813) and/or the [idea 💡 above](https://github.com/stellar/stellar-protocol/pull/489#issuecomment-567814294) would also be a way forward for it once these changes to SEP-10 are merged and start being picked up by anchors and others.

@pitsevich @dmgmyza Does this seem ok to you?

pselden:
> Maybe StellarGuard could detect the zero sequence number which all challenge transactions have and automatically not submit the transaction since it knows the transaction would be rejected? ❓

Yes I can implement that. I'm also planning to add an option where the user can choose whether or not to submit ANY transaction.

> If the JSON response the wallet receives when submitting the transaction to StellarGuard included a URL the wallet could use to poll for the signed transaction a backend-less wallet could poll for signed XDR. 💡

I actually have a SEP for this that languished because it paradoxically got too complicated AND didn't address enough use cases: https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0019.md

> Are you able to share any stats on what % of integrated wallets already use a callback?

I know that the actual usage of the callback is around 2% of all transactions submitted to StellarGuard -- so not many wallets are using it (or at least the more popular ones are not).

leighmcculloch:
I've completed the work on the Go SDK to add support for these changes in stellar/go#2071, and stellar/go#2072 adds an example of how this would be used in a server implementing. It does require a little more work on an implementer, but I think the SDKs will abstract most of it.

@tomerweller @msfeldstein If you think this change to the protocol is good and how it is demonstrated in code at those links is good, I'd like to validate that the SEP-10 change is good from the perspective of Polaris and the Integrations team by having the change made to Polaris too. Do you folks have capacity to help with that? cc @accordeiro

accordeiro:
I love the idea of building this into Polaris! Seems like a nice way to enable us to start implementing/testing things on the client side (like for Vega) too :)

pitsevich:
Thanks for making these changes. Think it's important to have multisig accounts play nicely with SEP-24 servers.

I think that it also makes sense to increase the recommended time bounds:
image

For accounts protected with multi-sig setting `max` to now() + 5 minutes could be too short timeframe.

-----------

On a separate note, we've recently seen issues with one of the anchors and a `min` time validation failing. 
The server was only accepting a signed challenge if we waited a few extra seconds before sending that to the server.

I assume, a server time on the anchor's side was off by a second or so (with server time being ahead of real time), which was resulting in a `min` time validation of the signed challenge failing.
This was fixed after anchor changed the min param and added some padding.

I'm not really sure if this issue was caused by the specifics of the implementation of this particular anchor and whether it would still apply to other implementations of this SEP, but mentioned this to make sure this SEP approach is not too sensitive to the server time.

leighmcculloch:
@pitsevich Thanks this is really helpful point to have surface. I think increasing the time limit from 5 minutes maybe to 15 minutes would be okay to provide time for multiple-signatures to be attached. Do you think that would be sufficient or do you have a idea for what time limit would be appropriate? The time period is also the period within which a challenge could be reused to get a new JWT so we should keep it as low as practically possible.

---

For the issue of dealing with clock skew between services, could you open an issue on this repository about the issue of clock skew in SEP-10 and we can discuss it separately? Thanks!

pitsevich:
@leighmcculloch I was thinking about 30 minutes but 15 minutes should be enough in most cases as well.

pitsevich:
@leighmcculloch 
> For wallets without a backend the callback won't be set. In this case the user would need to copy the resulting XDR out of the StellarGuard web UI, but because the callback isn't set the transaction will probably be submitted to the network which wouldn't be great because the signed transaction is essentially a time-limited password for authentication and shouldn't be broadcasted anywhere. Maybe StellarGuard could detect the zero sequence number which all challenge transactions have and automatically not submit the transaction since it knows the transaction would be rejected? ❓
> If the JSON response the wallet receives when submitting the transaction to StellarGuard included a URL the wallet could use to poll for the signed transaction a backend-less wallet could poll for signed XDR. 💡

This is quite important for us, since a lot of users are using multi-sig with StellarTerm.

We will be rolling out SEP-24 to StellarTerm shortly, so users who would want to make a withdrawal/deposit on StellarTerm and have multisig through Vault or StellarGuard would be having difficulties.

The wallets that have a backend could listen for the callback from services like Vault or Stellarguard and collect all signatures before sending to SEP-10 server for JWT token.
However, for backend-less wallets using callback is not an option

StellarTerm can't use the callback since it is a backend-less wallet.
So looks like the only possible way would be to implement the idea 💡 above.

In that case Vault would be able to provide a URL where the client (wallet) can track and fetch the signatures 

We would be willing to build something like this into Vault and StellarTerm. 
Hopefully we can agree on the same interface with StellarGuard to keep things standard. 
@pselden 👀

leighmcculloch:
>If the JSON response the wallet receives when submitting the transaction to StellarGuard included a URL the wallet could use to poll for the signed transaction a backend-less wallet could poll for signed XDR. 💡

@pselden @pitsevich Would you be up for exploring this idea independently of this change in parallel? Since both of your APIs are the same or very similar today, if you can add this capability in the same way and work with StellarTerm to integrate it, that might be a good point to document this API in a SEP. What do you think?

pselden:
Yes I'm up for that. We can use https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0019.md as a starting point or scrap it for something simpler.

leighmcculloch:
I'm planning on merging this change and locking it in on Monday January 27th 10am PST, along with the relevant Go SDK changes stellar/go#2071 and updating our reference implementation stellar/go#2072. @JakeUrban has started work on updating Polaris and it'll come after.

cc @tomerweller @tomquisel @rice2000

krudder64:
Please can anyone on here please remove vault signers from lobstr and stellar expert. Tell me how or how to do in laboratory. Step by step please…

leighmcculloch:
@krudder64 I recommend asking either on [StackOverflow](https://stellar.stackexchange.com/questions/ask) or [the #support Discord](https://discord.gg/fBTQDgAyCP).

PR #498, opened 2019-12-13
Update CAP-0023

jonjove:


core/cap-0023.md
```diff
@@ -239,146 +293,248 @@ default:
...
+9. Fail with `opTOO_MANY_SUBENTRIES` if `sourceAccount` is already at the
+   subentry limit
+10. Fail with `CLAIM_CLAIMABLE_BALANCE_LOW_RESERVE` if the `sourceAccount` does
+    not have sufficient available balance (see CAP-0003) to create a trust line
+11. Create a trust line with the following properties:
+    - `sourceAccount` as specified in the operation
+    - `asset` as specified in the `ClaimableBalanceEntry`
+    - `balance = 0`
+    - `limit = INT64_MAX`
+    - `flags = 0`
```
leighmcculloch: If an app is adding support for claiming balance entries it's probably well within their capabilities to create the trustline as a separate operation in the same transaction, which removes the assumption this process makes about the limit that should be set on the trustline, and would make this a smaller building block that could be combined with the existing trustline operations.

Is there a specific benefit in broadening the functionality provided by this operation to automatically create a trustline?

core/cap-0023.md
```diff
@@ -239,146 +293,248 @@ default:
...
+   than `threshold`.
+4. Fail with `CLAIM_CLAIMABLE_BALANCE_LINE_FULL` if the `sourceAccount` does not
+   have at least `reserve` available limit of native asset (see CAP-0003)
+5. Add `reserve` of native asset to the `sourceAccount`
+6. Skip to step 14 if `asset` is native
+7. Skip to step 16 if `sourceAccount` is the issuer for `asset`
+8. Skip to step 12 if `sourceAccount` has a trust line for `asset`
+9. Fail with `opTOO_MANY_SUBENTRIES` if `sourceAccount` is already at the
+   subentry limit
+10. Fail with `CLAIM_CLAIMABLE_BALANCE_LOW_RESERVE` if the `sourceAccount` does
```
leighmcculloch: If we fail at this point is the affect of step 5 something that has still taken place, or are all of these steps atomic?

core/cap-0023.md
```diff
@@ -239,146 +293,248 @@ default:
...
+
+1. Fail with `CLAIM_CLAIMABLE_BALANCE_DOES_NOT_EXIST` if there is no
+   `ClaimableBalanceEntry` matching `balanceID`.
+2. Calculate the total weight of `condition` with `destination = sourceAccount`
+   and satisfied predicates.
+3. Fail with `CLAIM_CLAIMABLE_BALANCE_CANNOT_CLAIM` if the total weight is less
+   than `threshold`.
+4. Fail with `CLAIM_CLAIMABLE_BALANCE_LINE_FULL` if the `sourceAccount` does not
+   have at least `reserve` available limit of native asset (see CAP-0003)
+5. Add `reserve` of native asset to the `sourceAccount`
```
leighmcculloch: Because the reserve is always transmitted to the recipient, the sender is losing native asset on every send. The cost of a claimable balance send is significantly more than the cost to send without a claimable. Is that correct? This seems like something that will prohibit its use. Is that intentional to limit the use of this op?

core/cap-0023.md
```diff
@@ -239,146 +293,248 @@ default:
...
-
-Suppose that we try to relax this requirement in order to avoid this downside. We
-could instead attach two public keys to the `BalanceEntry`, where the sender alone
-knows one of the private keys and the recipient alone knows the other. Which key is
-currently usable would be determined by the ledger close time. The operation to
-claim the `BalanceEntry` could then specify an account A and a signature over A
-from the key that is currently usable. This would allow the appropriate party to
-claim the `BalanceEntry` into any account that they control. But this would also
+   been claimed
+3. Receiving accounts always receive the reserve from the claimed
```
leighmcculloch: This makes using claims vs normal sends many times more expensive (I think 100,000x more expensive). The additional cost might be prohibitive?

core/cap-0023.md
```diff
@@ -239,146 +293,248 @@ default:
...
+does simplify the behavior of `ClaimableBalanceEntry` by guaranteeing that the
+receiving account always receives both the reserve and the amount of asset.
+
+### ClaimableBalanceEntry claimants are accounts
+For each `ClaimableBalanceEntry`, `condition` contains a finite and immutable
+list of accounts that could potentially claim the `ClaimableBalanceEntry`. Even
+if the conditions are satisfiable (which is not guaranteed), it is still
+possible for the `ClaimableBalanceEntry` to become stranded. If all of the
+accounts listed in `condition` are merged and none of the private keys are
+known, then the `ClaimableBalanceEntry` will no longer be claimable.
```
leighmcculloch: Dangling claimables seems problematic. This seems like something that could be footgun.

core/cap-0023.md
```diff
@@ -239,146 +293,248 @@ default:
...
+                type: CLAIM_PREDICATE_AFTER_ABSOLUTE_TIME
+                absAfter: T
+        threshold: 1
+```
 
 ## Backwards Incompatibilities
 All downstream systems will need updated XDR in order to recognize the new
 operations and ledger entries.
 
 ## Security Concerns
```
leighmcculloch: One use case this claimable capability could open up, if it didn't include the reserve, is the ability for payment products that send payments to give recipients the opportunity to decline the receipt of a payment. This isn't part of the motivation at the moment by the looks of it, but this operation would support that use case if the reserve was optional or was sent back.

MonsieurNicolas:
@leighmcculloch I merged it - let's discuss on the mailing list to avoid scattered feedback (@jonjove is going to create a "master thread CAP-0023" shortly)

PR #502, opened 2019-12-14
Create greetings.yml

xIITOXAIIx:


ire-and-curses:
This looks like spam - cut/paste from https://github.com/actions/starter-workflows/blob/master/automation/greetings.yml

PR #512, opened 2020-01-14
Add version: 2 to sep 24

msfeldstein:
Currently wallets are unsure whether any anchor supports SEP-6 or SEP-24, and need to hardcode that.  Now we add a `version: 2` flag to the /info endpoint to signal that the anchor supports all the new SEP-24 protocols.  If it's missing (and you haven't hardcoded that they support sep24 in your wallet) you can assume its still sep6.

### DIFF
--- a/ecosystem/sep-0024.md
+++ b/ecosystem/sep-0024.md
@@ -408,6 +408,7 @@ The response should be a JSON object like:
 
 ```json
 {
+  "version": 2,
   "deposit": {
     "USD": {
       "enabled": true,
@@ -446,7 +447,7 @@ The response should be a JSON object like:
 }
 ```
 
-The JSON object contains an entry for each asset that the anchor supports for deposit and/or withdrawal.
+The JSON object contains an entry for each asset that the anchor supports for deposit and/or withdrawal.  It also must include `version: 2` to signal to wallets that the anchor is SEP-24 aware, rather than the legacy SEP-6.
 
 #### For each deposit asset, response contains:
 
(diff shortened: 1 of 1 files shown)


JakeUrban:
Personally I think it would be better to add a field to SEP-1, like `TRANSFER_SEP=24` or `TRANSFER_PROTOCOL=sep6` instead of calling SEP-6 version 1 and sep-24 version 2. Note that we'll ultimately support SEP-26 as well, but SEP-26 isn't an upgraded version of SEP-24.

pitsevich:
Great to see this progressing, no matter if this will be recorded in the stellar.toml or in the /info endpoint. 

We currently have to manually mark the anchors as sep-6 or sep-24 compatible, and would love to remove that logic!

tomerweller:
I wholeheartedly support adding a version but not sure what's the right location/format. 
- In the future, an anchor might support multiple transfer server versions. This is common in long lasting APIs (`/v1`, `/v2`, etc. urls) and we should support this pattern. 
- Would be great to use the SEP versioning for this (could be just the major, or major and minor) rather than an additional versioning scheme.

A comprehensive version of this could look like a `stellar.toml` (breaking) change: 
```toml
[[TRANSFER_SERVERS]]
URL="https://api.domain.com/v1"
VERSION="1.1.0"

[[TRANSFER_SERVERS]]
URL="https://api.domain.com/v2"
VERSION="2.0.0"
```
Or, a bit less comprehensive (with only major version): 
```toml 
TRANSFER_SERVER_v1="https://api.domain.com/v1"
TRANSFER_SERVER_v2="https://api.domain.com/v2" 
```

I prefer 1 because it leaves more room for growth. Obviously this is less than ideal as it's a breaking change in `stellar.toml`, but we're long overdue for a SEP-0001 versioning discussion as well :)

msfeldstein:
@tomerweller if we're going with breaking changes, there's pretty robust support for this in SEP-27's Asset Info Server which we should use in the future (and SEP-26 does use). The goal here is to not be a breaking change, but just make it so people like @pitsevich don't need to manually track things.

msfeldstein:
It probably does belong in the TOML, and i can update that to be there.  However i don't want to try to broaden the scope of it to encompass anything more than whether this is SEP-6 or SEP-24.  We can call it TRANSFER_SEP if we want but the asset info server is the more robust version and i don't want to confuse the issue.  I'd even be fine calling it SUPPORTS_SEP24=true

leighmcculloch:
I think something like `TRANSFER_SERVER_SUPPORTS_SEP24=true` is intuitive and easily discovered alongside the existing `TRANSFER_SERVER`. It also means you can define `TRANSFER_SERVER_SUPPORTS_SEP6` alongside it if you offer both interactive and non-interactive. 

Would an anchor be able to specify `...SUPPORTS_SEP24` and `...SUPPORTS_SEP6` both as true and implement them on the same URL?

msfeldstein:
After talking this over we decided to support SEP-27 and not do any band-aid approaches

PR #522, opened 2020-01-28
[SEP24] Don't require auth for /info

msfeldstein:
Clarify that the /info endpoint, which provides publicly accessible information, shouldn't require an auth token, but all other endpoints should.

### DIFF
--- a/ecosystem/sep-0024.md
+++ b/ecosystem/sep-0024.md
@@ -43,7 +43,7 @@ To support this protocol an anchor acts as a server and implements the specified
 
 ## Authentication
 
-As stated, Anchors must support [SEP-10](sep-0010.md) web authentication to enable authenticated deposits, withdrawals, or transaction history lookups.  Clients must submit the JWT previously obtained from the anchor via the [SEP-10](sep-0010.md) authentication flow to all API endpoints that provide user data.  Currently `/info` should be unauthenticated, but all other endpoints will require a token.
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0024.md
```diff
@@ -43,7 +43,7 @@ To support this protocol an anchor acts as a server and implements the specified
 
 ## Authentication
 
-As stated, Anchors must support [SEP-10](sep-0010.md) web authentication to enable authenticated deposits, withdrawals, or transaction history lookups.  Clients must submit the JWT previously obtained from the anchor via the [SEP-10](sep-0010.md) authentication flow to all API endpoints.
+As stated, Anchors must support [SEP-10](sep-0010.md) web authentication to enable authenticated deposits, withdrawals, or transaction history lookups.  Clients must submit the JWT previously obtained from the anchor via the [SEP-10](sep-0010.md) authentication flow to all API endpoints that provide user data.  Currently `/info` should be unauthenticated, but all other endpoints will require a token.
```
leighmcculloch: If we're defining `/info` as unauthenticated, I don't think we should say `Currently`. This will be hard to change later on as it would be a breaking change, and currently makes it sound like we aren't really sure.

msfeldstein: Yeah the wording is bad, i meant that "Currently this is the only one, but there might be new endpoints later that can be unautneticated" not that we might require auth on /info.  I'll just remove that word

PR #526, opened 2020-02-03
Add account memo requirements SEP

orbitlens:


ecosystem/sep-0029.md
```diff
@@ -0,0 +1,84 @@
...
+whole ecosystem and custodial exchanges will require much more time, as well as
+development efforts. Specifying memo requirements in a form of data entry
+attached to the account introduces a simple way of enforcing transaction
+memo requirements without changing the underlying protocol. 
+
+The name of the data entry follows the namespaces format proposed in SEP-18.
+
+Memo type verification routines are intentionally omitted from this SEP in favor
+of a separate standard with extended definition of more granular control over
+the required memo type via a separate `config.memo_required_type` data entry
```
leighmcculloch: Can you make this less prescriptive of what any future standard should look like?
```suggestion
other attributes of memos via a separate data entries
```

ecosystem/sep-0029.md
```diff
@@ -0,0 +1,84 @@
...
+Users frequently forget to fill in a memo in their deposit transactions.
+Dealing with missing memos may result in lost funds and requires manual actions
+from the custodial support staff.
+
+We introduce a basic convention that allows defining memo requirements on the
+account level.
+
+Although this interoperability standard aims to deal with a missing memo problem
+in the first place, it can be also utilized by a wider class of Stellar-based
+applications to designate additional interaction or compliance prerequisites.
```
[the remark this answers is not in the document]
orbitlens: Done

ecosystem/sep-0029.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Abstract
+
+An account owner configures memo requirements by adding a data entry that
+specifies the whether a transaction memo is required when transferring funds to
+the account. Each wallet ensures that a memo is attached before submitting
+the transaction to the network.
```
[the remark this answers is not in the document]
orbitlens: Done

ecosystem/sep-0029.md
orbitlens: Rephrased. In any case, I left this in the rationale section only as an illustrative example.

ecosystem/sep-0029.md
```diff
@@ -0,0 +1,86 @@
...
+the transaction to the network. Horizon server performs the same validation
+routines before submission and returns an error to the client in case of
+requirements violation.
+
+## Specification
+
+To set up memo prerequisites for an account, the owner submits a transaction
+with `MANAGE_DATA` operation that adds a data entry with the name
+`config.memo_required` (in lower case following SEP-18 notation) and value `1`
+(a single character string "1" - ASCII code 49 or Unicode U+0031).
```
tomerweller: I'm concerned about the ASCII definition as the data entry value is an [opaque](https://github.com/stellar/stellar-core/blob/472f98f1e70d9e120d0befee3554f3f1d5386fa0/src/xdr/Stellar-ledger-entries.x#L16) structure. In fact, horizon returns a base64 encoding of these. So a user who sets this value to `1` with laboratory, for example, will later be presented with that value as `MQ==` by horizon. 

We can make our lives easier by ignoring the actual value of this data entry. If it exists - we treat is as `memo_required`, regardless of the value. And if not - we don't.

orbitlens: Fully agree with this. I proposed the same on the mailing list earlier.

tomerweller: Oh sorry, I missed that part of the discussion. I don't see any objection to this. 
@tomquisel @leighmcculloch ?

leighmcculloch: I don't think the fact the value is encoded as `MQ==` should be a deterrent. Horizon encodes all data entries and if we one day build utility functionality into the SDKs that help with parsing and reading account data sets like what is described in [SEP-18] then those tools can decode the data for applications. An application should expect to base64 decode any data entry value into an expected type (string, int, etc).

I think it is helpful if we specify what the value should be because Stellar will not let us set an empty value, so the question, "What should it be set to?" will be raised, and it is more intuitive and easier in some programming languages to look for a value set rather than a key set when inspecting deserialized JSON.

[SEP-18]: https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0018.md

tomquisel: For this change, I think it'd be easier to just check whether the key is set. However, I'm not sure it's a good long-term decision. If we have data entries, we should make both the keys and the values useable. While we can dodge that here since it's a boolean, I think it's better to meet the problem head on so there's a consistent pattern in the future for setting & getting keys & values out of data entries.

There's also the potential for confusion. If `config.memo_required` is set to `0`, someone will naturally think they've disabled it, but that won't be the case.

So I propose we move forward with setting this value to `1`.  We then need to mitigate the confusing `MQ==` in applications. I'd suggest the following:

* The lab and other tools that allow you to view the data entry values should try decoding the value as a string, and if it works, display that along with or instead of the encoded value.
* A flag feature could be added to Horizon for debugging where it attempts to decode the value itself, and returns the decoded result in its responses if successful.

PR #529, opened 2020-02-05
Update sep-0009.md

vladcentea:
Structured the information on KYC levels, added legal entities KYC information fields.

### DIFF
--- a/ecosystem/sep-0009.md
+++ b/ecosystem/sep-0009.md
@@ -16,12 +16,6 @@ This SEP defines a list of standard KYC and AML fields for use in Stellar ecosys
 
 This is a list of possible fields that may be necessary to handle many different use cases, there is no expectation that any particular fields be used for a particular application. The best fields to use in a particular case is determined by the needs of the KYC or AML application.
 
-## KYC Levels
-
-To improve user's privacy and to distinguish between various levels of risk that VASPs (Virtual Asset Service Providers) need to define in thier KYC/AML policy the KYC information will be groupped into distinct KYC Levels. This way each VASP (issuers, Banks...) can request or expose a certain level of KYC information depending on the risk profile of the customer, risk of the transaction and the risk appetite of the VASP itself.
-
-Note: some of the information marked as KYC2 level might be extracted from documents needed for KYC1 level (e.g. place of birth - might be on the ID document), but in the process of exchange information such information should not be exchanged between VASPs unless needed a higher level the KYC is needed for a certain operation. The main reason of using KYC level is to protect individual privacy and personal information. 
-Note2: This document is not describing the obligations of all parties involved in exchange of personal information but they need to comply with privacy laws. (like GDPR - Europe)
 
 ## Encodings
 
--- a/ecosystem/sep-0009.md
+++ b/ecosystem/sep-0009.md
@@ -16,6 +16,13 @@ This SEP defines a list of standard KYC and AML fields for use in Stellar ecosys
 
 This is a list of possible fields that may be necessary to handle many different use cases, there is no expectation that any particular fields be used for a particular application. The best fields to use in a particular case is determined by the needs of the KYC or AML application.
 
+## KYC Levels
+
+To improve user's privacy and to distinguish between various levels of risk that VASPs (Virtual Asset Service Providers) need to define in thier KYC/AML policy the KYC information will be groupped into distinct KYC Levels. This way each VASP (issuers, Banks...) can request or expose a certain level of KYC information depending on the risk profile of the customer, risk of the transaction and the risk appetite of the VASP itself.
+
+Note: some of the information marked as KYC2 level might be extracted from documents needed for KYC1 level (e.g. place of birth - might be on the ID document), but in the process of exchange information such information should not be exchanged between VASPs unless needed a higher level the KYC is needed for a certain operation. The main reason of using KYC level is to protect individual privacy and personal information. 
+Note2: This document is not describing the obligations of all parties involved in exchange of personal information but they need to comply with privacy laws. (like GDPR - Europe)
+
 ## Encodings
 
 ISO encodings are used for fields wherever possible. The table below lists the encodings used for different types of information.
--- a/ecosystem/sep-0009.md
+++ b/ecosystem/sep-0009.md
@@ -6,8 +6,8 @@ Title: Standard KYC / AML fields
 Author: stellar.org
 Status: Active
 Created: 2018-07-27
-Updated: 2018-09-11
-Version 1.1.0
+Updated: 2020-04-03
+Version 1.2.0
 ```
 
 ## Simple Summary
@@ -15,7 +15,6 @@ Version 1.1.0
 This SEP defines a list of standard KYC and AML fields for use in Stellar ecosystem protocols. Issuers, banks, and other entities on Stellar should use these fields when sending or requesting KYC / AML information with other parties on Stellar. This is an evolving list, so please suggest any missing fields that you use.
 
 This is a list of possible fields that may be necessary to handle many different use cases, there is no expectation that any particular fields be used for a particular application. The best fields to use in a particular case is determined by the needs of the KYC or AML application.
-
 
 ## Encodings
 
--- a/ecosystem/sep-0024.md
+++ b/ecosystem/sep-0024.md
@@ -436,12 +436,6 @@ The response should be a JSON object like:
   },
   "fee": {
     "enabled": false
-  },
-  "transactions": {
-    "enabled": true
-  },
-  "transaction": {
-    "enabled": false
   }
 }
 ```
@@ -468,8 +462,9 @@ The JSON object contains an entry for each asset that the anchor supports for de
 
 If `fee_fixed` or `fee_percent` are provided, the total fee is calculated as `(amount * fee_percent) + fee_fixed = fee_total`.  If the fee structure doesn't fit this model, omit them and provide the [`/fee`](#fee) endpoint instead.
 
-An anchor should also indicate in the `/info` response if they support the `fee`, `transactions` and `transaction` endpoints, by providing the following fields for each:
+An anchor should also indicate in the `/info` response if they support the `fee`endpoint, by providing the following fields:
 
+* `authentication_required`: `true` if client must be [authenticated](#authentication) before accessing the `fee` endpoint.
 * `enabled`: `true` if the endpoint is available.
 
 ## Fee


tomquisel:
@vladcentea thanks so much for this contribution! It seems like a good direction, and I'd like to learn more. Can you give a little more context on why this is a beneficial addition and what you're using it for? 

In particular, I'm curious:

* is the KYC level an international standard? Is there a source document you can link to?
* legal entity KYC fields make sense. Can you explain more about the use case where you're KYCing legal entities?

vladcentea:
Hello Tom,

1. About KYC levels.

It is more or less a common practice among the regulated entities 
(banks, payment institutions, electronic money institutions, funds, 
broker/dealers and so on) to put some sort of structure in their risk 
based approach. There are 2 main risk vectors (product risk and customer 
risk). This organizations have to balance between them and they usually 
have some compliance polices to take decisions. An example of product 
risk is when someone wants to make a big transaction and an example of 
customer risk is when the customer is a politically exposed person of a 
person from a sanctioned country or list. There are also some vague 
references in the legal frameworks in various countries such as 
(reporting of cross border transfers above X (e.g. 10k) amount or allow 
anonymous transactions below Y amount). So, in some cases they will need 
additional personal information or additional information about the 
transaction to clear it. The more information they have the lower will 
be risk. The the end the law is vague and generic and is generally 
saying that the regulated institutions have to take a risk based 
approach and pay for the broken glass if something bad happens like 
money laundry, terrorism financing and so on.

So, there is no actual standard for that since most of these a internal 
rules of the banks, exchanges, and so on, and since they are not 
exchanged customer information yet based on any protocol this was not 
necessary. However I believe that we're going into the direction of 
decentralized identities (https://w3c.github.io/did-use-cases) and 
verifiable claims(https://w3c.github.io/vc-use-cases/) so such standard 
would be useful for information exchange, and to protect as much as 
possible the more sensitive and more detailed information, unless there 
is a risk/reward justification for it.

2. Legal entity KYC

I see a lot of uses cases for Virtual Assets  and especially for FIAT 
tokenization and for digital securities that would benefit legal 
entities. From smallest one like the micro companies (like freelancers 
or very small companies) to SMEs. So I would expect that the anchors 
will have have to create also "corporate accounts". These also will help 
a lot to the market liquidity because what i noticed on the FX market is 
that usually in a specific country the natural persons (working abroad 
or remote) are more on the buying side of the local currency  and the 
companies are on the buying side of a foreign currency. (import).

Also another very important Stellar use case such as cross border 
payments it's very important for companies, so I expect that any anchor 
will have demand for corporate accounts.


If you need more information on these topics you can contact me directly 
by email/telegram. (vlad.centea@morfin.io / vcentea7)


Best regards,

Vlad.

ecosystem/sep-0009.md
```diff
@@ -30,39 +37,66 @@ occupation | 3 | [ISCO08](https://en.wikipedia.org/wiki/International_Standard_C
...
+'KYC1' | `id_issue_date` | date | ID issue date
+'KYC1' | `id_expiration_date` | date | ID expiration date
+'KYC1' | `id_number` | string | Passport or ID number
+'KYC1' | `photo_id_front` | binary | Image of front of user's photo ID or passport
+'KYC1' | `photo_id_back` | binary | Image of back of user's photo ID or passport
+'KYC3' | `notary_approval_of_photo_id` | binary | Image of notary's approval of photo ID or passport
+'KYC3' | `ip_address` | string | IP address of customer's computer
+'KYC2' | 'photo_proof_residence' | binary | Image of a utility bill, bank statement with the user's name and address
+
+Legal entity KYC fields
```
tomerweller: Maybe use KYB instead of KYB here?

ecosystem/sep-0009.md
```diff
@@ -30,39 +37,66 @@ occupation | 3 | [ISCO08](https://en.wikipedia.org/wiki/International_Standard_C
...
+'KYC0' | `company_VAT_number' | string | Company VAT number 
+'KYC0' | `company_registration_number' | string | Company registration number
+'KYC0' | `company_registered_address' | string | Company registered address
+'KYC1' | `number_of_shareholders' | number | Company shareholder number
+'KYC1' | `shareholder_name' | string | Can be a company or a person and should be queried recursively up to the ultimate beneficial owners (with KYC information for natural persons such as above)
+'KYC0' | `image_incorporation_doc' | string | Image of company incorporation documents
+'KYC1' | `image_proof_adress' | string | Image of a utility bill, bank statement with the company's name and address
+'KYC0' | `address_country_code` | country code | country code for current address
+'KYC0' | `state_or_province` | string | name of state/province/region/prefecture
+'KYC0' | `city` | string | name of city/town
```
tomerweller: I wonder if we should name-space entity fields to disambiguate. So something like `entity.city` instead of just `city`

ecosystem/sep-0009.md
vladcentea: In general in the financial sector they use KYC acronym for both private and business customers. Of course as the guys from Chainalysis invented the term KYT (know your transaction) it's not a problem to say KYB, but in general is not used this way in the industry, It does not mean is wrong or would do any harm using it.

ecosystem/sep-0009.md
vladcentea: That would indeed bring more clarity. Maybe we could call it company instead of entity (less abstract).
Your comment also makes me thinking if would be possible to have a kind of hierarchical organization of data, since in fact a company can have as shareholders other companies or persons, and during due diligence process, the financial institutions have to go recursively down the tree until the ultimate beneficial owners (which must be persons). So basically always a company KYC data must be linked to at least a person KYC data (and in many cases to several persons and companies). 
So for a business KYC it might be the case the after first request, the party who's interrogating the other party will send subsequent requests related to the first one. 
I don't want to overcomplicate things here i am just stating how this process is actually taking place in financial institutions. 
If you think it makes sense to explore this side we can go in details there too.

ecosystem/sep-0009.md
```diff
@@ -30,39 +37,66 @@ occupation | 3 | [ISCO08](https://en.wikipedia.org/wiki/International_Standard_C
...
+'KYC1' | `id_type` | string | `passport`, `drivers_license`, `id_card`, etc...
+'KYC1' | `id_country_code` | country code | country issuing passport or photo ID as ISO 3166-1 alpha-3 code
+'KYC1' | `id_issue_date` | date | ID issue date
+'KYC1' | `id_expiration_date` | date | ID expiration date
+'KYC1' | `id_number` | string | Passport or ID number
+'KYC1' | `photo_id_front` | binary | Image of front of user's photo ID or passport
+'KYC1' | `photo_id_back` | binary | Image of back of user's photo ID or passport
+'KYC3' | `notary_approval_of_photo_id` | binary | Image of notary's approval of photo ID or passport
+'KYC3' | `ip_address` | string | IP address of customer's computer
+'KYC2' | 'photo_proof_residence' | binary | Image of a utility bill, bank statement with the user's name and address
```
tomquisel: Minor nitpick, it looks like many of the single-quotes (`'`) should be back ticks (`) instead

ecosystem/sep-0009.md
tomquisel: I love the idea of prefixing all of these with a dot-notation namespace. I'd vote for `organization`, in keeping with [schema.org](https://schema.org/Organization).

I'd also hold off on more complex setups just yet, let's wait for the need to develop and then we'll have more info on how exactly to solve it.

ecosystem/sep-0009.md
tomquisel: What do you think about legal entity -> Organization?

ecosystem/sep-0009.md
```diff
@@ -16,6 +16,13 @@ This SEP defines a list of standard KYC and AML fields for use in Stellar ecosys
 
 This is a list of possible fields that may be necessary to handle many different use cases, there is no expectation that any particular fields be used for a particular application. The best fields to use in a particular case is determined by the needs of the KYC or AML application.
 
+## KYC Levels
```
tomquisel: I think KYC levels are an important concept, although it's tricky since there's no universally agreed-on or standard for which pieces are information are at which level. As a result, I would advocate removing the KYC level discussion from this doc. This SEP aims to define a set of names that can be used by automated systems to refer to particular kinds of KYC information. I'd recommend that protocols based on SEP-9 refer to each piece of information that's needed separately by name, rather than grouping them together and referring to them by KYC levels. Using specific names is more precise, and avoids any confusion when entities have different definitions of the levels internally.

ecosystem/sep-0009.md
```diff
@@ -30,39 +37,66 @@ occupation | 3 | [ISCO08](https://en.wikipedia.org/wiki/International_Standard_C
...
+KYC Level | Name | Type | Description
+-----------|-----|------|------------
+'KYC0' | `company_name' | string | Full company name as on the company incorporation papers
+'KYC0' | `company_VAT_number' | string | Company VAT number 
+'KYC0' | `company_registration_number' | string | Company registration number
+'KYC0' | `company_registered_address' | string | Company registered address
+'KYC1' | `number_of_shareholders' | number | Company shareholder number
+'KYC1' | `shareholder_name' | string | Can be a company or a person and should be queried recursively up to the ultimate beneficial owners (with KYC information for natural persons such as above)
+'KYC0' | `image_incorporation_doc' | string | Image of company incorporation documents
+'KYC1' | `image_proof_adress' | string | Image of a utility bill, bank statement with the company's name and address
```
tomquisel: Since this is the legal entity analog of `photo_proof_residence`, I'd try to make their names more consistent.

Also, since we already have a pattern of using `photo` instead of `image`, I'd stick to `photo` in these fields.

vladcentea:
Made some updated in line with the suggestions.

tomquisel:
Thanks @vladcentea! I suggested some tweaks on top: https://github.com/vladcentea/stellar-protocol/pull/1

leighmcculloch:
@vladcentea @tomquisel @tomerweller Is this PR still being actively iterated on?

tomquisel:
I think it's waiting on @vladcentea to decide what to do with https://github.com/vladcentea/stellar-protocol/pull/1. I'm happy with that change, though. I'd vote for merging this PR in after that other PR is merged in.

vladcentea:
I merged your PR. Sorry for the delay.

tomquisel:
@vladcentea it looks like that https://github.com/vladcentea/stellar-protocol/pull/1 PR went into your master. Can you merge it into your `patch-1` branch as well? That'll bring it into this PR. Then we can merge this and be done 😄

vladcentea:
Done. Please review, thanks!

tomquisel:
@tomerweller any final objections before I merge?

Issue #527, opened 2020-02-04
SEP24: withdraw amount in interactive URL


yuriescl:
In SEP24 section [Basic Anchor Implementation](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0024.md#basic-anchor-implementation) item **Interactive withdrawal**, one of the bullet items states:

_If an amount parameter is provided to the interactive URL, make sure to use that value to pre-populate the amount on the interactive form, since this greatly improves the user experience._

I couldn't find any mention of that `amount` parameter in the request fields for withdrawal. Is the `amount` supposed to be a request parameter for `/transactions/withdraw/interactive` or is it passed as an extra field to the interactive URL along with the `callback` field?

JakeUrban:
Based on how the SEP is written, it should be passed to the interactive URL just like the optional `callback` argument. 

The `POST /interactive` endpoints for both deposit and withdrawal are responsible for creating an `incomplete` transaction, and it's the interactive flow's responsibility to collect the KYC, amount, and bank/wallet account information.

I don't think this is a bug, but we could clarify that `amount` can be passed in the same way `callback` is.

msfeldstein:
> If an amount parameter is provided to the interactive URL, make sure to use that value to pre-populate the amount on the interactive form, since this greatly improves the user experience.

I think this line might be vestigial.  I wonder if we should remove it and encourage people to use the existing method of passing parameters to start a transaction, ie a `POST /transaction/withdraw/interactive` call parameter.  If so, we should still document it properly

JakeUrban:
There is no `amount` parameter listed for either `POST .../deposit` or `POST .../withdraw`, so this would be the only way to communicate the amount before rendering the form.

msfeldstein:
This was fixed in #550

PR #530, opened 2020-02-05
Add amount argument to interactive URL parameter list

JakeUrban:
In reaction to this issue: #527 

Referenced under the **Interactive withdrawal** section:
> If an amount parameter is provided to the interactive URL, make sure to use that value to pre-populate the amount on the interactive form, since this greatly improves the user experience.

However, the `amount` parameter isn't listed as a parameter for the interactive URL. I added it to make it more clear that this is permitted.


msfeldstein:
(copied from the issue)

> If an amount parameter is provided to the interactive URL, make sure to use that value to pre-populate the amount on the interactive form, since this greatly improves the user experience.

I wonder if we should remove that line and encourage people to use the existing method of passing parameters to start a transaction, ie a `POST /transaction/withdraw/interactive` call parameter.

msfeldstein:
added as a POST parameter no longer needed here

PR #540, opened 2020-02-19
Add changes to SEP-0018 according to the protocol team discussion

orbitlens:
Discussion follow-up:
- Removed the reference to a "flat.js" parser implementation and optional `convert` parser argument. The standard should provide a single, unified implementation that parses namespaces hierarchy and values unambiguously..
- Added the remark about the reserved "config" root namespace to prevent potential name conflicts in the future.

@MisterTicot Could you please review the changes and update the reference implementation to remove `convert` argument from `read()` method? It should internally use base64 converter that returns raw `Buffer` for values to avoid encoding conflicts across applications and different parser implementations. Horizon API always returns base64-encoded values.

cc @tomerweller

### DIFF
--- a/ecosystem/sep-0018.md
+++ b/ecosystem/sep-0018.md
@@ -6,6 +6,7 @@ Title: Data Entry Namespaces
 Author: MisterTicot <mister.ticot@cosmic.plus>
 Status: Draft
 Created: 2018-10-26
+Updated: 2020-02-18
 ```
 
 ## Simple Summary
@@ -35,7 +36,7 @@ libraries or utilities dealing with data entries.
 ## Specification
 
 Namespace hierarchy is encoded in the key using one or more terms separated by
-dots. Terms are made of lowercase letters, underscore and numbers and mustn't
+dots. Terms consist of lowercase letters, underscore, numbers, and mustn't
 start with a number.
 
 **Examples**
@@ -63,27 +64,30 @@ must be ignored.
 
 **Example: same data tree in JSON**
 
-```js
+```json
 {
-  config: {
-    two_factor: ...,
-    multisig: {
-      server: ...,
-      network: ...,
-      key: ...
+  "config": {
+    "two_factor": "",
+    "multisig": {
+      "server": "",
+      "network": "",
+      "key": ""
     }
   },
-  profile: {
-    name: ...,
-    language: ...
+  "profile": {
+    "name": "",
+    "language": ""
   },
-  wallet: {
-    btc: ...,
-    eth: ...
+  "wallet": {
+    "btc": "",
+    "eth": ""
   }
 }
 ```
 
+The "config" root namespace is reserved exclusively for SEP standards –
+applications shouldn't adopt values and sub-namespaces defined under this
+namespace for vendor-specific use-cases.
 
 ## Rationale
 
@@ -94,7 +98,7 @@ that is valid across a broad range of programming languages.
 
 This SEP uses JavaScript syntax because it is a well-known language. It is also
 familiar to anybody using JSON, which is a format widely used to pass data, and
-the one already used by horizon API.
+the one already used by Horizon API.
 
 Parsing the data tree must lead to the same namespace/key/value structure
 regardless of the programming language being used. This is the reason why term
@@ -133,23 +137,15 @@ data tree and they are still accessible as account data entries.
 Introduction of standard utilities to manage the data tree will provide
 an incentive to comply with the proposed semantic.
 
-## Reference utilities
+## Reference Implementation
 
-JavaScript:
+Reference JavaScript parser implementation can be found
+[here](https://github.com/MisterTicot/stellar-draft-0005/blob/master/tree.js).
+It operates with a tree-like structure representing namespaces and
+subnamespaces.
 
-* **Repository:** https://github.com/MisterTicot/stellar-draft-0005
-* **Minimal implementation:**
-  [flat.js](https://github.com/MisterTicot/stellar-draft-0005/blob/master/flat.js)
-  parse account data entries as a list of keys. It takes 45 lines of code so
-  it's nice for being compact.
-* **Recommended implementation:**
-  [tree.js](https://github.com/MisterTicot/stellar-draft-0005/blob/master/tree.js),
-  parse account data entries as a tree. It takes 70 lines of code. It is not
-  as compact and efficient, but it is dev-friendly.
+This implementation exposes the following methods:
 
-Both implementations expose the following methods:
-
-* `accountData.read(account, [convert])` parse valid namespace entries from
-  **account** and convert them to the desired format.
-* `accountData.write(account, accountData)` generate a transaction that change
-  **account** data entries to reflect **accountData**.
+* `accountData.read(account)` - parse valid namespace entries from an account.
+* `accountData.write(account, dataTree)` - generate a transaction that
+reflects account data entry changes.


MisterTicot:
Looks good - a couple of points to discuss though:

---

I don't get your point about data conversion. Do you mean that the helper should returns raw buffers instead of strings?

(btw I called the code "reference implementation" but at that point it's more a "sample")

---

As noted in the [mailing list thread](https://groups.google.com/forum/#!topic/stellar-dev/8Llk45EkPWw), I think we should discuss briefly canonical namespaces before validating this SEP, as it will be difficult to change those later on.

For now, the proposal is that application-specific data lives in their dedicated `vendor[name]` namespace, while root keys are for SEP-defined use-cases (the proposed `config`, `profile` and `wallet` can only make sense they are standardized across apps).

Giving each app its dedicated namespace is the only reliable way to prevent naming conflict.

MisterTicot:
Ah, wait... I just re-read my code.

So right now it's converting fetched base64 values into raw _Buffer_s by default, but provides the option to specify an alternative `converter`. The idea was to offer a simple way to get the values as UTF-8 strings.

On the encoding side (`toBase64`), it accepts either _Buffer_s or _String_s, in which case it assumes those are UTF-8 encoded.

One issue I see is that, so far, other SEPs making use of data entry namespace are all using string values that have to be somehow encoded into base64. To prevent the encoding conflict you mentioned, we need this SEP to specify a standard format (`ASCII` or `UTF-8`). From there, I'd rather have the helpers return/consume decoded _String_s rather than _Buffer_s.

The downside is that it will make it difficult to support binary data in the future, but I have nothing better at that point.

orbitlens:
> For now, the proposal is that application-specific data lives in their dedicated `vendor[name]` namespace, while root keys are for SEP-defined use-cases (the proposed `config`, `profile` and `wallet` can only make sense they are standardized across apps).

Well, I had the same question. But it seems like people don't care about the prefix. From what I see, app can use `[vendor_name]` root namespace by default. It is not specified in the standard, but we can add the note about using vendor-specific namespaces to make it more clear. 

As for predefined root namespaces, I'm ok with the 3 ns you proposed. There is no feedback or proposals for other root ns. Therefore, I can change the PR to include `profile` and `wallet` alongside with the `config` namespace to the reserved list. Is this ok?

> One issue I see is that, so far, other SEPs making use of data entry namespace are all using string values that have to be somehow encoded into base64. 

All the values are stored in the binary format on the ledger. It doesn't matter what format you are using in the app, the data on the ledger is binary. Base64 is only the text representation of binary data. Horizon uses it to represent binary data. If you provide a text value for a data entry value, it is automatically converted to byte array. It is easy to forget that you are always receive binary data from Horizon response, regardless of what format you used to modify the data entry. I made the same mistake myself.

> To prevent the encoding conflict you mentioned, we need this SEP to specify a standard format (ASCII or UTF-8).

First of all, UTF-8 is backwards compatible and shares the same binary representation with 7-bit ASCII. Different languages use different standards, for example, strings in JS are encoded in UTF-16 internally. In any case, specifying a string standard for encoding **binary** data seems wrong to me.

> From there, I'd rather have the helpers return/consume decoded String rather than Buffer.

If we imply that a vendor app or any future standard may store binary data there (hashes, UUIDs, icons, whatever), it means that using string-based converters may break it. Because conversions like `binary -> UTF-8 -> binary` or `binary -> UTF-16 -> binary` are not always safe, which means that you may end up with wrong data after decoding. 

My main point is that the reference implementation should be bulletproof and always return correct data on parsing. In this particular case, the conversion value to, say, UTF-8 string is quite straightforward: `value.toString('utf8')`.

MisterTicot:
>  From what I see, app can use [vendor_name] root namespace by default.

Seriously, it'll become messy if we mix root namespaces with vendor namespaces at the same level, as root namespace defines the _purpose_.

> All the values are stored in the binary format on the ledger. It doesn't matter what format you are using in the app

Indeed it doesn't. What matters is which encoding you use to map strings to their binary representation. If you say that, for example, `config.memo_type` can be either `true` or `false` or whatever, then you need to tell how those strings gets encoded into their binary representation.

If one implementator encodes `false` using UTF-8 and another decodes it using UTF-16, then the values won't be right and inter-operability brokes.

> If we imply that a vendor app or any future standard may store binary data there (hashes, UUIDs, icons, whatever), it means that using string-based converters may break it.

Vendor namespaces are not a big deal as they can encode things the way they want without compatibility issues.

orbitlens:
> Seriously, it'll become messy if we mix root namespaces with vendor namespaces at the same level, as root namespace defines the purpose.

Ok, how about `x` root namespace for vendors? Like `x.cosmic_plus.settings`?

> If one implementator encodes id using UTF-8 and another decodes it using UTF-16, then the value won't be right and inter-operability brokes. 
Vendor namespaces are not a big deal as they can encode things the way they want without compatibility issues.

I don't see how the custom decoder (`convert` argument) solves this issue. On the contrary, if the value was encoded as UTF-16, and I parse it using, say, ASCII converter, I'll definitely end up with wrong data on my side. If different SEP standards imply different encodings (i.e., text for SEP-29, and binary for some future SEP-100), then it means that I'll have to parse data entries **twice** using the reference implementation – first to retrieve text data, and then to parse binary entries. 

At the same time if are using the content-agnostic ref implementation, the result will be always binary, and I can interpret it as string, hash, or raw buffer depending on the key. 

As for the value comparison, to avoid potential problems it makes sense to use binary comparison, or base64-encoded strings if you like. This will be 100% unambiguous.

MisterTicot:
> Ok, how about x root namespace for vendors? Like x.cosmic_plus.settings?
 
Are you worried about the length? I'm fine with `x`. Maybe `app` would be more explicit.

> I don't see how the custom decoder (convert argument) solves this issue.

It doesn't, that's a separate feature that let you tell which kind of data you want to work with.

> On the contrary, if the value was encoded as UTF-16, and I parse it using, say, ASCII converter, I'll definitely end up with wrong data on my side. 

If we don't specify which encoding implementators must use to encode strings then it's precisely what is going to happen.

To be precise, `text` is not an encoding, UTF-8 is. `text` is a data type, UTF-8 gives a way to encode that data type into a binary values.

> At the same time if are using the content-agnostic ref implementation, the result will be always binary, and I can interpret it as string, hash, or raw buffer depending on the key.

I agree, what bugs me is that this solution isn't very dev friendly. I would prefer a generic interface that hides low-level considerations under the hood, rather than having devs check for the right encoding for each key.

> As for the value comparison, to avoid potential problems it makes sense to use binary comparison, or base64-encoded strings if you like. This will be 100% unambiguous.

Absolutely.

leighmcculloch:
This discussion is definitely highlighting that there is more than one way for applications to think about how they store data, and I think it is getting in the weeds on implementation details.

I think it is important we don't be prescriptive about how applications store data in data entries and focus this SEP on **how to use key names to structure/namespace data**.

It shouldn't matter if applications store data in different ways, or with different encodings. The data entries are binary. And an application ideally won't be reading the data of another application. SEPs that define a data entry will define exactly how that data entry will be encoded, just like how we have just done in [SEP-29].

A standard or library that defines a way to convert a set of data entries into a JSON structure, or native types, does sound really useful, and maybe we should even go as far as using something like JSON schema to make that a more wonderful and customizable experience, but I think we should tackle that independently of namespacing.

The main gain this SEP provides today is future proofing for how to namespace keys.

For that purpose it might be better not to have a reference implementation attached and instead to focus on how to define keys to form a key structure and to leave the rest up to applications.


@tomerweller @tomquisel Would value your thoughts on this ☝️?

[SEP-29]: https://github.com/stellar/stellar-protocol/blob/19db7bef97431d394a7262d901fd2d2363d70918/ecosystem/sep-0029.md#specification

leighmcculloch:
I have opened #544 to concretely demonstrate what I am proposing.

tomerweller:
I agree with @leighmcculloch. This SEP is about namespacing data entry *keys*. The discussion around implementation and data value encoding/decoding is counter productive.

tomerweller:
> I have opened #544 to concretely demonstrate what I am proposing.

This is great. @MisterTicot?

MisterTicot:
So you guys are fine having one SEP using UTF-8 strings and the next one using ASCIIs; and having to specify the encoding for strings within each SEP?

orbitlens:
I don't agree. The original @MisterTicot's proposal is comprehensive and covers a wide range of questions about how applications deal with data entries and store/parse data.
@leighmcculloch's version defines a set of keys reserved for a handful of SEP standards approved by SDF.

leighmcculloch:
>So you guys are fine having one SEP using UTF-8 strings and the next one using ASCIIs; and having to specify the encoding for strings within each SEP?

@MisterTicot In terms of data entry keys, SEP-18 already specifies a character set that is allowed to be used for namespaces and data entry keys is a subset of ASCII and UTF-8.

In terms of data entry values, the only data entry value we have defined in SEPs right now is for `config.memo_required` which contains only a single character `'1'` which has the same bytes representation in ASCII and UTF-8.

I expect SEPs will store text using the ASCII or UTF-8 encoding, but that is data entry value best practices, and not how to use data entry keys to namespace data entries. Applications will have good reasons to use other encodings, or to store data that is not text at all.

MisterTicot:
Ok you guys are right in saying we can't put heavy constraints on encoding here. Whats matters is that the encoding for each SEP data entry is explicit, whether SEP by SEP or globally in SEP-18.

Now that SEP-0029 has been edited I think we should go with @OrbitLens proposal that is keeps values datatype-agnostic - I see no need to strip out parsing rules & sample code from this setup as there's no debate left about how this should be implemented.

I'll make a separate proposal about mapping data type to specific encoding & I'll find another way to build the dev-friendly tool I'm aiming at.

leighmcculloch:
I still think it is worth separating these two concepts into two separate SEPs. It isn't removing these practices, but the first step in moving them into a separate SEP so that namespaces can stand alone. Namespaces are valuable enough that I think they are worthy of that.

>I'll make a separate proposal about mapping data type to specific encoding

The parsing rules that currently live in SEP-18 that go beyond defining the format of a key and discuss how data entries can be interpreted with JavaScript really belongs with that separate proposal I think.

MisterTicot:
I think we have a pretty solid yet concise proposal already and canonical namespaces is only a matter of a few lines. Splitting it seems overkill to me - especially as it comes to make a step backward by removing the core of that SEP which is defining how to represent data trees using data account entries.

tomquisel:
@MisterTicot I think #544 gets the core of the original SEP-18: defining hierarchical key names, which taken together represent a data tree.

I'm a fan of moving forward with #544 to get the uncontroversial part of this proposal finished. It lays the groundwork for a second SEP that can provide example implementations and information on how to deal with interpreting values. It's important to get a solution out the door. I think it's time to disagree and commit.

MisterTicot:
I hope that for future contributions I won't have to wait a year and a half for a review only to see someone cut half of my proposal then add his name on it.

leighmcculloch:
@MisterTicot 💯 That's a great point, thanks for raising. I removed my name in #544. I'm not adding any new concepts or new mechanisms and there's no need for it to be there. 👍 

@tomquisel @MisterTicot @orbitlens What's the best plan to move forward with re-adding the more complex parts of working with data entries in ecosystem apps in a separate SEP?

tomquisel:
@MisterTicot as a next step forward, are you up for drafting a second SEP that fully realizes your vision? That would be valuable.

MisterTicot:
@tomquisel @leighmcculloch 

I don't think I have enough content to justify a new SEP. Here's the remaining element to deal with, ordered by priority:

- Namespace restriction: Defining a space where vendor-specific data lives & reserving the rest of the tree for standard-defined usage. It will prevent situations where a SEP adds a root namespace that is also being used by a vendor.
- Implementator guidelines: corrupted data trees may break software logic in harmful ways, so it's worth thinking & specifying a standard way to recover from those - but that's a matter of two paragraphs so I don't see a separate SEP for this.
- Handling the variety of possible encoding: at that point, it seems clear that we'll need a library that keeps tracks of the expected data structure & encoding to make it easy to do the right thing & hard to do the wrong one.

leighmcculloch:
Namespace restriction is something that could be added to SEP-18.

I think there's plenty of other content in what you mentioned that is worthy of a SEP.

It could detail how an application takes a set of data entries and turns them into a tree data structure would be a good place to start. Assuming that other SEP uses JSON, I think that other SEP could be a good place to explore could something like JSON schema (or similar) be used to handle encoding/decoding concerns.

I also know @stanford-scs has encouraged us to include test cases in our SEPs too so that someone can look at a SEP and they have everything they need to write an implementation that is accurate, so even if the implementor guidelines can be described in a brief paragraph it will be good to include test cases for all the scenarios we're thinking are important. In the end this may not be that short.

MisterTicot:
> It could detail how an application takes a set of data entries and turns them into a tree data structure would be a good place to start. Assuming that other SEP uses JSON, I think that other SEP could be a good place to explore could something like JSON schema (or similar) be used to handle encoding/decoding concerns.

> I also know @stanford-scs has encouraged us to include test cases in our SEPs too so that someone can look at a SEP and they have everything they need to write an implementation that is accurate, so even if the implementor guidelines can be described in a brief paragraph it will be good to include test cases for all the scenarios we're thinking are important. In the end this may not be that short.

I think I'll let you lead that one as it looks like you have a clear direction & I can't afford to spend much more time on this proposal.

leighmcculloch:
This PR has had no activity for a few weeks and I think most of the issues it addresses have been addressed in #544 and there may be future SEPs or SEP changes discussed above that are likely going to be in a separate PR, so I'm going to close this one. @orbitlens Please feel free to reopen this or a new PR if you were still wanting to make changes.

PR #545, opened 2020-02-21
SEP-1: Include SEP-24 in the TRANSFER_SERVER field description

vitchor:


msfeldstein:
This is no longer valid since we have the separate field for sep24

PR #548, opened 2020-02-26
SEP-0018: Add canonical paths & restrictions

MisterTicot:
Defines a few canonical namespaces. The general idea is that root namespaces are tied to a function while child namespace may be tied to a scope, similarly to `/etc`, `/home` & `/lib` in Unix systems.

`profile` & `wallet` are not being used yet but are mentioned in case someone need them for future SEP.

### DIFF
--- a/ecosystem/sep-0018.md
+++ b/ecosystem/sep-0018.md
@@ -6,8 +6,8 @@ Title: Data Entry Namespaces
 Author: MisterTicot <mister.ticot@cosmic.plus>
 Status: Active
 Created: 2018-10-26
-Updated: 2020-02-20
-Version: 0.2.0
+Updated: 2020-02-26
+Version: 0.3.0
 ```
 
 ## Simple Summary
@@ -50,6 +50,20 @@ wallet.eth
 * **Term REGEXP:** `[_a-z][_a-z0-9]*`
 * **Key REGEXP:** `[_a-z][_a-z0-9]*(\.[_a-z][_a-z0-9]*)*`
 
+## Namespace Registry
+
+To guarantee interoperability and avoid naming conflicts, namespaces must be
+used as defined in SEP documents.
+
+- `app`: Space dedicated to vendor-specific keys. For example, "example.org" can
+  define its keys under `app.example_org`.
+- `config`: Space dedicated to ecosystem-wide configuration. Those are values
+  that impact software logic.
+  - `config.memo_required`: [SEP29](sep-0029.md)
+  - `config.multisig`: [SEP19](sep-0019.md)
+- `profile`: Space dedicated to account identification.
+- `wallet`: Space dedicated to external ledger addresses.
+
 ## Rationale
 
 **Syntax**
@@ -83,10 +97,3 @@ ensuring namespaces are not given values. Example:
 profile.language.language = en
 profile.language.region = US
 ```
-
-## Namespace Registry
-
-The following namespaces have special meanings and their use should be reserved
-for those meanings:
-
-- `config`: Keys defined to define behaviors or limitations of an account, for use in SEP standards.
--- a/ecosystem/sep-0018.md
+++ b/ecosystem/sep-0018.md
@@ -40,11 +40,9 @@ start with a number.
 **Examples**
 
 ```
+app.example_org.id
 config.memo_required
-profile.name
-profile.language
-wallet.btc
-wallet.eth
+config.multisig.coordinator
 ```
 
 * **Term REGEXP:** `[_a-z][_a-z0-9]*`
@@ -83,20 +81,20 @@ interpretation between case-sensitive and case-insensitive languages.
 **Identifier/namespace conflict**
 
 Conflict occurs when a term is used both as an identifier and a namespace. If
-for example a data entry with key `profile.language` exists, and a data entry
-with key `profile.language.region` exists, an application parsing the data
-entries into internal types or data structure will struggle to reconcile how to
-store the value assigned to `profile.language`.
+for example a data entry with key `app.example_org.language` exists, and a data
+entry with key `app.example_org.language.region` exists, an application parsing
+the data entries into internal types or data structure will struggle to
+reconcile how to store the value assigned to `app.example_org.language`.
 
 ```
-profile.language = en
-profile.language.region = US
+app.example_org.language = en
+app.example_org.language.region = US
 ```
 
 Applications should avoid defining namespaces that are also identifiers by
 ensuring namespaces are not given values. Example:
 
 ```
-profile.language.language = en
+app.example_org.language.language = en
 profile.language.region = US
 ```
--- a/ecosystem/sep-0018.md
+++ b/ecosystem/sep-0018.md
@@ -56,8 +56,8 @@ To guarantee interoperability and avoid naming conflicts, vendor-specific data
 must live under their reserved namespace:
 
 - `app`: Space dedicated to vendor-specific keys. Keys must be defined under
-  `app.{subdomain}`, with subdomain's dot(s) replaced by underscores. For
-  example, "example.org" defines its keys under `app.example_org`.
+  `app.{domain}`, with dot(s) replaced by underscores. For example,
+  "example.org" defines its keys under `app.example_org`.
 
 The other namespaces must be used according to their SEP documents:
 
--- a/ecosystem/sep-0018.md
+++ b/ecosystem/sep-0018.md
@@ -3,11 +3,11 @@
 ```
 SEP: 0018
 Title: Data Entry Namespaces
-Author: MisterTicot <mister.ticot@cosmic.plus>
-Status: Active
+Author: MisterTicot (@misterticot), Leigh McCulloch (@leighmcculloch)
+Status: Draft
 Created: 2018-10-26
-Updated: 2020-06-09
-Version: 0.5.0
+Updated: 2020-06-30
+Version: 0.3.0
 ```
 
 ## Simple Summary
--- a/ecosystem/sep-0018.md
+++ b/ecosystem/sep-0018.md
@@ -6,8 +6,8 @@ Title: Data Entry Namespaces
 Author: MisterTicot <mister.ticot@cosmic.plus>
 Status: Active
 Created: 2018-10-26
-Updated: 2020-02-26
-Version: 0.3.0
+Updated: 2020-05-06
+Version: 0.4.0
 ```
 
 ## Simple Summary
@@ -52,17 +52,24 @@ wallet.eth
 
 ## Namespace Registry
 
-To guarantee interoperability and avoid naming conflicts, namespaces must be
-used as defined in SEP documents.
+To guarantee interoperability and avoid naming conflicts, vendor-specific data
+must live under their reserved namespace:
 
-- `app`: Space dedicated to vendor-specific keys. For example, "example.org" can
-  define its keys under `app.example_org`.
+- `app`: Space dedicated to vendor-specific keys. Keys must be defined under
+  `app.{subdomain}`, with subdomain's dot(s) replaced by underscores. For
+  example, "example.org" defines its keys under `app.example_org`.
+
+The other namespaces must be used according to their SEP documents:
+
 - `config`: Space dedicated to ecosystem-wide configuration. Those are values
   that impact software logic.
   - `config.memo_required`: [SEP29](sep-0029.md)
   - `config.multisig`: [SEP19](sep-0019.md)
 - `profile`: Space dedicated to account identification.
 - `wallet`: Space dedicated to external ledger addresses.
+
+
+
 
 ## Rationale
 
--- a/ecosystem/sep-0018.md
+++ b/ecosystem/sep-0018.md
@@ -4,7 +4,7 @@
 SEP: 0018
 Title: Data Entry Namespaces
 Author: MisterTicot (@misterticot), Leigh McCulloch (@leighmcculloch)
-Status: Draft
+Status: Active
 Created: 2018-10-26
 Updated: 2020-06-30
 Version: 0.3.0
--- a/ecosystem/sep-0018.md
+++ b/ecosystem/sep-0018.md
@@ -96,5 +96,5 @@ ensuring namespaces are not given values. Example:
 
 ```
 app.example_org.language.language = en
-profile.language.region = US
+app.example_org.language.region = US
 ```
--- a/ecosystem/sep-0018.md
+++ b/ecosystem/sep-0018.md
@@ -53,15 +53,15 @@ config.multisig.coordinator
 To guarantee interoperability and avoid naming conflicts, app-specific data
 must live under their reserved namespace:
 
-- `app`: Space dedicated to app-specific keys. Keys must be defined under
+- `app`: Namespace available to app-specific keys. Keys must be defined under
   `app.{domain}`, with dot(s) replaced by underscores. For example,
   "example.org" defines its keys under `app.example_org`.
 
-The other namespaces must be used according to their SEP documents. Whenever a
-SEP makes use of account data entries, it must be listed here:
+Other namespaces have special meaning according to their SEP documents.
+When a SEP makes use of account data entries, it must be listed here and
+referenced:
 
-- `config`: Space dedicated to ecosystem-wide configuration. Those are values
-  that impact software logic.
+- `config`: Namespace available for use by ecosystem-wide configuration:
   - `config.memo_required`: [SEP29](sep-0029.md)
   - `config.multisig`: [SEP19](sep-0019.md)
 
--- a/ecosystem/sep-0018.md
+++ b/ecosystem/sep-0018.md
@@ -52,10 +52,10 @@ wallet.eth
 
 ## Namespace Registry
 
-To guarantee interoperability and avoid naming conflicts, vendor-specific data
+To guarantee interoperability and avoid naming conflicts, app-specific data
 must live under their reserved namespace:
 
-- `app`: Space dedicated to vendor-specific keys. Keys must be defined under
+- `app`: Space dedicated to app-specific keys. Keys must be defined under
   `app.{domain}`, with dot(s) replaced by underscores. For example,
   "example.org" defines its keys under `app.example_org`.
 
--- a/ecosystem/sep-0018.md
+++ b/ecosystem/sep-0018.md
@@ -6,8 +6,8 @@ Title: Data Entry Namespaces
 Author: MisterTicot <mister.ticot@cosmic.plus>
 Status: Active
 Created: 2018-10-26
-Updated: 2020-05-06
-Version: 0.4.0
+Updated: 2020-06-09
+Version: 0.5.0
 ```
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0018.md
```diff
@@ -50,6 +50,20 @@ wallet.eth
...
+used as defined in SEP documents.
+
+- `app`: Space dedicated to vendor-specific keys. For example, "example.org" can
+  define its keys under `app.example_org`.
+- `config`: Space dedicated to ecosystem-wide configuration. Those are values
+  that impact software logic.
+  - `config.memo_required`: [SEP29](sep-0029.md)
+  - `config.multisig`: [SEP19](sep-0019.md)
+- `profile`: Space dedicated to account identification.
+- `wallet`: Space dedicated to external ledger addresses.
```
leighmcculloch: Are there wallets or apps that are using these values today?

If not I think we should hold off adding namespaces until there are SEPs or use cases that are going to use them.

If so, could you point to some examples so it's easier to understand how they're using data entries and these fields?

MisterTicot: The point is precisely to prevent naming conflict by providing indication of how applications should use those namespaces (that is, putting vendor stuff under `app`). If we wait for them to implement their own things it's pointless.

leighmcculloch: Got it, that makes sense. I think that makes sense for `app`, I have some more questions about `profile` and `wallet` I'll post in a new thread.

ecosystem/sep-0018.md
```diff
@@ -50,6 +50,20 @@ wallet.eth
 * **Term REGEXP:** `[_a-z][_a-z0-9]*`
 * **Key REGEXP:** `[_a-z][_a-z0-9]*(\.[_a-z][_a-z0-9]*)*`
 
+## Namespace Registry
+
+To guarantee interoperability and avoid naming conflicts, namespaces must be
+used as defined in SEP documents.
+
+- `app`: Space dedicated to vendor-specific keys. For example, "example.org" can
+  define its keys under `app.example_org`.
```
leighmcculloch, line 59: 👍 I like that the `app` namespace has an example of how an app should use the namespace by specifying a sub-namespace that is a domain replacing `.`s with `_`. That seems simple and straight forward. Could we specify this?

With the 64-byte data entry key length limitation only apps with domains less than 60 characters in length will be supported. That's probably fine given 60 characters is very long and most websites or apps don't have domains that are that long.

ecosystem/sep-0018.md
leighmcculloch, line 65: It's not clear to me what fields should be within `profile` or `wallet`.

Could you share an example of what fields they'll be?

I'm guessing they'll be fields shared by applications so to prevent naming conflicts and duplication of data in those shared fields the fields in these two namespaces will need to be specified too. I think we'd benefit from having a wallet or service chime in on what fields they'd want in these namespaces and if they'd use them.

MisterTicot: The idea is that fields in those namespaces need to be part of a SEP before we use them. (as for any namespace other than `app`)

`Profile` would definitely require its own SEP. It could be a place where to put the usual profile things such as an icon & a pseudo to improve inter-operability/harmonization between Stellar-based services. For more serious things, it could be used for GPG keys, KYC or for proofs such as the ones Keybase uses for domains/accounts/...
Examples: `profile.icon`, `profile.gpg`, `profile.domain`, ...

`Wallet` is to link external wallets from other chains. This is meant to be used by anchors as well as social/tipping applications.
Examples: `wallet.btc`, `wallet.eth`, `wallet.xrp`...

> I think we'd benefit from having a wallet or service chime in on what fields they'd want in these namespaces and if they'd use them.

Definitely

ecosystem/sep-0018.md
MisterTicot: > Could we specify this?
Done

ecosystem/sep-0018.md
leighmcculloch: cc @tyvdh

ecosystem/sep-0018.md
```diff
@@ -50,6 +50,27 @@ wallet.eth
...
 * **Key REGEXP:** `[_a-z][_a-z0-9]*(\.[_a-z][_a-z0-9]*)*`
 
+## Namespace Registry
+
+To guarantee interoperability and avoid naming conflicts, vendor-specific data
+must live under their reserved namespace:
+
+- `app`: Space dedicated to vendor-specific keys. Keys must be defined under
+  `app.{subdomain}`, with subdomain's dot(s) replaced by underscores. For
+  example, "example.org" defines its keys under `app.example_org`.
```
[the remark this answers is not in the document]
MisterTicot: Subdomain is a well-defined term meaning that URLs such as `app.example.org` may be used as well. Subdomain is the base unit in web security, for example regarding local storage, cookies, SSL certificates, permissions.... Hence, it makes sense to use the same unit here.

As an example, GitHub pages relies on these properties to offer a well secured `stellar.github.io` page that is fully separated from activity on other `*.github.io` pages.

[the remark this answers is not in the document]
leighmcculloch: It is common to refer to a full domain as just a domain. `app.example.org` is a domain, and it is also a subdomain or `example.org`. `example.org` is a domain, and it is also a subdomain of `org`. These definitions come from [RFC1034 Section 3.1](https://tools.ietf.org/html/rfc1034#section-3.1). This [wikipedia page](https://en.wikipedia.org/wiki/Subdomain) makes this clear too when it says a subdomain is a domain.

If we specify `domain` we're saying the value could be `example.org` or `mysubdomain.example.org` or any other subdomain path like `my.subdomain.for.example.org` and we do not need to be more specific and require the value have to be a subdomain since that is just more specific definition of a domain.

Can we make that change above?

ecosystem/sep-0018.md
leighmcculloch: @MisterTicot I'm still interested on input from @tyvdh on this, but given that we agree it would be worthwhile to get input about `profile` and `wallet` from wallets, if we agree on adding the app namespace I think we should move profile and wallet to a separate PR and ideally have some input from others for that. What do you think?

tyvdh: I'm definitely on the side of not adding things which aren't being asked for or used in other forms elsewhere. I'm also ultimately not convinced adding lots of data entries is a better route than using an external reference like the TOML file. A standard format seems like a better SEP than a standard storage location. Storing lots of data attributes will get expensive for user's or wallets and with the fine grained options you're talking about, it seems to me like storing these options outside the ledger might be a better option.

leighmcculloch: Thanks @tyvdh, that's a great point. Do you think this applies to the `app` namespace too or just the `profile` and `wallet` ones?

tyvdh: I do yes. However it would seem that SEP-18 has already been accepted so it may not be worth transitioning the app namespace to an off-chain standard. My guess however is that we could probably move this all to a TOML storage solution without much if any issues.

leighmcculloch: It's already been accepted but the only registered namespace right now is the `config` namespace which is specifically for use in SEPs, e.g. [SEP-29](https://github.com/stellar/stellar-protocol/blob/f3768980b558501e66cf9bd28c38bae248ae2252/ecosystem/sep-0029.md#specification).
https://github.com/stellar/stellar-protocol/blob/f3768980b558501e66cf9bd28c38bae248ae2252/ecosystem/sep-0018.md#L87-L92
This PR is the first introduction of `app`, `profile`, `wallet` into the registry. Technically they are mentioned in an example in the existing version of the SEP but that's just an example demonstrating how the SEP could be used and they are not in the registry at the bottom of the document.

MisterTicot: Quickly passing by before we stray away.

About the on-chain vs off-chain thing, naturally, this SEP is relevant _iff_ a distributed consensus is required for an application to function.

That's the case of `isMemoRequired` or signature coordination setup (SEP-19). You simply can't provide an equal level of robustness by using a TOML file.

For centralized applications, it goes without saying that off-chain storage must be preferred.

MisterTicot: Back to the main point, I'd agree on dropping `wallet` & `profile` until it gets specified in another document. I think we need to keep `app` though, as a way to keep vendors data contained away from the global namespace.

ecosystem/sep-0018.md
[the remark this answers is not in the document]
MisterTicot: Thanks for the explanation. Let's do that then.

ecosystem/sep-0018.md
tyvdh: My fear is this SEP will never gain adoption if it locks up funds in order to comply. I don't see a real reason to include this stuff on chain. `isMemoRequired` is a very specific use case and a response to how wallets are already used. This feels proactive in a way it doesn't need to be. Unless there's conversation and requests out there I'm unaware of. Would it be possible to get some wallet makers to chip in here?

MisterTicot: @tyvdh You can check [sep19](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0019.md) & [sep21](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0021.md) for other example of this.

The question of whether or not we need data on-chain has been much debated prior to the acceptation on this SEP, which got accepted since `isMemoRequired` is based on it.

This SEP is meant as a way to prevent namespace conflict. It doesn't mean that every user out there is going to have dozens of data entries defined.

leighmcculloch: I think the reason to keep some data on-chain under the app namespace would be if a single service wanted accounts to register and store some information on their account, but if a single website/product is doing this they could also use SEP-10 to authenticate the account and then store the information off chain with the user authenticated.

The main value is where there is some common field that will be implemented with by many products at which point that interface is likely to become a SEP, and at that point the value could live under `config` like SEP-29.

As long as centralized apps keep their values outside the config namespace they are unlikely to clash with SEPs.

MisterTicot: The premise is that the global namespace (`config`, `wallet`, `profile`...`) is entirely reserved for SEPs. Not everything standard is meant to go under `config`.

This is similar to many computing system where the global namespace is a shared space that is meant to be kept unpolluted, while each modules gets its own dedicated space.

This is because vendor systems need to be compartmentalized anyway as else they would risk to conflict between each other.

MisterTicot: > I think the reason to keep some data on-chain under the app namespace would be if a single service wanted accounts to register and store some information on their account, but if a single website/product is doing this they could also use SEP-10 to authenticate the account and then store the information off chain with the user authenticated.

This is not the same: data stored on-chain can be mutated only with account signers consensus. That's something off-chain storage can't provide. on-chain storage retains the _trustless_ quality distributed ledgers offers, while off-chain storage introduces a single point of failure.

You need that on-chain storage for specific use case that are related to decentralized applications or account integrity.

For exemple, when it comes to multisig coordination, storing the setup on-chain is a matter of account integrity: it makes sense that signers consensus is required to required the signature coordination setup.

tyvdh: Hmm. I think describing it as a single point of change is better than single point of failure and your flow is only true in a multisig scenario as the data can be changed just as easily if not more so when on chain than in a toml file if a wallet only has one signer. 

I think I can be on board for the app domain specs but wallet and profile seem preemptive and I'm not convinced they're better on-chain. We have a spec for storing meta data in TOML files and it works great and is widely used, I don't think your concerns are founded unless there are others who can chime in to support the need for `profile` and `wallet` metadata being stored on-chain.

Who else is asking for this change? Do they also feel strongly that account metadata should be stored on-chain?

There's also a difference between signing for a change and pushing data to the ledger. You could store signed data off-chain which couldn't be changed without a valid signature from the account. Just because something isn't on chain doesn't mean you can't be sure an account signed for it. So there are alternatives to on-chain which accomplish alleviating your stated concerns.

I don't want to beat a dead horse but I don't think it's unreasonable to query for alternatives and seek out other voices who you envision implementing and making use of this feature to chime in so we don't push standards in isolation which ultimately would not be implemented.

tyvdh: Ultimately my vote is to simply include some clarifying language around cases where new namespaces may or should be created in conjunction with other SEPs. As other SEPs are made and wish to make use of on chain storage they should consider using the namespace which makes the most sense. So rather than creating an official list here preemptively we create the list as SEPs require it by looking at SEP-18 and realizing there can/should be other name spaces besides just `config.`

MisterTicot: Ok then, I edited the document accordingly.

tyvdh: Is there a SEP using `app` yet? @leighmcculloch would you be using `app` for the work you're doing? Curious if it makes sense to have that included in this PR vs just update this SEP-18 as other SEPs are or should actually be using name spaces besides `config`. A registry is a great idea once names are in use but proactively I'm not sure is the right move. Curious though as I definitely see everyone potentially just force fitting `config` to their use case if more aren't at least suggested first. Thoughts?

MisterTicot: `app` is not meant to be used by any SEP. That's specifically meant for applications that can't/won't/don't want to go through the standardization process.

For example, a service wants to trustlessly record accounts KYC sha256sum for interoperability between applications relying on the said service. Standardizing into a SEP may not be a priority for them - as their initial goal is to get that MVP out.

Then they can put that data under `app.awesomekyc_com.id`.

> Curious though as I definitely see everyone potentially just force fitting `config` to their use case if more aren't at least suggested first. Thoughts?

I think we can simply enter the discussion whenever that happens.

leighmcculloch: I'm not adding any data entries to any SEPs at the moment, but if we were adding them in a SEP we'd put them in config. I think @MisterTicot `app` is only useful in centralized use cases. Maybe having the namespace will just provide some organization so that folks who do want to use data entries for their application data will put it in one place that is clearly out of the way or anything else. I don't feel strongly about adding `app` but I'm not opposed to it. I agree we should defer `profile` and `wallet` for now.

ecosystem/sep-0018.md
```diff
@@ -50,6 +50,23 @@ wallet.eth
 * **Term REGEXP:** `[_a-z][_a-z0-9]*`
 * **Key REGEXP:** `[_a-z][_a-z0-9]*(\.[_a-z][_a-z0-9]*)*`
 
+## Namespace Registry
+
+To guarantee interoperability and avoid naming conflicts, vendor-specific data
+must live under their reserved namespace:
+
+- `app`: Space dedicated to vendor-specific keys. Keys must be defined under
```
leighmcculloch, line 58: Let's refer to this as app specific rather than vendor specific since the field is named app it means introducing less concepts and vocabulary that the reader must understand. It's a small factor.
```suggestion
To guarantee interoperability and avoid naming conflicts, app-specific data
must live under their reserved namespace:

- `app`: Space dedicated to app-specific keys. Keys must be defined under
```

ecosystem/sep-0018.md
```diff
@@ -83,10 +100,3 @@ ensuring namespaces are not given values. Example:
 profile.language.language = en
 profile.language.region = US
```
leighmcculloch, line 101: I think we should change this example to use the `app.example_org` namespace so that profile doesn't exist in this document.

ecosystem/sep-0018.md
```diff
@@ -40,16 +40,31 @@ start with a number.
...
+  `app.{domain}`, with dot(s) replaced by underscores. For example,
+  "example.org" defines its keys under `app.example_org`.
+
+The other namespaces must be used according to their SEP documents. Whenever a
+SEP makes use of account data entries, it must be listed here:
+
+- `config`: Space dedicated to ecosystem-wide configuration. Those are values
+  that impact software logic.
+  - `config.memo_required`: [SEP29](sep-0029.md)
+  - `config.multisig`: [SEP19](sep-0019.md)
```
leighmcculloch, line 66: I think the language of this section can be a bit softer as it was prior to this change. I've updated it to be similar to before, but retained the additional information that you're adding in this section.

```suggestion
The following namespaces have special meanings and their use should be reserved
for those meanings:

- `app.{domain}`: Namespace available to any domain operating an application and
  storing data entries. Data entry keys can be defined under `app.{domain}`, with
  dots in the domain replaced by underscores. For example, the website or app at
  domain `example.org` can define its data entry keys with namespace
  `app.example_org`.

- `config`: Namespace available for use by ecosystem-wide account configuration where data entry entries define behaviors or limitations of an account, for use in SEP standards. Known data entries in SEPs:
  - `config.memo_required`: [SEP29](sep-0029.md)
  - `config.multisig`: [SEP19](sep-0019.md)
```

ecosystem/sep-0018.md
```diff
@@ -66,27 +81,20 @@ interpretation between case-sensitive and case-insensitive languages.
...
+app.example_org.language.region = US
 ```
 
 Applications should avoid defining namespaces that are also identifiers by
 ensuring namespaces are not given values. Example:
 
 ```
-profile.language.language = en
+app.example_org.language.language = en
 profile.language.region = US
```
leighmcculloch, line 99:
```suggestion
app.example_org.language.region = US
```

leighmcculloch:
@MisterTicot Could you accept the suggestions above, and also merge master into this branch so we can see the final document here in this PR?

I was about to merge this SEP but it looks like this SEP is now active and no longer a draft. I think ideally the addition of `app` wouldn't happen in an Active SEP given it is a proposed namespace that nobody is currently using. It would be a separate draft SEP that adds a new namespace and defines that new namespace specific use. It would remain a draft SEP until there was some ecosystem use of the data key. There's also some downside to doing this as it would become disconnected from SEP-18, so I could also see the argument of just adding it here but I wonder how we do that in a way that we can remove `app` later if we find out nobody is using it.

ecosystem/sep-0018.md
MisterTicot: Haven't the words "must", "should" and "can" a precise meaning that must be respected when writing specifications?

leighmcculloch: You're right that's a good point, it's better to not soften the language and use those terms. I removed the changing of those terms from my suggestion here. Is this fine with you? If so can you apply this suggestion and I'll merge the PR.
```suggestion
To guarantee interoperability and avoid naming conflicts, app-specific data
must live under their reserved namespace:

- `app`: Namespace available to app-specific keys. Keys must be defined under
  `app.{domain}`, with dot(s) replaced by underscores. For example,
  "example.org" defines its keys under `app.example_org`.

Other namespaces have special meaning according to their SEP documents.
When a SEP makes use of account data entries, it must be listed here and
referenced:

- `config`: Namespace available for use by ecosystem-wide configuration:
  - `config.memo_required`: [SEP29](sep-0029.md)
  - `config.multisig`: [SEP19](sep-0019.md)
```

PR #549, opened 2020-02-26
[SEP-1] Add SEP-24 Transfer Server Field

vitchor:
Add "TRANSFER_SERVER_SEP0024" to the SEP-1 Account Information fields. This can be used by wallets to differentiate between anchors (or on/off ramps) that implement SEP-0006 and SEP-0024.


msfeldstein:
We should update this in the SEP24 spec as well

patrickxb:
This looks good to me.

PR #550, opened 2020-02-26
[SEP24] Add explicit amount field to api

msfeldstein:
To make a nice UX many wallets wish to collect the amount of deposit/withdrawal ahead of time to show fees before going through the whole interactive process.  We had originally loosely suggested providing it to the interactive url, but that doesn't help calculate fees ahead of time, and wasn't well documented.  Here we add it to the params expected in the initial POST calls.


JakeUrban:
Should anchors still show the form with the amount submitted in the POST request pre-populated, like we do when its passed to the interactive URL today? Or should it skip the form altogether?

PR #555, opened 2020-02-28
Simplify SEP process

rice2000:
**This is not ready to merge.  It's here to kick off a discussion.**

The goal of the PR is to simplify the SEP process so it's easier to understand and manage.   The biggest changes: 

* All drafts will remain as pull requests during discussion and revision.  That way, discussion is centralized in an obvious place, and authors can respond to feedback and easily push revisions.
* As a result, there's no longer a need for `Awaiting Decision`, `FCP`, or `Rejected` statuses. 
* A draft will only be merged and marked `Active` after there's a reference implementation.
* `Tracks` are out.  The distinction between `Informational` and `Standard` is difficult to grasp, and not all that useful.
* `Final` status is removed.  The ecosystem has yet to evolve to a state where we need it, and it corners us when we need to revise something based on real-world feedback.  For now, `Active` is the end state
* `Superseded` status is removed.  We'll just use `Deprecated` and note the superseding SEP.

There are still some open notes and questions:
* How do we handle versioning?
* What's the process for deprecation?
* What do we do with the SEPs that are currently in the `Draft` stage?

### DIFF
--- a/ecosystem/README.md
+++ b/ecosystem/README.md
@@ -1,171 +1,98 @@
 # Stellar Ecosystem Proposals (SEPs)
-
-## SEP Tracks
-* **Informational** — A SEP on the `Informational` track is one that is open to adoption by the
-  ecosystem, but has not been formally standardized by SDF, and is not endorsed by SDF for
-  adoption. Typically a SEP can start as `Informational` to gain traction within the ecosystem
-  before moving to the `Standards` track.
-* **Standard** — A SEP on the `Standards` track is one that aims for formal standardization and
-  endorsement by SDF for adoption. Typically a SEP Standard has a higher bar towards acceptance,
-  and it requires approval by 2 SDF members of the SEP Team.
+Stellar Ecosystem Proposals are open protocols that explain how to build interoperable infrastructure on top of the Stellar network. Each SEP addresses a specific problem, and walks through the motivation, specification, and design rationale for a suggested solution. By creating common standards and structures for implementation, SEPs allow broad interaction between ecosystem participants. 
 
 ## SEP Status Terms
-* **Draft** - A SEP that is currently open for consideration and actively being discussed.
-* **Awaiting Decision** — A mature and ready SEP that is ready for approval by the SEP
-  Team. If enough the approval requirements are met by SEP team members, the SEP will move towards 
-  `FCP`. Otherwise, it'll regress to a `Draft`.
-* **FCP** — A SEP that has entered a Final Comment Period (FCP). After one week has passed, during
-  which any new concerns should be addressed, the SEP's status will become `Active`.
-* **Active** - An actively maintained SEP that is intended for immediate adoption by the entire
-  ecosystem. Additional updates may be made without changing the SEP number.
-* **Final** - A finalized SEP will not be changed aside from minor errata. For a proposal to be a candidate to be made Final it must be being used in live products.
+* **Draft** - A proposal that is currently in discussion and undergoing revision.  Drafts remain as pull requests until approved.
+* **Active** - A proposal intended for adoption by the entire ecosystem.  Before a SEP is marked active, there has to be a functioning prototype or reference implementation.
+* **Deprecated** - A SEP that was previously Active, but is no longer suggested for use. 
 
-### Additional Statuses
-* **Deprecated** - A SEP that was previously on an active track but has been deprecated and is no longer suggested for use. There may be legacy usage of a deprecated SEP.
-* **Rejected** - A Standards SEP that has been formally rejected by the SEP Team, and will not be
-  implemented.
-* **Superseded: [New Final SEP]** - A SEP that which was previously final but has been superseded
-  by a new, final SEP. Both SEPs should reference each other.
+## Active Proposals
 
-## List of Proposals
+| Number | Title | Author |
+| --- | --- | --- |
+| [SEP-0001](sep-0001.md) | stellar.toml specification | SDF |
+| [SEP-0002](sep-0002.md) | Federation Protocol | SDF |
+| [SEP-0004](sep-0004.md) | Tx Status Endpoint | SDF |
+| [SEP-0005](sep-0005.md) | Key Derivation Methods for Stellar Accounts | SDF |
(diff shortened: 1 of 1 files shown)


ecosystem/README.md
```diff
@@ -1,171 +1,98 @@
...
-- Choose a track to propose your idea on. The bar for accepting an `Informational` SEP is much
-  lower than one for a `Standard`, and allows you to promote the SEP independently to gain feedback
-  and traction before creating a Standard out of it.
-- Gather feedback from discussion on the dev mailing list and other forums, and utilize it to begin
-  a draft proposal.
-- Follow the proposal process listed below. If you're having difficulty moving the proposal
-  forward, talk to the buddy that's assigned the SEP; they'll often have guidance on how to move
-  things forward, as well as feedback regarding feasibility and how the proposal does or does not
-  align with the Stellar Network's goals.
+* **Draft** - A proposal that is currently in discussion and undergoing revision.  Drafts remain as pull requests until approved.
```
leighmcculloch: I was really into this idea of keeping drafts in PRs when we talked about this a few weeks ago, but it occurred to me that having such a different process for SEPs compared to CAPs is confusing and likely to mean any process we document will not be something we follow consistently.

It's also really beneficial if we allow drafts to be iterated on in a central place that's accessible to everyone making contributions and while a PR makes that possible to a degree, it doesn't encourage it.

I like @stanford-scs's recent suggestion that we keep drafts fluid and quickly update-able. I think we absolutely can use PRs as the place to discuss a draft if the author chooses to by specifying a [Discussion URL](https://github.com/stellar/stellar-protocol/blob/072792efc9c0790742d0dc275a9ba63a39063603/sep-template.md#L10) like we already have in our templates. If we ensure we always use PRs for them, anyone can always comment on updates even if a PR has been merged, or if a PR merges too quickly they can even open a PR themselves proposing another direction for a SEP. Or an author can choose to have discussion in the mailing list or anywhere else of their choosing.

I think the issue that led, at least me, to think drafts should stay in PRs is that looking at the repo I've struggled to tell which SEPs/CAPs are drafts vs concrete things to rely on. But that's because for the longest time I didn't realize there was a README.md in the ecosystem directory (it is pushed so far down the page because the directory has so many files in it). #556 attempts to help others not make the same mistake I did.

We could make it clearer that a SEP is a draft by not giving it a number until it moves out of draft, this is something RFCs do. But that also has some downsides with git because renaming a file can make it harder to follow its history, so I don't think it's worth going that far.

cc @rice2000 @msfeldstein @tyvdh

ecosystem/README.md
```diff
@@ -1,171 +1,98 @@
...
+When the discussion tapers off, the SEP team will send out a notification to the stellar dev mailing list asking for final comments. 
+
+### Build Reference Implementation or Prototype
+Before a SEP is merged and deemed active, someone has to create a prototype or reference implementation demonstrating its viability.  **Add more info here about what that means**
 
+### Merge to Active 
+Once the SEP team has reviewed and approved the draft and implementation, they will merge the SEP, add a one-word name to end of file name, assign a number, and add it to the readme table.  At this point, a SEP is ready for others to implement.
+
+### Versioning
+**Note: need to figure out how to handle versioning**
```
leighmcculloch: Because we've removed Final there is no status that prevents a SEP from accepting breaking changes. I think this ties into us needing to figure out versioning. The RFC/IETF approach is after it moves out of draft it becomes final, that's it, nothing changes, not even typos. Breaking changes must be new drafts that can supersede or be an addendum to a previous RFC.

The semver approach might also work, but it can be really hard to quickly compare two major versions of a SEP in a browser or to communicate that it is okay to have both active at the same time.

I think we could use the semver approach but introduce major versions into SEP name/numbers in similar way that the Go community has embraced semantic import versioning where the major version past version 1 becomes part of the name. E.g., if we were to relive SEP-6/24, we could say:
- SEP-0006 (the v1 release of the transfer server)
- SEP-0006-v2 (the v2 release of the transfer server).

This approach has many advantages:
- Files stay organized and near each other as new major versions arrive.
- Major versions can co-exist _clearly_.
- Older major versions can continue to receive patches/improvements if needed.
- The existence of multiple major versions if easily discoverable by someone new to a SEP.
- Older major versions can be deprecated clearly by moving them to deprecated status.

cc @rice2000 @msfeldstein @tyvdh

rice2000: This suggestion makes a lot of sense to me, and I'm all for merging it into this PR.

stanford-scs: I like the idea of SEP-0006-v2 for minor, backwards compatible revisions.  However, I think the 0006 should be the "major number."  In the sense that if we switch to something completely incompatible, it should have a different number.  That would be a nice compromise between the IETF RFC process, where you can't even fix typos except by putting out separate errata documents, and full-on semver where the same software package could two completely different things between major versions.

I'm not crazy about using PRs for drafts, though, because it's kind of annoying to make PRs on PRs.  I suggest we just put drafts in a different directory, and move them to the core or ecosystem directory when they are accepted.  We could also name drafts like internet-drafts instead of assigning them numbers initially.

leighmcculloch: For minor backwards compatible changes we should just edit the existing doc I think. We do that already and it hasn't caused us any problems as far as I know? E.g. We added instructions for how to use phone numbers as Stellar federated address usernames to [SEP-2](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0002.md#phone-numbers).

The motivation for keeping new major versions alongside that are addressing the same problem and still in the same spirit of the original SEP is to make it easily discoverable. It'd still be a new number that just happens to have a `v` in it. We could drop the v? What problems do you see building on the same number to cause?

If a SEP was being changed in a huge way that it wasn't in the same spirit as the original we could go an entirely new number for sure.

I was thinking about the internet drafts idea too to be like the RFC process, but the thing I like about giving it a number upfront is we can start referring and linking to it before it's a standard and those links don't break when it becomes a SEP. Maybe that will just encourage us to get out of draft sooner. 🤔

leighmcculloch:
I'm closing this issue because I'm going to revisit these changes with input from others, after collecting more data.

PR #561, opened 2020-03-01
SEP7: remove `sequence=0` replacement

MisterTicot:
Removes the following sentence:

> If the source account is set and the sequence number is 0 then the URI handler should replace the sequence number before signing.

Now that the `replace` parameter has been added, this feature is not needed anymore. Worse, it conflicts with [SEP-0010](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0010.md) which requires the sequence to be 0, making it impossible to log-in through SEP-0007.

@nikhilsaraf @nebolsin @tomquisel @leighmcculloch


nikhilsaraf:
This change makes sense, thank you for proposing it @MisterTicot 

we will need wallets to chime in to see if there are any issues in having them implementing this. I think this is the correct change given the compatibility issue with SEP-10 and we should merge the change in the next few days once everyone has had a chance to share their opinion. 

@patrickxb @andywer @dmgmyza @pitsevich

Keybase has one related issue, which may or may not be addressed yet based on the conversation flow here: https://github.com/keybase/stellarnet/issues/85#issuecomment-592213772

andywer:
Fine from our site 👍

pselden:
This makes sense! LGTM. Best to change it now before it gets too widely implemented and is harder to change.

leighmcculloch:
I agree, sounds like this is not compatible with SEP-10 and it'd be good to remove this statement. This is technically a breaking change though and this SEP has a status of `Final`. Does this SEP have the wrong status? Should it be in Draft or Active state?

nikhilsaraf:
@leighmcculloch  I can change it to `Active` to account for any more changes that may come up. If we find there are no more changes then we can change it to Final at that point.

nikhilsaraf:
@MisterTicot can you give edit rights to maintainers for this PR?

MisterTicot:
@nikhilsaraf For some weird reason I can't, as the box doesn't appear for that repository.

nikhilsaraf:
@MisterTicot can you please change the following in the `Preamble` section?

```
Status: Active
Updated: 2020-03-03
Version: 2.0.0
```

Thank you 🙏

MisterTicot:
@nikhilsaraf Done. I found how to fix this edit rights issue for next PRs but unfortunately I can't do it for this one.

PR #564, opened 2020-03-03
Update SEP-25 with detail on lumen requirement

howardtw:
This PR adds the limitations section as well as corrects and adds the comments in the code snippet. It makes the logic on the provider side align with the one on the client side.

### DIFF
--- a/ecosystem/sep-0025.md
+++ b/ecosystem/sep-0025.md
@@ -39,7 +39,7 @@ After negotiating the off-chain deposit details, the Provider will create the ho
 ```
 # Providers Transaction:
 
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0025.md
```diff
@@ -39,7 +39,7 @@ After negotiating the off-chain deposit details, the Provider will create the ho
 ```
 # Providers Transaction:
 
-Create the holding account
+Create the holding account. HOLDING_LUMENS has to be at least 2 xlm in order to accommodate two entries, one for the extra signer, one for the trustline. If provoders would like be responsible for the balance required to establish the trustline for clients' accounts, HOLDING_LUMENS has to be at least 2 + 0.5 xlm + small amount of transaction fee.
```
msfeldstein: s/provoders/providers

PR #566, opened 2020-03-05
SEP-0030: Change the list of identities to be an arbitrary list

leighmcculloch:
### What
Change the fixed set of identities to be a list and make the roles arbitrary strings rather than a fixed set.

### Why
To allow for easier expansion in the future of both additional roles, multiple identities per role, and multiple values for each identity type (email, phone, etc).

### DIFF
--- a/ecosystem/sep-0030.md
+++ b/ecosystem/sep-0030.md
@@ -35,7 +35,7 @@ The registration flow is as follows:
 - The client uses [SEP-10] to get a [JWT] proving high threshold control of an account.
 - The client, authenticated as the account, registers with the server, providing:
   - One or more identities of individuals that can request the server sign transactions for the account, and for each identity:
-    - The identities role in the client's use case for later identifying an authenticated client's role in relation to the account.
+    - The identities role in the client's [use case].
     - The authentication methods that can be used to authenticate as this identity:
       - One or more other Stellar addresses.
       - And/or, one or more phone numbers.
@@ -609,6 +609,7 @@ Recoverysigner is implemented by the Stellar recoverysigner service, [recoverysi
 [`External`]: #authentication
 [Common Fields]: #common-fields
 [Motivation]: #motivation
+[use case]: #use-cases
 
 [SEP-10]: sep-0010.md
 [JWT]: https://tools.ietf.org/html/rfc7519
--- a/ecosystem/sep-0030.md
+++ b/ecosystem/sep-0030.md
@@ -7,6 +7,7 @@ Author: Leigh McCulloch <@leighmcculloch>, Lindsay Lin <@capybaraz>
 Track: Standard
 Status: Draft
 Created: 2019-12-20
+Updated: 2020-03-11
 Version: 0.2.0
 ```
 
--- a/ecosystem/sep-0030.md
+++ b/ecosystem/sep-0030.md
@@ -156,7 +156,7 @@ Name | Type | Description
   "identities": [
     {
       "auth_methods": [
-        { "type": "account", "value": "GDUA..." },
+        { "type": "stellar_address", "value": "GDUA..." },
         { "type": "phone_number", "value": "+10000000001" },
         { "type": "email", "value": "person1@example.com" }
       ]
@@ -181,7 +181,7 @@ identities.0.auth_methods.account=GDUA...&identities.0.auth_methods.email=person
     {
       "role": "sender",
       "auth_methods": [
-        { "type": "account", "value": "GDUA..." },
+        { "type": "stellar_address", "value": "GDUA..." },
         { "type": "phone_number", "value": "+10000000001" },
         { "type": "email", "value": "person1@example.com" }
       ]
@@ -189,7 +189,7 @@ identities.0.auth_methods.account=GDUA...&identities.0.auth_methods.email=person
     {
       "role": "receiver",
       "auth_methods": [
-        { "type": "account", "value": "GBEA..." },
+        { "type": "stellar_address", "value": "GBEA..." },
         { "type": "phone_number", "value": "+10000000002" },
         { "type": "email", "value": "person2@example.com" }
       ]
@@ -275,7 +275,7 @@ See [Common Fields].
     {
       "role": "sender",
       "auth_methods": [
-        { "type": "account", "value": "GDUA..." },
+        { "type": "stellar_address", "value": "GDUA..." },
         { "type": "phone_number", "value": "+10000000001" },
         { "type": "email", "value": "person1@example.com" }
       ]
@@ -283,7 +283,7 @@ See [Common Fields].
     {
       "role": "receiver",
       "auth_methods": [
-        { "type": "account", "value": "GBEA..." },
+        { "type": "stellar_address", "value": "GBEA..." },
         { "type": "phone_number", "value": "+10000000002" },
         { "type": "email", "value": "person2@example.com" }
       ]
@@ -341,7 +341,7 @@ See [Common Fields].
     {
       "role": "sender",
       "auth_methods": [
-        { "type": "account", "value": "GDUA..." },
+        { "type": "stellar_address", "value": "GDUA..." },
         { "type": "phone_number", "value": "+10000000001" },
         { "type": "email", "value": "person1@example.com" }
       ]
@@ -349,7 +349,7 @@ See [Common Fields].
     {
       "role": "receiver",
       "auth_methods": [
-        { "type": "account", "value": "GBEA..." },
+        { "type": "stellar_address", "value": "GBEA..." },
         { "type": "phone_number", "value": "+10000000002" },
         { "type": "email", "value": "person2@example.com" }
       ]
--- a/ecosystem/sep-0030.md
+++ b/ecosystem/sep-0030.md
@@ -140,29 +140,23 @@ A set of common fields are used to describe an account in requests and responses
 
 Name | Type | Description
 -----|------|------------
-`identities.owner.account` | string | The Stellar account ID of an another account that if authenticated will be given full permissions of this account.
-`identities.owner.phone_number` | string | The phone number that if authenticated will be given full permissions of this account. When encoding a phone number the ITU-T recommendations [E.123] and [E.164] should be followed. The phone number should be formatted in international notation with a leading `+` as demonstrated in E.123 Section 2.5 to ensure phone numbers are consistently encoded and are globally unique. Spaces should be omitted.
-`identities.owner.email` | string | The email that if authenticated will be given full permissions of this account.
-`identities.other.account` | string | Same as `identities.owner.account`, for the `other` identity.
-`identities.other.phone_number` | string | Same as `identities.owner.phone_number`, for the `other` identity.
-`identities.other.email` | string | Same as `identities.owner.email`, for the `other` identity.
+`identities` | array | A list of identities that if any one is authenticated will be given full permissions of this account.
+`identities[].role` | string | The role of the identity: the individual registering the account (`registrant`) or another party (`other`).
+`identities[].type` | string | The type of identity: `account`, `phone_number`, or `email`.
+`identities[].value` | string | The identity that if authenticated will be given full permissions of the account, either an account, phone number or email address respectively given the type. When encoding a phone number the ITU-T recommendations [E.123] and [E.164] should be followed. The phone number should be formatted in international notation with a leading `+` as demonstrated in E.123 Section 2.5 to ensure phone numbers are consistently encoded and are globally unique. Spaces should be omitted.
 
 ##### Example (JSON)
 
 ```json
 {
-  "identities": {
-    "owner": {
-      "account": "GDUA...",
-      "phone_number": "+10000000001",
-      "email": "person1@example.com"
-    },
-    "other": {
-      "account": "GBEA...",
-      "phone_number": "+10000000002",
-      "email": "person2@example.com"
-    }
-  }
+  "identities": [
+    { "role": "registrant", "type": "account", "value": "GDUA..." },
+    { "role": "registrant", "type": "phone_number", "value": "+10000000001" },
+    { "role": "registrant", "type": "email", "value": "person1@example.com" },
+    { "role": "other", "type": "account", "value": "GBEA..." },
+    { "role": "other", "type": "phone_number", "value": "+10000000002" },
+    { "role": "other", "type": "email", "value": "person2@example.com" },
+  ]
 }
 ```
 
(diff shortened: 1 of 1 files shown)


leighmcculloch:
### Why not
It makes the dataset more complicated and harder to interpret. The existing dataset is extremely simple. You can store data in a single row and easily get a clear snapshot of what identities are associated with an account without needing to join data.

This change would require storing the data in a way that is less clear. It's still just Postgres joins which are simple enough, but for anyone building this system it introduces more moving parts, which means more areas for bugs or mistakes, and higher possibility that in this multi-tenant application one tenants data could be associated with the wrong tenant and lead to allowing a user to control the wrong account.

leighmcculloch:
### Why
The issues I describe in the why-not ☝️ can be resolved while keeping the API as it is, if we decide to limit the number of identities for each type and role. If we say there is only one type/role allowed, we can easily store the data in a single table and the API can be changed to match this API that allows for future more complex use cases. It can be an implementation detail for how many of one type/role are allowed. An implementer preferring a simpler system may choose one or two of each type/role. Others may prefer to support an unlimited number.

leighmcculloch:
Thanks @accordeiro this really captures well the issues with the names given to these things both as they are now and in this PR.

I like proposal 1 because it reduces the set of responsibilities and concerns of SEP-30 down to the smallest set possible. The system becomes about solving one problem: signing transactions for an authenticated user. I especially like that it shifts the meta-data about why a transaction is being registered outside of this system, however I think we have to figure out where that place would be. It could be inside the wallet server but then this framework becomes limited to a single wallet. If someone wanted to recover in a new wallet they wouldn't be able to distinguish between why each account is registered and who it is shared with.

I like proposal 2 because it sets us up to make the most important use cases we're aware of right now work in a way that could work well cross-wallet.

I think we could meet the goals of proposal 2, but leverage the idea in proposal 1 that this additional information about why the account is being stored is meta-data that is meaningless to the service in a similar way that the network supports data entries but is unopinioated about their contents. This is what I'm thinking:

We change the identities so they can be grouped by some logical identifier. This gives us a way to model that identities do map to an individual and they could map to multiple individuals but we don't be anymore opinionated than that in SEP-30. We allow for storing metadata alongside but that be undefined.
```json
{
  "identities": {
    "<identifier-1>": [ // <identifier> is undefined in SEP-30
      { "type": "account", "value": "GDUA..." },
      { "type": "phone_number", "value": "+10000000001" },
      { "type": "email", "value": "person1@example.com" }
    ],
    "<identifier-2>": [
      { "type": "account", "value": "GBEA..." },
      { "type": "phone_number", "value": "+10000000002" },
      { "type": "email", "value": "person2@example.com" }
    ]
  },
  "metadata": {
    // undefined in SEP-30
  }
}
```

We can come up with common formats of metadata. Some example fields:

Account recovery probably wouldn't need any metadata:
```json
{}
```

Send / receive would use metadata:
```json
{
  "sender": "<identifier-1>",
  "receiver": "<identifier-2>"
}
```

leighmcculloch:
I think we'd also say that a server could define its own rules about the metadata. For our implementation I'm thinking max 5 fields, values can be whatever they want but there would be some size limits on the contents.

accordeiro:
Yeah, I've been thinking about arbitrary meta fields too. The thing that I'm struggling to convince myself is: does that actually increase interoperability? I think it'd only do if there is a consensus on **how** we use those meta fields across wallets. And if we'll specify how to use those fields in an interoperable way across the ecosystem, that would probably be through a SEP, at which point we might as well define it here.

I guess my point is: having several identities with names plus a metadata field is really powerful, but I'm afraid being too generic will increase entropy and will inevitably force us to come up with a protocol later on. Shouldn't we instead aim to start with something well-defined that is also extensible in the future? I feel like something based on Proposal 2 might achieve that.

leighmcculloch:
@accordeiro After giving it some more thought and further discussion with you yesterday I agree the metadata fields are a bit unwieldily and I think your second proposal with being more explicit will serve us better. The below are what I'm thinking now, and include some other changes too.

```javascript
{
  "identities": [
    {
      "role": "sender", // the contents of role would be unvalidated and undefined by the SEP
      "auth_methods": [
        { "type": "stellar_account" },
        { "type": "phone_number" },
        { "type": "email" }
      ]
    },
    {
      "role": "receiver",
      "auth_methods": [
        { "type": "stellar_account" },
        { "type": "phone_number" },
        { "type": "email" }
      ]
    },
    ...
  ]
}
```

One thing to notice about this request is that there are no IDs and a registered account can still only be updated using a PUT. I don't think we should put in the effort to define the PATCH just yet, however this message format will support updating individual identities. We'll be able to introduce an `id` field on each identity and we can support updating just a single identity by adding a new endpoint that uses that ID.

The response is easier to understand, follows the same format but removing `auth_methods` and adding `authenticated` which can indicate clearly which roles have been authenticated in the current session. I think this is easier to understand than the existing top-level `identity` field that I've had to explain to everyone who sees it.
```javascript
{
  "address": "GFDD...",
  "identities": [
    { "role": "sender", authenticated: true },
    { "role": "receiver" },
  ],
  "signer": "GADF..."
}
```

While the contents of the roles would be undefined by SEP-30, we could include them in examples as some common ways to implement the SEP. We could allow the `role` to be empty and optional. For a user just backing up their account it could be blank as it offers nothing. For accounts involving two parties in a send/receive style arrangement, the literal `sender` and `receiver` can be used as you suggested.

Let me know what you think.

accordeiro:
I really like that! I think this proposal addresses well our use case while maintaining the following properties:

- **Flexibility**, as roles are arbitrary, but not flexible to the point of allowing the SEP to be misused. Plus I think adding the `sender` and `receiver` examples would help clarify why people would use roles; and
- **Extensibility**, since we can later incorporate an `id` to the identities and make it possible for people to `PATCH` them individually – all without having to break the API

It's still not clear to me by just looking at the JSON whether we encourage people to aggregate several auth_methods that have the same `role` – even if they belong to different people/entities – under one single identity, or instead that they add several `identity` entries with the same role. I think it's the latter (thinking about the PATCH use case), but maybe it's worth making that clearer in the examples too? Your call whether we should be opinionated on that or not.

leighmcculloch:
The intention is that there would be one identity for each person/party. I'll make that clearer in the examples by having multiple identities with the same role. Thanks for pointing that out.

leighmcculloch:
@accordeiro Take a look at a020ea5, it updates just the limited request/response section with the new format. If you think that is good, I'll update the others. For the examples I've added an example of sender/receiver, but I'll probably see about adding a larger more detailed example section at the end of the doc.

Note: I'll change `account` => `stellar_account` in another PR rather than this one.

ecosystem/sep-0030.md
```diff
@@ -140,58 +140,93 @@ A set of common fields are used to describe an account in requests and responses
...
-    "has_other": true
-  },
-  "identity": "account",
+  "identities": [
+    {
+      "role": "sender",
+      "authenticated": true
+    },
+    {
+      "role": "receiver"
```
accordeiro: If I understood correctly, there's also a chance that you'd be authenticated for an identity that does not have a `role`, right? Should we also illustrate that here?

I'd also favor always returning all the possible fields, which means adding an `"authenticated": false` in this object.

leighmcculloch:
>Is there still anything preventing you from removing the draft status, besides finishing updating the other examples?

Draft status allows me to iterate on it and making wide breaking changes, it's more permissive than restrictive to keep it in this status. I'll keep it in draft until we have a running implementation that's somewhat proven, which is in line with the new changes we're making to the SEP process that's in #555 (WIP).

ecosystem/sep-0030.md
leighmcculloch: Correct to the first.

I can have authenticated: false returned in the other cases too 👍 .

leighmcculloch:
Actually, one thought @accordeiro for why to leave `"authenticated": false` as omitted, is that `"role": ""` is also omitted when a role is not provided because it's optional and it'd look pretty odd to return `"role": ""`. I feel like this makes the API responses communicate clearer when they are only shown when relevant. I can see the flip-side perspective though that it makes all the fields discoverable when all the fields are present, but I think we can supplement that with good documentation.

What do you think?

accordeiro:
I feel like I can argue both ways: it usually bugs me whenever I find a field I've never seen before in an API response (because its value defaults to something). But considering a mobile client, I think it's valid to optimize the payload size to minimize data consumption, and maybe that is a good enough reason to favor omitting defaults.

One quick follow-up thought: what would happen if there is no `role` and `authenticated` is `false`? Would we return an entry that looks like "`{}`"? Or would we omit it altogether? Maybe this edge case would help guide the decision process.

accordeiro:
My take is that omitting altogether is not a good idea – it might give a false impression that there are fewer identities than there actually are, while you could actually have several identities that are just role-less and not currently authenticated. I'm not a fan of `{}` either, but I'd either favor that or returning all visible fields (even if default) for all identities.

leighmcculloch:
I definitely think if there are no fields in the identity the identity should still be `{}` because I think the empty value of an object value is `null`, not `{}`, so we should display `{}`. That's what I've done here if you want to take a look at this test:
https://github.com/stellar/go/pull/2383/files#diff-3d3a2f8f712a84966724ba79351cfd15L33-R62

Overall I'm on the fence about omitting, but I found the [Google JSON Style Guide](https://google.github.io/styleguide/jsoncstyleguide.xml#Empty/Null_Property_Values) quite helpful even though it says very little. It's perspective is:

>If a property is optional or has an empty or null value, consider dropping the property from the JSON, **unless there's a strong semantic reason for its existence**.

If we consider both fields independently:

- `"role": ""` – The semantics of omitting this field are more correct, or we at least need to change its value to `null` if it isn't omitted. If a user doesn't provide a field and gets back `""`, that's odd. We could make it return `null`, that would be less odd, but it's also irrelevant since the identity has no role, so it brings attention to something that's irrelevant and that we don't want the developer to be encouraged to use unnecessarily. It's like an optional feature. The fact it's an optional features feels to me like we could easily omit it.
- `"authenticated": false` – The semantics of omitting this field is less clear to me. I could easily argue both ways. I think it keeps each individual API response smaller and a little clearer to what is important for the account, so I'm leaning slightly that way, but I could go either way.

The main reason I would lean towards including the fields is it makes the shape of the API responses predictable to a human looking at the response, but we're writing documentation for the API that should be the source of truth for a human and I'm seeing plenty of perspectives either against or that don't think it's important enough to stand behind that argument online, including Google's style guide.

I'll make them omitted, but stick with `{}` as you pointed out. I'm not a fan of `{}` either, but I think when we come to adding an `id` field to the identities it will look less weird, and an object isn't truly empty unless it is `null` and the identities are not `null`. I can probably add the `id` field soon as that doesn't really hurt anything.

accordeiro:
Agreed, the reasoning makes sense to me – let's stick with omitting the fields and using `"{}"` for now. Can we add a few examples with `"{}"`? I'm expecting that to happen quite frequently, so displaying it in the docs examples might be helpful for developers.

leighmcculloch:
I ended up making role required so that omission is irrelevant to it. Added in 6628c06.

howardtw:
👀

howardtw:
I really like the discussion in this thread, especially this `Shouldn't we instead aim to start with something well-defined that is also extensible in the future?`, and I'm very happy with the change.

One question: why don't we set the default value of `role` to `owner`? If I understand it correctly, when it shows an empty string, it means that the role is `owner`? Having something here is more expressive as well.

leighmcculloch:
Where we ended up settling (not entirely captured here) was that role was worth providing all the time and there are no assumptions made about it being empty meaning something in particular. The field is always returned in the response, it's always visible. A client could still set the empty string if they wanted to or not provide a value, but then the role will be an empty string and just mean that and only that.

In all the examples in the SEP we use a two different patterns, either:
- role = `owner`
- or, role = `sender` / `receiver`

They're the two use cases we have right now, but the SEP isn't constrained to them and someone could use other terms, and if that happens and is compelling to capture here we can add that as a use case in addition to the use cases here.

Setting a default value would constrain the SEP unnecessarily, and instead we should just encourage clients set an appropriate role. This will prevent people from using a default value when it is not the appropriate value to use and will help discover other valuable values assuming we hear about their use.

leighmcculloch:
You can see the role is always set in the examples and not specified as optional in change 6628c066d085585ac93ca179a03df13602bf179d.

ecosystem/sep-0030.md
```diff
@@ -76,6 +75,52 @@ A client can also use the endpoints in the specification to:
...
+Multiple identities should be specified each with a unique role so that when
+each individual is authenticated using an identities authentication methods
+they can recognize their role in relation to the account. In this case using
+roles such as `sender` and `receiver` are appropriate.
+
+Either individual can recover the account by authenticating with all servers and
+asking all servers to sign a transaction giving a new key signing weight
+sufficient to meet the high threshold of the account.
+
+The receiver can accept the account and either take ownership of it or merge it
```
howardtw: The current [sign](https://github.com/leighmcculloch/stellar--stellar-protocol/blob/6628c066d085585ac93ca179a03df13602bf179d/ecosystem/sep-0030.md#post-accountsaddresssign) endpoint is not expecting any operation other than `SET_OPTIONS`. Should we update it?

ecosystem/sep-0030.md
```diff
@@ -33,14 +34,12 @@ server, and later ask the server to sign transactions for the account.
...
 - The client, authenticated as the account, registers with the server, providing:
-  - Alternative identities for the owner of the account:
-    - Another Stellar address.
-    - And/or, a phone number.
-    - And/or, an email address.
-  - Alternative identities for another user of the account (optional):
-    - Another Stellar address.
-    - And/or, a phone number.
-    - And/or, an email address.
+  - One or more identities of individuals that can request the server sign transactions for the account, and for each identity:
```
howardtw: `request the server sign` -> `request the server to sign`

howardtw:
> Setting a default value would narrow the SEP unnecessarily, and instead we should just encourage clients set an appropriate role. This will prevent people from using a default value when it is not the appropriate value to use and will help discover other valuable values assuming we hear about their use.

In my opinion, having a default value makes the SEP well-defined to begin with, and it doesn't prevent people from setting to the values they want as they can simply overwrite it. As a matter of fact, we can still encourage clients to set it to a value that better suit their needs. Having an empty role is a little confusing to me.

leighmcculloch:
>Having an empty role is a little confusing to me.

We're not encouraging an empty role. We did discuss that as a possibility above, but that was removed in https://github.com/stellar/stellar-protocol/pull/566#issuecomment-598443030 because I agree, having the role be optional was confusing.

leighmcculloch:
FYI the code doesn't yet error if role isn't provided because the code doesn't yet do a ton of request validation (TODO), but the intention in 6628c06 is that role is required.

ecosystem/sep-0030.md
leighmcculloch: >The transaction can contain any operations

I think I'm missing something about the request. The sign endpoint can accept any operation. What at this line should we update?

howardtw: I'm not sure how we plan on implementing this endpoint. According to

> The transaction must only contain operations for the account. The transaction can contain any operations, but it is anticipated for the use cases discussed earlier in this protocol that the transaction will contain an operation to add a new signer to the account and possibly to remove an old signer.

It is not clear to me what it means by `contain operations for the account`? Does receiving a payment from other account or sending payment to another account consider operations for only the account?

On top of that, given that this service is currently called recoverysigner, I think it might be more explicit that it only does adding and removing signers of a specific account. And we will let the account decide what to do with the signer power.

howardtw: > It is not clear to me what it means by contain operations for the account? Does receiving a payment from other account or sending payment to another account consider operations for only the account?

Nvm. I found the answer in the code: https://github.com/stellar/go/pull/2383/files#diff-28eb4357c2c210cfb8f53d9dbc938aa3R112-R127

We might just want to make it clear what we mean by `only contain operations for the account` by saying `the source account of the transaction and the source account of all operations in the transaction must match the account in the request.`

leighmcculloch: => #572

PR #568, opened 2020-03-05
Reject cap13

msfeldstein:
This seems to be superceded by CAP23, moving this to rejected to prevent confusion


MonsieurNicolas:
weird I thought we did that already. Thanks!

PR #572, opened 2020-03-16
SEP-0030: Make what transactions can be signed clearer

leighmcculloch:
### What
State what transactions will be signed by a SEP-30 server.

### Why
It isn't clear exactly what transactions a SEP-30 server will sign. The document states that transactions for an account will be signed, but we should be more explicit about what that means.

This is a follow up to https://github.com/stellar/stellar-protocol/pull/566#discussion_r392453939.


leighmcculloch:
I added more details in 11bd38e. Thanks!

PR #574, opened 2020-03-17
[SEP24] Update lang param description

msfeldstein:
Expand the scope of the `lang` param when starting a transaction.  This can also be used to localize the interactive flow UI, as well as anything else that might be shown to the user in the context of this transaction


vitchor:
I believe there are other places in SEP-24 where this also needs to be updated.

msfeldstein:
updated it for withdraw as well as deposit thanks victor

PR #579, opened 2020-03-28
Reactivating SEP-6 as a non-interactive anchor/wallet interface alongside SEP-24.

marcinx:
Hey! @tomerweller @msfeldstein @gdini2003 @vitchor @rice2000 & everyone.

As discussed via email the objective of this PR is to bring back a non-interactive anchor/wallet interface alongside SEP-24. We do this by deprecating all interactive elements in SEP-6, thus making it explicitly non-interactive, bumping the version and setting it to active.

Non-interactive SEP-6 is currently used by numerous anchors and integrated into most wallets that interact with anchors. Our own service heavily relies on it as well. It can peacefully co-exist with SEP-24 until something more advanced comes along (SEP-26 looks like a good potential candidate after some additional grooming). 

With this PR anchors will again/still have the choice between providing a low-level interface based on backend APIs (SEP-6, + optional SEP-12) or providing a higher level interface based on browser-based user interfaces with interactive KYC verification when needed (SEP-24).

I put my comments inline and am looking forward to your thoughts and feedback. Thanks!

PS: If this is too much too fast then we can tone it back down and simply change the deprecation message to apply to only the interactivity part in SEP-6. Just let me know in that case and I'll edit it accordingly.


tomerweller:
I completely agree with the rational here. SEP6 is widely used and doesn't have any viable (non-interactive) alternatives yet so it should be re-activated.

I'm worried about completely removing the interactive portion from the spec as it's used by some anchors/wallets.

How about, for now, we reactivate and put a deprecation notice on the interactive portion but without removing it? It's definitely not as clean but a bit more friendly to current consumers.

Also, for the sake of pr/commit scope, let's move other SEP/status/naming changes into their own PRs and keep this one focused on SEP6.

marcinx:
Thanks, @tomerweller !

> I'm worried about completely removing the interactive portion from the spec as it's used by some anchors/wallets.
Yeah, I thought that might be an issue. 

> How about, for now, we reactivate and put a deprecation notice on the interactive portion but without removing it? It's definitely not as clean but a bit more friendly to current consumers.
Okay. 

> Also, for the sake of pr/commit scope, let's move other SEP/status/naming changes into their own PRs and keep this one focused on SEP6.
Okay. I'll break it down into smaller chunks with their own PRs where needed.

Issue #484, opened 2019-12-04
Discussion: SEP26 Non-interactive Anchor/Wallet Asset Transfer


msfeldstein:
SEP26 was submitted as the non-interactive counterpoint to SEP24's interactive transfer protocol.  It uses some more forward thinking infrastructure, such as SEP27's Asset Information Server

Please discuss thoughts on this proposal here

https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0026.md

msfeldstein:
Additional discussion: https://github.com/stellar/stellar-protocol/pull/460

s-a-y:
1. I still don't understand why /deposit endpoint doesn't require an amount. But at the same time, there is a "fee" field in response (calculated fee). How can you calculate it without the amount?
We could do it the same way as for withdrawal.
2. Maybe responses for /deposit and /withdraw should have `id` field analogous to sep24. Just to make it easier for wallets to handle it in the same way (keep id to query /transaction endpoint later)

s-a-y:
Regarding 1. Fritz explained to me in private, that you can request amount using optional form field using sep27. The fee is optional in response, therefore most likely won't be present unless anchor can actually calculate the specific amount.

dolcalmi:
please reference SEP26 in SEP006.... and next time it would be nice not to deprecate before finalizing the specification

andywer:
Hey there!

At SatoshiPay we would like to help move SEP-26 forward. Currently we need to stick to the anchors' SEP-6 implementations for our new B2B platform as anchors don't seem to implement SEP-26 yet and we need a non-interactive flow.

**I have got one question up-front, though:**

Why has the description of custom fields been moved from the `/info` endpoint to the asset info one? I would like to argue that this makes it harder for anchors to implement as they now need to implement the SEP-27 asset info endpoint in order to have a usable SEP-26 implementation.

Furthermore that information is only useful for a SEP-26 setting, hence it's not quite clear why it would reside in a general-purpose asset info endpoint instead of in one of the transfer server's.

The benefit if this decision is not clear to me as a reader of those specs. Especially since including those fields in the `/info` response doesn't interfere with SEP-24 and the format outlined in SEP-27 seems to be backwards-compatible to SEP-6 on first glance.

Maybe one of you guys can shine a light on that 😉

andywer:
Hi @msfeldstein! I would like to propose some updates to the spec as a PR. Those are the first low-hanging fruits from my point of view:

[...]

PR #580, opened 2020-03-30
SEP-26 - Make document easier to read

andywer:
Make large example responses collapsible using `<details>`/`<summary>` elements. The doc as of now looks super intimidating, even to SEP-6 / SEP-24 veterans. This should make it much easier to keep an overview.

[**Before**](https://github.com/stellar/stellar-protocol/blob/fcc397b6cc3dbbeb3ef44323581395322e907816/ecosystem/sep-0026.md) • [**After**](https://github.com/andywer/stellar-protocol/blob/sep-26/make-easier-to-read/ecosystem/sep-0026.md)

Related to #484.

### DIFF
--- a/ecosystem/sep-0026.md
+++ b/ecosystem/sep-0026.md
@@ -560,7 +560,6 @@ If the given account does not exist, or if the account doesn't have a trust line
 ```
 POST TRANSFER_SERVER/deposit
 Content-Type: multipart/form-data
-
 ```
 
 Request Parameters:
@@ -610,8 +609,10 @@ Name | Type | Description
 `fee` | float | (optional) Calculated fee (if any). In units of the deposited asset.
 `extra_info` | array | (optional) JSON array with additional information about the deposit process. Each element in the array is formatted as follows {key: KEY, value: VALUE}. Wallets are encouraged to present extra_info in a tabular manner and enable easy copy to clipboard for each line value.
 
-
-Bitcoin response example:
+<details>
+<summary>
+  <b>Bitcoin response example</b>
+</summary>
 
 ```json
 {
@@ -620,7 +621,12 @@ Bitcoin response example:
 }
 ```
 
-Ripple response example:
+</details>
+
+<details>
+<summary>
+  <b>Ripple response example</b>
+</summary>
 
 ```json
 {
@@ -640,13 +646,16 @@ Ripple response example:
       "key" : "Tag",
       "value": "88"
     }
-
   ]
-
 }
 ```
 
-Mexican peso (MXN) response example:
+</details>
+
+<details>
+<summary>
+  <b>Mexican peso (MXN) response example</b>
+</summary>
 
 ```json
 {
@@ -654,7 +663,6 @@ Mexican peso (MXN) response example:
   "eta": 1800,
   "fee": 100,
   "extra_info": [
-
     {
       "key" : "Bank",
       "value": "STP"
@@ -670,6 +678,8 @@ Mexican peso (MXN) response example:
   ]
 }
 ```
+
+</details>
 
 ##### Special Cases
 ###### Stellar account does not exist
@@ -755,7 +765,10 @@ Name | Type | Description
 `extra_info` | array | (optional) JSON array with additional information about the withdrawal process. Each element in the array is formatted as follows {key: KEY, value: VALUE}. Wallets are encouraged to present extra_info in a tabular manner and enable easy copy to clipboard for each line value.
 
 
-Example:
+<details>
+<summary>
+  <b>Example response</b>
+</summary>
 
 ```json
 {
@@ -776,6 +789,8 @@ Example:
 }
 ```
 
+</details>
+
 ## Deposit and Withdraw shared responses
 
 ### 1. Interactive customer information needed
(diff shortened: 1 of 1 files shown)


marcinx:
@andywer Thanks for making these edits to SEP-26. 

How do you feel about the arbitrary `fee` string in the SEP-26 fee attribute? I put my thoughts here: https://github.com/stellar/stellar-protocol/pull/460 but basically I feel like it would be beneficial to keep the concept of fee_fixed and fee_percent in order for services to compute output amounts/fees programmatically. 

An arbitrary string seems like an invite to all cause of problems. There should be a mechanism to compute fees before invoking /deposit or /withdraw and fee_fixed/fee_percent seems to be covering most use cases. For more complex fee structures we could re-introduce the /fee endpoint from SEP-6 later on (although I don't think I've seen anyone use it so far).

andywer:
Thanks, @marcinx.

The `fee` string… Yeah, not a big fan either. It might be ok-ish in so far as the transfer transaction will tell the client the exact fee amount for this particular deposit/withdrawal before any funds are submitted, anyway. The `/fee` endpoint would probably be a good idea, though (never used it myself, though, tbh).

In fact, I was questioning the design decision to move metadata that's critical for the transfer server and never used for anything but the transfer server to the SEP-27 endpoint in the first place: https://github.com/stellar/stellar-protocol/issues/484#issuecomment-600215172

Would love to hear the reasoning, but so far no one replied.
Btw, I saw your PR, but I don't know what you guys discussed via email 😉 I guess if it wasn't for SEP-27, the switch from SEP-6 to SEP-26 would require few changes and thus be fairly quick & easy.

I might be wrong here, though, as there could be more changes to SEP-26 to come that I am just not aware of. But if integration with SEP-27 was optional, I think there would be a fair chance to get SEP-26 to a point where you could easily migrate from SEP-6 within a short amount of time.

marcinx:
> I might be wrong here, though, as there could be more changes to SEP-26 to come that I am just not aware of.
I guess the "changes" are up to you, myself and everyone else involved and interested :) The SEP is in draft state and nothing is set in stone yet. Now is a good time to make the most important improvements before it (possibly) matures to active state. The draft was originally composed and submitted by @efritze and @cameroon and is up for discussion right now.

> The fee string… Yeah, not a big fan either. It might be ok-ish in so far as the transfer transaction will tell the client the exact fee amount for this particular deposit/withdrawal before any funds are submitted, anyway.
I think that is the current rationale. However, it would invite much better UX design when output amounts can be calculated programmatically and displayed beforehand, especially if end users are to compare several anchoring options. When the output amounts are computable automatically beforehand, we can display to users that they will receive 1.15 on anchor A, 1.20 on anchor B and so on. 

This is how our platform currently does it and it's an important factor in the decision making process for where to withdraw. You can't expect users to have a calculator ready to compute themselves what they're going to get. Human readable strings for display can be generated from numeric attributes, but calculating from arbitrary strings is basically impossible.

> The /fee endpoint would probably be a good idea, though (never used it myself, though, tbh).
Yeah, haven't come across any anchor who uses that either. Might as well leave it out in the initial version and go with fee_fixed+fee_percent only. This probably covers 99% of business models. If it absolutely becomes necessary for some anchor to have a more complex fee structure, /fee could be added to a future version of the SEP.

> But if integration with SEP-27 was optional, I think there would be a fair chance to get SEP-26 to a point where you could easily migrate from SEP-6 within a short amount of time.
I am also in favor of using TRANSFER_SERVER (SEP-6), TRANSFER_SERVER_SEP0024, TRANSFER_SERVER_SEP0026 and so on. No need to introduce additional complexity and redundancy as proposed in SEP-27. You made a very good point earlier today when you pointed out that the asset issuer information is redundant across all the assets with the proposed model.

Another thing important to consider is the KYC/AML integration in SEP-26. The way it currently exists in the draft makes it basically interactive as it involves web views and graphical user interfaces. I think that's a blocker for a non-interactive SEP. Probably better to explicitly specify the data transmission interface with the use of SEP-12 and SEP-9 or an improvement of those. That way SEP-26 will be truly non-interactive.

> Btw, I saw your PR, but I don't know what you guys discussed via email
We discussed via email that there are services that heavily rely on purely non-interactive communication with anchors, for example because they are backend-API based. The point of the PR is to keep SEP-6 (without the interactivity parts) as the non-interactive standard for anchor/wallet communication until we have something more advanced (SEP-26 with some more grooming?). SEP-24 depends on a browser window and doesn't work for all use cases. At the same time all current anchors already support non-interactive SEP-6 and our own platform heavily relies on it. The SEP works well and is up and running. There is no reason to retire it yet. 

If we introduce SEP-26 later on, then anchors and services can gracefully upgrade from SEP-6 to SEP-26 and/or SEP-24. I think SEP-6 (without all the interactivity components) works extraordinarily well and is currently the de-facto standard for anchor/wallet communication.

The vison/idea for timeline in my head is: Re-activate SEP-6 (and version out all the interactivity which now lives in SEP-24, see my PR)... then groom SEP-26 into something that is production ready... then keep SEP-6 and SEP-26 in parallel for a while and then potentially deprecate SEP-6.

@tomerweller @msfeldstein @rice2000 @gdini2003

msfeldstein:
omg yes such a good change

> : Re-activate SEP-6 (and version out all the interactivity which now lives in SEP-24, see my PR)... then groom SEP-26 into something that is production ready... then keep SEP-6 and SEP-26 in parallel for a while and then potentially deprecate SEP-6.

👍  I agree that people are currently finding good usage with sep6 at the moment so there's no reason to rush anyone to change

PR #581, opened 2020-03-31
Update SEP-26 endpoint route paths

andywer:
Make them consistent with SEP-24 route paths. My VS code also auto-fixed a bunch of trailing spaces.

- `/deposit` => `/transactions/deposit/non-interactive`
- `/withdraw` => `/transactions/withdraw/non-interactive`

I mentioned the intended change in #484, but got no reply so far. So here is the PR as it's straight-forward. If you disagree with this change, let's discuss it here or over at #484 🙂

### DIFF
--- a/ecosystem/sep-0026.md
+++ b/ecosystem/sep-0026.md
@@ -155,7 +155,7 @@ To support this protocol, an anchor acts as a server and implements the specifie
             "name": "Facebook Pay",
             "fee": "0.1% + 1 USD",
             "logo": "https://fb.com/pay.png",
-            
+
             "fields": [
               {
                 "id": "dest",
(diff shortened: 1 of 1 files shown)


msfeldstein:
I don't feel strongly about this either way.  One of the nice things about using SEP-27 here is that each protocol can have its own URL prefix, so you could say

```
{
  "transfer_protocols": {
    "sep26": {

      "transfer_server": "https://api.examplebtcanchor.com/non-interactive",
    },
    "sep24": {
      "transfer_server": "https://api.examplebtcanchor.com/interactive",
    }
}```

and then urls would look like `https://api.examplebtcanchor.com/interactive/deposit` and `https://api.examplebtcanchor.com/non-interactive/deposit`

Does that make it more difficult to developer your anchor software that way?

andywer:
@msfeldstein I think it makes sense to discuss #484 first.

I just went ahead and opened this PR as it looked like the spec has just not been updated yet after the fork from SEP-6 and I didn't get a reply in #484 either.

PR #588, opened 2020-04-06
Updating SEP-12 to play well with non-interactive SEP-6

### LINKED ISSUE #587: SEP-12 deprecation and SEP6/SEP24 extra info
If Anchors want to allow Wallets to pass user information non-interactively, they'd probably want to use SEP-12, but SEP-12 is deprecated in favor of interactive web pages.

What's the reason behind SEP-12 deprecation, and how to proceed if the Anchor still wants to collect user information non-interactively from the Wallets (for both SEP6 and SEP24)?

Also, if the Anchor wants to allow extra fields to be passed in SEP6/SEP24 (example: destination first name, or some options for the SEP24 interactive flow), there's currently no documentation regarding this. How to proceed about that?

marcinx:
* Updating SEP-12 to play well with non-interactive SEP-6
* Typos
* Resolves #587 
* Follow-Up to #582

### DIFF
--- a/ecosystem/sep-0006.md
+++ b/ecosystem/sep-0006.md
@@ -4,7 +4,7 @@
 SEP: 0006
 Title: Anchor/Client interoperability
 Author: SDF
-Status: Active (Interactive components are deprecated in favor of [SEP-24](sep-0024.md))
+Status: Active (Interactive components are deprecated in favor of SEP-24)
 Created: 2017-10-30
 Updated: 2020-04-01
 Version 3.1.1
@@ -339,7 +339,7 @@ Example:
 ```json
 {
   "type": "non_interactive_customer_info_needed",
-  "fields" : ["family_name", "given_name", "address", "tax_id"],
+  "fields" : ["family_name", "given_name", "address", "tax_id"]
 }
 ```
 
--- a/ecosystem/sep-0012.md
+++ b/ecosystem/sep-0012.md
@@ -6,17 +6,13 @@ Title: Anchor/Client customer info transfer
 Author: Interstellar
 Status: Active
 Created: 2018-09-11
-Updated: 2018-09-11
-Version 1.1.0
+Updated: 2020-04-06
+Version 1.1.1
 ```
-
-## Deprecation warning
-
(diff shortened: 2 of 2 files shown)


ecosystem/sep-0006.md
```diff
@@ -339,7 +339,7 @@ Example:
 ```json
 {
   "type": "non_interactive_customer_info_needed",
-  "fields" : ["family_name", "given_name", "address", "tax_id"],
+  "fields" : ["family_name", "given_name", "address", "tax_id"]
```
marcinx, line 342: Just a typo fix.

ecosystem/sep-0006.md
```diff
@@ -4,7 +4,7 @@
 SEP: 0006
 Title: Anchor/Client interoperability
 Author: SDF
-Status: Active (Interactive components are deprecated in favor of [SEP-24](sep-0024.md))
+Status: Active (Interactive components are deprecated in favor of SEP-24)
```
marcinx, line 7: Code blocks don't support Markdown links.

ecosystem/sep-0012.md
```diff
@@ -6,17 +6,13 @@ Title: Anchor/Client customer info transfer
...
 ```
 
-## Deprecation warning
-
-We're currently [recommending](sep-0006.md#recommendations) that anchors and wallets use the SEP-6 interactive flow for transmitting KYC information, rather than SEP-12. Therefore, SEP-12 is **deprecated** for anyone using it as part of a [SEP-6](sep-0006.md) implementation.
-
 ## Abstract
 
-This SEP defines a standard way for stellar wallets to upload KYC (or other) information to anchors who need it.
+This SEP defines a standard way for stellar wallets to upload KYC (or other) information to anchors that implement non-interactive SEP-6.
```
marcinx, line 15: Now that SEP-6 is active for non-interactive flows, SEP-12 should be used to transfer KYC/AML information between service and anchors.

msfeldstein: I'm not sure we want to tie this directly to sep-6 as it can be used outside of it, but i think this is fine for now.  If we start to use sep-12 more we can loosen this up, imo always better to keep things too strict and loosen than the opposite.

msfeldstein:
thanks!

ecosystem/sep-0012.md
marcinx: Okay, thanks. Good point! SEP-12 will probably also be interesting for SEP-26 because there is currently no non-interactive KYC/AML transfer defined there yet.

PR #592, opened 2020-04-10
Direct Payment SEP

msfeldstein:
In order to enable direct fiat-to-fiat payments facilitate by two anchors, we propose this Direct Payment SEP.  A lot of discussion has gone into the general shape of this protocol, but this creates an explicit definition for everything.

This is based on the SEP6 spec that people have found useful as a rough guide. It changes requests to a POST request (instead of GET for security), and creates a separate endpoint since this is distinctly different from a `deposit`.

It also adds the ability for the sender to request KYC information about the receiving client that the receiving anchor can pull from their banking partnerships.

It removes the need for SEP-12 and just passes everything along in one request.

It removes unneeded endpoints, statuses, and fields, and adds missing one (ie `pending_sender`and `pending_receiver` since `pending_anchor` is ambiguous now.


ecosystem/sep-direct-payment.md
```diff
@@ -0,0 +1,318 @@
...
+1. The sending anchor can query the status of this transaction via the [`/transaction`](#transaction) endpoint, and should communicate updates to the sending client as that progresses.
+
+
+
+## API Endpoints
+
+* [`GET /info`](#info): required
+* [`POST /send`](#send): required
+* [`GET /transactions`](#transaction-history): required
+* [`GET /transaction`](#single-historical-transaction): required
```
msfeldstein, line 96: Do we need the `/transactions` endpoint?  It was mostly for creating UIs for wallets and feels unnecessary here

ecosystem/sep-direct-payment.md
```diff
@@ -0,0 +1,318 @@
...
+   "receive":{
+      "USD":{
+         "enabled":true,
+         "fee_fixed":5,
+         "fee_percent":1,
+         "min_amount":0.1,
+         "max_amount":1000,
+         "fields":{
+            "sender":{
+               "first_name":"The sender's first name",
```
msfeldstein, line 128: Does it make sense to split up 'fields' like this?  If we don't, we can't disambiguate between the SEP-9 first_name field for receiver vs sender.  This allows us to keep using SEP-9 fields without having to change to `sender_first_name` and `receiver_first_name`

ecosystem/sep-direct-payment.md
```diff
@@ -0,0 +1,318 @@
...
+* `fee_fixed`: Optional fixed (flat) fee for deposit. In units of the received asset. Leave blank if there is no fee or the fee schedule is complex.
+* `fee_percent`: Optional percentage fee for deposit. In percentage points. Leave blank if there is no fee or the fee schedule is complex.
+* `fields` object as explained below.
+
+The `fields` object allows an anchor to describe fields that must be passed into `POST /send`. It should include any KYC fields required for the sender or receiver, account information, and anything else required by the receiving anchor in order to complete a transaction.  Fields are broken out by `sender`, `receiver`, and `transaction`.  `sender` and `receiver` contain KYC requests of values from SEP-9 while `transacton` contains transaction specific information requested.
+
+Each `fields` sub-object contains a key for each field name and an object with the following fields as the value:
+
+* `description`: description of field to show to user.
+* `choices`: list of possible values for the field.
```
msfeldstein, line 172: I removed the ability to mark fields as 'optional'.  Do you think optional fields are useful?

ecosystem/sep-direct-payment.md
JakeUrban: I think it makes sense.

ecosystem/sep-direct-payment.md
```diff
@@ -0,0 +1,318 @@
...
+```json
+{
+   "receive":{
+      "USD":{
+         "enabled":true,
+         "fee_fixed":5,
+         "fee_percent":1,
+         "min_amount":0.1,
+         "max_amount":1000,
+         "fields":{
```
JakeUrban, line 126: If all fields listed are from a SEP, there's no reason to have a description for each field. Anchors can keep a list of descriptions in their native language instead of relying on the other anchor to return a translation of the their own description. 

Then, the value for `sender`, `receiver`, and `transaction` could be a list of SEP9 field names.

ecosystem/sep-direct-payment.md
```diff
@@ -0,0 +1,318 @@
...
+               "first_name":"The sender's first name",
+               "last_name":"The sender's last name"
+            },
+            "receiver":{
+               "first_name":"The receiver's first name",
+               "last_name":"The receiver's last name",
+               "email_address":"The receiver's email address"
+            },
+            "transaction":{
+               "routing_number":{
```
JakeUrban, line 137: Are the fields here from a SEP? I don't see it in SEP-9, but I see `bank_number`. Ideally anchors would only use fields from a SEP, otherwise Anchor A could list `routing_number` but Anchor B could list `routing_num` and the sending Anchor C would have to know how to handle both, which would get really messy quickly.

ecosystem/sep-direct-payment.md
JakeUrban: I'd keep it if unless we're sure it wouldn't be useful.

ecosystem/sep-direct-payment.md
```diff
@@ -0,0 +1,318 @@
...
+```
+
+The JSON object contains an entry for each asset that the anchor supports for receiving and completing a direct payment.
+
+#### For each asset available for receiving, response contains:
+
+* `min_amount`: Optional minimum amount. No limit if not specified.
+* `max_amount`: Optional maximum amount. No limit if not specified.
+* `fee_fixed`: Optional fixed (flat) fee for deposit. In units of the received asset. Leave blank if there is no fee or the fee schedule is complex.
+* `fee_percent`: Optional percentage fee for deposit. In percentage points. Leave blank if there is no fee or the fee schedule is complex.
```
JakeUrban, line 164: If the fee schedule is complex, how is it communicated?

ecosystem/sep-direct-payment.md
```diff
@@ -0,0 +1,318 @@
...
+
+Name | Type | Description
+-----|------|------------
+`id` | string | Persistent identifier to check the status of this payment
+`stellar_account_id` | string | Stellar account to send payment to
+`stellar_memo_type` | string | Type of memo to attach to the stellar payment `(text | hash | id)`
+`stellar_memo` | string | The memo to attach to the stellar payment
+`receiver_info` | object | Key-value pairs of the information the sender requested via `require_receiver_info`
+
+##### Customer Info Needed (403 forbidden)
```
JakeUrban, line 230: I don't think 403 is appropriate here, 400 seems like the natural fit.

ecosystem/sep-direct-payment.md
```diff
@@ -0,0 +1,318 @@
...
+
+In the case where the sending anchor didn't provide all the information requested in `/info`, or if the transacton requires extra information, the request should fail with a 403 status code and the following body in JSON format.  The sender should then retry the entire request including all the previously sent fields plus the fields described in the response.
+
+Name | Type | Description
+-----|------|------------
+`error`| string | `customer_info_needed`
+`fields` | object | A key-value pair of missing fields in the same format as fields described in [`/info`](#info)
+
+##### Error (403 Forbidden)
```
JakeUrban, line 239: Same here.

ecosystem/sep-direct-payment.md
```diff
@@ -0,0 +1,318 @@
...
+`receiver_info` | object | Key-value pairs of the information the sender requested via `require_receiver_info`
+
+##### Customer Info Needed (403 forbidden)
+
+In the case where the sending anchor didn't provide all the information requested in `/info`, or if the transacton requires extra information, the request should fail with a 403 status code and the following body in JSON format.  The sender should then retry the entire request including all the previously sent fields plus the fields described in the response.
+
+Name | Type | Description
+-----|------|------------
+`error`| string | `customer_info_needed`
+`fields` | object | A key-value pair of missing fields in the same format as fields described in [`/info`](#info)
```
JakeUrban, line 237: `fields` could also just be a list of SEP9 field names.

ecosystem/sep-direct-payment.md
```diff
@@ -0,0 +1,318 @@
...
+Author: SDF
+Status: Draft
+Created: 2020-04-07
+Updated: 2020-04-07
+Version 1.0.0
+```
+
+## Simple Summary
+
+This SEP defines a protocol for enabling fiat to fiat payments between two real world accounts, facilitated by two different anchors.
```
JakeUrban, line 16: Are there other "real world accounts" other than bank accounts? If not I'd replace the "real world" description to "bank".

ecosystem/sep-direct-payment.md
```diff
@@ -0,0 +1,318 @@
...
+The JWT should be included in all requests as request header:
+```
+Authorization: Bearer <JWT>
+```
+
+Any API request the fails to meet proper authentication should return a 403 forbidden response with the error payload:
+
+```
+{
+  "type": "authentication_required"
```
JakeUrban, line 48: Is there a reason we need this specific message? If we change the endpoints below to return 400 instead of 403 on a bad request, then a 403 can indicate bad authentication.

ecosystem/sep-direct-payment.md
JakeUrban: I think it makes sense.

ecosystem/sep-direct-payment.md
JakeUrban: I'd keep it if unless we're sure it wouldn't be useful.

ecosystem/sep-direct-payment.md
gaclaudiu: Hi Guys , 
I am Claudiu (from FinClusive) . 
Do you guys plan to add to SEP-9 also the fields for entities ? I think right now SEP-9 targets only individuals. 
Or this should be each anchor responsibility to create an entity fields definition and agree it with any anchor they are integrating with ?

ecosystem/sep-direct-payment.md
```diff
@@ -0,0 +1,318 @@
...
+            "sender":{
+               "first_name":"The sender's first name",
+               "last_name":"The sender's last name"
+            },
+            "receiver":{
+               "first_name":"The receiver's first name",
+               "last_name":"The receiver's last name",
+               "email_address":"The receiver's email address"
+            },
+            "transaction":{
```
gaclaudiu, line 136: For transaction fields , can we (as an anchor) add custom fields and agree them with the anchors we are integrating with ?  
Let me give you and example : 
When we (FinClusive) do ledgering we have for some of our clients virtual accounts ( which do no heave a real bank account behind) and we plan for this type of clients to instruct them to share their account id  as they don't have an account number or routing number.

ecosystem/sep-direct-payment.md
```diff
@@ -0,0 +1,318 @@
...
+         "fee_fixed":5,
+         "fee_percent":1,
+         "min_amount":0.1,
+         "max_amount":1000,
+         "fields":{
+            "sender":{
+               "first_name":"The sender's first name",
+               "last_name":"The sender's last name"
+            },
+            "receiver":{
```
gaclaudiu, line 131: In the receiver section ca we add custom fields ? 
Each of our clients are identified by a unique value called FinClusiveID , for us it is enough for the sending anchor to pass this into the sent payment request in the receiver section .

ecosystem/sep-direct-payment.md
```diff
@@ -0,0 +1,318 @@
...
+               "email_address":"The receiver's email address"
+            },
+            "transaction":{
+               "routing_number":{
+                  "description":"routing number of the destination bank account"
+               },
+               "account_number":{
+                  "description":"bank account number of the destination"
+               },
+               "type":{
```
gaclaudiu, line 143: is this section for the sending anchor or for receiving anchor ? 
Is this a required section ?

ecosystem/sep-direct-payment.md
```diff
@@ -0,0 +1,318 @@
...
+This is the successful case where a receiving anchor confirms that they can fulfill this payment as described. The response body should be a JSON object with the following values
+
+Name | Type | Description
+-----|------|------------
+`id` | string | Persistent identifier to check the status of this payment
+`stellar_account_id` | string | Stellar account to send payment to
+`stellar_memo_type` | string | Type of memo to attach to the stellar payment `(text | hash | id)`
+`stellar_memo` | string | The memo to attach to the stellar payment
+`receiver_info` | object | Key-value pairs of the information the sender requested via `require_receiver_info`
+
```
gaclaudiu, line 229: I am curios here if there is a way for the receiving anchor to generate the stellar transaction id which should be used by the sending anchor when publish the transaction with the payment on the Stellar Network?
I am asking this because we plan to tie our AML checks on a Stellar Payment to a Stellar transaction id.

ecosystem/sep-direct-payment.md
msfeldstein: I'd like to either augment SEP-9 or define them in this SEP, I really don't want to make parties have to come to NxN agreements on field names

ecosystem/sep-direct-payment.md
msfeldstein: i was actually really confused by what i thought 403 meant.  
http://w3.org/Protocols/rfc2616/rfc2616%20-sec10.html
"The server understood the request, but is refusing to fulfill it. Authorization will not help and the request SHOULD NOT be repeated"

My understanding was that 403 meant 'needs authorization' but now this says its not.  

400 says 
"The request could not be understood by the server due to malformed syntax"
which doesn't make sense here for sure

ecosystem/sep-direct-payment.md
msfeldstein: cash out at a place like moneygram for example

ecosystem/sep-direct-payment.md
msfeldstein: There's no way to pre-generate a stellar tx id, but what you're describing is the reason for the stellar_memo field.  You can tie your AML checks or whatever you want to the stellar transaction that has the pre-generated memo.

ecosystem/sep-direct-payment.md
JakeUrban: That link gave me a 404, but check out the [spec](https://tools.ietf.org/html/rfc7231#section-6.5.1)

I would consider missing arguments a malformed request.

andywer:
Quick question, but fundamental: What we have got here now is essentially an alternative API that from a technical perspective serves the same purpose as SEP-26 (= non-interactive crypto to fiat offramp), right?

Couldn't we just turn this into SEP-26 and then

**a)** have separate endpoints for end-user withdrawal and direct payment
or
**b)** abstract the authentication a bit more, so that you can either authenticate for a withdrawal as an end-user or as an anchor (maybe extend SEP-10 by an "authenticate me as" parameter (as anchor or as end-user)?

Would love to here your thoughts on that or why non-interactive SEP-6 / SEP-26 cannot be used for this purpose (would also be great to document that in the SEP). @msfeldstein

msfeldstein:
@andywer technically we could shoehorn this usecase into sep-6 (and some parties have) but i think that's going to conflate SEP-6 into doing too many things.  This is a fundamentally different thing, even if it looks similar technically, and i want to keep both specs as simple and flexible as possible.  If we start mixing bilateral and unilateral use-cases i think we'll see things getting confusing quickly.  I think this is similar to how we mixed the interactive and non-interactive flows in SEP-6 originally when everyone's use case used one or the other.

tomerweller:
@andywer when we deprecated the legacy compliance protocol (sep-0003) the working assumption was that we could compose a send/receive transaction (b2b, remittance, etc) using a composition of a wallet deposit and a withdraw (sep6 or other).

However, we were wrong. We learned the hard way that a send/receive transaction has two parties and two anchors, and these need to be able to exchange information between them. The SEP6 protocol and language all assume a "unilateral" transaction - the user is moving money from one representation to another, using the anchor. No other parties involved.

Can we extend SEP6 or SEP26 to accommodate this? Probably. Should we? (That's a good question for the round table tomorrow @msfeldstein) I don't know. I think @msfeldstein's approach of separating different business flows to different SEPs is the right way forward as it doesn't create monster specs. 

Obviously there's not one solution - a lot of things can work. We, as an ecosystem, just need to agree on something that doesn't suck too much and push forward. Good thing we've got talented wallet developers that can make the magic happen ;)

andywer:
@tomerweller Haha, thanks for the quick response! I am also not fundamentally opposed to the approach, I just felt there is reason to challenge the assumptions it is based on, especially as they are not addressed explicitly in the spec right now.

Looking forward to the discussion tonight.

msfeldstein:
I've added an `/update` endpoint to allow for the sender to update information requested by the receiver, in the case of a bank or cash pickup saying the receivers identification doesn't match
https://github.com/stellar/stellar-protocol/pull/592/commits/b3d56ba4c00eba3552fdcda00dd1d0d955591821

ecosystem/sep-direct-payment.md
```diff
@@ -0,0 +1,394 @@
...
+If the transaction cannot be found, the endpoint should return a `404 NOT FOUND` result.
+
+#### Pending info update
+
+In certain cases the receiver might need to request updated information, for example if the bank tells them that the provided receiver name is incorrect, or missing a middle initial.  In this case, the transaction should go into the `pending_info_update` state until the sender provides updates.
+
+### Update
+
+The `/update` endpoint allows for updating certain pieces of information that need to be corrected.  For example a bank may reject a deposit because a name is mis-spelled, or a middle initial is missing.  This allows transactions to be updated instead of errored and refunded.
```
andywer, line 354: I think a second paragraph outlining *when* the sender should be able to use this endpoint would be helpful. Should anchors implement the endpoint so that it will always fail if not in `pending_info_update` state or should the sender be allowed to update eagerly?

(Not sure how realistic that scenario is in the real world, but at least technically this edge case seems feasible)

marcinx:
Just, out of curiosity. Why did SEP-12 not fit in here? It seems like such an elegant and RESTy way to decouple payment requests from customer information. 

Wouldn't it be easier for receiving anchors to receive POST /customer information once and then only a reference/id to that customer/receiver in the direct-payment request? Would it be easier for anchors to maintain customer data in in relational DBs with SEP-12? They wouldn't need to compare a whole set of customer information against their own records every time a direct-payment request comes in.

ecosystem/sep-direct-payment.md
```diff
@@ -0,0 +1,394 @@
...
+1. The sending anchor can query the status of this transaction via the [`/transaction`](#transaction) endpoint, and should communicate updates to the sending client as that progresses.  If the status is `pending_info_update` it should request that info from the sending client and provide it to the receiving anchor via the [`/update`](#update) endpoint.
+1. Once the [`/transaction`](#transaction) endpoint returns a `completed` status the transaction has been completed.
+
+
+
+## API Endpoints
+
+* [`GET /info`](#info)
+* [`POST /send`](#send)
+* [`GET /transaction`](#single-historical-transaction)
```
vitchor, line 97: /transaction and /update links are broken

ecosystem/sep-direct-payment.md
tomerweller: I agree that 400 makes a bit more sense here (it's also a kitchen sink for any client side error): 
>   The 400 (Bad Request) status code indicates that the server cannot or
   **will not process the request due to something that is perceived to be
   a client error** (e.g., malformed request syntax, invalid request
   message framing, or deceptive request routing).

leighmcculloch: According to the RFCs I think 403 works, but practically it will be confusing because the first thing many think of when we see that error code is credentials, even if the specs don't require it to be related to credentials. Note, there's also two definitions of 403 in two different RFCs, which I think contributes to the confusion around when to use it. [RFC7231](https://tools.ietf.org/html/rfc7231#section-6.5.3) obsoleted [RFC2616](https://tools.ietf.org/html/rfc2616#section-10.4.4).

We could also consider 422. 400 was defined to mean the request was incorrect in and of itself (e.g. format, wrong field, missing field, wrong type in a field, etc), and 422 [was added for WebDAV](https://tools.ietf.org/html/rfc4918#section-11.2) and is referenced in at least one other [non-WebDAV RFC](https://tools.ietf.org/html/rfc5789#section-2.2) and indicates there is something semantically incorrect with sending that request at this time. I think this more accurately reflects the issue here since the request parses correctly.

I think it's easy to argue that the request is bad because it's invalid given the state of the account since it is missing information, but the downside with using a 400 is a developer might jump to "what did I put wrong in this request" rather than "what about the state of the account is incomplete."

Stripe uses 402. GitHub uses 422.

I think 400 is ok. There might be some benefits to distinguishing the case with 422, but that might just replace one type of confusion with a different kind of confusion so I don't think it is a clear win.

ecosystem/sep-direct-payment.md
JakeUrban: Ok good point, how about "non-stellar account"? What I'm trying to get at is we should probably not reinforce the idea that cryptocurrency isn't "real money" by including it in our SEPs.

leighmcculloch: How about:
```suggestion
This SEP defines a protocol for enabling fiat to fiat payments between two financial accounts that exist outside of the Stellar network. The payments are facilitated by two anchors.
```

ecosystem/sep-direct-payment.md
JakeUrban: I don't think missing key-value pairs in the `"fields"` object is a valid use of 422, since the request is syntactically incorrect, not semantically incorrect. 

A good example for when 422 would make a lot of sense is where SEP-6 currently uses [403 Customer Info Needed](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0006.md#2-customer-information-needed-non-interactive). This case differs from the one we're discussing (even though they share the same name) because in SEP-6 the `"fields"` returned in the 403 response need to be sent to a _different endpoint_, namely [SEP-12's `/customer`](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0012.md#customer-put). 422 should be used in this case because the SEP-6 request was valid but couldn't be processed without a prior request to `/customer`, i.e. the semantics are incorrect.

ecosystem/sep-direct-payment.md
msfeldstein: Yeah that is no problem you can request whatever you want (as long as the sending anchor knows how to provide it)

ecosystem/sep-direct-payment.md
msfeldstein: yep, no problem

ecosystem/sep-direct-payment.md
msfeldstein: There's no SEP for transfer information but we can either create one, or add such fields in this SEP

ecosystem/sep-direct-payment.md
msfeldstein: This section being "type"?  No, all of these fields are just an example of what you could put in here. The only requirement is that there is a sender, receiver, and transaction section.  The fields you put in them are totally up to you.

ecosystem/sep-direct-payment.md
msfeldstein: I don't think we can require that they need to be from a SEP.  Anchors can request whatever they want.

ecosystem/sep-direct-payment.md
msfeldstein: Moved to 400

ecosystem/sep-direct-payment.md
msfeldstein: Added so that if the sender tries to update when the receiver doesn't allow it it should error

ecosystem/sep-direct-payment.md
msfeldstein: fixed

ecosystem/sep-direct-payment.md
msfeldstein: I think we can not worry about it, and just find out the final fee once the transaction is created (but before any money is sent).  This was more important for interactive flow stuff because it makes a nicer UX to see it up front but i think in this case, especially since both ends might take a cut, there's no need to add a /fee endpoint back yet. 

Honestly we might even remove these fee values.

msfeldstein:
@marcinx sorry i missed this comment before landing.

We actually have a proposal for registering users up front, and i think it might make sense to use this in this sep as well.  You can see it here: https://docs.google.com/document/d/1LuTKU5E_cOgQYlWBUol1dwUVO5FrVpUE4XZSXRuoNqA/edit

marcinx:
@msfeldstein I noticed that SEP-31 uses `POST /send` and `PUT /update` to create and edit payments. Would it make sense to rename those to `POST /payment` and `PUT /payment` so that they are more conform to the REST model? I think this would be the normalized way to create and update resources RESTfully.

JakeUrban:
In addition to @marcinx 's comment we could change `/transaction` to `GET /payment`.

JakeUrban:
> Just, out of curiosity. Why did SEP-12 not fit in here? It seems like such an elegant and RESTy way to decouple payment requests from customer information.

@marcinx , you have way of seeing the future  :-)

We may in fact also make the change you requested in your last comment, moving all endpoints (except `/info`) to a single, RESTful `/transactions` or `/payments` endpoint. Are you still interested in this change?

PR #596, opened 2020-04-15
[sep24] Add asset_issuer

### LINKED ISSUE #443: SEP-24: Lumen deposit/withdrawal
We have a couple of use cases on our minds that would require / benefit from the SEP-24 spec being extended to cover the native lumen asset as well.

Potential side effect: The change proposed here might also allow depositing/withdrawing assets that are not issued by that anchor.

## Proposed change

Let's define a standard way to refer to the native XLM asset in the transfer server endpoints. As far as I can see right now, that should be about the only thing blocking the feature right now.

Right now there is only the `asset_code` property which is supposed to denote the asset issued by that anchor that has this very code. Hence, there is no way to reference an asset that is not issued by this anchor.

### Option 1: Add `asset_issuer` and `asset_type`

The data structure could made consistent with the returned by Horizon, allowing the transfer server to reference any Stellar asset.

Would add a bit of bloat to the data structures, though. Might be mitigated by making `asset_issuer` and `asset_type` optional, which would also ensure backwards compatibility.

### Option 2: Magic `asset_code` value

An alternative approach might be to define a special value that indicates that we mean the native XLM asset. Such a value might be `$XLM` or similar.

It avoids adding any new props, but is less versatile, as you would not be able to reference any non-native assets issued by other anchors.

## Motivation

One motivation for this feature is to extend SEP-24 so that it serves as an open API standard for lumen onramps (like Coinbase or Moonpay). Another one is to allow offramps where you can pay in XLM, USD, EURT, ..., but without the need to issue yet another USD, EUR token.

Happy to hear your thoughts! :)

msfeldstein:
Fixes #443 
Adds an asset_issuer field to the initial POST requests when creating a transaction.  This serves 2 purposes:
1. Allows anchors to deposit/withdraw assets they didn't issue
1. Allows deposit/withdraw of lumens

It also changes the definition of asset_code to refer to the stellar asset for deposit as well as withdrawal, instead of being different for deposit and withdrawal.  This was a subtle difference that caused a lot of confusion.

### DIFF
--- a/ecosystem/sep-0024.md
+++ b/ecosystem/sep-0024.md
@@ -140,7 +140,7 @@ Request Parameters:
 
 Name | Type | Description
 -----|------|------------
-`asset_code` | string | The code of the stellar asset the user wants to receive for their deposit with the anchor.
+`asset_code` | string | The code of the stellar asset the user wants to receive for their deposit with the anchor.  If the user wants to receive lumens use the code `XLM`.
 `asset_issuer` | string | (optional) The issuer of the stellar asset the user wants to receive for their deposit with the anchor.  If the user wants to receive lumens use the issuer `native`.  If asset_issuer is not provided, the anchor should use the asset issued by themselves as described in their TOML file.
 `amount` | number | (optional) Amount of asset requested to deposit.  If this is not provided it will be collected in the interactive flow.
 `account` | `G...` string | The stellar account ID of the user that wants to deposit. This is where the asset token will be sent.
@@ -208,7 +208,7 @@ Request parameters:
 
 Name | Type | Description
 -----|------|------------
-`asset_code` | string | Code of the asset the user wants to withdraw. This must match the asset code issued by the anchor. Ex if a user withdraws MyBTC tokens and receives native BTC, the `asset_code` must be MyBTC.
+`asset_code` | string | Code of the asset the user wants to withdraw. This must match the asset code issued by the anchor. Ex if a user withdraws MyBTC tokens and receives native BTC, the `asset_code` must be MyBTC.  If the user wants to withdraw lumens use the code `XLM`.
 `asset_issuer` | string | (optional) The issuer of the stellar asset the user wants to deposit with the anchor.  If the user wants to withdraw lumens use the issuer `native`.  If asset_issuer is not provided, the anchor should use the asset issued by themselves as described in their TOML file.
 `amount` | number | (optional) Amount of asset requested to withdraw.  If this is not provided it will be collected in the interactive flow.
 `account` | `G...` string | (optional) The stellar account ID of the user that wants to do the withdrawal. This is only needed if the anchor requires KYC information for withdrawal. The anchor can use `account` to look up the user's KYC information.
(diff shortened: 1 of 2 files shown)


ecosystem/sep-0024.md
```diff
@@ -207,7 +208,8 @@ Request parameters:
 
 Name | Type | Description
 -----|------|------------
-`asset_code` | string | Code of the asset the user wants to withdraw. This must match the asset code issued by the anchor. Ex if a user withdraws MyBTC tokens and receives BTC, the `asset_code` must be MyBTC.
+`asset_code` | string | Code of the asset the user wants to withdraw. This must match the asset code issued by the anchor. Ex if a user withdraws MyBTC tokens and receives native BTC, the `asset_code` must be MyBTC.
```
JakeUrban, line 211: > This must match the asset code issued by the anchor.

This should be removed right? The anchor can back an asset it didn't issue

ecosystem/sep-0024.md
```diff
@@ -207,7 +208,8 @@ Request parameters:
...
+`asset_issuer` | string | (optional) The issuer of the stellar asset the user wants to deposit with the anchor.  If the user wants to withdraw lumens use the issuer `native`.  If asset_issuer is not provided, the anchor should use the asset issued by themselves as described in their TOML file.
```
JakeUrban, line 212:
```suggestion
`asset_issuer` | string | (optional) The issuer of the stellar asset the user wants to withdraw with the anchor.  If the user wants to withdraw lumens use the issuer `native`.  If asset_issuer is not provided, the anchor should use the asset issued by themselves as described in their TOML file.
```

ecosystem/sep-0024.md
msfeldstein: yep

msfeldstein:
@andywer does this solve your needs?

andywer:
Thanks, @msfeldstein!

Yes, that looks spot-on from our point of view 👍

tomerweller:
LGTM

Small nit - what's the expected value of `asset_code` in case `asset_issuer` is `native`? Does it become optional? null? I don't have a strong opinion - but let's be explicit about it. 

In terms of discovery - do we expect anchors to list supported currencies in their TOML [Currency Documentation](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0001.md#currency-documentation)? I think it makes sense but we need to change the language of the spec to clarify that the currency documentation is not reserved only for issues currencies (the way it is now)

andywer:
@tomerweller Completely agree, but also no strong preference here 💯

My first intuition would have been `asset_issuer = "native", asset_code = "XLM"`, but again, that was only my first thought, it's not a real preference.

> In terms of discovery - do we expect anchors to list supported currencies in their TOML Currency Documentation?

I think it really comes down to clarifying the semantics of the `stellar.toml` file. Should it list *supported* assets or does it list *issued* assets? I think up until now it felt more like a list of *issued assets*, but that doesn't mean it's the best decision.

No strong preference here either, but I'd like to sleep a night about the second question.

andywer:
PS: Regarding the `asset_code`… I think my gut feeling might have been based on the fact that the JS Stellar SDK returns `"XLM"` for `Asset.native().getCode()`. Not sure if this is consistent between the different SDKs, though.

vitchor:
LGTM

PR #604, opened 2020-04-25
Green: A Language for Building Stellar Ledger Scenarios

bartekn:


Synesso:
I love the sound of this. 👍

ire-and-curses:
This is pretty interesting! Some quick initial comments:
1) Can you fix the markdown table layout? I'm reading this in raw markdown and it's a bit annoying
2) The generality of these concepts is interesting. I wonder if any of them are derivable from the XDR, or are they higher level?
3) It's worth a discussion about whether txrep can be extended in a way that would cover the use cases covered here, or if yaml is more advantageous.

tomerweller:
I think a DSL for this purpose is great a idea. Some initial comments in no particular order:  
- YAML is great and FWIW [I argued](https://github.com/stellar/stellar-protocol/issues/159#issuecomment-418548386) for the original TxRep to be YAML. You can see dm's [reply](https://github.com/stellar/stellar-protocol/issues/159#issuecomment-418554040). One thing to be mindful of is that long nested YAML files become less readable very fast because you lose context. That's one thing that the verbose nature of txrep is pretty good at. Looking at you `tx.operations[4].body.paymentOp.amount: 400004000`. 
- @ire-and-curses this seems outside the scope of TxRep as it's not a direct mapping to XDR. Specifically things like creating keypairs and helper functions to create single-op txs. 
- If I understand correctly it doesn't seem like there's any separation between ledgers. I think that for various simulations we'd want to explicitly say which transactions makes it to which ledger to build a custom history. Though not sure what's the right YAMLy way to do that (maybe replace the `commands` array with an array of `ledgers`?)
- How about an option for an explicit `network_passphrase`? That will make it easier for the resulting network to be usable outside the scope of building the scenario (can default to something auto-generated).
- I don't fully understand the idea behind the `print` commands. Querying the ledger for info sounds like it belongs in a different tool. I'd focus on using this just to setup a scenario.

bartekn:
> Can you fix the markdown table layout? I'm reading this in raw markdown and it's a bit annoying

Fixed.

> I wonder if any of them are derivable from the XDR, or are they higher level?

I don't fully understand. If you mean commands, `transaction` and one-op commands are almost 1:1 XDR mapping (but hide low level account/asset representation from user), the rest of the commands are like `add_signer` are higher level (the concept is taken directly from `scc`).

> It's worth a discussion about whether txrep can be extended in a way that would cover the use cases covered here, or if yaml is more advantageous.

As I mentioned in the doc txrep serves a different purpose and it's too verbose. YAML gives you many language constructs that avoid repetition (that often happen when writing recipes).

> One thing to be mindful of is that long nested YAML files become less readable very fast because you lose context.

I think that it will still be more readable that the same set of instructions written in one of SDKs. And it doesn't mean there's something wrong with our SDKs - it's just each language has a specific syntax you need to follow and it has to support writing all kinds of applications, not only Stellar transactions. Whereas YAML is more a serialization language that is great for this specific purpose. It's flow mappings and map extensions are really great to make the code clear and simple (see Design Rationale).

> If I understand correctly it doesn't seem like there's any separation between ledgers. I think that for various simulations we'd want to explicitly say which transactions makes it to which ledger to build a custom history.

Correct, `scc` has `close_ledger` command to serve this purpose. I think we can easily add it here.

> How about an option for an explicit `network_passphrase`? That will make it easier for the resulting network to be usable outside the scope of building the scenario (can default to something auto-generated).

I think we can add that but recipes will be run against a specific stellar-core instance which already has network passphrase set.

> I don't fully understand the idea behind the `print` commands. Querying the ledger for info sounds like it belongs in a different tool. I'd focus on using this just to setup a scenario.

It's not perfect I agree. The idea here was to allow playgrounds to print info about the ledger at any moment (not only after all the txs are submitted). This will allow better understanding what's happening to specific ledger entries. But maybe instead of this, playgrounds should allow setting breakpoints and users can inspect ledger and history while in break point? Probably makes more sense.

Synesso:
Another option to consider is HOCON
<https://github.com/lightbend/config/blob/master/HOCON.md>. It is a
superset of JSON (all valid JSON is valid HOCON) with a focus on
human-read/writability. It supports merging.

Supported in java, go, js, ruby, python, rust, scala, c# at a minimum.

Here's an example
<https://github.com/akka/akka-samples/blob/2.6/akka-sample-cqrs-scala/src/main/resources/application.conf>
of a HOCON file that I find in the wild. Very easy to read and write.

graydon:
Generally support the idea, happy to have a simpler format for tests and playpen use-cases!

Unfortunately I'm also pretty anti-YAML -- security, complexity, incompatibility, difficulty editing, etc. Others have repeated the concerns above, but I feel them too. It'd be near the bottom of a list of options for me.

Being a critic is easy though, so for a constructive suggestion I would maybe recommend building on [Dhall](https://dhall-lang.org/)

(There are other contenders in this space -- [Starlark](https://github.com/bazelbuild/starlark) and [Jsonnet](https://jsonnet.org/) and [Cue](https://cuelang.org/) and such -- but I think Dhall draws lines at mostly the right places: includes types, total functions and freezable imports; excludes everything else.)

bartekn:
Thanks for the comments everyone! We're probably months from the first playground implementations so I have plenty of time to process everything.

Honestly, I wasn't aware that some YAML parsers allowed command execution. This sounds bad and we should take this into account when thinking about the spec, or, if we use YAML, specify which parsers we recommend.

At the same time I think we should separate security concerns of underlying technologies and SEP itself. What I mean is this SEP assumes that users will use well-tested YAML parser. It's possible that XDR parsers have some nasty bugs inside, yet we still use XDR in Stellar and we don't mention possible XDR bugs in CAPs. In fact, we found such bugs in our Go and C++ parsers and in some of community maintained SDKs. Or, we don't include info about [possible JSON parser bugs](http://seriot.ch/parsing_json.php) in SEPs that are built on top of JSON.

MonsieurNicolas:
Note that core supports the building blocks for a "playground" mode as of [v14.1.0](https://github.com/stellar/stellar-core/releases/tag/v14.1.0).

Related to this proposal.

We found that to craft the ledger state it was actually much easier to use transactions than to use ledger entries as allowing arbitrary ledger entries to be created came with a new problem: overall invalid ledger state. For example you can image offers being created without corresponding trustline or account.

PR #610, opened 2020-04-28
[SEP24] Remove interactive call no-info-needed response

### LINKED ISSUE #539: SEP-24: Remove 'Success: no additional information needed' section
TL;DR: We should remove the `/interactive` [success](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0024.md#1-success-no-additional-information-needed) response since we now have SEP-26 for non-interactive flows and we changed the `/interactive` endpoints from `GET` to `POST`.

Originally SEP-24's `/interactive` endpoints used to be `GET` requests, and had two different types of responses; one where more information is needed and another success response.

On the 'success' response:
> This is the correct response if the anchor is able to execute the withdrawal and needs no additional information about the user.

This sentence implies that SEP-24 anchors should support a non-interactive case, but SEP-26 was developed to address the need for non-interactive flows.

Furthermore, these endpoints were changed from `GET` requests to `POST` requests. `POST` requests should always be non-idempotent, but the SEP outlines non-idempotent behavior:
> It should also be used if the anchor requires information about the user, but the information has previously been submitted and accepted.

Thanks @andywer for pointing this out on stellar/django-polaris#149

msfeldstein:
SEP24 always has an interactive portion so we shouldn't provide a case that was from SEP6 where no information was needed and it was safe to not do any interactive portion.  
Fixes #539

### DIFF
--- a/ecosystem/sep-0024.md
+++ b/ecosystem/sep-0024.md
@@ -237,40 +237,9 @@ There are several possible kinds of response, depending on whether the anchor ne
 
 The first response, the success response, is explained below. The other possible responses are shared with the deposit endpoint, and are explained in the [Deposit and Withdraw shared responses](#deposit-and-withdraw-shared-responses) section directly below.
 
-#### 1. Success: no additional information needed
-
-Response code: `200 OK`
-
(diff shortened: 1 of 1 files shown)


JakeUrban:
Lets bump this. @tomerweller does this change still look good?

JakeUrban:
@tomerweller bump

PR #625, opened 2020-06-01
[SEP-24] Additive changes for XLM deposit and withdrawals

JakeUrban:
The changes implemented provide clarification for supporting XLM deposits and withdraws via SEP-24. This is one way to support this functionality, other ideas are welcome.

One open question:

Should I add clarifications for the `amount_in`, `amount_out`, and `amount_fee` columns returned from `/transaction(s)` for XLM deposit/withdraws?

### DIFF
--- a/ecosystem/sep-0024.md
+++ b/ecosystem/sep-0024.md
@@ -142,7 +142,7 @@ Name | Type | Description
 -----|------|------------
 `asset_code` | string | The code of the stellar asset the user wants to receive for their deposit with the anchor.  If the user wants to receive lumens use the code `XLM`.
 `asset_issuer` | string | (optional) The issuer of the stellar asset the user wants to receive for their deposit with the anchor.  If the user wants to receive lumens use the issuer `native`.  If asset_issuer is not provided, the anchor should use the asset issued by themselves as described in their TOML file.
-`amount` | number | (optional) Amount of asset requested to deposit.  If this is not provided it will be collected in the interactive flow.
+`amount` | number | (optional) Amount of asset requested to deposit.  If this is not provided it will be collected in the interactive flow.  If the user wants to receive lumens this argument should be ignored.
 `account` | `G...` string | The stellar account ID of the user that wants to deposit. This is where the asset token will be sent.
 `memo_type` | string | (optional) Type of memo that anchor should attach to the Stellar payment transaction, one of `text`, `id` or `hash`.
 `memo` | string | (optional) Value of memo to attach to transaction, for `hash` this should be base64-encoded.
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0024.md
```diff
@@ -142,7 +142,7 @@ Name | Type | Description
 -----|------|------------
 `asset_code` | string | The code of the stellar asset the user wants to receive for their deposit with the anchor.  If the user wants to receive lumens use the code `XLM`.
 `asset_issuer` | string | (optional) The issuer of the stellar asset the user wants to receive for their deposit with the anchor.  If the user wants to receive lumens use the issuer `native`.  If asset_issuer is not provided, the anchor should use the asset issued by themselves as described in their TOML file.
-`amount` | number | (optional) Amount of asset requested to deposit.  If this is not provided it will be collected in the interactive flow.
+`amount` | number | (optional) Amount of asset requested to deposit.  If this is not provided it will be collected in the interactive flow.  If the user wants to receive lumens this argument should be ignored.
```
JakeUrban, line 145: `amount` should be ignored if `asset_code` is `XLM` because the amount of XLM deposited to the user's stellar account is dependent on the amount sent to the anchor's non-stellar account and the rate of exchange between XLM and the non-stellar asset.

ecosystem/sep-0024.md
```diff
@@ -483,7 +516,8 @@ Name | Type | Description
 `operation` | string | Kind of operation (`deposit` or `withdraw`).
 `type` | string | (optional) Type of deposit or withdrawal (`SEPA`, `bank_account`, `cash`, etc...).
 `asset_code` | string | Asset code.
-`amount` | float | Amount of the asset that will be deposited/withdrawn.
+`external_asset_code` | string | The asset code of the external asset the user will send or receive, if the value of `asset_code` is `XLM`.
+`amount` | float | Amount of the asset that will be deposited or withdrawn. If `asset_code` is `XLM`, `amount` should be in units of `external_asset_code` if `operation` is `deposit` and in units of XLM if `operation` is `withdraw`.
```
JakeUrban, line 520: The units of `amount` should depend on the asset the user is sending to the anchor. For example, the user sends the anchor XLM for a withdraw, but a non-stellar asset for a deposit.

tomerweller:
The "If XLM then..." clauses feel like a a spec smell. First of all, it's not just XLM, it's any case in which you deposit one asset in exchange of another. It could also be EURO->BTC for that matter. Second, it adds a lot of rope for developers to hang themselves with every time an `amount` field shows up. 

If this was just the info/fees I'd say let's have an option for an anchor to not specify any fees - they can do that in the interactive portion.  

The bigger problem is that we're drifting away from deposit/withdraw into the realm of purchase/exchange and providing off-chain liquidity. If you squint hard they might look the same but the semantics are different and I'm afraid they will make sep24 more complicated for everyone.  

I'd love to hear @andywer's thoughts as he brought this up initially: https://github.com/stellar/stellar-protocol/issues/443

andywer:
Yes, I would say so, too. Don't see the need for a separate `external_asset_code` and would also argue rather strongly against it (also for consistency's sake).

I guess the fees endpoint was just forgotten when adding `asset_issuer` to the spec before.

@JakeUrban Was there any other reason for adding the `external_asset_code` or do you agree that we can simply replace it by an `asset_issuer` that might be set to `"native"` like in the other endpoints?

JakeUrban:
@andywer unless the anchor expresses the fee in units of XLM it needs `external_asset_code`. Fees should always be described in terms of the units the user is actually sending/receiving from the anchor.

For example, if a user wants to send 100 USD to Anchor A, Anchor A should tell the user the fee is 2 USD, not 16 XLM. In the same way, when the user wants to withdraw 800 XLM for 100 USD, the anchor should express the fee in USD, not XLM.

But regardless, I agree with @tomerweller's assessment that SEP-24 isn't a great fit for what we're trying to accomplish. It is close, but the team is considering creating a new SEP for this use case instead of making adjustments to SEP-24. What are your thoughts on that?

andywer:
Hmmm, I see your point now… 🙈

The XLM deposit/withdrawal case really only seems to make sense with a conversion to another asset on the fly and that then requires the `external_asset_code`. Drats.

A separate SEP sounds reasonable at this point (up to you guys, of course), but I am also a bit concerned that we are heading towards a small explosion of SEPs that are 90% the same here. Not sure how to tackle this.

As no one has implemented XLM deposit/withdrawal yet AFAIK, maybe it's still a good time to drop that feature from SEP-24 again and re-open the original feature request for the time being to figure out what to do instead?

PR #635, opened 2020-06-10
ALL: Remove use of the terms whitelist/blacklist

leighmcculloch:
### What
Remove use of the terms whitelist and blacklist, replacing them with other descriptions of what functionality is intended or with other terms allowlist and denylist.

### Why
The terms whitelist and blacklist make many people feel unwelcome and can trigger racial stereotyping. A similar change was made on the Go repository which prompted me to check if we used these terms in our documentation or code. golang/go@608cdcaede1e7133dc994b5e8894272c2dce744b

The terms whitelist and blacklist are not better at communicating intent than simply describing the behavior we're looking for or than using another term. This seems like an easy change to make with no cost that improves the general language of our documentation. An example of how this generally improves the documentation is this sentence where we state:

>...test them with your wallet to be sure they work before allowing them to be used with your wallet.

Which is much clearer and easier to read than the original:

>...test them with your wallet to be sure they work before whitelisting them.

This changes previously accepted documents. I don't think that is particularly an issue given than the meaning of the documents remains the same.

### DIFF
--- a/ecosystem/sep-0006.md
+++ b/ecosystem/sep-0006.md
@@ -80,7 +80,7 @@ SEP-6 lays out many options for how deposit and withdrawal can work. These are r
 
 ### Basic Wallet Implementation
 
-* Identify anchors you want to support manually, and test them with your wallet to be sure they work before whitelisting them. We encourage you to support as many anchors as possible.
+* Identify anchors you want to support manually, and test them with your wallet to be sure they work before allowing them to be used with your wallet. We encourage you to support as many anchors as possible.
 * For each anchor, use information from its [`stellar.toml`](sep-0001.md) file and its [`/info`](#info) endpoint to display useful information to the user about the asset they've picked.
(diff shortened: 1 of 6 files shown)


ecosystem/sep-0006.md
```diff
@@ -80,7 +80,7 @@ SEP-6 lays out many options for how deposit and withdrawal can work. These are r
 
 ### Basic Wallet Implementation
 
-* Identify anchors you want to support manually, and test them with your wallet to be sure they work before whitelisting them. We encourage you to support as many anchors as possible.
+* Identify anchors you want to support manually, and test them with your wallet to be sure they work before allowing them to be used with your wallet. We encourage you to support as many anchors as possible.
```
ire-and-curses, line 83: In several places: "Authorising" rather than "allowing" mores exactly replicates the original intent (it's less woolly). I don't feel strongly about this though.

leighmcculloch: I avoided the word authorizing here because the term is used in relation to trustlines, where-as here we're talking about an application choosing to support an anchor in its product.

PR #636, opened 2020-06-10
[SEP31] Add callback for transactions

msfeldstein:
Currently in SEP31 sending anchors need to poll each transactions status to see when it's changed, in case it wants to send email updates to their clients.  This can be a lot of network requests, so it is useful to allow for a callback URL so the receiving anchor can push updates to them rather than polling.

### DIFF
--- a/ecosystem/sep-0031.md
+++ b/ecosystem/sep-0031.md
@@ -202,7 +202,7 @@ Content-Type: application/json
       "type": "SEPA"
     }
   },
-  "callback": "https://myanchor.com/sep31/callback/423253235"
+  "callback": "https://sendinganchor.com/sep31/callback/423253235"
 }
 ```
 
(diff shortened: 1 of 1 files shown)


gaclaudiu:
> Don't we need to describe the POST request schema for the callback? For example, what if the status changes to `customer_info_update` instead of `completed`?
> 
> I would suggest having a simple `status` attribute in the JSON body:
> 
> ```
> {
>     "status": "pending_info_update"
> }
> ```
> 
> The changes in this PR also imply that the sending anchor must specify a callback on a per-transaction basis, which I think makes sense, but another way to do this would be to have a standard callback URL with the ID in the POST body:
> 
> ```
> {
>     "id": 423253235,
>     "status": "completed"
> }
> ```

I like the idea to have a standard callback URL rather then specify the callback url in each transaction send request. 
I think it's less error prone. 

And the callback can be discover through the stellar toml file , something like Direct_Payment_Server/callback.
So if an anchor has this callback defined in toml file it will get notifications  if is not defined they have to use polling on the get transaction endpoint. 

What do you guys think ?

msfeldstein:
I think a standard URL is fine, will update.

The schema of the callback is defined to be the same as the /transaction endpoint, so 
```
{
  "transaction": {
      "id": "82fhs729f63dh0v4",
      "status": "pending_info_update",
      "status_eta": 3600,
      "external_transaction_id": "ABCDEFG1234567890",
      "amount_in": "18.34",
      "amount_out": "18.24",
      "amount_fee": "0.1",
      "started_at": "2017-03-20T17:05:32Z",
      "required_info_message": "The bank reported an incorrect name for the receiver, please ensure the name matches legal documents",
      "required_info_updates": {
         "receiver": {
            "first_name":"The receiver's first name",
            "last_name":"The receiver's last name"
         }
      }
    }
}
```
This has the ID and status in it.  I can make this more clear

JakeUrban:
@gaclaudiu we have some reservations about introducing callback functionality to the SEP, since it would require the sending anchor to have a public-facing web server. The sending anchor would also probably _still poll_ in case the receiving anchor doesn't or can't make the callback request.

So we'd like to explore improvements to the polling strategy before requiring more from the sending anchor.

Another solution worth considering is changing the `/transaction` endpoint in such a way to allow sending anchors to easily poll for updates to multiple transactions. 

We could change `/transaction` to `/transactions` with a `sort_by=updated_at` argument. The receiving anchor would return a page-worth of the most-recently updated transactions created by the sending anchor.

```
GET /transactions?account=<sending anchor address>&sort_by=updated_at&page_size=100
```

This way, sending anchors wouldn't need to make 1 request per transaction. This makes the polling mechanism much more scalable for the receiving anchor.

PR #639, opened 2020-06-11
SEP-0032: Add proposal for user-friendly asset addresses and process for verifying them

leighmcculloch:
### What
Add a proposal for user-friendly asset addresses that includes a process for verifying them.

### Why
There does not seem to be a consistent user-friendly approach to referencing or resolving unique assets in products that are built on Stellar. Each wallet or application I use has a different approach and
it is an unpleasant experience to find assets using their issuer addresses.

### Discussion
Please take discussion to the [stellar-dev mailing list](https://groups.google.com/d/topic/stellar-dev/rWThI_BIs0Y). Any minor quick feedback such as typos, etc feel free to leave here in the PR.

### DIFF
--- a/ecosystem/sep-0032.md
+++ b/ecosystem/sep-0032.md
@@ -5,7 +5,7 @@ SEP: TBA
 Title: Asset Address
 Author: Leigh McCulloch (@leighmcculloch)
 Status: Draft
-Discussion: TBA
+Discussion: https://groups.google.com/d/topic/stellar-dev/rWThI_BIs0Y
 Created: 2020-06-11
 Version 0.1.0
 ```
--- a/ecosystem/sep-0032.md
+++ b/ecosystem/sep-0032.md
@@ -71,7 +71,7 @@ a string with the parts separated by a colon:
 ```
 
 Where:
-- `{asset-code}` is the 1-12 alphanumeric [Stellar asset code].
+- `{asset-code}` is the 1-12 alphanumeric Stellar asset code.
 - `{asset-domain}` the domain name that is hosting a `stellar.toml` file that
   contains a [`CURRENCIES`] section for the asset, as defined by [SEP-1].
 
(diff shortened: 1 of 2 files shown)


ecosystem/sep-0032.md
```diff
@@ -0,0 +1,147 @@
...
+- The client retrieves the `stellar.toml` for the domain located at
+  `https://{asset-domain}/.well-known/stellar.toml`.
+- The client parses the `stellar.toml` and finds a `CURRENCIES` section with
+  `code` equal to `{asset-code}`.
+- The client retrieves any currency.toml files as reference by the
+  `stellar.toml` as required.
+- If none is found, the client finds a `CURRENCIES` section with
+  `code_template` that matches `{asset-code}`.
+- If none is found, the client fails to resolve the asset address.
+- If one is found, the client continues to resolve the asset's issuer address.
```
abuiles, line 104: @leighmcculloch by reading SEP001, it seems possible to have two assets with the same `CODE`

leighmcculloch:
TODO:
- [ ] Add specification section about reverse lookup from asset code and asset issuer to asset address (code + domain)
- [ ] Add rational section about why the colon-separated format is selected
- [ ] Add notes to specification sections stating that applications should treat addresses as ephemeral and never store them, instead storing what they resolve to, since addresses could change at anytime. Applications can use reverse lookup to then turn what they have stored into valid names.

ecosystem/sep-0032.md
leighmcculloch: Thanks, that's a good point. I don't think there's anyway to support assets that are defined multiple times unless we say that the first currency defined with an asset code is the one resolved to, or we add a field that signals which is selected by the address address. Both of these options complicate the issue. I'd be interested to know how many meaningful asset issuers are doing this, I might spend sometime using the BigQuery data set and some scripts to see if I can figure out how many issuers do this. I also think this is a practice that should probably be discouraged though since an issuing organization reusing the same code is confusing for consumers in general. Asset addresses are geared towards consumers so we could just say it doesn't provide a capability to reference multiple when they are multiple. I added a section about that to the proposal for now in 2e62e03.

leighmcculloch:
I've merged so that we can discuss further on the [stellar-dev mailing list](https://groups.google.com/d/topic/stellar-dev/rWThI_BIs0Y).

PR #643, opened 2020-06-12
SEP31: Add asset_code and asset_issuer to /send request parameters

JakeUrban:
These changes are not strictly necessary because SEP31 requires the 2 participating anchors to come to an agreement prior to facilitating transactions on behalf of their users. This agreement would presumably include a list of acceptable assets and their respective issuers.

However, it is better for the receiving anchor to reject `/send` requests when they made rather than having to refund payments of an unaccepted asset. It also allows receiving anchors to create transaction records in their database including the asset expected.

### DIFF
--- a/ecosystem/sep-0031.md
+++ b/ecosystem/sep-0031.md
@@ -227,7 +227,7 @@ Name | Type | Description
 `require_receiver_info` | array | A list of [SEP-9](sep-0009.md) values requested for the receiving client
 `amount` | number | Amount of payment in destination currency
 `asset_code` | string | Code of the asset the sending anchor intends to send. This must match one of the entries listed in the receiving anchor's `/info` endpoint.
-`asset_issuer` | string | optional. The issuer of the stellar asset the sending anchor intends to send. If not specified, the the asset sent must be issued by the receiving anchor.
+`asset_issuer` | string | (optional) The issuer of the stellar asset the sending anchor intends to send. If not specified, the the asset sent must be issued by the receiving anchor.
 `fields` | object | A key-pair object containing the values requested by the receiving anchor in their `/info` endpoint, broken out by sender, receiver, and transacton.
 `lang` | string | (optional) Defaults to `en`. Language code specified using [ISO 639-1](https://en.wikipedia.org/wiki/ISO_639-1).  Any human readable error codes or field descriptions will be returned in this language.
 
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0031.md
```diff
@@ -194,6 +194,8 @@ Content-Type: application/json
 
 {
   "amount": 100,
+  "asset_code": "USD",
+  "asset_issuer": "GDRHDSTZ4PK6VI3WL224XBJFEB6CUXQESTQPXYIB3KGITRLL7XVE4NWV",
```
leighmcculloch, line 198: Did you folks also consider using the single field string representation that's defined in SEP-11? We've started using that in some places instead of separate fields.

JakeUrban: We didn't, but I don't I see the value in using that notation. The same info is communicated without requiring the receiver to parse the string. And the SEP-11 notation requires the client specify an issuer, where in this spec the issuer is optional.

abuiles: @leighmcculloch @jacekn was there a reason for not using `asset: assetcode:assetissuer` which is the canonical representation for assets? we are moving towards this representation in Horizon

abuiles: btw, we are trying to push for SEP-11 notation on Horizon and it would be great if we start doing the same in the SEPs

leighmcculloch: The main value is consistency so that assets are represented the same everywhere. It means that when integrating with different SEPs developers can pass the same strings without needing to format differently. But the ability to specify only the code makes it interesting and complicated, SEP-11 doesn't support that well because you wouldn't be able to distinguish between native the asset code or native the asset type.

abuiles: @leighmcculloch 
> SEP-11 doesn't support that well because you wouldn't be able to distinguish between native the asset code or native the asset type.

Why not?

leighmcculloch: @abuiles If @JakeUrban's user wishes to specify exactly the native asset they would specify it as `"asset": "native"` according to SEP-11. If the user wishes to specify any issued asset that has the code `native` and any issuer and omitted the `:<issuer>` they would also specify it as `"asset": "native"`. There'd be no way for the server to distinguish if the request was specifying the native asset or an issued asset with code `native`.

abuiles: @leighmcculloch ok, makes sense! I missed the last bit on Jake's comment. :)

PR #646, opened 2020-06-16
SEP31: Exclude /info from endpoints requiring SEP10 JWT

JakeUrban:


ecosystem/sep-0031.md
```diff
@@ -34,20 +34,14 @@ NigeriaPay will utilize its European rail, enabled with EuroPay Anchor service,
...
 Authorization: Bearer <JWT>
 ```
 
-Any API request the fails to meet proper authentication should return a 401 Unauthorized response with the error payload:
-
-```
-{
-  "type": "authentication_required"
-}
-```
```
JakeUrban, line 50: We use this response in SEP-6 because there are other responses defined that share the same status code. For example, both SEP-6 bad auth and customer_info_needed responses use 403 and include the `type` attribute to disambiguate.

In SEP-31, the 401 status code is only used for bad auth, so there is no need to specify a `type`.

PR #647, opened 2020-06-16
SEP-0005: Replace uses of the term account with key

leighmcculloch:
### What
Replace uses of the term account with key in all cases where the SEP is likely to be used for a Stellar key that isn't necessarily a Stellar account.

### Why
A Stellar account is a key which has led to some of our documentation confusingly referencing an account when in fact the SEP is intended to work with any key. This is an important distinction for accounts that use multi-sig. SEP-5 is appropriate for multi-sig accounts and for generating keys that will be used as signers for an account where that signer is just a key and not an account.

This SEP is in final status but this change clarifies the intent and makes no functional change to the specification and so I don't think it is worthy of a new SEP to make this distinction.

This change was proposed by @tomquisel.


leighmcculloch:
>Should the mentions of `Accounts` in the test cases section be converted to `keys` as well?

@tomquisel Good catch, thanks! Fixed in 9c71b7e.

PR #648, opened 2020-06-18
Change SEP-8 to submit transactions from server

msfeldstein:
One of the complicating factors of SEP-8 is keeping an accurate state of the world.  Currently we have the wallet/client submit the transaction, but this adds the ability for a wallet to hold on to a transaction and submit another transaction for approval.  The server needs to keep track of all outstanding, signed transactions which makes servers a lot more complicated and error prone.  This approach means the server only needs to check the state of the ledger as the source of truth.

### DIFF
--- a/ecosystem/sep-0008.md
+++ b/ecosystem/sep-0008.md
@@ -30,11 +30,12 @@ Implementing a Regulated Asset consists of these parts:
 ## Regulated Assets Approval Flow
 
 1. User creates and signs a transaction.
-2. Wallet resolves asset information and detects that it's a regulated asset.
-3. Wallet sends the transaction to the approval server.
-4. Approval server determines whether the transaction is compliant based on the current state of the ledger, known pending transactions, and its compliance ruleset.
-5. Wallet handles approval response:
-    1. *Success?* Transaction has been approved and signed by issuer. Submit to Horizon.
+1. Wallet resolves asset information and detects that it's a regulated asset.
+1. Wallet sends the transaction to the approval server.
+1. Approval server determines whether the transaction is compliant based on the current state of the ledger, known pending transactions, and its compliance ruleset.
+1. If the transaction is approved by the approval server it gets signed and submitted to horizon
+1. Wallet handles approval response:
+    1. *Success?* Transaction has been approved, signed, and submitted to horizon by the issuer.
     2. *Revised?* Transaction has been revised to be made compliant, and signed by the issuer. Wallet will show the changes to the user and ask them to sign the new transaction. Submit to Horizon once signed.
     3. *Pending?* The issuer will need to asynchronously approve this transaction. Wallet should send the same transaction to the approval service after a specified timeout (Return to 3).
     4. *Action Required?* Transaction requires a user action to be completed. Wallet will present the attached action url as a clickable link, along with the attached message and an option to resubmit once the action has been taken.
@@ -64,7 +65,7 @@ approval_criteria="The goat approval server will ensure that transactions are co
 The transaction approval server receives a signed transaction, checks for compliance and signs if possible.
 
 ### Possible results
-- Success: Transaction was found compliant and signed by service.
+- Success: Transaction was found compliant and signed and submitted by service.
 - Revised: Transaction was modified to be compliant and signed by service. It should be re-signed by the client.
 - Pending: The issuer will asynchronously validate the transaction and respond later.
 - Action Required: The user must complete an action in order to have the transaction approved.
@@ -144,7 +145,6 @@ error|string|A human readable string explaining why the transaction is not compl
 - Core operations shouldn't be modified as that can be confusing and misleading. For example, if the user wishes to put an offer for 1000 GOATS but due to velocity limits they can only put an offer for 500 GOATS, it is better to error with a message than to change the amount.
 - Adding an upper timebound to a transaction can help the issuer ensure that their view of the world does not get out of sync.
 - Issuers can enforce additional fees by adding additional operations. For example, any transaction involving goats, will also send 0.1 GOAT to Boris’ account.
-- Once a transaction has been signed, the issuer should mark it as pending and take it into account, as long as it hasn’t been timed out, so that they can have a accurate view of the world.
 
 ## Account Setup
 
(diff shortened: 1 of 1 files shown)


msfeldstein:
The original solution was to create a short timebounds to ensure things couldn't hang around, but this still leaves the possibility that the transaction will never be submitted, so the anchor still doesn't know what the impact of a transaction it signed will be.  It becomes a Schroedingers Transaction.

ecosystem/sep-0008.md
```diff
@@ -159,8 +159,8 @@ In the future, this requirement can be replaced by introducing protocol level [C
...
 Alternatively, some other available options are to utilize [SEP006](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0006.md) in order to perform KYC on deposit and withdraw, and/or use [AUTHORIZATION_REQUIRED](https://www.stellar.org/developers/guides/issuing-assets.html#requiring-or-revoking-authorization) which allows issuers to authorize specific accounts to hold their issued assets.
 
-### Why doesn’t the approval service submit transactions to the network?
+### Why does the approval service submit transactions to the network?
 
-- Separation of concerns between signing transactions and submitting them.
-- Transactions might require more signatures from further regulated asset issuers.
+- The complicating factor of having approved, but not submitted transactions makes it difficult for approval servers to keep track of the state of the world.
+- One downside is that transactions involving multiple regulated assets are no longer possible.
```
JakeUrban, line 165: Why is this the case?

leighmcculloch:
>the anchor still doesn't know what the impact of a transaction it signed will be. It becomes a Schroedingers Transaction.

If the implementation will just be relying on Horizon for the state of the world then the anchor won't know what the impact of a transaction is other than by looking at Horizon, which is similar to handing the transaction back to the client and not knowing and just looking at the network.

>The server needs to keep track of all outstanding, signed transactions which makes servers a lot more complicated and error prone.

Would it be sufficient for the SEP to state that the server can assume the transaction will be submitted immediately by the client and any subsequent requests to the server will be as if the transaction has not taken place if it is not submitted?

leighmcculloch:
Something to consider: It is more difficult to scale services that make long running outgoing requests.

ecosystem/sep-0008.md
msfeldstein: This should actually say transactions involving multiple issuers, if there's two regulated assets by the same approval server its no problem.

If the transaction has 2 regulated assets it means it needs to be approved by 2 different approval servers.  If we have assets REGA and REGB, the REGA server wants to approve it but its not signed by REGB server, and vice versa.  In the original implementation we could get approved by REGA, get the tx signed by them, and then submit it to REGB server for signing.

msfeldstein:
> If the implementation will be relying on Horizon for the state of the world then the anchor won't know what the impact of a transaction is other than by looking at Horizon, which is similar to handing the transaction back to the client and not knowing and just looking at Horizon.

Yeah but at least you will know within 5 seconds.  I guess this isnt so different from the time-bounded transaction scheme that we currently have. 

@IlyaShkapo do you have any thoughts on which scheme would be better for a company that actually needs to maintain this?

tomerweller:
The question of who submits is a tough one. 

I tend to agree with @leighmcculloch that having the server submit the transaction doesn't smell great from a system architecture perspective. In fact, we had some discussions around this topic when SEP8 was first published and agreed that leaving it to the client would make implementing approval servers easier. 

At the end, it boils down to how we expect approval servers to behave. I think that the simplest behavior is the naive approach of setting up a global lock when approving a TX and unlocking as soon as that transaction is published. It limits approval rates to once per ledger (not great, but we're talking securities style assets, they shouldn't move all that much). You can move from a global lock to a per user lock but that becomes somewhat complex with global policy rules ("global volume per day" style restrictions), not to mention orderbook operations. 

As @leighmcculloch If you go down this route with client submissions you loose control of how long it takes to unlock. 

> Would it be sufficient for the SEP to state that the server can assume the transaction will be submitted immediately by the client and any subsequent requests to the server will be as if the transaction has not taken place if it is not submitted?

@leighmcculloch can you unpack this? do you mean that new transactions that are approved will somehow invalidate previous transactions that have been approved? Or multiple "approved but not submitted" transactions can live together in the wild?

leighmcculloch:
>@leighmcculloch can you unpack this?

I was meaning the server only looks at the network to understand current state. If a first transaction is approved and not submitted yet, a second transaction would be considered on the basis the first transaction doesn't exist. The invalidation of previously approved transactions might be important but that'll be difficult so I'd expect the server to only approve the transaction if the sequence number was +1 of the accounts current sequence number on the network which will have the effect of only allowing one of the approved transactions to proceed which isn't quite the same as saying it will invalidate all previous transactions.

I'm suggesting this not confident I know all the use cases though. It'd be useful to have a write up of all the known use cases that transactions might be approved or denied.

msfeldstein:
Given that we need to wait for a ledger to close to understand the effects of the transaction, it might make sense to leave the original scheme of "make signed transactions fast-expiring".  We're already in the realm of needing a lock for single-second timeframes, so it wouldn't make too much of a difference.

It also gives implementors the ability to do much more complicated logic if they actually care to while allowing others to keep the "lock the account(s) for N<10 seconds" scheme

msfeldstein:
I'm going to close this for now.  For implementors, you'll need to build an N-Second time-based lock on accounts anyway so it shouldn't change the architecture too much if we change this later, and we might as well leave it more flexible to see if we end up needing multi-regulated-asset or multi-signer support

tomerweller:
@leighmcculloch account seq numbers are a good tool for synchronization under the assumption that your rules don't cross account boundaries. With regulated assets they'll often do. 

@msfeldstein I don't fully understand your last comment. What's the expeced behavior of an approval server when it receives approval requests while there are in-flight transactions?

msfeldstein:
It's the same behavior whether the server or client submits the transaction, it rejects it with a 'try again in 30 seconds' message.  If the server submits the transaction you know there is a transaction in flight for 5 seconds.  If the client submits the transaction you know there is a transaction in flight for 10 seconds (5s-ledger + 5s of timelock allowing the client to hold on to it before submitting).

PR #650, opened 2020-06-22
CAP-0033 Clarifications and Fixes

jonjove:


core/cap-0033.md
```diff
@@ -68,8 +68,14 @@ typedef AccountID* SponsorshipDescriptor;
 
 struct AccountEntryExtensionV2
 {
-    uint32 numSponsored;
-    uint32 numSponsoring;
+    uint32 numSponsored;  // Number of reserves sponsored by this account
+                          // This reduces the base reserve requirement
+    uint32 numSponsoring; // Number of reserves sponsored for this account
+                          // This increases the base reserve requirement
```
ire-and-curses, line 74: In our working group we said
>numSponsoring: how many (reserves) I’m sponsoring (plus sign in the base reserve formula)
numSponsored:  how many of my reserves are sponsored by someone else (minus sign in the base reserve formula)

This seems like the opposite of the comment in this PR? Which is correct?

leighmcculloch: I'm still very confused by these descriptions. I thought I understand until I read the second comment on each. Is `numSponsored` reducing the base reserve of this account or another account?

core/cap-0033.md
```diff
@@ -603,6 +605,34 @@ ecosystem. Storing the sponsoring IDs for signers separately avoids this problem
...
+So how would a transfer work if `UpdateSponsorshipOp` were to obey the typical
+rules of sponsorship? Let `S1` be the current sponsor, `S2` be the new sponsor,
+`K` be the `LedgerKey` for an entry sponsored by `S1`, and `A` be the account
+which controls the entry identified by `K`. We would need a transaction with
+
+1. `S2` is source for `SponsorFutureReservesOp` with `sponsoredID = A`
+2. `S1` is source for `UpdateSponsorshipOp` with `ledgerKey = K`
+3. `A` is source for `ConfirmAndClearSponsorshipOp`
+
+But this clearly requires a signatures from the sponsored account, so if we are
```
ire-and-curses, line 631:
```suggestion
But this clearly requires a signature from the sponsored account, so if we are
```

core/cap-0033.md
```diff
@@ -181,6 +189,7 @@ enum OperationType
 
 struct SponsorFutureReservesOp
 {
+    // Source account will sponsor future reserves for this account
```
leighmcculloch, line 192: This comment should lead with the field so that it is clear who the field is in the comment. Because it leads with the source account that is not in this struct it is easy to think the source account is the sponsoredID here.

This might be a bit wordy for you, but:
```suggestion
    // SponsoredID references the account for which future reserves will be sponsored by the source account of the operation this SponsoreFutureReservesOp is specified within.
```

core/cap-0033.md
```diff
@@ -275,11 +290,12 @@ enum UpdateSponsorshipResultCode
...
 
     // codes considered as "failure" for the operation
-    UPDATE_SPONSORSHIP_DOES_NOT_EXIST = -1,
-    UPDATE_SPONSORSHIP_NOT_SPONSOR = -2,
-    UPDATE_SPONSORSHIP_LINE_FULL = -3,
-    UPDATE_SPONSORSHIP_LOW_RESERVE = -4,
-    UPDATE_SPONSORSHIP_ONLY_TRANSFERABLE = -5
+    UPDATE_SPONSORSHIP_DOES_NOT_EXIST = -1,   // specified entry does not exist
+    UPDATE_SPONSORSHIP_NOT_SPONSOR = -2,      // not sponsor of specified entry
+    UPDATE_SPONSORSHIP_LOW_RESERVE = -3,      // owner cannot afford this entry
```
leighmcculloch, line 295: The term owner is new. There's no field owner that I can see, can we reference who cannot afford the entry specifically?

core/cap-0033.md
```diff
@@ -275,11 +290,12 @@ enum UpdateSponsorshipResultCode
...
+    UPDATE_SPONSORSHIP_ONLY_TRANSFERABLE = -4 // sponsorship of ClaimableBalance
+                                              // must be transferred to another
+                                              // account
```
leighmcculloch, line 298: Does this apply to only claimable balance? I would think this could happen for any sponsorship?

core/cap-0033.md
```diff
@@ -375,7 +391,12 @@ If op.sponsoredID = op.sourceAccount
 The behavior of `SponsorFutureReservesOp op` is:
 ```
 If an account is sponsoring future reserves for op.sponsoredID
-    Fail with SPONSOR_FUTURE_RESERVES_ALREADY_EXISTS
+    Fail with SPONSOR_FUTURE_RESERVES_ALREADY_SPONSORED
```
leighmcculloch, line 394: This refers to already being sponsored just in the ephemeral sense inside this transaction, is that right?

core/cap-0033.md
jonjove: The comments from the working group are correct. Thanks, fixed.

core/cap-0033.md
jonjove: Good eye. Fixed.

core/cap-0033.md
```diff
@@ -499,6 +522,10 @@ Similarly, `AccountMergeOp` will fail with `ACCOUNT_MERGE_IS_SPONSOR` if
 attempting to merge an account that has non-zero `numSponsoring`. This is
 required for sponsorship bookkeeping.
 
+#### CreateAccountOp
+`CreateAccountOp` will now be valid if `startingBalance >= 0`, whereas prior to
+this proposal it was valid if `startingBalance > 0`.
```
leighmcculloch, line 527: 💯 :shipit:

core/cap-0033.md
```diff
@@ -651,10 +717,71 @@ operations[5]:
...
+operations[0]:
+    sourceAccount: S2
+    body:
+        type: SPONSOR_FUTURE_RESERVES
+        sponsoredID: S1
+operations[1]:
+    sourceAccount: S1
+    body:
+        type: UPDATE_SPONSORSHIP
+        ledgerKey: K
```
leighmcculloch, line 765: It's difficult from looking at these two example transactions in isolation to understand what the update sponsorship is doing. Would it be feasible to give these specific names:

- `REMOVE_SPONSORSHIP`
- `TRANSFER_SPONSORSHIP`

The term transfer is what is used in the document so it sounds like the right vocab to use for the operation.

core/cap-0033.md
```diff
@@ -651,10 +717,71 @@ operations[5]:
...
+fee: <FEE>
+seqNum: <SEQ_NUM>
+timeBounds: <TIME_BOUNDS>
+memo:
+    type: MEMO_NONE
+operations[0]:
+    sourceAccount: S2
+    body:
+        type: SPONSOR_FUTURE_RESERVES
+        sponsoredID: S1
```
leighmcculloch, line 760: So when transferring sponsorship the sponsored account becomes the current sponsor? 🤯 This is very unintuitive to me.

core/cap-0033.md
jonjove: Clarified, let me know if you think it can use more work. I really want to be precise about this.

core/cap-0033.md
jonjove: Fixed, did something close to what you wrote. Thanks!

core/cap-0033.md
jonjove: Clarified.

core/cap-0033.md
jonjove: Right now this can only happen for `ClaimableBalanceEntry`. Basically, every type of entry other than claimable balance has a default sponsor (the account that controls that entry, which is the account that reflects that entry in the `numSubEntries`). But claimable balance has no default, it always has to have an explicit sponsor.

core/cap-0033.md
jonjove: By "these" do you mean the operations? I agree that would clarify intentions, but it would lead to more errors because fundamentally remove and transfer are the same operation. So it is my opinion that adding additional restrictions on which one to use would just lead to people using the wrong one and having their transactions fail.

core/cap-0033.md
```diff
@@ -569,6 +598,45 @@ Already there are discussions of ideas which may mitigate some of these issues
...
+signers, the sponsoring ID is stored separately. Why take different approaches?
+The reason is that `LedgerEntry` has an extension point but `Signer` does not.
+So if we wanted to store the sponsoring ID in the signer, then we would need to
+extend the `SignerKey` union. But this would force any downstream system that
+interacts with signers to update as well, introducing complexity throughout the
+ecosystem. Storing the sponsoring IDs for signers separately avoids this problem
+entirely, but it is perhaps slightly less elegant. Still, the trade-off is more
+than justified.
+
+### Why Doesn't UpdateSponsorshipOp Obey Typical Rules of Sponsorship?
```
abuiles, line 612: @jonjove which are the `Typical Rules of Sponsorship`?

rice2000:
Looks like @jonjove addressed all the open questions.  Merging!

PR #653, opened 2020-06-24
[SEP12] Add GET method for customers

msfeldstein:
resolves stellar/integration-meta#146

Many anchors need the ability to KYC customers ahead of time, before doing any transactions with them.  This is especially important for B2B where kyc is manual and takes a while.  SEP-12 starts to provide this, but is underpowered, particularly in that the client has no understanding of what the anchor requires for KYC, or what additional fields it might need, it only has the ability to send fields blindly.

This adds a `GET /customer` method in order to check the status of an existing users KYC, or to find out what fields are needed to KYC a new user, or complete an in progress KYC attempt.

Discussion here: https://docs.google.com/document/d/1LuTKU5E_cOgQYlWBUol1dwUVO5FrVpUE4XZSXRuoNqA/edit

### DIFF
--- a/ecosystem/sep-0012.md
+++ b/ecosystem/sep-0012.md
@@ -152,16 +152,7 @@ Property | Type | Description
 
 #### Errors
 
-If the client specifies an ID of a non-existent customer, return a 404 error response.
-For example:
-
-```json
-{
-   "error": "customer not found"
-}
-```
-
-For invalid requests, return a 400 error response.
+For invalid requests, return an error describing the issue.
 For example:
 
 ```json
--- a/ecosystem/sep-0012.md
+++ b/ecosystem/sep-0012.md
@@ -66,7 +66,7 @@ Name | Type | Description
 `account` | `G...` string | The Stellar account ID used to identify this user
 `memo` | string | (optional) Unique identifier for user being identified, necessary for custodial, multi-user accounts.
 `memo_type` | string | (optional) type of `memo`. One of `text`, `id` or `hash`
-`lang` | string | (optional) Defaults to `en`.  Language code specified using [ISO 639-1](https://en.wikipedia.org/wiki/ISO_639-1).  Human readable descriptions and messages should be in this language.
+`lang` | string | (optional) Defaults to `en`.  Language code specified using [ISO 639-1](https://en.wikipedia.org/wiki/ISO_639-1).  Human readable descriptions, choices, and messages should be in this language.
 
 ### Response
 
@@ -87,13 +87,26 @@ Name | Type | Description
 // The case when an anchor requires more info a bout a user
 {
    "status": "NEEDS_INFO",
-   "fields": [
-      {
-         "name": "email_address",
+   "fields": {
+      "email_address": {
          "description": "Email address of the user",
          "type": "string"
+         "optional": true
+      },
+      "id_type": {
+         "description": "Government issued ID",
+         "type": "string",
+         "choices": [
+            "Passport",
+            "Drivers License",
+            "State ID"
+         ]
+      },
+      "photo_id_front": {
+         "description": "A clear photo of the front of the government issued ID",
+         "type": "binary"
       }
-   ]
+   }
 }
 ```
 
@@ -116,13 +129,14 @@ REJECTED | This users KYC has failed and will never succeed.  The `message` must
 
 #### Fields
 
-The fields entry is required for the `NEEDS_INFO` status.  Fields should be specified as an array of objects containing the following properties:
+The fields entry is required for the `NEEDS_INFO` status.  Fields should be specified as an object with keys representing field names required, preferably from SEP-9.
 
 Property | Description
 ---------|------------
-`name` | Field name, ideally from SEP-9 if available
 `type` | `string | binary | number | date`
 `description` | A user readable description of this field, especially important if this is not a SEP-9 field.
+`choices` | An array of valid values for this field.
+`optional` | A boolean whether this field is required to proceed or not.
 
 
 ## Customer PUT
--- a/ecosystem/sep-0012.md
+++ b/ecosystem/sep-0012.md
@@ -53,7 +53,7 @@ Alternatively, if the client cannot add the authorization header. The JWT should
 
 Checks the status of a customers information. This allows clients to check whether the customers information was accepted, rejected, or still needs more info.  Lookups are done via the stellar account ID and memo that were used for the initial PUT request.
 
-This also gives you the information needed as far as what information a KYC provider requires to KYC a user.  If you fetch an account that has never been KYC'd the server should respond with `NEEDS_INFO` and a list of all the fields you will need to provide.
+This also provides the information needed from a KYC provider to KYC a user.  If a new account that has never been KYC'd is fetched,the server should respond with `NEEDS_INFO` and a list of all the fields the client will need to provide.
 
 ### Request
 
--- a/ecosystem/sep-0012.md
+++ b/ecosystem/sep-0012.md
@@ -139,10 +139,10 @@ The fields entry is required for the `NEEDS_INFO` status.  Fields should be spec
 
 Property | Type | Description
 -----|------|------------
-`type` | `string | binary | number | date` | The data type of the field value
-`description` | `string` | A user readable description of this field, especially important if this is not a SEP-9 field.
-`choices` | `array` | (optional) An array of valid values for this field.
-`optional` | `boolean` | (optional) A boolean whether this field is required to proceed or not. Defaults to false.
+`type` | enum | The data type of the field value. Can be `string`, `binary`, `number`, or `date`
+`description` | string | A user readable description of this field, especially important if this is not a SEP-9 field.
+`choices` | array | (optional) An array of valid values for this field.
+`optional` | boolean | (optional) A boolean whether this field is required to proceed or not. Defaults to false.
 
 
 ## Customer PUT
--- a/ecosystem/sep-0012.md
+++ b/ecosystem/sep-0012.md
@@ -152,7 +152,7 @@ Property | Type | Description
 
 #### Errors
 
-For invalid requests, return an error describing the issue.
+For invalid requests, return a 400 response describing the issue.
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0012.md
```diff
@@ -48,6 +49,82 @@ Alternatively, if the client cannot add the authorization header. The JWT should
 ?jwt=<token>
 ```
 
+## Customer GET
+
+Checks the status of a customers information. This allows clients to check whether the customers information was accepted, rejected, or still needs more info.  Lookups are done via the stellar account ID and memo that were used for the initial PUT request.
+
+This also gives you the information needed as far as what information a KYC provider requires to KYC a user.  If you fetch an account that has never been KYC'd the server should respond with `NEEDS_INFO` and a list of all the fields you will need to provide.
```
JakeUrban, line 56: I'd personally remove mentions of "you". 

`This endpoint also describes the KYC information the anchor requires of a user.`

ecosystem/sep-0012.md
```diff
@@ -48,6 +49,82 @@ Alternatively, if the client cannot add the authorization header. The JWT should
...
+`memo_type` | string | (optional) type of `memo`. One of `text`, `id` or `hash`
+`lang` | string | (optional) Defaults to `en`.  Language code specified using [ISO 639-1](https://en.wikipedia.org/wiki/ISO_639-1).  Human readable descriptions and reasons should be in this language.
+
+### Response
+
+Name | Type | Description
+-----|------|------------
+`status` | string | Status of the customers KYC process.
+`fields` | array<object> | An array of fields still needed to finish KYC Process
+`reason` | string | Human readable reason for rejection
```
JakeUrban, line 77: Maybe `description` or `message` instead of `reason`?

ecosystem/sep-0012.md
```diff
@@ -48,6 +49,82 @@ Alternatively, if the client cannot add the authorization header. The JWT should
...
+{
+   "status": "ACCEPTED",
+}
+```
+
+```js
+// The case when an anchor requires more info a bout a user
+{
+   "status": "NEEDS_INFO",
+   "fields": [
```
JakeUrban, line 90: So in SEP-31 `/info` the format for describing required fields is different. Most notably, the field name is a key. We don't _need_ to have the same format but I think its better if it does.

I like the `type` attribute here though, maybe we can add that to the SEP-31 `/info` endpoint. But I would definitely add the `optional` and `choices` fields to this schema.

ecosystem/sep-0012.md
```diff
@@ -48,6 +49,82 @@ Alternatively, if the client cannot add the authorization header. The JWT should
...
+}
+```
+
+#### Status
+
+Status | Description
+-------|------------
+ACCEPTED | KYC has been accepted and user can be identified by stellar account and optional memo in future transactions
+PROCESSING | KYC process is in flight and client can check again in the future to see if any further info is needed
+NEEDS_INFO | More info needs to be provided to finish KYC for this user.  The `fields` entry is required in this case.
```
JakeUrban, line 114: We should make a comment about how an originally `ACCEPTED` user can move to `NEEDS_INFO` and `PROCESSING` again if the anchor realizes more info is necessary.

ecosystem/sep-0012.md
msfeldstein: Good call, updated.

marcinx:
Love it!

marcinx:
I really like the idea of using a receiver id. Should it be called `customer_id` though to be more intuitive since it relates to `/customer` endpoints?

We've been discussing this same topic with a fiat anchor for a non-interactive SEP-6/SEP-12 integration by the way. They were just as confused as I originally was about the account and memo query parameters which link SEP-6 /withdrawal flows with SEP-12 /customer flows. 

Could this also make its way into SEP-6 if we go with anchors specifying a unique customer id? Instead of clients providing the customer identification account+memo as GET parameters to SEP-6 /withdraw, the new and unique receiver_id (customer_id) could be sent instead. It would probably be more intuitive to send a `customer_id` than an account/memo pair to identify the relevant customer for a withdrawal or deposit.

marcinx:
> 1. Sending anchor A makes an authenticated GET request to /customer
> 1. Receiving anchor B responds with the required fields for a PUT request

Can we introduce a `GET /customer-info` endpoint for SEP-12 instead? The only purpose of this endpoint would be to provide the required fields for PUT /customer. It would be equivalent to the `/info` endpoint we have for transfer servers. 

Gbubemi and I were planning to submit a PR to SEP-12 for that. Thoughts?

JakeUrban:
@marcinx Regarding your second comment, we could have an `/info`, /customer/info`, or simply `/customer` endpoint, but I don't like the `/customer-info` name.

Addressing your first comment, `receiver_id` wouldn't be used in SEP-12, only SEP-31. In SEP-12, all ID's would use the `id` field. And the goal of this change is to use it in SEP-6 and SEP-31.

marcinx:
@JakeUrban Yes, much better name with `/customer/info` :+1: All sounds great. Good progress.

msfeldstein:
> it doesn't make sense to use a stellar account and memo to identify a user, since SEP31 users don't have a stellar account.
They don't need a stellar account, they can be identified by the memo, and the account can always just be the sending anchors account, its basically how we do custodial ids

that said it is confusing and you're probably right, but its a breaking change to the protocol

hm

msfeldstein:
@marcinx 
> Can we introduce a GET /customer-info endpoint for SEP-12 instead? The only purpose of this endpoint would be to provide the required fields for PUT /customer. It would be equivalent to the /info endpoint we have for transfer servers.

Whats the difference between `GET /customer-info` and the `GET /customer` call i add in this PR which will tell you what fields are needed for any particular user?

JakeUrban:
@msfeldstein That would work but the downside is that the client specifies the memo ID for the user, not the server. If this user uses different clients, we'll have different ID's for the same user.

msfeldstein:
> If this user uses different clients, we'll have different ID's for the same user.

Is there any way around this?  Are you imagining that the server will understand that this is the same user and return the same customer id?

JakeUrban:
Actually, I think it is better if the client specifies an ID for the user. Receiving anchors shouldn't let different clients edit the same user's info, each client should have its own user data it manages with the receiving user.

JakeUrban:
I think the main issue we need to work out is how the `sender` info will be communicated. Will clients make a SEP-12 `PUT` request for the sending user too? How does the client specify to the receiving anchor that a user is a sender and not a receiver?

msfeldstein:
>  Receiving anchors shouldn't let different clients edit the same user's info, each client should have its own user data it manages with the receiving user.

Right the goal of sep12 is for users to not have to re-enter data between anchors, not to share data between wallets

msfeldstein:
> I think the main issue we need to work out is how the SEP-31 `sender` info will be communicated. Will clients make a SEP-12 `PUT` request for the sending user too? How does the client specify to the receiving anchor that a user is a sender and not a receiver?

why would it need to differentiate?

JakeUrban:
Right now SEP-12 has the assumption that the info needed for all users is the same, but thats not the case with SEP-31 participants.

Lets step back for a second. Why does the receiving anchor need info on the sending user?

msfeldstein:
This is all for pre-KYCing people so i guess its up to the SEP31 implementors to tell us why they want to pre-kyc and who they want to pre-kyc.  We might leave this new endpoint as a "how to fully KYC someone" so sending anchors can pre-kyc the receivers, but still use sep31 fields to just ask for the name of the senders

ecosystem/sep-0012.md
```diff
@@ -48,6 +49,102 @@ Alternatively, if the client cannot add the authorization header. The JWT should
...
+REJECTED | This users KYC has failed and will never succeed.  The `message` must be supplied in this case.
+
+#### Fields
+
+The fields entry is required for the `NEEDS_INFO` status.  Fields should be specified as an object with keys representing field names required, preferably from SEP-9.
+
+Property | Description
+---------|------------
+`type` | `string | binary | number | date`
```
JakeUrban, line 142: fix the formatting here

ecosystem/sep-0012.md
```diff
@@ -48,6 +49,102 @@ Alternatively, if the client cannot add the authorization header. The JWT should
...
+
+Different types of actions may have different kyc requirements.  The type parameter is used to signify what type of transaction this user needs to be KYC'd for.  For example, if a user is being KYC'd as a SEP31 sender, they may only require full name, but a SEP31 receiver needs to be fully KYCd with comprehensive data.  It can also be used to specify if this is a business being KYC'd or a person which can require different fields.  The value used for `type` is up to the implementor to signify the types of transactions their services support, and is optional.
+
+### Response
+
+Name | Type | Description
+-----|------|------------
+`status` | string | Status of the customers KYC process.
+`fields` | array<object> | An array of fields still needed to finish KYC Process for the given KYC type
+`message` | string | Human readable reason for rejection
```
JakeUrban, line 83: This should be `(optional)`

ecosystem/sep-0012.md
```diff
@@ -48,6 +49,102 @@ Alternatively, if the client cannot add the authorization header. The JWT should
...
+#### Type specification
+
+Different types of actions may have different kyc requirements.  The type parameter is used to signify what type of transaction this user needs to be KYC'd for.  For example, if a user is being KYC'd as a SEP31 sender, they may only require full name, but a SEP31 receiver needs to be fully KYCd with comprehensive data.  It can also be used to specify if this is a business being KYC'd or a person which can require different fields.  The value used for `type` is up to the implementor to signify the types of transactions their services support, and is optional.
+
+### Response
+
+Name | Type | Description
+-----|------|------------
+`status` | string | Status of the customers KYC process.
+`fields` | array<object> | An array of fields still needed to finish KYC Process for the given KYC type
```
JakeUrban, line 82: also this is not an array

JakeUrban:
@tomerweller The SEP-12 changes are ready to merge. Do you want to take a look before I do?

PR #655, opened 2020-06-29
SEP-0030: Add security section about external authentication methods

leighmcculloch:
### What
Add security section about external authentication methods and highlighting that a SEP-30 implementation chooses what external authentication methods to support and clients choose which to use when registering accounts.

### Why
The intention is that the external authentication methods are flexible and the methods use make sense for the use case, both functionally, and from the perspective of security and risk. The document already takes this position but this section clarifies that intent. 

cc @capybaraz @accordeiro

### DIFF
--- a/ecosystem/sep-0030.md
+++ b/ecosystem/sep-0030.md
@@ -41,6 +41,7 @@ The registration flow is as follows:
       - One or more other Stellar addresses.
       - And/or, one or more phone numbers.
       - And/or, one or more email addresses.
+      - And/or, one or more other external authentication methods.
 - The server verifies the [JWT] is valid according to [SEP-10].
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0030.md
```diff
@@ -679,6 +680,28 @@ A client can query the server for what signers the server has available for an
 account and should replace the signer on the account with the most recently
 generated signing key at its earliest convenience.
 
+### External Authentication
+
+Implementers choose the external authentication methods supported by their
+implementation.
+
+Implementers should consider the risks posed by the external authentication
+methods that are supported by their implementation. An implementation, and the
```
ire-and-curses, line 689: I think implementations can't make security evaluations
```suggestion
methods that are supported by their implementation. An implementer, and the
```

PR #656, opened 2020-06-29
added language parameter

antb321:


JakeUrban:
@antb321 I don't see a `description` attribute in the response, so `lang` is currently not useful. Did you mean to add a `description` field to the response as well?

PR #657, opened 2020-06-30
SEP31 - Remove require_receiver_info

msfeldstein:
`require_receiver_info` is a controversial feature of SEP31.  It is used when a sending anchor requires information about a receiving client, and assumes the receiving anchor can fetch that information from their banking rails or other source, and can legally share it with them.  

Given that no one has a hard requirement for it right now, and it seems unlikely too many anchors will have this capability or want to share this kind of user info, I propose we remove this until someone actually has a pressing need for it, then we can put it back.

### DIFF
--- a/ecosystem/sep-0031.md
+++ b/ecosystem/sep-0031.md
@@ -190,7 +190,6 @@ Content-Type: application/json
   "amount": 100,
   "asset_code": "USD",
   "asset_issuer": "GDRHDSTZ4PK6VI3WL224XBJFEB6CUXQESTQPXYIB3KGITRLL7XVE4NWV",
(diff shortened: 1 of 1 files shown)


msfeldstein:
@tempofr @JakeUrban @tomerweller

JakeUrban:
I support removing `require_receiver_info`.

It seems strange that the sending anchor has to pass a majority of the receiving user's info to the receiving anchor but could still request additional info on the receiving user from the receiving anchor. The receiving anchor will likely not have the info requested at the time of the sending anchor's request, especially if it is the first time the receiving user gets a payment.

msfeldstein:
Got some offline feedback that this is good to go

PR #660, opened 2020-07-01
SEP31: Added SEP-12 logic

JakeUrban:
resolves stellar/integration-meta#147

Added the ability to use SEP-12 for KYC'ing senders and receivers of SEP-31 transactions.

- Changes the format of the `GET /info` response to include `sender_sep12_type` and `receiver_sep12_type`
- Removed fields other than `memo` in the `sender` and `receiver` objects for `GET /info` and `POST /send`
- Changed the `customer_info_needed` error response to signal a required update to the SEP-12 `PUT /customer` endpoint
- Renamed the original `customer_info_needed` error response to `transaction_info_needed`
- Changed `pending_info_update` `Transaction` status to `pending_transaction_info_update`
- Added `pending_customer_info_update` `Transaction` status for updates to SEP-12 `PUT /customer`

### DIFF
--- a/ecosystem/sep-0031.md
+++ b/ecosystem/sep-0031.md
@@ -343,7 +343,7 @@ If the transaction cannot be found, the endpoint should return a `404 NOT FOUND`
 
 #### Pending customer info update
 
-In certain cases the receiver might need to request updated information.  For example, if the bank tells them that the provided receiver name is incorrect, or missing a middle initial.  Since this information was sent via SEP-12, the transaction should go into the `pending_customer_info_update` state until the sender makes another `PUT /customer` request.
+In certain cases the receiver might need to request updated information.  For example, if the bank tells the anchor that the provided receiver last name is incorrect, or missing a middle initial.  Since this information was sent via SEP-12, the transaction should go into the `pending_customer_info_update` state until the sender makes another `PUT /customer` request to update. The sending anchor can check which fields need to be updated by making a `GET /customer` request including the `id` or `account` & `memo` parameters. The receiving anchor should respond with a `NEEDS_INFO` status and `last_name` included in the fields described.
 
 #### Pending transaction info update
 
--- a/ecosystem/sep-0031.md
+++ b/ecosystem/sep-0031.md
@@ -126,13 +126,13 @@ The response should be a JSON object like:
          "receiver_sep12_type": "sep31-receiver",
          "fields":{
             "sender": {
-               "memo": {
-                  "description": "The memo used in the SEP-12 PUT /customer request for the sender",
+               "id": {
+                  "description": "The id returned in the SEP-12 PUT /customer response for the sender",
                }
             },
             "receiver": {
-               "memo": {
-                  "description": "The memo used in the SEP-12 PUT /customer request for the receiver",
+               "id": {
+                  "description": "The id returned in the SEP-12 PUT /customer response for the receiver",
                }
             },
             "transaction":{
@@ -168,7 +168,7 @@ The JSON object contains an entry for each asset that the anchor supports for re
 * `receiver_sep12_type`: The value of the `type` parameter the sending anchor should use for a `SEP-12 GET /customer` request.
 * `fields`: as explained below.
 
-The `fields` object allows an anchor to describe fields that must be passed into `POST /send`.  Fields are broken out by `sender`, `receiver`, and `transaction`.  `sender` and `receiver` contain the memo used for in the `SEP-12 PUT /customer` requests, while `transaction` contains transaction specific information.
+The `fields` object allows an anchor to describe fields that must be passed into `POST /send`.  Fields are broken out by `sender`, `receiver`, and `transaction`.  `sender` and `receiver` contain the id returned in the `SEP-12 PUT /customer` responses, while `transaction` contains transaction specific information.
 
 Each `fields` sub-object contains a key for each field name and an object with the following fields as the value:
 
@@ -190,10 +190,10 @@ Content-Type: application/json
   "asset_issuer": "GDRHDSTZ4PK6VI3WL224XBJFEB6CUXQESTQPXYIB3KGITRLL7XVE4NWV",
   "fields": {
     "sender": {
-       "memo": 123
+       "id": 123
     },
     "receiver": {
-       "memo": 456
+       "id": 456
     },
     "transaction": {
       "receiver_routing_number": "442928834",
--- a/ecosystem/sep-0031.md
+++ b/ecosystem/sep-0031.md
@@ -222,6 +222,7 @@ In the case where the sending anchor didn't provide all the KYC information requ
 Name | Type | Description
 -----|------|------------
 `error`| string | `customer_info_needed`
+`type` | string | (optional) A string for the `type` URL argument the sending anchor should use when making the `GET /customer` request
 
 ##### Transaction Info Needed (400 Bad Request)
 
--- a/ecosystem/sep-0031.md
+++ b/ecosystem/sep-0031.md
@@ -65,22 +65,25 @@ This protocol involves the transfer of value, and so HTTPS is required for all e
 ### User Onboarding
 
 1. User onboarding happens with the sending anchor, and is out of the scope of this spec.
-1. Sending anchor should use this to collect any KYC about the receiver it will need so it doesn't have to do it at every transaction.
+2. The sending anchor should, at minimum, collect the KYC information of the sending and receiving users specified by the receiving anchor's [SEP-12](sep-0012) implementation.
 
 ### Payment Flow
 
 1. Payment initiation happens in the sending anchors interface, and is also out of the scope of this spec.
 1. The sending client chooses a destination region.
 1. The sending anchor fetches the TOML file to find the DIRECT_PAYMENT_SERVER API root and [`/info`](#info) endpoint fields for the receiving anchor for this region.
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0031.md
```diff
@@ -65,21 +65,24 @@ This protocol involves the transfer of value, and so HTTPS is required for all e
...
+1. For every `*_sep12_type` value specified in the relevant asset's [`/info`](#info) object, the sending anchor makes a [`SEP-12 GET /customer`](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0012.md#customer-get) request.
+1. The sending anchor collects all information specified in the `GET /customer` response(s).
+1. The sending anchor PUTs all that information to the [`/customer`](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0012.md#customer-put) endpoint of the receiving anchor.
+1. The sending anchor POSTs the transaction information to the [/send](#send) endpoint of the receiving anchor.
+1. If information is missing, or the receiving anchor needs extra information to complete this transaction, `POST /send` will return `400 Customer Information Needed` or `400 Transaction Information Needed`.  The sending anchor should collect the necessary information and retry the entire `POST /send` call, until it succeeds.
```
msfeldstein, line 79: Its unclear what 'collect the necessary information and retry the entire post send call' means.  It makes sense for the transaction information, but for customer info do we need to do extra SEP12 PUTs, or do we pass it into the /send call?

I can imagine doing another SEP12 PUT with a new `type`, for example if we find out the amount is very large there might be a kyc_type `sep31-receiver-large-amount` and the sep12 implementation will know what info to collect for large amounts

msfeldstein: alternatively it might be fine to just put the fields needed directly here, and either ask you to send them to sep12, or to let you send them directly to /send

ecosystem/sep-0031.md
```diff
@@ -65,21 +65,24 @@ This protocol involves the transfer of value, and so HTTPS is required for all e
...
 
 ### Payment Flow
 
 1. Payment initiation happens in the sending anchors interface, and is also out of the scope of this spec.
 1. The sending client chooses a destination region.
 1. The sending anchor fetches the TOML file to find the DIRECT_PAYMENT_SERVER API root and [`/info`](#info) endpoint fields for the receiving anchor for this region.
-1. Using the fields fetched from /info, the sending anchor collects all needed information from the sending client.
-1. The sending anchor POSTs all that information to the [/send](#send) endpoint of the receiving anchor.
-1. If information is missing, or the receiving anchor needs extra information to complete this transaction, `POST /send` will return `400 Customer Information Needed` along with the missing fields. The sending anchor should collect these and retry the entire `POST /send` call, until it succeeds.
+1. For every `*_sep12_type` value specified in the relevant asset's [`/info`](#info) object, the sending anchor makes a [`SEP-12 GET /customer`](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0012.md#customer-get) request.
```
msfeldstein, line 75: we should add some small blurb about how the _type works, that its an arbitrary key that the anchor can use in order to inform the SEP12 endpoint what set of fields we need.  We can link to the sep12 section on it as well once that lands.

ecosystem/sep-0031.md
```diff
@@ -119,31 +122,24 @@ The response should be a JSON object like:
...
          "fields":{
-            "sender":{
-               "first_name": {
-                  "description": "The sender's first name"
-               },
-               "last_name": {
-                  "description": "The sender's last name"
+            "sender": {
+               "id": {
+                  "description": "The id returned in the SEP-12 PUT /customer response for the sender",
```
msfeldstein, line 130: I'm not sure i like having the sender and receiver objects here like this, it implies that you can have other fields.  It seems like we're strongly opinionated that all kyc fields go through sep12 so we may want to remove these here and just prescribe that you'll add sender_id and receiver_id to the /send call

ecosystem/sep-0031.md
```diff
@@ -192,17 +190,14 @@ Content-Type: application/json
   "asset_issuer": "GDRHDSTZ4PK6VI3WL224XBJFEB6CUXQESTQPXYIB3KGITRLL7XVE4NWV",
   "fields": {
     "sender": {
-      "first_name": "Alice",
-      "last_name": "Jones"
+       "id": 123
```
msfeldstein, line 193: maybe add a little note here like `// This is the customer ID returned from SEP12`

ecosystem/sep-0031.md
```diff
@@ -237,11 +232,19 @@ Name | Type | Description
 
 ##### Customer Info Needed (400 Bad Request)
 
-In the case where the sending anchor didn't provide all the information requested in `/info`, or if the transacton requires extra information, the request should fail with a 400 status code and the following body in JSON format.  The sender should then retry the entire request including all the previously sent fields plus the fields described in the response.
+In the case where the sending anchor didn't provide all the KYC information requested in `GET /customer`, or where the receiver requires additional KYC information after learning of the transaction `amount`, the request should fail with a 400 status code and the following body in JSON format. The sender should then retry both the `GET /customer` request to collect the additional fields and the `PUT /customer` request including all fields described in the `GET /customer` response.
```
msfeldstein, line 235: this handles the case where the sending client just didn't fill out the fields in the first place, but it doesn't handle the case where extra info above the defaults is needed, for example if the amount is large.  See my comment above about passing new types to sep12

ecosystem/sep-0031.md
```diff
@@ -351,15 +356,17 @@ Example response:
 
 If the transaction cannot be found, the endpoint should return a `404 NOT FOUND` result.
 
-#### Pending info update
+#### Pending customer info update
 
-In certain cases the receiver might need to request updated information, for example if the bank tells them that the provided receiver name is incorrect, or missing a middle initial.  In this case, the transaction should go into the `pending_info_update` state until the sender provides updates.
+In certain cases the receiver might need to request updated information.  For example, if the bank tells them that the provided receiver name is incorrect, or missing a middle initial.  Since this information was sent via SEP-12, the transaction should go into the `pending_customer_info_update` state until the sender makes another `PUT /customer` request.
```
msfeldstein, line 361: How does the client know which fields need to be updated?

ecosystem/sep-0031.md
JakeUrban: The instruction is intentionally vague because I didn't want to get into the differences between the two 400 statuses in this section. I may just take this instruction out since its for the failure case anyway and just describe it in detail below.

Regarding your second comment, the client-side developer in me likes only needing 1 request for KYC and transaction info, but I think including parameters described in SEP-12 in a SEP-31 request is confusing and disorganized. I don't think its too much overhead to make 2 requests.

ecosystem/sep-0031.md
JakeUrban: I agree, I'll make the change.

ecosystem/sep-0031.md
JakeUrban: I think it handles both. The client can pass just a regular `type=sep31-sender` parameter and then have a large `amount` value in the `POST /send` request. The anchor would return this response (400 Customer Info Needed) even though the client has already completed it for a regular sized transaction.

The client can again check the info required from `GET /customer`, which should be changed now that the anchor knows the customer is making a large transaction.

However I agree its way easier to have different `type` values for large transactions that require different KYC. The point is that if they don't, this flow can still work.

JakeUrban: Maybe we should allow a `choices` field in the `sender_sep12_type` `/info` response.

```
{
    "sender_sep12_type": {
        "description": "A description of this field",
        "choices": ["sep31-sender", "sep31-large-sender"]
    },
    "receiver_sep12_type": {
        "description": "A description of this field",
        "choices": ["sep31-receiver", "sep31-large-receiver"]
    },
    "fields": {...}
}

File: ecosystem/sep-0031.md (hunk 6 above)
Comment (JakeUrban) [reply ecosystem/sep-0031.md:361]: The `/transaction` response status should be `pending_customer_info_update`, so they'll know an update is needed. They'll make an additional request to `GET /customer` to see which fields are still needed.

**Comment by JakeUrban** [comment]
This isn't ready yet. We'll likely wait for the [SEP-12 changes](https://github.com/stellar/stellar-protocol/pull/653) to land before merging this, and there are still some adjustments we're considering.

File: ecosystem/sep-0031.md (hunk 6 above)
Comment (JakeUrban) [reply ecosystem/sep-0031.md]: Updated to include how the client determine which fields need updating.

File: ecosystem/sep-0031.md (hunk 2 above)
Comment (JakeUrban) [reply ecosystem/sep-0031.md]: done

File: ecosystem/sep-0031.md (hunk 5 above)
Comment (JakeUrban) [reply ecosystem/sep-0031.md:220]: Continuing on this idea ^, I actually think they should be included in the `fields` section in the `/info` response, since they're included there in the `POST /send` request.

```
{
    "fields": {
        "sender_sep12_type": {
            "description": "A description of this field",
            "choices": ["sep31-sender", "sep31-large-sender"]
        },
        "receiver_sep12_type": {
            "description": "A description of this field",
            "choices": ["sep31-receiver", "sep31-large-receiver"]
        },
        "transaction": {...}
    }
}
```

JakeUrban: Nevermind, the sender won't know which `type` to use.

PR #663, opened 2020-07-02
SEP-0031: Add to ecosystem README list of Active SEPs

leighmcculloch:
### What
Add SEP-31 to the ecosystem README's list of Active SEPs.

### Why
The SEP is missing from the list but exists in the repo.


JakeUrban:
@leighmcculloch We're considering some [significant changes](https://github.com/stellar/stellar-protocol/pull/660) to SEP-31 right now. I'm wondering if it might be more appropriate to put it in the draft state.

leighmcculloch:
The SEP is already marked as Active inside the SEP document. I'm just adding the SEP to the list, can you open a separate PR changing its status if the status is incorrect? I don't want to put the document in a list that is inconsistent with the document itself.

Something we can discuss on that other PR is if SEP-31 was in `Active` state and didn't turn out, we maybe should mark it as `Deprecated` and create a new SEP, unless that would cause some undesirable friction.

Issue #644, opened 2020-06-14
SEP-007 (license to curl) ambiguities in `replace` parameter


Synesso:
### What version are you using?

SEP-0007 v2.0.0

### Problem

No guidance is given to valid structure of _reference_identifier_ and _hint_ fields in the `replace` parameter of the SEP-0007 URL. This value is logically structured with invariants that are ambiguous.  It is perhaps easiest to define this problem via examples.

**reference_identifer mismatch**
`replace=tx.sourceAccount:X;Y:Source Account`
The reference identifier set in the _field to id_ half should match the reference identifier set in the _id to hint_ half exactly. In this example, `{X} != {Y}`

**Separators in hints**
`replace=tx.sourceAccount:Signing Account,operations[0].sourceAccount:Paying Account;Signing Account:If you please, the signing source account goes here,Paying Account:The account Id for the payer`

Commas are important delineators in the structure, but also useful characters in the hints (and perhaps also labels). There should be guidance on how to differentiate between separators and content. 

`replace=tx.sourceAccount:Signing Account,operations[0].sourceAccount:Paying Account;Signing Account:The signing source account,Paying Account:The account Id: paying account`

Similarly, colons can be employed in hints.

Semi-colons in hints might also cause confusion.

### What did you expect to see?

Ideally a defined structure that avoids ambiguity and redundancy; or more detailed parsing implementation guides in the SEP.

JSON might have worked well as the value here. If it's too late to tighten up that value, then perhaps the SEP can be modified to include parsing guidelines.

nikhilsaraf:
Hi! Thanks for raising these issues. I'll try and address them as best as I can and if we conclude that the SEP needs modification then that's on the table.

#### reference_identifer mismatch
>  replace=tx.sourceAccount:X;Y:Source Account
The reference identifier set in the field to id half should match the reference identifier set in the id to hint half exactly. In this example, {X} != {Y}

The example quoted above is invalid, for the reason your mentioned, i.e. {X} != {Y}

We should use the [example specified in the SEP](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0007.md#operation-tx), quoted for reference:
> sourceAccount:X,operations[0].sourceAccount:Y,operations[1].destination:Y;X:account from where you want to pay fees,Y:account that needs the trustline and which will receive the new tokens

[...]

PR #664, opened 2020-07-02
Update sep-0007.md

Synesso:
Describe `tx.replace` constraint around balanced reference identifiers. #644


nikhilsaraf:
Thanks for adding this. approved 👍

Issue #622, opened 2020-05-22
Protocol Request: change how the close time is used when applying transactions


MonsieurNicolas:
### What problem does your feature solve?

Right now when applying transactions, we use the close time decided during consensus, which is a different close time than what was used to validate transactions. This creates the possibility that a transaction gets accepted in a ledger only to fail later. This can break certain smart contracts if transactions are submitted at exactly the wrong time.

### What would you like to see?

close time should be set for the ledger *after* transactions get applied, in the same way the protocol version is updated after.

### What alternatives are there?

we could try to move the checks upstream, but this may be challenging as "combine candidates" (SCP) can modify the transaction set and close time independently.

MonsieurNicolas:
Expanding on this a bit, on why this is annoying.

For regular transactions submitted by SDKs that set expiration time to make it easier to know when they can safely re-submit (with a potentially new sequence number) a transaction. This is trying to guard people against clients running in parallel and issues caused by mem pool & network congestion/surge pricing. With this bug, they're charged a transaction fee for no good reason.

The larger issue is with smart contracts that today are written as chains of pre-signed transaction, some have "branches" with 2 mutually exclusive "paths" based on time.

For example, one of the transaction is only valid before a certain time (let's call it `A`), and the other is only valid after a certain time (`B`).

With this bug in the protocol, it's possible for a transaction to be accepted in the candidate tx set even though it will fail during consensus because a future timestamp is used when executing that transaction.

So basically `A` is valid, but then fails, yet consumes a sequence number, therefore unlocking transactions that are child of `A`.
If that transaction was anything other than a "no-op", this can have very bad effects on the smart contract (not just unlocking the branch, there is no way to "retry" that transaction).

Running into this in the context of smart contracts is not super likely as the timescale for expiration for those contracts are in the order of hours or days (or even years); but nevertheless, we should close the possibility of failure.

MonsieurNicolas:
Actually there might be a different way to fix this:
what we can do is change the combine candidates function to remove transactions from the transaction set (and dependent transactions) based on the new close time.

See:
https://github.com/stellar/stellar-core/blob/c097a3fe36f108bcbb52e4995b81bfe617f3109a/src/herder/HerderSCPDriver.cpp#L724

Right now this function is a no-op; only downside with this is that when we change the transaction set here, that means that we introduce a new value on the network - that said, as it's a deterministic function so nodes that are in sync should actually construct the new transaction set without having to rely on peers.

[...]

PR #670, opened 2020-07-08
Draft of CAP-0034 (change closetime semantics)

rokopt:
First draft of CAP-0034, intended to address
stellar-protocol issue #622, by changing the semantics of
the combination of candidates by the SCP ballot protocol
from producing an unsigned form of union of the candidates
(taking, for example, the maximum closetime of all
candidates as the closetime to externalize) to selecting
one particular signed, nominated StellarValue (comprising
transaction set, closetime, and upgrades) as the one to
externalize.

See also the [draft implementation](https://github.com/rokopt/stellar-core/tree/proto-622-closetime).


rokopt:
Removing PR to move code to a branch of my fork.

PR #677, opened 2020-07-09
SEP1: properly name home_domain

msfeldstein:
The SEP lists the accounts `home_domain` property as `homedomain`.  This is important to fix because its not googleable if someone wants to find where we talk about `home_domain`

### DIFF
--- a/ecosystem/sep-0001.md
+++ b/ecosystem/sep-0001.md
@@ -12,7 +12,7 @@ Version: 2.1.0
 
 ## Simple Summary
 
-The `stellar.toml` file is used to provide a common place where the Internet can find information about your organization’s Stellar integration. By setting the homedomain of your Stellar account to the domain that hosts your `stellar.toml`, you can create a definitive link between this information and that account. Any website can publish Stellar network information, and the `stellar.toml` is designed to be readable by both humans and machines.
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0001.md
```diff
@@ -12,7 +12,7 @@ Version: 2.1.0
 
 ## Simple Summary
 
-The `stellar.toml` file is used to provide a common place where the Internet can find information about your organization’s Stellar integration. By setting the homedomain of your Stellar account to the domain that hosts your `stellar.toml`, you can create a definitive link between this information and that account. Any website can publish Stellar network information, and the `stellar.toml` is designed to be readable by both humans and machines.
+The `stellar.toml` file is used to provide a common place where the Internet can find information about your organization’s Stellar integration. By setting the home_domain of your Stellar account to the domain that hosts your `stellar.toml`, you can create a definitive link between this information and that account. Any website can publish Stellar network information, and the `stellar.toml` is designed to be readable by both humans and machines.
```
leighmcculloch, line 15: I think this should be `home domain`. This sentence is talking about the concept of the home domain and not any specific code or API implementation.

I assume the `home_domain` you're referring to is that Horizon's API outputs the home domain in a field named `home_domain`, but this SEP shouldn't be bound to Horizon. It's `homeDomain` in the XDR, and has different formats depending on the SDK also.

PR #689, opened 2020-07-20
[SEP12] Update example URL for GET

msfeldstein:
It looks like we had a mismatched example and request parameter.


JakeUrban:
No longer needed because the change was made in another commit

PR #692, opened 2020-07-27
[SEP31] Update URL schema for transactions and replace update HTTP method with a PATCH.

abuiles:
This pull-request updates the URL schema for transactions to make it look more like a resource rather than use custom URLs for different interactions with the same resource. It also makes it more clear how to collect/update a given resource.

The following are the changes included in this PR and the rationale behind:

1. Pluralize resource URL to `/transactions`: this could look more like bikeshedding but since I was going to propose a change on the URL schema I went for plural, keeping the doors open to eventually allow a new end-point to query transactions in bulk.
2. Change `GET /transaction` to `GET /transactions/:id`: The current SEP is not clear on how the `ID` should be pass is it in the URL is it as a query param? This change makes it explicit that the resource ID should be pass as part of the URL
3. Remove `/update` in favor of `PATCH /transactions/:id`: Using `/update` to update a resource is counter-intuitive to API design, using the resource identifier help a `PATCH` instead of a `PUT` conveys the intention behind this end-point which is updating specific parts of the given transactions instead of changing the whole thing.

The changes here could be considered breaking, but we are still at a point where this is fine since nobody has implemented yet this in production. We know there are a couple of anchors who are almost ready with this, I reached out to them and they support this change. So let's get this which is a low-hanging fruit and will leave us with a better API.

### DIFF
--- a/ecosystem/sep-0031.md
+++ b/ecosystem/sep-0031.md
@@ -58,9 +58,9 @@ This protocol involves the transfer of value, and so HTTPS is required for all e
 
 ### Setting up rails
 
-1. To create a rail find a counterparty who implmements this SEP in the region you wish to provide access to, and agrees to do business with you.
+1. To create a rail find a counterparty who implements this SEP in the region you wish to provide access to, and agrees to do business with you.
 1. Trade public keys with each other in order to identify and securely interoperate with each other
-1. Keep a mapping of region to home_domain, using home_domain as your intial entry point to interoperating.
+1. Keep a mapping of region to home_domain, using home_domain as your initial entry point to interoperating.
 
 ### User Onboarding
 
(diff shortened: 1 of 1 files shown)


marcinx:
Good change. Would it make sense to change `POST /send` into `POST /transaction` as well? This would make the design of the SEP even more consistent from a REST point of view as the resource type would then be defined uniformly in the endpoint path and the action in the verb.

abuiles:
@marcinx 
> Good change. Would it make sense to change POST /send into POST /transaction as well?

I agree with this too - I thought about it but wanted to avoid adding more changes on this PR -- however, if this is something we all agree, I'll be happy to update the PR to include this change too.

abuiles:
@jacekn > Let's do that in another issue so we can discuss it separately.

Sounds good! I'll open a new PR in abit

abuiles:
@marcinx @jacekn follow up PR here https://github.com/stellar/stellar-protocol/pull/693

PR #693, opened 2020-07-27
[SEP31] Change URL end-point to create a new transactions.

abuiles:
This is a follow up on https://github.com/stellar/stellar-protocol/pull/692 -- the proposal is to create new transactions through a `POST` request on `/transactions`.

This PR also updates the response code to a 201, meaning a new transaction has been created and can be processed.


JakeUrban:
Can we make the associated change with `PUT /update` too? It should be `PATCH /transaction`

abuiles:
@JakeUrban that's on the other PR :)

JakeUrban:
You're right.. nvm

JakeUrban:
We should wait on merging this though until we get community feedback.

PR #696, opened 2020-07-29
Add "embed" field to SEP-0007 tx ops

nikhilsaraf:
Allows tracking the chain of SEP-0007 requests that triggered the creation of this request


nikhilsaraf:
fyi @andywer @pselden @pitsevich @dmgmyza @patrickxb 

This field will be used in an upcoming SEP to improve and standardize multisig coordination efforts.
If I have not reached out to you already I will reach out to you in the next few weeks with more information about this new effort.

andywer:
Looks good. Might make sense to clarify the basic anatomy of the use case a little more: Use this parameter to wrap an existing SEP-7 request in a new one with altered parameters like a different `callback` URL. You cannot mutate the source SEP-7 request as it would break its signature, so you create a new SEP-7 URI and embed the source URI.

ecosystem/sep-0007.md
```diff
@@ -36,6 +36,7 @@ The `tx` operation represents a request to sign a specific XDR `TransactionEnvel
...
 1. `pubkey` (optional) - Specify which public key you want the URI handler to sign for. Useful with the `callback` parameter above for example with multisig coordination.
+1. `chain` (optional) -  There can be an optional `chain` query param to include a single SEP-0007 request that spawned or triggered the creation of this SEP-0007 request. This will be a URL-encoded value. The goal of this field is to be informational only and can be used in the case of forwarding SEP-0007 requests. The embedded SEP-0007 request of course can itself contain another embedded SEP-0007 request in its own `chain` field, but caution should be exercised when doing so since this can easily explode the length of the final SEP-0007 request. To prevent attacks against this field, wallets can expect this field to have no more than 7 nested levels.
```
leighmcculloch, line 39: Could you add an example of how this would be used, and what purpose it would serve?

>The goal of this field is to be informational only

Informational in what way? Who would view the field, and how would they use it?

If the goal is to link requests, we could introduce the concept of a hash or an identifier. This will be easier to use to reconcile across in other systems like logs.

nikhilsaraf: will add a blurb on this.
using hashes would be limiting because some or all of the SEP-7 requests could be internal-only to the systems that spawned the given request.

in the future if we want to add support for hashes we can always do that since this format is forward compatible, for example:
`chain=web+stellar:` vs `chain=hash:`

nikhilsaraf:
@leighmcculloch @andywer updated to include your suggestions.

PR #701, opened 2020-08-01
SEP-0011: Add Flutter SDK implementations

christian-rogobete:
Please add the Flutter SDK to the list of implementations of SEP-0011


ecosystem/sep-0011.md
```diff
@@ -264,3 +264,7 @@ Txrep is implemented by the Stellar transaction compiler, [stc](https://github.c
 ### JavaScript
 
 Txrep is implemented as a JavaScript library for browser or NodeJS, [@stellarguard/txrep](https://github.com/stellarguard/txrep).
+
+### Flutter SDK
+
+Txrep is implemented by the Flutter SDK. See [example](https://github.com/Soneso/stellar_flutter_sdk/blob/master/documentation/sdk_examples/sep-0011-txrep.md).
```
leighmcculloch, line 270:
```suggestion
Txrep is implemented by the Flutter SDK. See [Soneso/stellar_flutter_sdk](https://github.com/Soneso/stellar_flutter_sdk/blob/master/documentation/sdk_examples/sep-0011-txrep.md).
```

leighmcculloch (approved):
I think we can add this, with one change below.

@stanford-scs You're the author of this SEP, can this implementation be added?

ecosystem/sep-0011.md
christian-rogobete: thank you @leighmcculloch

leighmcculloch (commented):
@stanford-scs On second thought I think it might be useful to designate a reference implementation separately to the other implementations. We do that in some other SEPs, like [SEP-5]. The separation makes it clear which implementations have strong ties to the definition of the SEP and possibly its author, and can be used as a reference. Thoughts?

[SEP-5]: https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0005.md#reference-implementation

PR #708, opened 2020-08-14
SEP-10: Make SIGNING_KEY required and change manage data op value

JakeUrban:
resolves stellar/integration-meta#180

Makes a SEP-10 service's `SIGNING_KEY` required in their stellar.toml file. Also changes the value of the manage data operation from a random 64-bit value to the service's home domain.

If the home domain specified in the manage data operation doesn't match the home domain request by the client, or the `SIGNING_KEY` specified on the TOML file did not sign the transaction, the challenge is invalid and the client should not sign it.

### DIFF
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -7,7 +7,7 @@ Author: Sergey Nebolsin <sergey@mobius.network>, Tom Quisel <tom.quisel@gmail.co
 Status: Active
 Created: 2018-07-31
 Updated: 2020-08-14
-Version 1.3.1
+Version 2.0.0
 ```
 
 ## Simple Summary
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -96,19 +96,21 @@ On success the endpoint must return `200 OK` HTTP status code and a JSON object 
   * source account set to server's signing account
   * invalid sequence number (set to 0) so the transaction cannot be run on the Stellar network
   * time bounds: `{min: now(), max: now() + 900 }` (we recommend expiration of 15 minutes to give user time to sign transaction)
-  * operations: `manage_data(source: client_account, key: '<anchor name> auth', value: '<home domain>')`
+  * operations: `manage_data(source: client_account, key: '<home domain> auth', value: random_nonce())`
+    * The value of key is the signing server's [fully qualified domain name](https://en.wikipedia.org/wiki/Fully_qualified_domain_name) followed by `auth`. It can be at most 48 characters (64 bytes after encoding).
+    * The value must be 64 bytes long. It contains a 48 byte cryptographic-quality random string encoded using base64 (for a total of 64 bytes after encoding).
   * signature by the service's stellar.toml `SIGNING_KEY`
 * `network_passphrase`: (optional but recommended) Stellar network passphrase used by the server. This allows the client to verify that it's using the correct passphrase when signing and is useful for identifying when a client or server have been configured incorrectly.
 
 Example:
 ```json
 {
-  "transaction": "AAAAAgAAAAC2c0axdmfUQ/Dq5fRRlvtybUBghi3vdOe3nHnMS6bxHgAAAGQAAAAAAAAAAAAAAAEAAAAAXza9AQAAAABfNhe8AAAAAAAAAAEAAAABAAAAAN1RWGc1gsEPHI/V0igO6YRsOYcwN8m/TuuGysVv4vz3AAAACgAAAAhTREYgYXV0aAAAAAEAAAAWdGVzdGFuY2hvci5zdGVsbGFyLm9yZwAAAAAAAAAAAAFLpvEeAAAAQHbHXU77KC1q+LIt/TffCV/wSlo/2rwuVgIfbV+Bb0yJ9Bd8qk744JxZov8T1iwd4XiN9zlKCHwN2brf1Ka1nA8=",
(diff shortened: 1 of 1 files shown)


JakeUrban:
@nebolsin

ecosystem/sep-0010.md
```diff
@@ -94,21 +96,19 @@ On success the endpoint must return `200 OK` HTTP status code and a JSON object
   * source account set to server's signing account
   * invalid sequence number (set to 0) so the transaction cannot be run on the Stellar network
   * time bounds: `{min: now(), max: now() + 900 }` (we recommend expiration of 15 minutes to give user time to sign transaction)
-  * operations: `manage_data(source: client_account, key: '<anchor name> auth', value: random_nonce())`
-    * The value of key is not important, but can be the name of the anchor followed by `auth`. It can be at most 64 bytes.
-    * The value must be 64 bytes long. It contains a 48 byte cryptographic-quality random string encoded using base64 (for a total of 64 bytes after encoding).
-  * signature by the web service signing account
+  * operations: `manage_data(source: client_account, key: '<anchor name> auth', value: '<home domain>')`
```
nebolsin, line 99: I don't think we should put home domain in place of the random nonce. First, the cryptographic-quality random nonce is important to keep challenge transactions unique for every request, without it challenge becomes deterministic and susceptible to different attacks. Strictly speaking, for proper mutual authentication there should be also a client nonce, which is originally sent to the server and is included in the transaction payload.

Another thing I want to point out, is that `key` of the manage data operation is defined in XDR as 64-byte string and is considered human-readable, while the value is 64-byte opaque binary value and mostly represented in  base64-encoded form. Placing home domain into the key will make it possible to visually verify the payload, for example before signing on a hardware wallet. I would suggest using key of `<home domain> auth` and keeping the value as a random nonce.

ecosystem/sep-0010.md
```diff
@@ -3,10 +3,10 @@
 ```
 SEP: 0010
 Title: Stellar Web Authentication
-Author: Sergey Nebolsin <sergey@mobius.network>, Tom Quisel <tom.quisel@gmail.com>, Leigh McCulloch <@leighmcculloch>
+Author: Sergey Nebolsin <sergey@mobius.network>, Tom Quisel <tom.quisel@gmail.com>, Leigh McCulloch <@leighmcculloch>, Jake Urban <jake@stellar.org>
 Status: Active
 Created: 2018-07-31
-Updated: 2020-04-03
+Updated: 2020-08-14
 Version 1.3.1
```
nebolsin, line 10: I think we need to bump the version, since this introduces a breaking change to the protocol.

ecosystem/sep-0010.md
overcat, line 99: Should the length limit for `home_domain` be stated here? I think this limit may be triggered.

ecosystem/sep-0010.md
JakeUrban: I'm unfamiliar with the versioning scheme. Would this constitute a patch, minor, or major version bump?

ecosystem/sep-0010.md
JakeUrban: It's possible but I would think unlikely. Here is a string that would reach the 64-byte limit:
```
>>> from base64 import b64encode
>>> len(b64encode(b"thisisatest.sandbox.anchor.anchordomain.com auth"))
64
```

I'll add a note about the limit though.

JakeUrban: I agree, adjusted.

ecosystem/sep-0010.md
accordeiro: Agreed we should bump this to 2.0, since clients who implement this will reject outdated server-side SEP-10 transactions.

JakeUrban:
@accordeiro no problem! Regarding your points:

1. Its actually _not_ a backwards compatible change. The `anchorName` parameter is being swapped for `homeDomain`, so the SDK's need to update their JWT generation helpers. SDK's, then server-side users of the SDK's, still need to update first.

2. I would image SEP-10 implementations slowly migrate to version 2.0 as we raise awareness, so I think clients will likely support both SEP-10 versions for a time. The other option is for anchors to communicate directly with known wallets and notify them of the SEP-10 update server-side, planning a coordinated effort to migrate at the same time.

Regarding your proposed `stellar.toml` variable, I think this can be detected by clients from the challenge. The updated SDK's will likely throw an exception if the `Manage Data` op contains an `anchorName` and not a `homeDomain`, which will signal to the clients which version is being used. However, I could be convinced it's worth it to have something on the `stellar.toml`.

accordeiro:
> 1. Its actually not a backwards compatible change. The anchorName parameter is being swapped for homeDomain, so the SDK's need to update their JWT generation helpers. SDK's, then server-side users of the SDK's, still need to update first.

I agree, backwards compatible might've been the wrong term to use. What I meant to say was that the value of the key in the `manage_data` op [was considered unimportant](https://github.com/stellar/stellar-protocol/pull/708/files#diff-72480d0894ae2d1a31c229b64ad4c37bL98), so from the client perspective whatever data was inside the op couldn't have been used as validation criteria in the past.

So even if the backends update their implementations (i.e. we make our producers output more strict), it shouldn't **break** any consumers as long as they don't update their client SDKs (i.e. as long as we don't make our consumers validation more strict). Does that make sense? So the overall strategy of updating servers before clients seems sound to me.

> 2. I would imagine SEP-10 implementations slowly migrate to version 2.0 as we raise awareness, so I think clients will likely support both SEP-10 versions for a time. The other option is for anchors to communicate directly with known wallets and notify them of the SEP-10 update server-side, planning a coordinated effort to migrate at the same time.
>
> Regarding your proposed stellar.toml variable, I think this can be detected by clients from the challenge. The updated SDK's will likely throw an exception if the Manage Data op contains an anchorName and not a homeDomain, which will signal to the clients which version is being used. However, I could be convinced it's worth it to have something on the stellar.toml.

Makes sense to me. I agree they can easily tell which version by looking at the `manage_data` op, so the `stellar.toml` change is redundant. We can defer that to whenever we make any changes that aren't easily detectable, if that happens in the future.

ecosystem/sep-0010.md
```diff
@@ -94,21 +96,21 @@ On success the endpoint must return `200 OK` HTTP status code and a JSON object
   * source account set to server's signing account
   * invalid sequence number (set to 0) so the transaction cannot be run on the Stellar network
   * time bounds: `{min: now(), max: now() + 900 }` (we recommend expiration of 15 minutes to give user time to sign transaction)
-  * operations: `manage_data(source: client_account, key: '<anchor name> auth', value: random_nonce())`
-    * The value of key is not important, but can be the name of the anchor followed by `auth`. It can be at most 64 bytes.
+  * operations: `manage_data(source: client_account, key: '<home domain> auth', value: random_nonce())`
+    * The value of key is the signing server's [fully qualified domain name](https://en.wikipedia.org/wiki/Fully_qualified_domain_name) followed by `auth`. It can be at most 48 characters (64 bytes after encoding).
```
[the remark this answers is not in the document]
JakeUrban: We can add an optional argument for the web services home domain, but it shouldn't be made mandatory. Signing servers that serve multiple other web services can use a default if not specified.

Making the changes regarding `<home domain>` encoding

nebolsin (approved):
Looks good to me :+1:

Issue #685, opened 2020-07-15
SEP-24: Wallet Attribution - verifying users' client applications


JakeUrban:
**Problem Statement**
There has been ongoing discussion among wallets and anchors about how to verify that a particular user, or rather a particular stellar account, uses a specific client wallet application when making SEP-24 deposits and withdrawals.

The benefits of this would be:
- improved BI analytics for anchors
- the ability to process users from specific wallet applications differently

For example, Anchor A could decide to not charge any fees for users from Wallet B.

**Solution**
One solution being discussed uses [SEP-2](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0002.md), specifically reverse federation lookup. The main idea is that when a user (through a wallet) makes a deposit request with an anchor, the anchor can call the wallet's federation server to verify that the stellar address used in the deposit request belongs to a user of the wallet application.

This solution would require two things from the wallet organization:
- Every stellar account the wallet acts on behalf of must have its `home_domain` option set to the home domain of the wallet's [stellar.toml](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0001.md)
- The wallet must maintain a SEP-2 federation server that allows authenticated clients to make requests using the `id` parameter

The flow would be as follows:
1. The user (using Wallet A) initiates a deposit with Anchor B
1. Anchor B requests the account information from Horizon
1. Anchor B uses the `home_domain` option on the account to find the `FEDERATION_SERVER` URL for the wallet
1. Anchor B makes a request to the server for the account requesting a deposit
1. Wallet A responds with a valid SEP-2 response containing information on the user

Anchor B could then not charge a fee for the user, or simply update its usage analytics.

**Discussion**
This issue is a place for interested parties to weigh in on this solution. If accepted by the ecosystem, SEP-24 (and potentially other relevant SEPs) will be updated to include a section on wallet's hosting reverse federation lookup servers.

leighmcculloch:
I think one issue is that this relies on the optional reverse lookup feature of a federation server, looking up a username from a Stellar address. Not all federation servers will implement this. Some will not want to support reverse lookup to prevent someone from discovering usernames for addresses.

We could address this concern without any changes to SEP-2, by saying that a federation server that wishes to prevent exposure of a username with reverse lookup should still return a valid response for reverse lookup, but instead of the real username, use the Stellar address as the username.

Example:
```
[...]

PR #709, opened 2020-08-17
SEP-34: Wallet Attribution

JakeUrban:
resolves stellar/integration-meta#158
resolves stellar/integration-meta#182

This protocol outlines how a client can prove to a web server that the application requesting the relevant operation is owned by the client. Prior discussion can be found on #685

This PR also adds wallet attribution sections to SEP-24 and SEP-6.


ecosystem/sep-0034.md
```diff
@@ -0,0 +1,84 @@
...
+## Abstract
+This following protocol involves 3 parties:
+
+- the client application: the party requesting an operation from the attribution server 
+- the client's signing server: the web server that claims ownership of the client application
+- the attribution server: the web server that verifies the signing server's claims
+
+At a high level, the signing server generates a [JSON Web Signature](https://www.rfc-editor.org/rfc/rfc7515.html) (JWS) that the client receives and includes in future requests to the attribution server. The attribution server verifies that the token included in the request was signed by the signing server, proving that the signing server takes ownership for the client requesting the operation from the verfication server.
+
+How the client application communicates with the signing server and receives the JWS is out of the scope of this document. The only requirement is for the signing server to return a JWS in the format descibed in the next section.
```
morleyzhi, line 30: Why doesn't the SEP define this? If each anchor has their own API for accepting this JWS, it's going to be hard for clients to implement this feature.

JakeUrban: The client communicates with its own server to obtain the JWS, not the anchor it is trying to verify its identity with. We don't need to define the protocol between the client application and the client's signing server as long as the JWS is in the correct format. 

We could describe what a flow between the client application and signing server could look like, but it is useless to define a standard and expect it to be used since there is no need for interoperability.

ecosystem/sep-0034.md
```diff
@@ -0,0 +1,84 @@
...
+## Motivation
+In general, web services may want to attribute incoming requests to client applications on a per-operation basis. Because requests headers like `User-Agent` can be spoofed, anchors require a cryptographically secure method of verifying a client's identity.
+
+For example, an anchor server may not charge fees for transactions created using a particular wallet. In this case, the operation is processing a deposit or withdrawal transaction, but the operation could be any action related to different use cases.
+
+## Abstract
+This following protocol involves 3 parties:
+
+- the client application: the party requesting an operation from the attribution server 
+- the client's signing server: the web server that claims ownership of the client application
```
tomerweller, line 25: I think that Morley's question is indicative that it's not very clear that `the client application` and `the client's signing server` are both operated by the same entity. 
Maybe `Wallet` and `Wallet Backend`?

ecosystem/sep-0034.md
```diff
@@ -0,0 +1,84 @@
...
+- the attribution server: the web server that verifies the signing server's claims
```
tomerweller, line 26: This is the Anchor Service, right?

ecosystem/sep-0034.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Token
+
+The JWS can be generated using two JSON objects and a secret if the HMAC SHA-256 algorithm is used.
+
+The first JSON object is the [JWS Protected Header](https://www.rfc-editor.org/rfc/rfc7515.html#section-4):
+```json
+{
+    "typ": "JWT",
+    "alg": "HS256",
```
tomerweller, line 40: let's stick to Ed25519 so that we don't add new crypto algorithms to our sdks.

ecosystem/sep-0034.md
```diff
@@ -0,0 +1,84 @@
...
+.
+eyJpc3MiOiAiaHR0cHM6Ly9hbmNob3Jkb21haW4uY29tL2F0dHJpYnV0aW9uIiwic3ViIjogImFhNzc5ODNhLWU1NTAtNGQ5MC04Y2MyLWQ2NjFkN2YwYjhmNiIsImV4cCI6ICIxNTk3Nzg5ODAxIiwiaWF0IjogIjE1OTc3MDMzNzUifQ
+.
+mEUU0apnMrlwiNER-Wac6F5xuXDWuZjAsmqy7Ap02C4
+```
+
+The signing key used to generate the third part of the JWS can also be used to verify it. 
+
+## Verifying Requests
+Other proposals will specify how the JWS returned by the client's signing server should be sent to the attribution server. Once sent, the attribution server should check the signing server's [stellar.toml](sep-0001.md) for a `SIGNING_KEY` variable and validate that the value found was used to encrypt the JWS.
```
tomerweller, line 75: We should link to concrete proposals once these are live

ecosystem/sep-0034.md
```diff
@@ -0,0 +1,84 @@
+## Preamble
+
+```
+SEP: To Be Assigned
+Title: Client Application Attribution
```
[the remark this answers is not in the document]
JakeUrban: I'll take up your suggestions to make the phrasing more concrete. We don't have another use case at the moment.

JakeUrban:
This SEP is still being drafted. Comments and reviews are welcome but keep in mind this is not the final product.

At this point in the draft's progress, and considering the comments that have already been made, the following changes should be made:
- Make the actors more concrete (wallet, wallet signing server, anchor server)
    - A note should be made that this can be generalized
- Define what "operation" (we really need a synonym) means in the context of SEP-34
    - SEP-10 attributes _accounts to requests_
    - SEP-34 attributes _resources/actions to clients_ (For SEP-24, the resource would be a transaction)
- Make it clear that attribution can only happen _after_ the resource has been created or action taken
    - Attribution verification requests can only be made _after_ a resource or action ID has been generated by the anchor
    - The decoded JWS token must contain a JWT payload that includes the resource or action ID

Does anyone have concerns about the above points?

cnusrini:
Did you guys decided the below : 
How is the wallet type decided ?
If fee is waved then how to use feeStats() or will it be totally deprecated ?

JakeUrban:
@cnusrini 
>How is the wallet type decided ?

I'm not sure what you mean by wallet type.

> If fee is waved then how to use feeStats() or will it be totally deprecated ?

Fee Stats can still be used because the wallet will still need to pay stellar network fees for withdrawal transactions, nothing in SEP-34 changes that. The fees SEP-34 could wave are the ones charged by the anchor itself. See SEP-24's [info](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0024.md#info) and [fee](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0024.md#fee) endpoints.

ecosystem/sep-0006.md
```diff
@@ -838,6 +839,36 @@ For example:
 }
 ```
 
+## Wallet Attribution
```
JakeUrban, line 842: I need to make a comment about when to make this requests

ecosystem/sep-0006.md
```diff
@@ -838,6 +839,36 @@ For example:
...
+Request parameters:
+
+Name | Type | Description
+-----|------|------------
+`id` | string | The id of the transaction.
+`token` | string | The JWS token returned by the wallet's [signing server](sep-0034.md)
+
+See [Verifying Requests](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0034.md#verifying-requests) for more information about validating the tokens passed.
+
+### Response
```
JakeUrban, line 861: If the request is valid but the wallet has has already sent on-chain funds to the anchor or the user has sent off-chain funds to the anchor, the anchor shouldn't process the request. Transaction that can be attributed to clients must be in the `incomplete` status, meaning the user or client has not been tasked with sending funds to the anchor.

Anchors should return a `422 Unprocessable Entity` status, meaning the anchor does not refund fees paid that were attributed to a wallet discount or sponsorship policy after the transaction funds were sent and fees paid.

JakeUrban:
I'm going to remove the SEP-24 and SEP-6 changes. Its premature to design the API contract for attribution.

ecosystem/sep-0034.md
```diff
@@ -0,0 +1,160 @@
...
+In general, web services may want to attribute resources created or actions taken on their service to client applications. Because requests headers like `User-Agent` can be spoofed, anchors require a cryptographically secure method of verifying a client's identity.
+
+For example, an anchor server may not charge fees for transactions created using a particular wallet. In this case, the resource is the deposit or withdrawal transaction, but the resource or action could be related to different use cases.
+
+## Abstract
+This following protocol involves 3 parties:
+
+- the wallet: the application requesting the transaction with the anchor server 
+- the wallet signing server: the web server that claims ownership of the wallet
+- the anchor server: the web server that verifies the signing server's claims
```
tomerweller, line 26:
```suggestion
- The Anchor Server: the web server that verifies the wallet's claims
```

ecosystem/sep-0034.md
```diff
@@ -0,0 +1,160 @@
...
+
+At a high level, the signing server generates a [JSON Web Signature](https://www.rfc-editor.org/rfc/rfc7515.html) (JWS) that the wallet application receives and includes in future requests to the anchor server. The anchor server verifies that the token included in the request was signed by the signing server, proving that the signing server takes ownership for the client requesting the transaction with the anchor server.
```
tomerweller, line 28:
```suggestion
At a high level, the wallet server generates a [JSON Web Signature](https://www.rfc-editor.org/rfc/rfc7515.html) (JWS) that the wallet application receives and includes in future requests to the anchor server. The anchor server verifies that the token included in the request was signed by the wallet server, proving that the wallet takes ownership of the request being made.
```

ecosystem/sep-0034.md
```diff
@@ -0,0 +1,160 @@
...
+
+How the client application communicates with the signing server and receives the JWS is out of the scope of this document. The only requirement is for the signing server to return a JWS in the format descibed in the next section. Wallet signing servers should not accept requests for tokens if the client cannot be authenticated or if the request omits the transaction ID the wallet received from the anchor.
```
tomerweller, line 30:
```suggestion
How the wallet communicates with it's server and receives the JWS is out of the scope of this document. The only requirement is for the wallet server to return a JWS in the format described in the next section. Wallet servers should not accept requests for tokens if the client cannot be authenticated or if the request omits the transaction ID the wallet received from the anchor.
```

ecosystem/sep-0034.md
```diff
@@ -0,0 +1,160 @@
...
+
+## Token
+
+The JWS can be generated using two JSON objects and a JSON Web Key ([JWK](https://www.rfc-editor.org/rfc/rfc7517.html)). The JWK itself is generated using the wallet signing server's `SIGNING_KEY` and associated secret key listed in the server's [`stellar.toml`](sep-0001.md).
```
tomerweller, line 34:
```suggestion
The JWS is generated using two JSON objects and a JSON Web Key ([JWK](https://www.rfc-editor.org/rfc/rfc7517.html)). The JWK itself is generated using the wallet signing server's `SIGNING_KEY` and associated secret key listed in the server's [`stellar.toml`](sep-0001.md).
```

ecosystem/sep-0034.md
```diff
@@ -0,0 +1,160 @@
...
+* `exp` - the [timestamp](https://tools.ietf.org/html/rfc7519#section-4.1.4) for the time on or after which the JWT must not be accepted for processing
+
+The two JSON objects should be base64-encoded and concatenated with a `.` character. 
+```
+eyJhbGciOiJFZERTQSJ9
+.
+eyJhdWQiOiJodHRwczovL2FuY2hvcnNlcnZlci5jb20iLCJleHAiOiIxNTk3Nzg5ODAxIiwiaWF0IjoiMTU5NzcwMzM3NSIsImlzcyI6Imh0dHBzOi8vd2FsbGV0c2lnbmluZ3NlcnZlci5jb20iLCJqdGkiOiJhYTc3OTgzYS1lNTUwLTRkOTAtOGNjMi1kNjYxZDdmMGI4ZjYiLCJzdWIiOiJHQ0NYS0NIVkVESDZNR1pZSUhER0lGRDVMSEdEM1JDQVo3TkZMSUFZNk5aQ0ZZQlZTNk1RUEZSNyJ9
+```
+
+The resulting string should then be signed by the server's JSON Web Key (JWK). JWK's can be constructed by using the Stellar public and private keys used by client.
```
tomerweller, line 80: I think the following code, both the js portion and the python are a bit confusing. 

I suggest not having any code here, just a step by step example process to getting from a stellar keypair to a signed request. Also, if we avoid using random keypairs we can remove statements of the "this *may* look something like" kind.

Once we have SDK support we can retrofit this SEP with proper code examples

ecosystem/sep-0034.md
```diff
@@ -0,0 +1,160 @@
...
+.
+eyJhdWQiOiJodHRwczovL2FuY2hvcnNlcnZlci5jb20iLCJleHAiOiIxNTk3Nzg5ODAxIiwiaWF0IjoiMTU5NzcwMzM3NSIsImlzcyI6Imh0dHBzOi8vd2FsbGV0c2lnbmluZ3NlcnZlci5jb20iLCJqdGkiOiJhYTc3OTgzYS1lNTUwLTRkOTAtOGNjMi1kNjYxZDdmMGI4ZjYiLCJzdWIiOiJHQ0NYS0NIVkVESDZNR1pZSUhER0lGRDVMSEdEM1JDQVo3TkZMSUFZNk5aQ0ZZQlZTNk1RUEZSNyJ9
+.
+xZi9TX_KI9sXA1XJlm_O5t6FLxs0sx9suBf5Kv4SVqL8Glo_zL4EbPZ6SgjeqBd31xnVQ-b5q4rwpiPIz278AQ
+```
+
+The public key used to generate the third part of the JWS can also be used to verify the integrity of the first and second part. This is what the anchor server would do when a wallet makes a request containing the JWS token. 
+
+## Verifying Requests
+Other proposals will specify how the JWS returned by the client's signing server should be sent to the attribution server. The anchor can keep the public keys of known wallet signing servers, but if the client is unknown, the anchor should request the signing server's [stellar.toml](sep-0001.md) and validate that the value found was used to encrypt the JWS.
```
tomerweller, line 151:
```suggestion
Other proposals will specify how the JWS returned by the Wallet Server should be sent to the attribution server. The anchor can keep the public keys of known wallet  servers, but if the client is unknown, the anchor should request the wallet server's [stellar.toml](sep-0001.md) and validate that the value found was used to encrypt the JWS.
```

ecosystem/sep-0034.md
```diff
@@ -0,0 +1,160 @@
...
+
+Wallet signing servers could also rotate their `SIGNING_KEY`. Anchors receiving JWS tokens from these clients may have to update their local copies of the wallets' public keys.
+
+Once verified, the anchor server can use the JWT attributes with confidence that the integrity of the data has been maintained and that the wallet requesting the transaction and passing the JWS belongs to the wallet's signing server.
+
+### Expiration
+Signing servers should select an expiration time for the JWT that is appropriate for the assumptions and risk of the interactions a user can perform with it. A client may own a transaction at the time the JWT is issued but they may not at some point in the future.
```
tomerweller, line 158:
```suggestion
Wallet servers should select an expiration time for the JWT that is appropriate for the assumptions and risk of the interactions a user can perform with it. A wallet may own a request at the time the JWT is issued but they may not at some point in the future. 
```

ecosystem/sep-0034.md
```diff
@@ -0,0 +1,160 @@
...
+
+Wallet signing servers could also rotate their `SIGNING_KEY`. Anchors receiving JWS tokens from these clients may have to update their local copies of the wallets' public keys.
+
+Once verified, the anchor server can use the JWT attributes with confidence that the integrity of the data has been maintained and that the wallet requesting the transaction and passing the JWS belongs to the wallet's signing server.
+
+### Expiration
+Signing servers should select an expiration time for the JWT that is appropriate for the assumptions and risk of the interactions a user can perform with it. A client may own a transaction at the time the JWT is issued but they may not at some point in the future. 
+
+Expiration times that are too long increase the risk that control of the transaction request will change. Expiration times that are too short increase the number of times authentication must reoccur, and a user using a hardware signing device or who must complete a complex signing process could have a poor user experience.
```
tomerweller, line 160: Why is the user signing device relevant here?

ecosystem/sep-0034.md
```diff
@@ -0,0 +1,160 @@
...
+Status: Draft
+Created: 2020-08-17
+Discussion: https://github.com/stellar/stellar-protocol/issues/685
+```
+
+## Simple Summary
+This protocol defines how anchors can cryptographically attribute transaction requests to wallet applications. 
+
+## Motivation
+In general, web services may want to attribute resources created or actions taken on their service to client applications. Because requests headers like `User-Agent` can be spoofed, anchors require a cryptographically secure method of verifying a client's identity.
```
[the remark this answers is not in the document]
JakeUrban: I think the majority of the document is specific to wallets and anchors, but I'd like to keep at least one mention of the fact that this _could_ be generalized. I don't feel too strongly on this though.

ecosystem/sep-0034.md
```diff
@@ -0,0 +1,160 @@
...
+
+For example, an anchor server may not charge fees for transactions created using a particular wallet. In this case, the resource is the deposit or withdrawal transaction, but the resource or action could be related to different use cases.
```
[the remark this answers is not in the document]
JakeUrban: If you're of the opinon that there should be no mention of other use cases thats fine, but I image there will be a future conversation about generalizing this SEP for a different identity verification use case.

ecosystem/sep-0034.md
```diff
@@ -0,0 +1,160 @@
...
+
+## Abstract
+This following protocol involves 3 parties:
+
+- the wallet: the application requesting the transaction with the anchor server
```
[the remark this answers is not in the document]
JakeUrban: This is not helpful, because this JWS token isn't used for all requests like the SEP-10 token. I vote to keep this the way it is.

ecosystem/sep-0034.md
JakeUrban: This makes it seem like the `request` is not an HTTP request, but a general request for a transaction. This should be disambiguated.

ecosystem/sep-0034.md
JakeUrban: Good point its not. This is copied from SEP-10

ecosystem/sep-0034.md
JakeUrban: How would you describe the process of decomposing a stellar keypair into the raw bytes and into the base64 representation? The only way I know how to communicate that process is via an SDK example.

I agree I'd like to not have code in the SEP, but some of these processes are really hard to explain without examples, like how to sign a JWS with a JWK. How do I give steps for this without getting into the RFC details?

ecosystem/sep-0034.md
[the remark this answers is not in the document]
JakeUrban: I've updated this section, let me know if you think it still needs work.

ecosystem/sep-0034.md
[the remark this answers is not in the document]
JakeUrban: I've updated the section to better explain the SEP's main use case but also leave room for generalization.

ecosystem/sep-0034.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Abstract
+This following protocol involves 3 parties:
+
+- The Wallet: the application that makes HTTPS requests to the anchor server for a resource or action
+- The Wallet Server: the web server that claims ownership of the wallet
+    - Note that frontend-only wallets _must_ add a backend web server to comply with this protocol
+- The Anchor Server: the web server that verifies the wallet's claims
+
+At a high level, the Wallet Server generates a [JSON Web Signature](https://www.rfc-editor.org/rfc/rfc7515.html) (JWS) that the Wallet receives and includes in future requests to the Anchor Server. The Anchor Server verifies that the JWS included in the request was signed by the Wallet Server and that the transaction (or resource) ID was included in the decoded [JWT](https://tools.ietf.org/html/rfc7519) payload. These steps prove that the Wallet Server takes ownership of the requests being made for a particular transaction.
```
lydiat, line 29: could JWT be defined as JWS is, i.e. JSON Web Token (JWT)

JakeUrban: Yes, good idea

lydiat:
- `the user's stellar account for which to deposit or withdraw funds`
should be 
`the user's Stellar account that funds will be deposited into/withdrawn from`

- `with it's Wallet Server` should be its

- Stellar should be consistently capitalized

- AP Style/Chicago Manual of Style differ in how high numbers should be written, but both agree they should be spelled out up to at least nine

- Consistency in bullet points: either all or none end in periods

- In other SEPs I noticed that mentions of SEPs were linked i.e. SEP-10 links to https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0010.md

- JWS payload bullets use both en and em dashes

- JWK should only be defined once

- keypair should be consistently lowercase

- `JWS Signature` should just be JWS

- Consistency in how base64/base 64/base-64 is written

- The phrasing of "Wallet Servers could also rotate their SIGNING_KEY” makes it sound like it's an alternative suggestion on how to verify requests - is it an alternative to the proposals yet to be written?

- `A Wallet application may own the request for a deposit or withdrawal transaction at the time the JWT is issued but it may not at some point in the future. ` 
 ->
 `A Wallet application that owns the request for a deposit or withdrawal transaction at the time the JWT is issued may not own it in the future` [because]? (don’t know if it’s necessary to say why, but the sentence is less passive this way at least)

- `Expiration times that are too short increase the number of times authentication must reoccur, and requiring repeated reauthentication could have a poor effect on user experience.`
 ->
 `Expiration times that are too short increase the number of reauthentication attempts, which creates a poor user experience.`

JakeUrban:
Thanks @lydiat ! Hopefully I've adjusted everything you mentioned. There is one thing that I had a hard time finding:

> AP Style/Chicago Manual of Style differ in how high numbers should be written, but both agree they should be spelled out up to at least nine

Where am I using 1-9?

JakeUrban:
Found one, let me know if I missed others

PR #713, opened 2020-08-26
SEP-24: design /transaction/attribution endpoint for SEP-34

JakeUrban:
resolves stellar/integration-meta#182
resolves stellar/stellar-protocol#685

The diagram below outlines the beginning of a SEP-24 transaction flow that uses SEP-34 for wallet attribution.

sep34

The main idea is that:
- the wallet receives the transaction `id` and interactive `url`
- the wallet authenticates with the wallet server
- the wallet receives a JWS token
- the wallet passes the token to the anchor's `/transaction/attribution` endpoint
- the anchor validates the token
- the user opens the interactive flow and is charged no fees

### DIFF
--- a/ecosystem/sep-0024.md
+++ b/ecosystem/sep-0024.md
@@ -704,7 +704,7 @@ Anchors can then attribute future requests related to that particular transactio
 
 ### Request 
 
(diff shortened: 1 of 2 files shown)


lydiat:
LGTM, at least from what I understand of the new protocol. 
If we're including this with the doc, it might be helpful to relabel `Anchor` as `Anchor Server`. 
Also, will the details of `transaction/attribution` endpoint/204 response be added to the SEP doc later?

JakeUrban:
We're not moving forward with this design if we can make an additive change to SEP-10 that verifies the wallet identity by signing the challenge transaction along with the wallet application.

PR #718, opened 2020-08-31
Add SEP-31's DIRECT_PAYMENT_SERVER attribute to the TOML attribute list

JakeUrban:
Is `DIRECT_PAYMENT_SERVER` the way we want to go? Its assumed in Polaris, but thats the only released software that uses it currently.


tomerweller (approved):
I can live with this

Issue #716, opened 2020-08-28
SEP31: add dynamic fee endpoint


tomerweller:
The SEP31 `/info` endpoint already supports fixed fees and percentage fees. However, often times the receiver might have a more complex fee structure in place.

This is already supported in SEP24 through the `/fees` endpoint.

I propose adding a similar dynamic fee endpoint to the SEP31 spec. In SEP31 we might want this to be an authenticated endpoint so that the receiver can give different quotes based on the identity of the sender anchor.

JakeUrban:
bump

leighmcculloch:
Could we make it optionally authenticated, so if possible fees can be exposed regardless like SEP-24, then user specific fees can be served when authenticated? Will it be possible for the fee endpoint to maintain the same format and structure as SEP-24?

JakeUrban:
@leighmcculloch 

> Could we make it optionally authenticated

Yes

> so if possible fees can be exposed regardless like SEP-24

Sure. Do we want to make authentication required per asset though? Otherwise I see a world where anchors return 400 HTTP codes from `/fee` with a message like `"not authenticated to request fees for asset"`.

> then user specific fees can be served when authenticated?

Interesting idea. This would work for SEP-31 because the sending Stellar account always exists at this point in the flow, but it wouldn't work for SEP-24 because anchors are expected to sponsor account creation for users.

JakeUrban:
> Will it be possible for the fee endpoint to maintain the same format and structure as SEP-24?

The fee endpoint already maintains the same structure.

JakeUrban:
I have a draft PR, can we discuss there?

JakeUrban:
Not stale, just related to a larger problem we haven't addressed

PR #719, opened 2020-09-01
SEP-31: Add fee endpoint and info attributes

JakeUrban:
resolves stellar/stellar-protocol#716

Adds a `fee_types` dictionary to the `GET /info` response as well as the SEP-24 `/fee` endpoint, minus the `operation` parameter.

### DIFF
--- a/ecosystem/sep-0031.md
+++ b/ecosystem/sep-0031.md
@@ -72,7 +72,10 @@ This protocol involves the transfer of value, and so HTTPS is required for all e
 1. The sending client chooses a destination region.
 1. The sending anchor fetches the TOML file to find the DIRECT_PAYMENT_SERVER API root and [`/info`](#info) endpoint fields for the receiving anchor for this region.
 1. For each `_sep12_type` value specified in the relevant asset's [`/info`](#info) object, the sending anchor makes a [`SEP-12 GET /customer`](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0012.md#customer-get) request.
-1. The sending anchor collects all information specified in the `GET /customer` response(s).
+1. The sending anchor collects the amount of the selected asset to send to the receiving client.
+1. If the [`/fee`](#fee) endpoint is enabled, the sending anchor uses the amount collected and appropriate key value from `/info`'s `fee_types` response object to request the fee the receiving anchor will charge the sending client.
+1. Optionally, the sending anchor verifies the `fee` returned with the sending client before proceeding.
+1. The sending anchor collects all information specified in the `GET /customer` response(s) and the `fields` returned from `/info`.
 1. The sending anchor PUTs all that information to the [`/customer`](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0012.md#customer-put) endpoint of the receiving anchor.
 1. The sending anchor POSTs the transaction information to the [/transactions](#transactions) endpoint of the receiving anchor.
 1. If some of the information POSTed was missing or incorrect, or the `amount` value is large enough to require additional information, the receiving anchor returns a `400` error response using either the `transaction_info_needed` or `customer_info_needed` status strings. More information about these responses can be found in the [/transactions](#transactions) section.
--- a/ecosystem/sep-0031.md
+++ b/ecosystem/sep-0031.md
@@ -31,7 +31,6 @@ NigeriaPay will utilize its European rail, enabled with EuroPay Anchor service, 
 * Anchors will create bi-lateral agreements to interoperate with each other.  This differs from other protocols in that there is no concept of 'discoverability'.  Anchors should keep a mapping of their partnerships in different regions to their home domains.
 * Each anchor registers a Stellar key with each counterparty anchor they interact with in order to identify themselves via [SEP-10 Web Authentication](sep-0010.md).
 
-
 ## Authentication
 
 Anchors should support [SEP-10](sep-0010.md) web authentication to ensure the counterparty they're interoperating with is actually who they say they are.  Clients must submit the JWT obtained via the SEP-10 authentication flow to all API endpoints except `/info`.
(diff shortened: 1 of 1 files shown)


tomerweller (commented):
I don't understand how the `fee_types` dictionary should be used. Can you add that to the payment flow?

JakeUrban:
Done @tomerweller

JakeUrban:
@leighmcculloch this is lower-priority, but since you took some interest in the issue I thought we could continue discussion here.

JakeUrban:
@yuriescl do you have thoughts here?

yuriescl:
It looks like `type` is always a required parameter (if not passed, it must have a default) for both calculating fees and creating transactions, so maybe it's better to have a required field `type` in `/transaction` instead of defining the types on `fee_types`.
This would also apply to SEP-6 and SEP-24, in which `type` is a custom field defined by the Anchor. But since it's listed in `/fees` parameters, it should not be a custom defined field, only the `type` choices should be customizable:

This idea also relates to this [issue](https://github.com/stellar/stellar-protocol/issues/743).
So `min_amount`, `max_amount`, `fee_fixed`, `fee_percent`, and complex fees all seem to depend on the payment type on SEP-6, SEP-24 and SEP-31.
And `sender_sep12_type` and `receiver_sep12_type` might depend on the payment type on SEP-31 (different documents required for different payment types)
So maybe there could be a structure like this for defining those values:

```
GET /info response
{
  ...
  types: {
    SEPA: {
        description: "SEPA transfer, this is the default type",
        min_amount: 15.0,
        max_amount: 10000.0,
        fee_fixed: 1.0,
        fee_percent: 1.0,
        sender_sep12_type: "sep31-sepa-sender",
        receiver_sep12_type: "sep31-sepa-receiver",
    },
    ...
  },
```

JakeUrban:
We would have to go about this in an additive way but I like the idea of defining these attributes per rail type.

ecosystem/sep-0031.md
```diff
@@ -137,8 +140,19 @@ The response should be a JSON object like:
                   ]
                }
             }
+         },
+         "fee_types": {
+            "cash_pickup": {
```
leighmcculloch, line 145: ❓ Are these fee types one-to-one with the values in the `choices` list? Or are they different?

ecosystem/sep-0031.md
```diff
@@ -249,6 +268,51 @@ In the case where the transaction just cannot be completed, return an error resp
...
+GET DIRECT_PAYMENT_SERVER/fee
+```
+
+Request parameters:
+
+Name | Type | Description
+-----|------|------------
+`type` | string | (optional) Type of payment method, optionally described in the `GET /info` endpoint
+`asset_code` | string | Asset code of the payment
+`amount` | float | Amount of the asset that would be sent to the receiving user
```
leighmcculloch, line 285: ❗ I think we should make the `amount` request parameter a string, and note in the description it should be a decimal. That would be consistent with the other amount fields that already exist in SEP-31, for e.g.:
https://github.com/stellar/stellar-protocol/blob/e195a267aa3f2c27a78c12e85676ed33a806debb/ecosystem/sep-0031.md#L303-L305

Why: The consistency will make it less taxing on the developer consuming the API if they can assume amounts are represented and consumed in the same way everywhere. A 'float' comes with some baggage. Floats can be represented with an exponent sometimes by some programming languages when serializing and I don't think we'd want to complicate parsing. Also some languages have types that are much better at naturally handling money than floats.

```suggestion
`amount` | string | Amount of the asset that would be sent to the receiving user as a string with up to 7 decimals.
```

Would 7 decimals suffice?

ecosystem/sep-0031.md
```diff
@@ -249,6 +268,51 @@ In the case where the transaction just cannot be completed, return an error resp
...
+```
+
+Every HTTP status code other than `200 OK` will be considered an error. The body should contain error details.
+For example:
+
+```json
+{
+   "error": "This anchor doesn't support the given currency code: USD"
+}
+```
```
leighmcculloch, line 314: ❓ I don't see mention of this here so I'm not sure if it is the case, but is this endpoint authenticated or not? There could be times where it being authenticated would allow different fee structures for different users based on their existing volumes, or agreements, or the wallet they're using. This might be out of scope and I'm not asking for it, just thinking out loud.

ecosystem/sep-0031.md
```diff
@@ -249,6 +268,51 @@ In the case where the transaction just cannot be completed, return an error resp
...
+
+Name | Type | Description
+-----|------|------------
+`fee` | float | The total fee (in units of the asset sent) that would be charged on top of the specified `amount` of `asset_code` sent to the user.
+
+Example response:
+
+```json
+{
+  "fee": 0.013
```
leighmcculloch, line 303: ❗ Similar to the above and for the same reasons I think we should make this a string, e.g.:

```suggestion
  "fee": "0.013"
```

leighmcculloch (commented):
Some questions (❓ ), ideas/suggestions (💡 ), and some concerns (❗ ) noted below and inline.

---

Also some general questions. I don't have a particular issue with any of these things I'm curious the context behind the API decisions:

- It's interesting to me that some of the details about fee types are in the `/info` endpoint, i.e. the description, and some is in the `/fee` endpoint, i.e. the calculated fee. What's the rationale behind the distinction of putting some information in one spot and some in another? 

- What's the rationale behind having the `/fee` endpoint returning a single type of fee rather than all the fees for a given transaction? Are clients likely to only display one fee type option to a user after they choose their type of withdraw/deposit, or is a client likely to call the endpoint repeatedly for every fee type then display info to the user?

- The `/fee` endpoint doesn't distinguish between deposit and withdrawal. Is it used for both or only withdrawal? Do we need to distinguish?

- Are there other elements of a planned deposit/withdrawal beyond the amount that could impact the fee? For example, could a service provider wish to charge different fees for a deposit/withdrawal to different banks within the SEPA network?

---

Sorry, lots of questions! Let me know if I'm just missing context that I can get myself to answer them.

JakeUrban:
This PR has not been merged because we need to support different fees for different rails/countries. There are a handful of issues associated with SEP-6 & 24 as well related to this problem. I'll be taking this problem on in Q1 or Q2.

PR #722, opened 2020-09-02
SEP-12: added 404 response to SEP-12 GET /customer

JakeUrban:
This PR adds a 404 response to the `GET /customer` endpoint which should be used when the `id` passed is for a nonexistent customer or for a customer created by another anchor, determined by the `sub` JWT key.

This response code will allow clients to determine if a customer exists as opposed to getting 400 back, which could mean the request was bad.

### DIFF
--- a/ecosystem/sep-0012.md
+++ b/ecosystem/sep-0012.md
@@ -169,7 +169,14 @@ Property | Type | Description
 
 #### Errors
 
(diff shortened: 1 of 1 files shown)


JakeUrban:
@yuriescl

tomerweller (approved):
LGTM

PR #726, opened 2020-09-17
SEP-0010: Add Memo Support

### LINKED ISSUE #727: SEP-10: Support pooled accounts

JakeUrban:
This PR has been closed because `JakeUrban:sep10-pooled-accounts` will be merged into `stellar:sep10-pooled-accounts` instead of `stellar:master`. Changes to other SEPs will be made to support the memo support added to SEP-10 in this branch, and ultimately merged all together to `stellar:master`.

---

resolves #727 

Add support for clients that share a Stellar account and use memos to distinguish users or accounts for payments.

Changes:
- Adds optional `memo` and `memo_type` arguments to the `GET <WEB_AUTH_ENDPOINT>` request
- Adds optional memos on challenge transactions derived from the passed `memo` and `memo_type`
    - This allows the anchor to verify the user identified by the memo and source account in the `POST <WEB_AUTH_ENDPOINT>`
- Modifies the contents of the token returned by allowing anchors to add the `memo` and `memo_type` string values to the decoded JWT object


JakeUrban:
CC @tomerweller does the direction of these changes seem in-line with whats required for pooled-account support?

JakeUrban:
CC @ire-and-curses

JakeUrban:
[
Screen Shot 2020-09-21 at 1 44 14 PM
](url)

tomerweller:
Isn't the above diagram missing the bit in which the wallet asks the custodian to sign the challenge transaction. 

I don't find this too offensive. But would love to hear @leighmcculloch's thoughts.

JakeUrban:
This is still a draft.. by no means would I use this diagram anywhere. We're still at the point where we need to decide if we're implementing this. But yes, the Wallet would send the challenge to the Wallet Backend to sign before returning the signed challenge to the anchor.

JakeUrban:
It is, I can update it.

ecosystem/sep-0010.md
```diff
@@ -121,6 +126,16 @@ Every other HTTP status code will be considered an error. For example:
 }
 ```
 
+### Pooled Accounts
+
+Before defining pooled accounts, we must first define custodial accounts. In the context of Stellar, custodial accounts are Stellar accounts that are held and managed by an entity that is _not_ the owner of the funds within the account. Specifically, custodial entities hold the secret key(s) of the accounts they control. These custodial entities can create individual accounts for each user, but a much more common practice is to hold the funds of many users in a single _pooled_ account.
```
leighmcculloch, line 131: I don't think we can use the term custodial here. I believe it is a legal term, which means it could be interpreted differently in different regions of the world, and we maybe don't want to get into that in a technical specification. Can we define what we need to define here without using the term custodial?

ecosystem/sep-0010.md
```diff
@@ -149,13 +164,17 @@ To validate the challenge transaction the following steps are performed by the s
...
 
 Upon successful verification, service responds with a session JWT, containing the following claims:
 
 * `iss` (the principal that issued a token, [RFC7519, Section 4.1.1](https://tools.ietf.org/html/rfc7519#section-4.1.1)) — a [Uniform Resource Identifier (URI)] for the issuer (`https://example.com` or `https://example.com/G...`)
-* `sub` (the principal that is the subject of the JWT, [RFC7519, Section 4.1.2](https://tools.ietf.org/html/rfc7519#section-4.1.2)) — the public key of the authenticating Stellar account (`G...`)
+* `sub` (the principal that is the subject of the JWT, [RFC7519, Section 4.1.2](https://tools.ietf.org/html/rfc7519#section-4.1.2)) — the public key of the authenticating Stellar account potentially contcatenated with the `memo` value originally specified by the client using an `-` character
+  * For example, when `account` is pooled, the user can be identified using the 32-byte hash memo: `GAMJ4W6QDEKJWZRMAWGD7ETKPIMKUX4NCZ3JBEYPILZSOWTW7NL6TDQ4-yQ2eD5xHRSWUFJfCZqggCS5peHFkb9SXJ3hBjjZjHnc=`
+  * Web services that are passed a JWT token can use the `sub` memo value to internally identify the user who owns the funds held by the client on each request
```
leighmcculloch, line 177: We need to define how each memo type is formatted explicitly.

leighmcculloch, line 177: Unfortunately the `sub` field must be a string limiting us to the delimiter approach, but we do not need to use `sub`. This could be a good time to make this a breaking change where by we abandon the use of the `sub` field and define new fields like `account` and `memo`. Something like:

Standard account:
```
"account": "G..."
```

Pooled or multiplexed account:
```
"account": "M..."
```
or
```
"account": {
    "account": "G...",
    "memo": "1234"
}
```

leighmcculloch (commented):
Are we able to hold off on this until multiplexed accounts are in use? They will make pooled accounts smoother I think if we do not need this immediately.

JakeUrban:
> Are we able to hold off on this until multiplexed accounts are in use? They will make pooled accounts smoother I think if we do not need this immediately.

This is the much larger question. After talking with Tomer and Eric we're exploring this option because the Stellar ecosystem is familiar with this pattern (memo-based User-IDs) and there is _a lot_ of work to do to actually implement M-addresses, which largely affects Eric and his team.

The idea is that this would be replaced with a strategy that does use M-addresses once they're implemented.

ecosystem/sep-0010.md
JakeUrban: I'm usually not one to back down from a breaking change but SEP-10 is one of the most widely used standards in the Stellar ecosystem, and given the current state of SEP-10 regarding the 2.0 SDK updates, I think we should try not to break things until we define a better transition process.

leighmcculloch: Yeah that's a good point.

I worry that delimiting the data in `sub` will come back to bite us in ways we don't expect, but I don't have any concrete examples.

I like that with this proposal the SDKs we know of should break if they see the hyphen and haven't been upgraded to support a memo. So they shouldn't accidentally validate a JWT as proving control of an account when it only proves alignment with a memo value.

Some other options:
- We could also define the `sub` to be encoded in some other format, such as JSON, so it is JSON stringified. This would make it predictable using common technology to build and parse. The value of sub would become:
   ```
   "sub": "{\"acc\":\"G...\",\"memo\":\"1234\"}"
   ```
   This makes it clear for any developer looking at the data at what they have because it is self-describing.
- We could use the strkey multiplexed format in SEP-10 without supporting them generally in other ways. This would allow us to adopt SEP-10 in the direction of that standard without waiting for any other part of it being implemented. We'd only use it to encode the data we have. This would also limit the memo types to the 64-bit ID.

JakeUrban: I'm not familiar with the strkey multiplexed format I'll take a look at it and come back with comments/questions.

Regarding the first 'other' option, it would certainly work, but I think it may confuse clients. If they've already deserialized the JSON string from the response body, it seems strange to deserialize a specific attribute within the JWT.

Personally I think the current design is the way to go. The one concern I have is that the current design doesn't provide the `memo_type` value in the `sub` field of the JWT. I think we need a better format for encoding this info.

The SDK's would be updated to validate this format if present, otherwise the original Stellar public key value would be expected.

leighmcculloch: We can avoid needing to include the memo type if we limit the memo types allowed to the memo types allowed by multiplexed addresses / muxed accounts, which is the memo id type, a 64-bit integer. If we're thinking ahead to when multiplexed addresses are in use we'll be limited to those memo types anyway. My understanding is the research @stanford-scs did highlighted that most folks only needed that memo type. @tomerweller Thoughts?

JakeUrban: @tomerweller bump, starting to think about this again

tomerweller: Yes. We discussed this during last year's meridian protocol session and there was a wide agreement that int64s are enough. Which will make transitioning to muxed accounts easier.

ecosystem/sep-0010.md
JakeUrban: Is there a better term you can think of?

ecosystem/sep-0010.md
JakeUrban: This is resolved now that memos are only ID's

ecosystem/sep-0010.md
JakeUrban: Because not all custodial accounts are pooled accounts, so we need a synonym or we need to define a term here in the SEP.

JakeUrban:
This PR has been updated to remove `memo_type` and require all `memo`'s to be unsigned 64-bit integers. Other than the custodial branding/wording, are there any technical changes that need to happen?

ecosystem/sep-0010.md
```diff
@@ -149,13 +163,17 @@ To validate the challenge transaction the following steps are performed by the s
...
 
 Upon successful verification, service responds with a session JWT, containing the following claims:
 
 * `iss` (the principal that issued a token, [RFC7519, Section 4.1.1](https://tools.ietf.org/html/rfc7519#section-4.1.1)) — a [Uniform Resource Identifier (URI)] for the issuer (`https://example.com` or `https://example.com/G...`)
-* `sub` (the principal that is the subject of the JWT, [RFC7519, Section 4.1.2](https://tools.ietf.org/html/rfc7519#section-4.1.2)) — the public key of the authenticating Stellar account (`G...`)
+* `sub` (the principal that is the subject of the JWT, [RFC7519, Section 4.1.2](https://tools.ietf.org/html/rfc7519#section-4.1.2)) — the public key of the authenticating Stellar account potentially contcatenated with the `memo` value originally specified by the client using an `-` character
+  * For example, when `account` is pooled, the user can be identified using the 32-byte hash memo: `GAMJ4W6QDEKJWZRMAWGD7ETKPIMKUX4NCZ3JBEYPILZSOWTW7NL6TDQ4-yQ2eD5xHRSWUFJfCZqggCS5peHFkb9SXJ3hBjjZjHnc=`
```
leighmcculloch, line 175: I agree we don't know yet if there will be more effort immediately put into rolling out multiplexed accounts in all areas of the SDKs, Horizon, etc, however I think if we are willing to limit pooled account memo support here to int64s then we may as well leverage [SEP-23](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0023.md) and use the `M` strkey format for a multiplexed address as the format that we encode the account and the memo ID.

If multiplexed addresses get their moment then we'll already be using the appropriate address format here.

In the event multiplexed addresses never take off in other ways using the `M` strkey encoding here won't be any worse than us creating our own format.

By using the `M` strkey format we avoid adding a new format here that is unique to SEP-10.

@JakeUrban @tomerweller @ire-and-curses Thoughts?

leighmcculloch, line 175: I think we need two examples here, one without a pooled account and one with, so that it is clearer memo is optional.

ecosystem/sep-0010.md
```diff
@@ -14,6 +14,8 @@ Version 2.0.0
 
 This SEP defines the standard way for clients such as wallets or exchanges to create authenticated web sessions on behalf of a user who holds a Stellar account. A wallet may want to authenticate with any web service which requires a Stellar account ownership verification, for example, to upload KYC information to an anchor in an authenticated way as described in [SEP-12](sep-0012.md).
 
+This SEP also supports clients who control [pooled accounts](#pooled-accounts) containing funds owned by _many_ users. These clients are _custodians_ of their users' Stellar accounts and possess the secret keys for those accounts.
```
leighmcculloch, line 17: I think we should avoid the use custodian. That term has specific legal meaning which is outside the scope of this technical spec. I also think we're being unnecessarily prescriptive to discuss secret keys of accounts.
```suggestion
This SEP also supports clients who control [pooled accounts](#pooled-accounts) containing funds owned by _many_ users.
```

ecosystem/sep-0010.md
```diff
@@ -82,6 +84,7 @@ Name      | Type          | Description
 ----------|---------------|------------
 `account` | `G...` string | The stellar account that the wallet wishes to authenticate with the server
 `home_domain` | string | (optional) The [fully qualified domain name](https://en.wikipedia.org/wiki/Fully_qualified_domain_name) of the service requiring authentication. SEP-10 signing servers used to generate tokens for multiple different web services can use this parameter to create valid tokens for each service. If not provided by the client, a default should be used.
+`memo` | integer | (optional) The 64-bit unsigned integer used to identify the user in the case of a pooled `account`. Note that the `memo_type` field is not needed because all values are of type `MEMO_ID`.
```
leighmcculloch, line 87: See note further down about leveraging the `M` address format to encode the memo and account together. I think we could leverage that format for just SEP-10.

ecosystem/sep-0010.md
leighmcculloch: I think we can start with "Pooled accounts are..." and define what they are without referencing custodial accounts. What drives the need to discuss custodial?

leighmcculloch (commented):
A little controversial, but even if multiplexed addresses aren't a thing for a long time, I think we should leverage their `M` strkey definition and use them within SEP-10 only to encode the memo ID along with the account ID.

ecosystem/sep-0010.md
JakeUrban: I'm concerned the presence of `M` address in SEP-10 will imply wider ecosystem support. Won't developers be confused to see a new type of Stellar account without being able to query Horizon for its `/account` JSON? I don't want to conflate the _need_ to encode a memo with the _desire_ to use `M` addresses.

JakeUrban:
> What drives the need to discuss custodial?

Custodians are brought up because we're always talking directly with them. If they have this type of approach then this issue is the first limiting factor. I think we should do our due diligence on the `custodial/custodian` term before ruling it out, because I think it actually makes what we're trying to accomplish a lot clearer.

ecosystem/sep-0010.md
leighmcculloch: I think the fact the M accounts cannot be queried via Horizon endpoints is what allows us to use it in a limited capacity. It will be no different if someone tries to query Horizon with `GAMJ4W6QDEKJWZRMAWGD7ETKPIMKUX4NCZ3JBEYPILZSOWTW7NL6TDQ4-yQ2eD5xHRSWUFJfCZqggCS5peHFkb9SXJ3hBjjZjHnc=`, they'll see the same error.

JakeUrban:
Update, Candace agree with you @leighmcculloch , I'll update the language to remove mention of custodians.

ecosystem/sep-0010.md
JakeUrban: What do we gain from using the M-address format? Memos are the tried and true solution to this problem without M-addresses, and without full `M`-address support in Horizon/SDK's I don't see a reason to use them.

ecosystem/sep-0010.md
JakeUrban: What do we gain from using the M-address format? Memos are the tried and true solution to this problem without M-addresses, and without full M-address support in Horizon/SDK's I don't see a reason to use them.

JakeUrban: I can also see future issues where the `M`-addresses used here complicate the process for full `M`-address support. For example, how will the `memo` value be encoded into the `G`-address to get an `M`-address? It will likely be a different algorithm than the algorithm used to generate real `M`-addresses.

JakeUrban:
Other than the `M-addresses` vs. Memo debate, this PR is ready for review and merge. @tomerweller you might have to break the tie here.

leighmcculloch:
I'd like to do another pass before we merge it independent of the M-address vs memo debate, it's on my queue. Also we need to think about timing of merging this change with other SEP-10 changes in the pipeline. Let me know if there's a deadline we're trying to hit with this.

ecosystem/sep-0010.md
```diff
@@ -121,6 +125,14 @@ Every other HTTP status code will be considered an error. For example:
 }
 ```
 
+### Pooled Accounts
+
+In the context of Stellar, pooled accounts are Stellar accounts that are held and managed by an entity that is _not_ the owner of the funds within the account. An alternative approach to pooled accounts that still allows the client entity to manage and hold account secret keys is to create a separate Stellar account for each unique user. While this is possible, a much more common practice is to hold the funds of many users in a single _pooled_ account.
+
+The `memo` parameter supported in `GET <WEB_AUTH_ENDPOINT>` API calls can be used to create `account`-`memo` composite identifiers for the web service's users. The decoded contents of the token returned from `POST <WEB_AUTH_ENDPOINT>` calls should contain both the `account` and `memo` values so the service requiring SEP-10 authentication can tie each request to the user.
+
+In the future, there may be a more streamlined way to specify users within pooled accounts. This section and the API specification will be updated when that change happens.
```
leighmcculloch, line 134: I think this section needs to be used to the bottom of the document because it is informational discussion. Right now it is sitting in between the challenge and token sections which should be back to back.

I also think we should remove the discussion about fund ownership, secret keys, unique user's, the new concept of "pooled accounts", and the theories about what might happen in the future. I think we should call out in this section why memos are relevant and when they can be used. This aligns more specifically with the raw constructs.

```suggestion
### Memos

Stellar transactions may be specified with a memo and services that share a single account for many users typically use the memo as a method to distinguishing one user from another, or one internal off-network account from another.

The `memo_type` and `memo` parameters supported in `GET <WEB_AUTH_ENDPOINT>` API calls can be used to communicate to the server that the client intends to authenticate a Stellar account with an additional claim that the authenticating user is the provided `memo`.
```

ecosystem/sep-0010.md
```diff
@@ -14,6 +14,8 @@ Version 2.0.0
 
 This SEP defines the standard way for clients such as wallets or exchanges to create authenticated web sessions on behalf of a user who holds a Stellar account. A wallet may want to authenticate with any web service which requires a Stellar account ownership verification, for example, to upload KYC information to an anchor in an authenticated way as described in [SEP-12](sep-0012.md).
 
+This SEP also supports clients who control [pooled accounts](#pooled-accounts) containing funds owned by _many_ users. These clients manage their users' Stellar accounts and hold the secret keys for those accounts.
```
leighmcculloch, line 17: Suggested change to align with the other comment about pooled accounts:
```suggestion
This SEP also supports authenticating clients who will prove a claim to a [memo](#memos) value of a Stellar account that is useful for authenticating clients who share a Stellar account.
```

leighmcculloch (commented):
The way I was thinking about this change when I left some prior comments was conflating muxed addresses and transaction-memo support as the same thing. @JakeUrban I think an earlier proposal you made was better than where I took us.

Transaction-memos are conceptually the same thing as the newer muxed addresses, but also not really. They don't behave the same way and given muxed addresses are not completely implemented we also don't completely know how they will impact or behave in all products. For this reason I think it is more useful to think of this as I think @JakeUrban had already been thinking about this, authenticating a G address that has an additional claim, a memo.

When muxed addresses come along they will be the authentication of an M address with no additional claim and will be supported in an entirely different way.

This was probably obvious to others, but this just dawned on me.

I make a couple requests below, that I think are likely asking we revert back to a prior proposal that is similar but slightly different to 5a145d49975459cffddd33acc1bd20fb5fddbfa1. @JakeUrban I can chat more synchronously about these tomorrow afternoon if you're available.

ecosystem/sep-0010.md
```diff
@@ -149,13 +161,18 @@ To validate the challenge transaction the following steps are performed by the s
...
+  * For example, when `account` is pooled, the user can be identified using the 32-byte hash memo: `GAMJ4W6QDEKJWZRMAWGD7ETKPIMKUX4NCZ3JBEYPILZSOWTW7NL6TDQ4-yQ2eD5xHRSWUFJfCZqggCS5peHFkb9SXJ3hBjjZjHnc=`
+  * The `sub` field without a `memo` value would be a plain Stellar public key: `GAMJ4W6QDEKJWZRMAWGD7ETKPIMKUX4NCZ3JBEYPILZSOWTW7NL6TDQ4`
+  * Web services that are passed a JWT token can use the `sub` memo value to internally identify the user who owns the funds held by the client on each request
 * `iat` (the time at which the JWT was issued [RFC7519, Section 4.1.6](https://tools.ietf.org/html/rfc7519#section-4.1.6)) — current timestamp (`1530644093`)
 * `exp` (the expiration time on or after which the JWT must not be accepted for processing, [RFC7519, Section 4.1.4](https://tools.ietf.org/html/rfc7519#section-4.1.4)) — a server can pick its own expiration period for the token (`1530730493`)
```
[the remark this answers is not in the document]
JakeUrban: I agree, this is the route to go. It wouldn't surprise me if wallet-anchor services are already adding non-standard fields to their JWT's.

ecosystem/sep-0010.md
JakeUrban: I think its important to have a section header that hints at what it enables. We don't need to strictly defined Pooled Accounts, but a person looking for how a pooled account is authenticated will appreciate the more direct label.

ecosystem/sep-0010.md
JakeUrban: `who will prove a claim to a memo value` doesn't read well. What about:
```suggestion
This SEP also supports authenticating clients that share a Stellar account. Each unique client should be identified with a [memo](#memos) value included in the initial GET request as well as the JWT body.
```

leighmcculloch: That's clearer. I think we should leave out technical details in this paragraph, and instead of stating what a client should do, we don't need to be prescriptive, and can highlight who it relates to. 
```suggestion
This SEP also supports authenticating clients that share a Stellar account and use [memos](#memos) to distinguish users or accounts for payments.
```

ecosystem/sep-0010.md
leighmcculloch: Multi-user accounts is ambiguous with multi-signer, which is a different. I think some of the issues with names like multi-user, pooled, custodial, shared, etc is that they are not terms used by anyone else. In our other SEPs that ellude to the same use cases we discuss only memos, e.g. SEP-2. For this reason I think titling this as memos is sufficient, especially with the link from the summary.

leighmcculloch: Also if muxed addresses take off ultimately it will become the term people use and this current functionality will fade into the background.

ecosystem/sep-0010.md
```diff
@@ -121,6 +126,12 @@ Every other HTTP status code will be considered an error. For example:
 }
 ```
 
+### Memos
+
+Stellar transactions may be specified with a memo and services that share a single account for many users typically use the memo as a method to distinguishing one user from another, or one internal off-network account from another.
+
+The `memo_type` and `memo` parameters supported in `GET <WEB_AUTH_ENDPOINT>` API calls can be used to communicate to the server that the client intends to authenticate a Stellar account with an additional claim that the authenticating user has control of or is possession of the provided `memo`.
+
```
leighmcculloch, line 134: Can we move this to the bottom of the document after the `JWT Expiration` section?

ecosystem/sep-0010.md
```diff
@@ -262,12 +277,6 @@ This convention ensures that SEP-10 clients and servers can use the same passphr
 
 The client can examine the `network_passphrase` (if defined) that the server includes in its response from the challenge endpoint to be sure it's using the correct passphrase and is connecting to the server that it expected.
 
-## JWT best practices
```
leighmcculloch, line 265: Could we do the JWT expiration/best practices change in a separate PR? It seems unrelated to the rest of this change.

leighmcculloch (commented):
I think this looks good. One minor thing mentioned below.

Before merging, this PR will also require version update, and we need to consider when this could be released and what version to release it in. 

Do we have any isolated use cases who could experiment with it before we merge?

ecosystem/sep-0010.md
JakeUrban: It is unrelated but not meaningful, just combines the two JWT sections since the `JWT best practices` was one sentence.

JakeUrban: I can undo the change if we don't want to include it

JakeUrban:
We could merge this into it's own branch on stellar/stellar-protocol, so it's not in a PR and won't accidentally be merged before it should be. I remember thinking this would be included in a SEP-10 4.0 change.

leighmcculloch (commented):
I think we have agreement and so I have marked this change proposed to SEP-10 as accepted and as needing a pilot to confirm the functionality meets the needs we've set out to fulfill. We'll hold off releasing/merging this change until the pilot occurs.

JakeUrban:
**DON'T MERGE THIS YET**

JakeUrban:
I've implemented the changes described here in Polaris, including the related changes to SEP-6, SEP-12, and SEP-24: [PR](https://github.com/stellar/django-polaris/pull/379)

Linopena318:
How do we make more coins

JakeUrban:
Closing, but opening a new PR from `JakeUrban:sep10-pooled-accounts` to `stellar:sep10-pooled-accounts`.

PR #735, opened 2020-09-30
SEP-12: DELETE Content-type and Request Body Args

JakeUrban:
resolves stellar/stellar-protocol#734

### DIFF
--- a/ecosystem/sep-0012.md
+++ b/ecosystem/sep-0012.md
@@ -243,12 +243,13 @@ For example:
 
 ## Customer DELETE
 
-Delete all personal information that the anchor has stored about a given customer. `[account]` is the Stellar account ID (`G...`) of the customer to delete. This request must be authenticated (via [SEP-10](sep-0010.md)) as coming from the owner of the account that will be deleted.
+Delete all personal information that the anchor has stored about a given customer. `[account]` is the Stellar account ID (`G...`) of the customer to delete. This request must be authenticated (via [SEP-10](sep-0010.md)) as coming from the owner of the account that will be deleted. If `account` does not uniquely identify an individual customer (a shared account), the client should use the `memo` and `memo_type` fields in the request body.
 
 ### Request
 
 ```
 DELETE [KYC_SERVER || TRANSFER_SERVER]/customer/[account]
+Content-Type: multipart/form-data
 ```
 
 Name | Type | Description
--- a/ecosystem/sep-0012.md
+++ b/ecosystem/sep-0012.md
@@ -243,7 +243,7 @@ For example:
 
 ## Customer DELETE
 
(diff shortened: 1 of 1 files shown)


JakeUrban:
@gaclaudiu

ecosystem/sep-0012.md
```diff
@@ -243,12 +243,13 @@ For example:
...
 ## Customer DELETE
 
-Delete all personal information that the anchor has stored about a given customer. `[account]` is the Stellar account ID (`G...`) of the customer to delete. This request must be authenticated (via [SEP-10](sep-0010.md)) as coming from the owner of the account that will be deleted.
+Delete all personal information that the anchor has stored about a given customer. `[account]` is the Stellar account ID (`G...`) of the customer to delete. This request must be authenticated (via [SEP-10](sep-0010.md)) as coming from the owner of the account that will be deleted. If `account` does not uniquely identify an individual customer (a shared account), the client should use the `memo` and `memo_type` fields in the request body.
 
 ### Request
 
 ```
 DELETE [KYC_SERVER || TRANSFER_SERVER]/customer/[account]
+Content-Type: multipart/form-data
```
leighmcculloch, line 252: ❓ Are the `memo` and `memo_type` fields passed in the query parameters, or passed in the form data? We could do with some request examples in this SEP to make it clearer.

Also what other form data is being passed, will there be any files? If no files, why multipart? If we're passing only form data  we should use `application/x-www-form-urlencoded`.

Or is the goal to make all the endpoints use the same content type and some accept files?

ecosystem/sep-0012.md
```diff
@@ -243,12 +243,13 @@ For example:
 
 ## Customer DELETE
 
-Delete all personal information that the anchor has stored about a given customer. `[account]` is the Stellar account ID (`G...`) of the customer to delete. This request must be authenticated (via [SEP-10](sep-0010.md)) as coming from the owner of the account that will be deleted.
+Delete all personal information that the anchor has stored about a given customer. `[account]` is the Stellar account ID (`G...`) of the customer to delete. This request must be authenticated (via [SEP-10](sep-0010.md)) as coming from the owner of the account that will be deleted. If `account` does not uniquely identify an individual customer (a shared account), the client should use the `memo` and `memo_type` fields in the request body.
```
leighmcculloch, line 246: 💭 
```suggestion
Delete all personal information that the anchor has stored about a given customer. `[account]` is the Stellar account ID (`G...`) of the customer to delete. This request must be authenticated (via [SEP-10](sep-0010.md)) as coming from the owner of the account that will be deleted. If `account` does not uniquely identify an individual customer (a shared account), the client should include the `memo` and `memo_type` fields in the request body.
```

leighmcculloch (approved):

ecosystem/sep-0012.md
JakeUrban: They're passed in the request body as `multipart/form-data`, not query parameters. This is because `PUT` is done the same way.

> Or is the goal to make all the endpoints use the same content type and some accept files?

Yes.

leighmcculloch (approved):

leighmcculloch:
Independent of this PR we could do with some request examples in SEP-12. There are JSON response examples, but no request examples for each endpoint.

PR #739, opened 2020-10-06
SEP-24: Communicate KYC status

morleyzhi:
Wallets need visibility into the user's KYC status to alert users when they can continue or re-request their deposit / withdrawal intent after being approved. This adds two things that help do this:

- A boolean `kyc_verified` field on transaction objects
- A `on_change_callback` that calls back each time the transaction object's `status` or `kyc_verified` fields change

### DIFF
--- a/ecosystem/sep-0024.md
+++ b/ecosystem/sep-0024.md
@@ -337,7 +337,7 @@ As an alternative to using the `callback` parameter, the wallet can poll the tra
 
 Anchors should only send responses to `callback` when the user is finished with the interactive flow, and it's safe for the wallet to close the window. It should only be called zero or one time.
 
-`on_change_callback` should be called each time the `status` or `kyc_verified` properties of the transaction should change. It can be called any number of times.
+`on_change_callback` should be called each time the `status` or `kyc_verified` properties of the transaction change. It can be called any number of times.
 
 
 #### Guidance for wallets: completing an interactive withdrawal
--- a/ecosystem/sep-0024.md
+++ b/ecosystem/sep-0024.md
@@ -593,7 +593,6 @@ Name | Type | Description
 `status` should be one of:
 
 * `incomplete` -- there is not yet enough information for this transaction to be initiated. Perhaps the user has not yet entered necessary info in an interactive flow.
-* `pending_kyc` -- the user has submitted their KYC information and the anchor is verifying it.
 * `pending_user_transfer_start` -- the user has not yet initiated their transfer to the anchor. This is the necessary first step in any deposit or withdrawal flow.
 * `pending_external` -- deposit/withdrawal has been submitted to external network, but is not yet confirmed. This is the status when waiting on Bitcoin or other external crypto network to complete a transaction, or when waiting on a bank transfer.
 * `pending_anchor` -- deposit/withdrawal is being processed internally by anchor. This can also be used when the anchor must verify KYC information prior to deposit/withdrawal.
--- a/ecosystem/sep-0024.md
+++ b/ecosystem/sep-0024.md
@@ -305,7 +305,9 @@ The basic parameters are summarized in the table below.
 
 Name | Type | Description
 -----|------|------------
-`callback` | string | (optional) A URL that the anchor should `POST` a JSON message to when the `status` or `kyc_verified` properties change. Can also be set to `postMessage`.
+`callback` | string | (optional) A URL that the anchor should `POST` a JSON message to when the user successfully completes the interactive flow. Can also be set to `postMessage`. 
+`on_change_callback` | string | (optional) A URL that the anchor should `POST` a JSON message to when the `status` or `kyc_verified` properties change. Can also be set to `postMessage`.
+
 
 **`callback` details**
 
@@ -330,6 +332,13 @@ target.postMessage({
 ```
 
 As an alternative to using the `callback` parameter, the wallet can poll the transaction endpoint [`/transaction`](#single-historical-transaction) with the request `id` to check the status of the request.
+
+**Differences between `callback` and `on_change_callback`**
+
+Anchors should only send responses to `callback` when the user is finished with the interactive flow, and it's safe for the wallet to close the window. It should only be called zero or one time.
+
+`on_change_callback` should be called each time the `status` or `kyc_verified` properties of the transaction should change. It can be called any number of times.
+
 
 #### Guidance for wallets: completing an interactive withdrawal
 
--- a/ecosystem/sep-0024.md
+++ b/ecosystem/sep-0024.md
@@ -549,6 +549,7 @@ Name | Type | Description
 `kind` | string | `deposit` or `withdrawal`.
 `status` | string | Processing status of deposit/withdrawal.
 `status_eta` | number | (optional) Estimated number of seconds until a status change is expected.
+`kyc_verified` | boolean | (optional) Whether or not the deposit/withdraw subject has been verified by the anchor.
 `more_info_url` | string | A URL that is opened by wallets after the interactive flow is complete. It can include banking information for users to start deposits, the status of the transaction, or any other information the user might need to know about the transaction. 
 `amount_in` | string | Amount received by anchor at start of transaction as a string with up to 7 decimals. Excludes any fees charged before the anchor received the funds.
 `amount_out` | string | Amount sent by anchor to user at end of transaction as a string with up to 7 decimals. Excludes amount converted to XLM to fund account and any external fees.
--- a/ecosystem/sep-0024.md
+++ b/ecosystem/sep-0024.md
@@ -313,7 +313,15 @@ Name | Type | Description
 
 If the wallet wants to be notified that the user has completed the anchor's interactive flow (either success or failure), it can add this parameter to the URL. If the user abandons the process, the anchor does not need to report anything to the wallet. If the `callback` value is a URL, the anchor must `POST` to it with a JSON message as the body. If `callback=postMessage` is passed, the anchor must post a JSON message to `window.opener` via the Javascript [`Window.postMessage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) method. If `window.opener` is undefined, the message must be posted to `window.parent` instead.
 
-In either case, the JSON message should be identical to the response format for the [/transaction](#single-historical-transaction) endpoint.
+**Differences between `callback` and `on_change_callback`**
+
+Anchors should only send responses to `callback` when the user is finished with the interactive flow, and it's safe for the wallet to close the window. It should only be called zero or one time.
+
+`on_change_callback` should be called each time the `status` or `kyc_verified` properties of the transaction change. It can be called any number of times.
+
+**`callback` / `on_change_callback` example**
+
+The JSON message should be identical to the response format for the [/transaction](#single-historical-transaction) endpoint.
 
 ```javascript
 // Example postMessage callback at the end of an interactive withdraw, indicating that the anchor is waiting for the wallet to send a payment in the amount of 80 of the asset in question. 
@@ -331,13 +339,7 @@ target.postMessage({
 }, "*");
 ```
 
-As an alternative to using the `callback` parameter, the wallet can poll the transaction endpoint [`/transaction`](#single-historical-transaction) with the request `id` to check the status of the request.
-
-**Differences between `callback` and `on_change_callback`**
-
-Anchors should only send responses to `callback` when the user is finished with the interactive flow, and it's safe for the wallet to close the window. It should only be called zero or one time.
-
-`on_change_callback` should be called each time the `status` or `kyc_verified` properties of the transaction change. It can be called any number of times.
+As an alternative to using the `callback` or `on_change_callback` parameters, the wallet can poll the transaction endpoint [`/transaction`](#single-historical-transaction) with the request `id` to check the status of the request.
 
 #### Guidance for wallets: completing an interactive withdrawal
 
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0024.md
```diff
@@ -305,7 +305,7 @@ The basic parameters are summarized in the table below.
 
 Name | Type | Description
 -----|------|------------
-`callback` | string | (optional) A URL that the anchor should `POST` a JSON message to when the user successfully completes the interactive flow. Can also be set to `postMessage`.
+`callback` | string | (optional) A URL that the anchor should `POST` a JSON message to when the `status` property changes, or when the `kyc_verified` property changes and the `status` property is incomplete or pending. Can also be set to `postMessage`.
```
JakeUrban, line 308:
```suggestion
`callback` | string | (optional) A URL that the anchor should `POST` a JSON message to when the `status` or `kyc_verified` properties change. Can also be set to `postMessage`.
```

ecosystem/sep-0024.md
```diff
@@ -549,6 +549,7 @@ Name | Type | Description
 `kind` | string | `deposit` or `withdrawal`.
 `status` | string | Processing status of deposit/withdrawal.
 `status_eta` | number | (optional) Estimated number of seconds until a status change is expected.
+`kyc_verified` | boolean | (optional) Whether or not the deposit/withdraw subject has been verified by the anchor.
```
JakeUrban, line 552:
```suggestion
`kyc_verified` | boolean | (optional) Whether or not the user's KYC information has been verified by the anchor.
```

JakeUrban (commented):
Small changes in phrasing but otherwise LGTM

ecosystem/sep-0024.md
leighmcculloch: Prior to this change if the client uses `postMessage` as the callback they expect to get a single call to postMessage when the interactive flow has changed. With this change an implementer might think they should call postMessage during the interactive flow resulting in multiple `postMessage` calls. It's likely because of the previous definition that a client has implemented their application to close the window containing the interactive flow when the first postMessage is received.

I think because of this we need to retain some of the original language so that when postMessage is used the first call to postMessage indicates not only a change in status or properties but that the interactive flow has completed.

JakeUrban: +1 to Leigh's comment, I didn't look at the original language before making my comment.

morleyzhi: @leighmcculloch I'm not quite understanding what you're suggesting? It seems like our proposed change will be a breaking one for your specific case, and I don't think there's any way around it. Unless you're saying that postMessage would behave differently than callback URLs?

JakeUrban: @morleyzhi I think what Leigh is saying is that the original purpose to the `postMessage` call was for the client to close the window. There's no reason to remove this language, since the change we're making is additive. The first `postMessage` callback can signal that the flow is complete AND subsequent callbacks can signal a change in `status` or `kyc_verified`.

JakeUrban: Note that when the flow completes, the `status` will likely change too, so the distinction between the two purposes is actually not needed from a functional standpoint.

morleyzhi: The problem is that there may be some status changes before the flow is complete: the incomplete -> pending_kyc one comes to mind.

morleyzhi: If we're concerned about backwards compatibility, we can add a second callback property.

JakeUrban: That may be necessary, because if you look at the SEP, `callback` is stated to be used to signal the end of the interactive flow all over the place. Going back on this would cause issues for wallets expecting a single callback request.

leighmcculloch: `callback_on_change` sounds good to me.

ecosystem/sep-0024.md
```diff
@@ -583,6 +584,7 @@ Name | Type | Description
 `status` should be one of:
 
 * `incomplete` -- there is not yet enough information for this transaction to be initiated. Perhaps the user has not yet entered necessary info in an interactive flow.
+* `pending_kyc` -- the user has submitted their KYC information and the anchor is verifying it.
 * `pending_user_transfer_start` -- the user has not yet initiated their transfer to the anchor. This is the necessary first step in any deposit or withdrawal flow.
 * `pending_external` -- deposit/withdrawal has been submitted to external network, but is not yet confirmed. This is the status when waiting on Bitcoin or other external crypto network to complete a transaction, or when waiting on a bank transfer.
 * `pending_anchor` -- deposit/withdrawal is being processed internally by anchor. This can also be used when the anchor must verify KYC information prior to deposit/withdrawal.
```
tomerweller, line 590: Are we keeping the KYC language on `pending_anhcor`?

leighmcculloch: It does sound like pending_anchor already does the thing we want.

JakeUrban: Good catch Tomer. This second sentence should be removed, however existing anchors will likely continue to use `pending_anchor` while verifying KYC. No change for them is needed unless they work with a client that wants to see the `pending_kyc` status.

JakeUrban: @leighmcculloch it doesn't because its a wider umbrella then just KYC. SEP-24 needs have a way to communicate that the anchor is specifically working on KYC validation

ecosystem/sep-0024.md
```diff
@@ -549,6 +549,7 @@ Name | Type | Description
 `kind` | string | `deposit` or `withdrawal`.
 `status` | string | Processing status of deposit/withdrawal.
 `status_eta` | number | (optional) Estimated number of seconds until a status change is expected.
+`kyc_verified` | boolean | (optional) Whether or not the user's KYC information has been verified by the anchor.
```
tomerweller, line 552: Should we add a qualifications here that KYC has been verified for this specific transaction? The current language might conflate the status of the user with the status of the transaction. KYC requirements can change and different tiers can be invoked in the future so a wallet should make a concise decision wrt what this status means in the context of a user.

ecosystem/sep-0024.md
tomerweller: @JakeUrban, what other states are under that wider umbrela of `pending_anchor`?

JakeUrban: `pending_anchor` is not completely covered by sub-statuses, really there is no such thing other than this one case.

Meaning, `pending_anchor` historically has been a catch-all for "things the anchor needs to do that it doesn't have a better status for". Now that there's `pending_kyc`, the anchor should use that whenever its relevant in place of `pending_anchor`, however `pending_anchor` will still be necessary for other situations.

JakeUrban: There are no other statuses that could be seen as substitutes for. `pending_anchor`. Its possible some anchors use `pending_anchor` when really its `pending_external`, but thats a minor thing and has no affect on functionality.

tomerweller: @JakeUrban can you give some examples of these other situations that `pending_anchor` will be necessary for?

It is my understanding that this PR attempts to de-complect kyc verification from other potential scenarios in which the wallet is waiting on the anchor. I just don't have a full understanding of what those scenarios are. 

My mental model of `pending_anchor` has been that someone needs to take a look at the KYC information, transaction details and make a yes/no decision. If you take that way from `pending_anchor` - what is left? 

This is a breaking change. We need to have strong conviction here.

ecosystem/sep-0024.md
morleyzhi: Yeah, that's a good idea.

ecosystem/sep-0024.md
morleyzhi: We'll use `pending_anchor` for now.

JakeUrban: Other than KYC, `pending_anchor` is used when 

- Polaris is constructing the Stellar transaction envelope to be submitted to the network
- Polaris has detected a withdrawal payment to its distribution account and is sending the funds off chain

Directly after the first case listed, the status will be `pending_stellar`. Directly after the second case listed, the status will be `pending_external` if still pending or `completed` if the user received the funds off-chain.

tomerweller: > Polaris has detected a deposit payment to the anchor's bank account and is constructing the Stellar transaction envelope to be submitted to the network

tomerweller: Jake, this is super helpful. In both these specific cases it sounds like `pending_anchor` is a very short ephemeral status - can we just skip it?

Though I imagine that with multi-sig distribution transactions your first example can take a longer...

JakeUrban: > can we just skip it?

From a UX perspective, the same rationale could be used for `pending_stellar`, since it's also very temporary. The answer depends on if you want clients or users to know either of those two things happen. If not, you can just notify of success or failure when it happens. Until then, it would be in `pending_user_transfer_start`.

If the transaction requires signatures from distributed systems, then being `pending_anchor` is much more significant since it could take a long time.

morleyzhi: To get this moving, I removed pending_kyc. Anchors that don't support any change can use `pending_anchor` for now.

ecosystem/sep-0024.md
```diff
@@ -331,6 +333,13 @@ target.postMessage({
 
 As an alternative to using the `callback` parameter, the wallet can poll the transaction endpoint [`/transaction`](#single-historical-transaction) with the request `id` to check the status of the request.
 
+**Differences between `callback` and `on_change_callback`**
+
+Anchors should only send responses to `callback` when the user is finished with the interactive flow, and it's safe for the wallet to close the window. It should only be called zero or one time.
+
+`on_change_callback` should be called each time the `status` or `kyc_verified` properties of the transaction should change. It can be called any number of times.
```
leighmcculloch, line 340:
```suggestion
`on_change_callback` should be called each time the `status` or `kyc_verified` properties of the transaction change. It can be called any number of times.
```

ecosystem/sep-0024.md
```diff
@@ -549,6 +558,7 @@ Name | Type | Description
 `kind` | string | `deposit` or `withdrawal`.
 `status` | string | Processing status of deposit/withdrawal.
 `status_eta` | number | (optional) Estimated number of seconds until a status change is expected.
+`kyc_verified` | boolean | (optional) Whether or not the anchor has verified the user's KYC information and has authorized them for this transaction.
```
leighmcculloch, line 561: We should avoid the term authorized here, because the term has particular meaning in payments products and our overloaded use of it may cause confusion for folks in financial industries. I think its use here is unnecessary anyway and can be simply deleted.
```suggestion
`kyc_verified` | boolean | (optional) True if the anchor has verified the user's KYC information for this transaction.
```

ecosystem/sep-0024.md
```diff
@@ -331,6 +333,13 @@ target.postMessage({
 
 As an alternative to using the `callback` parameter, the wallet can poll the transaction endpoint [`/transaction`](#single-historical-transaction) with the request `id` to check the status of the request.
 
+**Differences between `callback` and `on_change_callback`**
+
+Anchors should only send responses to `callback` when the user is finished with the interactive flow, and it's safe for the wallet to close the window. It should only be called zero or one time.
+
+`on_change_callback` should be called each time the `status` or `kyc_verified` properties of the transaction change. It can be called any number of times.
+
+
```
leighmcculloch, line 342: This document appears to use single line separation between sections. Let's drop the double.
```suggestion
```

ecosystem/sep-0024.md
```diff
@@ -331,6 +333,13 @@ target.postMessage({
```
leighmcculloch, line 333: Immediately above here is a good section that details the callback parameter and what data is posted, etc. I think we need a similar section for the new parameter. You could mostly copy paste it and note how the `on_change_callback` is posted to on every change.

JakeUrban (approved):
LGTM

leighmcculloch (approved):

JakeUrban:
Should we add this to SEP-6 for consistency?

ecosystem/sep-0024.md
```diff
@@ -305,13 +305,23 @@ The basic parameters are summarized in the table below.
...
+
 
 **`callback` details**
 
 If the wallet wants to be notified that the user has completed the anchor's interactive flow (either success or failure), it can add this parameter to the URL. If the user abandons the process, the anchor does not need to report anything to the wallet. If the `callback` value is a URL, the anchor must `POST` to it with a JSON message as the body. If `callback=postMessage` is passed, the anchor must post a JSON message to `window.opener` via the Javascript [`Window.postMessage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) method. If `window.opener` is undefined, the message must be posted to `window.parent` instead.
 
-In either case, the JSON message should be identical to the response format for the [/transaction](#single-historical-transaction) endpoint.
+**Differences between `callback` and `on_change_callback`**
+
+Anchors should only send responses to `callback` when the user is finished with the interactive flow, and it's safe for the wallet to close the window. It should only be called zero or one time.
```
JakeUrban, line 318:
```suggestion
Anchors may make _at most one request_ containing a `/transaction` response body to `callback` when the user is finished with the interactive flow _and_ the anchor has verified that it is safe for the wallet to close the popup window. 
```

ezequielsettle (approved):

JakeUrban:
@morleyzhi @marcelosalloum @leighmcculloch are we waiting on anything here? If not I'll merge it.

PR #740, opened 2020-10-07
SEP-0010: Revert home domain verification and relax manage data operation constraints (v2.1.0)

leighmcculloch:
### What
Revert home domain verification and relax manage data operation constraints allowing additional manage data operations as long as their source account is the server account.

### Why
There's been some ambiguity about the intended contents of home domain and in the interest of encouraging compatibility between servers and clients I think we need to consider reverting verification the SEP and SDKs are requiring on that field.

The relaxing of the manage data operations so that we can add a new manage data operation in <SEP-10 v3.0.0 PR> that adds a new hostname for verification. Allowing additional manage data operations with the server account as the source account will allow future releases to add additional fields of data without breaking older clients. Requiring the source account to be the server account ensures that if there is a need in the future to add new fields that the client should confirm and sign for, that older clients won't blindly sign those challenge transactions.

This change is backwards compatible with v2.0.0.

### Example Implementations

https://github.com/stellar/go/pull/3108
https://github.com/stellar/js-stellar-sdk/pull/580

### DIFF
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -6,8 +6,8 @@ Title: Stellar Web Authentication
 Author: Sergey Nebolsin <sergey@mobius.network>, Tom Quisel <tom.quisel@gmail.com>, Leigh McCulloch <@leighmcculloch>, Jake Urban <jake@stellar.org>
 Status: Active
 Created: 2018-07-31
-Updated: 2020-08-14
-Version 2.0.0
+Updated: 2020-??-??
+Version 2.1.0
 ```
 
 ## Simple Summary
@@ -23,7 +23,8 @@ The authentication flow is as follows:
 1. The client obtains a unique [`challenge`](#challenge), which is represented as specially formed Stellar transaction
 1. The client verifies that the transaction has an invalid sequence number 0.  This is extremely important to ensure the transaction isn't malicious.
 1. The client verifies that the transaction is signed by the `SIGNING_KEY` specified by the requested service's [SEP-1 stellar.toml](sep-0001.md).
-1. The client verifies that the transaction includes a Manage Data operation, and that the value contains the home domain used to request the challenge.
+1. The client verifies that the transaction has a single Manage Data operation containing the user's account and a nonce.
+1. The client verifies that if the transaction has other Manage Data operations they all contain the server's account.
 1. The client signs the transaction using the secret key(s) of signers for the user's Stellar account
 1. The client submits the signed challenge back to the server using [`token`](#token) endpoint
 1. The server checks that the user's account exists
@@ -68,7 +69,7 @@ In order for browsers-based wallets to validate the CORS headers, as [specified 
 
 ### Challenge
 
-This endpoint must respond with a Stellar transaction signed by the server that has an invalid sequence number (0) and thus cannot be executed on the Stellar network. The client can then sign the transaction using standard Stellar libraries and submit it to [`token`](#token) endpoint to prove that they control their account. This approach is compatible with hardware wallets such as Ledger. The client must also verify the server's signature to be sure the challenge is signed by the `SIGNING_KEY` from the server's [`stellar.toml`](sep-0001.md). Additionally, the client must ensure that the challenge transaction was created by the requested service by checking the value of the Manage Data operation.
+This endpoint must respond with a Stellar transaction signed by the server that has an invalid sequence number (0) and thus cannot be executed on the Stellar network. The client can then sign the transaction using standard Stellar libraries and submit it to [`token`](#token) endpoint to prove that they control their account. This approach is compatible with hardware wallets such as Ledger. The client must also verify the server's signature to be sure the challenge is signed by the `SIGNING_KEY` from the server's [`stellar.toml`](sep-0001.md).
 
 #### Request
 
@@ -97,9 +98,11 @@ On success the endpoint must return `200 OK` HTTP status code and a JSON object 
   * source account set to server's signing account
   * invalid sequence number (set to 0) so the transaction cannot be run on the Stellar network
   * time bounds: `{min: now(), max: now() + 900 }` (we recommend expiration of 15 minutes to give user time to sign transaction)
-  * operations: `manage_data(source: client_account, key: '<home domain> auth', value: random_nonce())`
-    * The value of key is the fully qualified domain name of the web service requiring authentication followed by `auth`. It can be at most 64 characters.
-    * The value must be 64 bytes long. It contains a 48 byte cryptographic-quality random string encoded using base64 (for a total of 64 bytes after encoding).
+  * operations:
+    * `manage_data(source: client_account, key: '<home domain> auth', value: random_nonce())`
+      * The value of key is the fully qualified domain name of the web service requiring authentication followed by `auth`. It can be at most 64 characters.
+      * The value must be 64 bytes long. It contains a 48 byte cryptographic-quality random string encoded using base64 (for a total of 64 bytes after encoding).
+    * zero or more `manage_data(source: server_account, ...)` reserved for future use
   * signature by the service's stellar.toml `SIGNING_KEY`
 * `network_passphrase`: (optional but recommended) Stellar network passphrase used by the server. This allows the client to verify that it's using the correct passphrase when signing and is useful for identifying when a client or server have been configured incorrectly.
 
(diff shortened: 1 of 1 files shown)


JakeUrban:
@nebolsin Can we get your thoughts here? Thanks for being patient with us as we develop a plan for address the vulnerability. The Ruby SDK already has this functionality correct? It was good intuition on your part.

JakeUrban:
@nebolsin Can we get your thoughts here? Thanks for being patient with us as we develop a plan for these changes. The Ruby SDK already has this functionality correct? It was good intuition on your part.

ecosystem/sep-0010.md
```diff
@@ -6,8 +6,8 @@ Title: Stellar Web Authentication
 Author: Sergey Nebolsin <sergey@mobius.network>, Tom Quisel <tom.quisel@gmail.com>, Leigh McCulloch <@leighmcculloch>, Jake Urban <jake@stellar.org>
 Status: Active
 Created: 2018-07-31
-Updated: 2020-08-14
-Version 2.0.0
+Updated: 2020-??-??
```
JakeUrban, line 9: I'd just mark it as today. This field has always been funny to be since this doc lives on github.

ecosystem/sep-0010.md
```diff
@@ -23,7 +23,8 @@ The authentication flow is as follows:
 1. The client obtains a unique [`challenge`](#challenge), which is represented as specially formed Stellar transaction
 1. The client verifies that the transaction has an invalid sequence number 0.  This is extremely important to ensure the transaction isn't malicious.
 1. The client verifies that the transaction is signed by the `SIGNING_KEY` specified by the requested service's [SEP-1 stellar.toml](sep-0001.md).
-1. The client verifies that the transaction includes a Manage Data operation, and that the value contains the home domain used to request the challenge.
+1. The client verifies that the transaction has a single Manage Data operation containing the user's account and a nonce.
```
JakeUrban, line 26:
```suggestion
1. The client verifies that the transaction has a single Manage Data operation with its source account set to the user's account. The the operation must also contain a nonce value.
```

ecosystem/sep-0010.md
```diff
@@ -23,7 +23,8 @@ The authentication flow is as follows:
...
+1. The client verifies that if the transaction has other Manage Data operations they all contain the server's account.
```
JakeUrban, line 27:
```suggestion
1. The client verifies that if the transaction has other Manage Data operations they all have their source accounts set to the the server's account.
```

JakeUrban (approved):
A couple suggestions but looks good

leighmcculloch:
@JakeUrban Thanks. I made those changes to the protocol.

JakeUrban:
I'm slightly curious/concerned that allowing N Manage Data Ops will give room for anchors-client pairs to start using that functionality to pass non-standardized key-value pairs. I don't know how to feel about that, what do you think? @leighmcculloch

leighmcculloch:
@JakeUrban That's possible. It's not clear to me that it'd be a problem if that happened. As long as a server doesn't include operations that are intended for the client to sign/authorize for, which is prevented by the requirement that the source accounts of any additional ops be the server account.

An anchor-client pair could do this today without this change too. And in both cases they need to modify the SDKs since the challenge transaction build functions don't allow adding arbitrary manage data ops.

An alternative would be to add the `server_hostname` manage data op that we're adding in SEP-10 v3.0.0 as a reserved for future use manage data op in this change. I think that is unnecessarily specific and binds the releases together.

JakeUrban:
> And in both cases they need to modify the SDKs

But that won't be the case with this change. A specific client-server pair could have 2 additional manage data ops to store account data, say. They expect the home domain op first as specified by the standard, then the 2 custom ops. With SEP-10 3.0, the order changes. The second Manage Data op will be something different than it used to be, which would likely require them to adjust the expectations in their code.

ecosystem/sep-0010.md
```diff
@@ -23,7 +23,8 @@ The authentication flow is as follows:
 1. The client obtains a unique [`challenge`](#challenge), which is represented as specially formed Stellar transaction
 1. The client verifies that the transaction has an invalid sequence number 0.  This is extremely important to ensure the transaction isn't malicious.
 1. The client verifies that the transaction is signed by the `SIGNING_KEY` specified by the requested service's [SEP-1 stellar.toml](sep-0001.md).
-1. The client verifies that the transaction includes a Manage Data operation, and that the value contains the home domain used to request the challenge.
+1. The client verifies that the transaction has a single Manage Data operation with its source account set to the user's account and value set to a nonce value.
```
accordeiro, line 26: Should we make it explicit that the client should **not** validate the contents of the `key` param? I want to avoid any confusion, since later on in the spec we give more strict instructions for the server-side implementation.

leighmcculloch:
In v3.0.0 the order of additional ops after the first op isn't intended to matter. Let's discuss on #741 how we can clarify that in v3.0.0.

ecosystem/sep-0010.md
```diff
@@ -97,9 +98,11 @@ On success the endpoint must return `200 OK` HTTP status code and a JSON object
   * source account set to server's signing account
   * invalid sequence number (set to 0) so the transaction cannot be run on the Stellar network
   * time bounds: `{min: now(), max: now() + 900 }` (we recommend expiration of 15 minutes to give user time to sign transaction)
-  * operations: `manage_data(source: client_account, key: '<home domain> auth', value: random_nonce())`
-    * The value of key is the fully qualified domain name of the web service requiring authentication followed by `auth`. It can be at most 64 characters.
-    * The value must be 64 bytes long. It contains a 48 byte cryptographic-quality random string encoded using base64 (for a total of 64 bytes after encoding).
+  * operations:
+    * `manage_data(source: client_account, key: '<home domain> auth', value: random_nonce())`
```
accordeiro, line 102: I don't feel strongly about this one, but should we remove the spec for what's in the `key` altogether now? I guess it doesn't hurt, but it's also not necessary.

ecosystem/sep-0010.md
```diff
@@ -136,8 +139,7 @@ To validate the challenge transaction the following steps are performed by the s
 * verify that transaction source account is equal to the server's signing key;
 * verify that transaction has time bounds set, and that current time is between the minimum and maximum bounds;
 * verify that transaction contains a single [Manage Data](https://www.stellar.org/developers/guides/concepts/list-of-operations.html#manage-data) operation that:
-  * has a non-null source account 
-  * has the service's home domain as the key of the operation
+  * has a non-null source account
```
accordeiro, line 142: Same point as above, about being explicit about what shouldn't be validated.

accordeiro (commented):
The changes look good – I left a few suggestions to reduce ambiguity 👇

ecosystem/sep-0010.md
leighmcculloch: I think we should defer the decision of what to do with home domain field in general and separate it from the much smaller change we're making now to stop requiring it be verified. It's the path to the most compatibility since a server that stops populating it and who was successfully integrated with a client who was verifying it will sudennly see their auth fail.

ecosystem/sep-0010.md
leighmcculloch: We could make it more explicit, but I also think we are being sufficiently explicit that the home domain is no longer included in what should be verified. This returns the specification back to essentially what it said prior to this change. Let me know if you feel strongly about this, or can you make a suggestion for how to reword this?

leighmcculloch: What do you think of this?
```suggestion
1. The client verifies that the transaction has a single Manage Data operation with its source account set to the user's account and value set to a nonce value. The client ignores the home domain included.
```

ecosystem/sep-0010.md
leighmcculloch: I think if we add this statement we can leave this section as is: https://github.com/stellar/stellar-protocol/pull/740#discussion_r502068312

Also FYI this section is actually what the server must do to validate the challenge transaction passed back.

ecosystem/sep-0010.md
accordeiro: Looks good!

accordeiro (approved):

gekpp:
Hi everybody.
On behalf of Tempo Money transfer, EURT asset issuer I'd like to vote for this PR.

Let me describe why.
Recently we faced the limitation of having single manage data operation in the challenge transaction.
We wanted to add more key-value pairs for being included in the resulting JWT. More specifically, we wanted to add auth token expiration time and ACL. The idea was the wallet backend knows a way to learn user's permissions, the wallet backend interacts with payment system backend via SEP-XX protocols. We want to implement the flow of issuing SEP-10 token as follows:
1) Wallet backend learns user's access rights from passport service.
2) Wallet backend learns the expiration time for particular use-case. It might depend on front-end application using the wallet backend.
3) Wallet backend requests a challenge via `GET /challenge` sending extra query parameters (required expiration time and ACL) along with `account`.
4) Payment System Backend verifies extra parameters and adds them as manage data operations to transaction, signs and replies to the requestor.
5) Wallet backend validates that `manage data` equal to what was requested, signs a transaction and POST it for issuing the JWT.
6) Payment System Backend iterates over manage data operations and adds them as claims into JWT.


This approach will give the possibility to extend resulting JWT with any data the parties agreed on.
There are different approaches of implementing adding extra params to JWT but all of them require introducing more connection between backend components and we would like to avoid it if possible.

PR #741, opened 2020-10-07
SEP-0010: Add new manage data op containing server hostname (v3.1.0)

leighmcculloch:
### What
Add a new manage data op containing server hostname.

### Why
To use in place of the home domain verification changed in #740.

### DIFF
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -7,7 +7,7 @@ Author: Sergey Nebolsin <@nebolsin>, Tom Quisel <tom.quisel@gmail.com>, Leigh Mc
 Status: Active
 Created: 2018-07-31
 Updated: 2020-10-07
-Version 2.1.0
+Version 3.0.0
 ```
 
 ## Simple Summary
@@ -23,7 +23,8 @@ The authentication flow is as follows:
 1. The client obtains a unique [`challenge`](#challenge), which is represented as specially formed Stellar transaction
 1. The client verifies that the transaction has an invalid sequence number 0.  This is extremely important to ensure the transaction isn't malicious.
 1. The client verifies that the transaction is signed by the `SIGNING_KEY` specified by the requested service's [SEP-1 stellar.toml](sep-0001.md).
-1. The client verifies that the transaction has a single Manage Data operation with its source account set to the user's account and value set to a nonce value. The client ignores the home domain included.
+1. The client verifies that the transaction has exactly one Manage Data operation containing the user's account and a nonce. This operation must be the first operation. The client ignores the home domain included.
+1. The client verifies that the transaction has at most one Manage Data operation with key `server_hostname` with the server's account as the source account. This operation may appear in any position after the first operation. If present, the client verifies that this operation contains the hostname of the SEP-10 server from which the client requested the challenge.
 1. The client verifies that if the transaction has other Manage Data operations they all have their source accounts set to the the server's account.
 1. The client signs the transaction using the secret key(s) of signers for the user's Stellar account
 1. The client submits the signed challenge back to the server using [`token`](#token) endpoint
@@ -102,7 +103,10 @@ On success the endpoint must return `200 OK` HTTP status code and a JSON object 
     * `manage_data(source: client_account, key: '<home domain> auth', value: random_nonce())`
       * The value of key is the fully qualified domain name of the web service requiring authentication followed by `auth`. It can be at most 64 characters.
       * The value must be 64 bytes long. It contains a 48 byte cryptographic-quality random string encoded using base64 (for a total of 64 bytes after encoding).
-    * zero or more `manage_data(source: server_account, ...)` reserved for future use
+    * subsequent operations, order unimportant:
+      * `manage_data(source: server_account, key: 'server_hostname', value: server_hostname)`
+        * The value is the fully qualified domain name of the SEP-10 server that issued the challenge transaction. It can be at most 64 characters.
+      * zero or more `manage_data(source: server_account, ...)` reserved for future use
   * signature by the service's stellar.toml `SIGNING_KEY`
 * `network_passphrase`: (optional but recommended) Stellar network passphrase used by the server. This allows the client to verify that it's using the correct passphrase when signing and is useful for identifying when a client or server have been configured incorrectly.
 
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -7,7 +7,7 @@ Author: Sergey Nebolsin <@nebolsin>, Tom Quisel <tom.quisel@gmail.com>, Leigh Mc
 Status: Active
 Created: 2018-07-31
 Updated: 2020-10-07
-Version 3.0.0
+Version 3.1.0
 ```
 
 ## Simple Summary


ecosystem/sep-0010.md
```diff
@@ -97,12 +99,20 @@ On success the endpoint must return `200 OK` HTTP status code and a JSON object
...
   * invalid sequence number (set to 0) so the transaction cannot be run on the Stellar network
   * time bounds: `{min: now(), max: now() + 900 }` (we recommend expiration of 15 minutes to give user time to sign transaction)
-  * operations: `manage_data(source: client_account, key: '<home domain> auth', value: random_nonce())`
-    * The value of key is the fully qualified domain name of the web service requiring authentication followed by `auth`. It can be at most 64 characters.
-    * The value must be 64 bytes long. It contains a 48 byte cryptographic-quality random string encoded using base64 (for a total of 64 bytes after encoding).
+  * operations:
+    * `manage_data(source: client_account, key: '<home domain> auth', value: random_nonce())`
+      * The value of key is the fully qualified domain name of the web service requiring authentication followed by `auth`. It can be at most 64 characters.
+      * The value must be 64 bytes long. It contains a 48 byte cryptographic-quality random string encoded using base64 (for a total of 64 bytes after encoding).
+    * `manage_data(source: server_account, key: 'server_hostname', value: server_hostname)`
```
JakeUrban, line 106: So @leighmcculloch as it is currently, the server hostname op must be 2nd. If your intention is to treat the operations more like an unordered set, we need to be a lot more clear on that.

I can imagine most developers accessing these ops will do so with explicit indexes. Currently, I'm sure many do something like:
```
challenge.operations[0]
```

ecosystem/sep-0010.md
```diff
@@ -23,7 +23,9 @@ The authentication flow is as follows:
 1. The client obtains a unique [`challenge`](#challenge), which is represented as specially formed Stellar transaction
 1. The client verifies that the transaction has an invalid sequence number 0.  This is extremely important to ensure the transaction isn't malicious.
 1. The client verifies that the transaction is signed by the `SIGNING_KEY` specified by the requested service's [SEP-1 stellar.toml](sep-0001.md).
-1. The client verifies that the transaction includes a Manage Data operation, and that the value contains the home domain used to request the challenge.
+1. The client verifies that the transaction has a single Manage Data operation containing the user's account and a nonce.
+1. The client verifies that if the transaction has a single Manage Data operation with key `server_hostname` it contains the hostname of the SEP-10 server that the client requested the challenge.
+1. The client verifies that if the transaction has other Manage Data operations they all contain the server's account.
```
accordeiro, line 28: I feel like the wording here might be a bit confusing – you could interpret as if there should be only one Manage Data operation, whereas we're actually expecting a few ones. Here's my suggestion:

```suggestion
1. The client verifies that the transaction has **exactly** one Manage Data operation containing the user's account and a nonce.
1. The client verifies the presence of **at most one** Manage Data operation with key `server_hostname` within the transaction. If present, the client verifies that this operation contains the hostname of the SEP-10 server from which the client requested the challenge.
1. The client verifies that if the transaction has other Manage Data operations they all contain the server's account.
```

ecosystem/sep-0010.md
```diff
@@ -97,12 +99,20 @@ On success the endpoint must return `200 OK` HTTP status code and a JSON object
   * source account set to server's signing account
   * invalid sequence number (set to 0) so the transaction cannot be run on the Stellar network
   * time bounds: `{min: now(), max: now() + 900 }` (we recommend expiration of 15 minutes to give user time to sign transaction)
-  * operations: `manage_data(source: client_account, key: '<home domain> auth', value: random_nonce())`
-    * The value of key is the fully qualified domain name of the web service requiring authentication followed by `auth`. It can be at most 64 characters.
-    * The value must be 64 bytes long. It contains a 48 byte cryptographic-quality random string encoded using base64 (for a total of 64 bytes after encoding).
+  * operations:
+    * `manage_data(source: client_account, key: '<home domain> auth', value: random_nonce())`
+      * The value of key is the fully qualified domain name of the web service requiring authentication followed by `auth`. It can be at most 64 characters.
```
accordeiro, line 104: Note: if we decide to remove the restriction of what goes in the "key" param on SEP v2.1.0, we should also port that change here.

ecosystem/sep-0010.md
```diff
@@ -97,12 +99,20 @@ On success the endpoint must return `200 OK` HTTP status code and a JSON object
...
+      * The value must be 64 bytes long. It contains a 48 byte cryptographic-quality random string encoded using base64 (for a total of 64 bytes after encoding).
+    * `manage_data(source: server_account, key: 'server_hostname', value: server_hostname)`
+      * The value is the fully qualified domain name of the SEP-10 server that issued the challenge transaction. It can be at most 64 characters.
+    * zero or more `manage_data(source: server_account, ...)` reserved for future use
   * signature by the service's stellar.toml `SIGNING_KEY`
 * `network_passphrase`: (optional but recommended) Stellar network passphrase used by the server. This allows the client to verify that it's using the correct passphrase when signing and is useful for identifying when a client or server have been configured incorrectly.
 
+Notes:
```
JakeUrban, line 112: I can see the argument that the 2nd manage data op listed above is grouped with the `other` ops that are unordered. But most people will look at the `transaction` challenge description and see the explicit 2nd op and be confused.

accordeiro (commented):
Looks good, left a few suggestions :)

ecosystem/sep-0010.md
leighmcculloch: Yup, anything after the first op is essentially an unordered set of key-value pairs.
https://github.com/stellar/stellar-protocol/blob/0e537c56ff99edc47af175783a17fc32b7fd0da6/ecosystem/sep-0010.md#L112-L114

JakeUrban: Ok, then I don't think we should have the 2nd manage data operation listed in the `transaction` section explicitly as the 2nd op.

ecosystem/sep-0010.md
leighmcculloch: 💯 I've modified it slightly here to remove the emphasis (bold) since there's lots of important text here and these aren't necessarily the more important than other word here. I also added a statement about the source account being set to the server account which is important.
```suggestion
1. The client verifies that the transaction has exactly one Manage Data operation containing the user's account and a nonce.
1. The client verifies the the transaction has at most one Manage Data operation with key `server_hostname` with the server's account as the source account. If present, the client verifies that this operation contains the hostname of the SEP-10 server from which the client requested the challenge.
1. The client verifies that if the transaction has other Manage Data operations they all contain the server's account.
```

ecosystem/sep-0010.md
leighmcculloch: I've removed these notes, let's continue to discuss the issue here: https://github.com/stellar/stellar-protocol/pull/741#discussion_r502078391.

ecosystem/sep-0010.md
leighmcculloch: I've made the unordered and ordered nature of the ops explicit in two places, how does that look? e0fbb13 ff7a51a

leighmcculloch:
This change was superseded by #746.

leighmcculloch:
This change is being reopened to be a sequel to #746 instead.

leighmcculloch:
Actually, this change needs separating into two releases, so I'll close it and open two new PRs.

PR #744, opened 2020-10-10
SEP-0030: Make compatible with CAP-33

### LINKED ISSUE #668: SEP-0030: Make compatible with CAP-33
### What
Make SEP-30 compatible with CAP-33 by allowing as source account of any transaction/operation being signed to be an account as defined in an configuration provided to the application during startup.

### Why
SEP-30 requires an implementation to only sign transactions where the source account of a transaction and all its operations are the registered account. CAP-33 adds sponsorship to the core protocol and will requires a mix of source accounts in set options operations where the change of signers and the new signer ledger entry are sponsored. A SEP-30 implementation will reject such a transaction. The transaction should be allowed as long as the other accounts, the sponsoring account, is not a registered account of the server to make this SEP compatible with CAP-33.

At some future point if it becomes possible to sign only specific operations on a transaction this change will have been a subset of that future change and so this change is unlikely to become incompatible with some evolution of related functionality.

Related reference implementation change: stellar/go#2796.

cc @stellar/nebula @accordeiro @jonjove

leighmcculloch:
### What
Allow SEP-30 servers to sign transactions that contain source accounts other than the authenticated account, at the discretion of the SEP-30 server operator.

### Why
SEP-30 is not compatible with CAP-33 because SEP-30 requires servers to reject transactions that contain other accounts as source accounts. For accounts that have signers sponsored by a sponsoring account as supported by CAP-33, changing a sponsor requires operations with the sponsoring account as a source account. This means it is not possible to change a sponsored signer via a SEP-30 server.

The limitation within SEP-30 is to address the confused deputy problem. The SEP-30 server is signing a transaction, potentially with a shared key, and it must ensure that it only signs a transaction for a user that relate to accounts that the user is authenticated for.

In the case that a SEP-30 server operator is also the operator of a solution that is utilizing CAP-33, it is highly unlikely that the operator needs the sponsoring account to be registered with the SEP-30 server. In that case it is simple for the operator to ensure that the signing keys used by the SEP-30 server are never signers of the sponsoring account and therefore it doesn't matter if the SEP-30 server signs transactions containing the sponsoring account.

There may be other features of Stellar that make it easier to sign for specific operations on transactions and not an entire transaction, and if they ever eventuate this change would become redundant and could be deprecated. At the moment this appears to be the simplest path forward.

If this change is accepted, then the reference implementation should be updated to match. The issue for that work is stellar/go#2796.

Close #668

### DIFF
--- a/ecosystem/sep-0030.md
+++ b/ecosystem/sep-0030.md
@@ -486,7 +486,7 @@ need to sign transactions that contain other accounts, like a sponsoring
 account used with CAP-33. In those cases a server may be implemented to allow
 transactions to contain that specific sponsoring account as a source account,
 but the server operator must ensure that the signing keys used by the server
-are never capable of authorizing operations for the sponsoring account.
+are never signers of the sponsoring account.
 
 The transaction can contain any operations, but it is anticipated for the use
 cases discussed earlier in this protocol that the transaction will contain an
--- a/ecosystem/sep-0030.md
+++ b/ecosystem/sep-0030.md
@@ -8,8 +8,8 @@ Track: Standard
 Status: Draft
 Discussion: https://groups.google.com/forum/#!topic/stellar-dev/SFr2dHBZlsY
 Created: 2019-12-20
-Updated: 2020-06-30
-Version: 0.5.0
+Updated: 2020-10-09
+Version: 0.6.0
 ```
 
 ## Summary
@@ -59,8 +59,13 @@ The signing flow is as follows:
 - The client, authenticated as the external authentication method (e.g. phone, email, etc) or another account, requests the server to sign a transaction.
 - The server verifies that the [JWT] is valid according to the authentication provider's specification.
 - The server verifies that the external authentication method (e.g. phone, email, etc) or account in the [JWT] is an alternative identity of the account.
-- The server verifies that the source account of the transaction is the account.
-- The server verifies that the source account of all operations on the transaction are not set or set to the account.
+- The server verifies that the source account of the transaction is one of:
+  - The account the client has authenticated for.
+  - In some unique cases an account the server knows definitively for the validity period of the transaction will not be authorized by the signing key.<sup>[1](#note1)</sup>
+- The server verifies that the source account of each operation in the transaction are one of:
+  - Not set.
+  - The account the client has authenticated for.
+  - In some unique cases an account the server knows definitively for the validity period of the transaction will not be authorized by the signing key.<sup>[1](#note1)</sup>
 - The server signs the transaction with the signing secret key.
 - The server responds to the client with the signature.
 
@@ -77,6 +82,9 @@ A client can also use the endpoints in the specification to:
 - Update the identities associated with an account.
 - Find out if a new signing key is available due to signing key rotation.
 - Delete accounts.
+
+Notes:
+1. <a name="note1"></a>See [Endpoints `POST /accounts/<address>/sign/<signing-address>`](#post-accountsaddresssignsigning-address) for more details.
 
 ## Use Cases
 
@@ -467,9 +475,18 @@ available for signing. An account should check what signers are available
 periodically and update the signer on the account to be the most recently
 generated signer.
 
-The transaction must only contain operations for the account. The source
-account of the transaction and the source account of all operations in the
-transaction must match the account in the request.
+A server must only authorize operations in signed transactions that operate on
+the registered account that the client has authenticated for. This is to ensure
+that the server does not authorize operations for an account the client is not
+authorized. Generally to achieve this the transaction must only be signed if it
+contain operations for the authenticated account. The server can verify that
+the source account of the transaction and the source account of all operations
+in the transaction match the account in the request. In some cases servers may
+need to sign transactions that contain other accounts, like a sponsoring
+account used with CAP-33. In those cases a server may be implemented to allow
+transactions to contain that specific sponsoring account as a source account,
+but the server operator must ensure that the signing keys used by the server
+are never capable of authorizing operations for the sponsoring account.
 
 The transaction can contain any operations, but it is anticipated for the use
 cases discussed earlier in this protocol that the transaction will contain an
--- a/ecosystem/sep-0030.md
+++ b/ecosystem/sep-0030.md
@@ -467,6 +467,7 @@ See [Common Fields].
 This endpoint signs a transaction that has operations for the account using the
 signing key associated with the signing address specified in the path.
 
+##### `<signing-address>`
 The signing address specified must be one of the signing keys that the server
 has provided as a signer. The client should use the signing address that has
 been added as a signer to the account. The server may generate new signing keys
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0030.md
```diff
@@ -467,9 +475,18 @@ available for signing. An account should check what signers are available
...
+that the server does not authorize operations for an account the client is not
+authorized. Generally to achieve this the transaction must only be signed if it
+contain operations for the authenticated account. The server can verify that
+the source account of the transaction and the source account of all operations
+in the transaction match the account in the request. In some cases servers may
+need to sign transactions that contain other accounts, like a sponsoring
+account used with CAP-33. In those cases a server may be implemented to allow
+transactions to contain that specific sponsoring account as a source account,
+but the server operator must ensure that the signing keys used by the server
+are never capable of authorizing operations for the sponsoring account.
```
howardtw, line 489: What exactly is the scenario you are trying to prevent? Is it that a bad user might include a `BeginSponsoringFutureReserves` operation and if the recoverysigner signs the transaction without the sponsoring account's consent, the client can exploit the sponsoring account? Just want to make sure I understand the case you have in mind. If my understanding is correct, when we add the unique signing key feature, this can be prevented, right?

Finally, when you said `authorizing`, partially authorizing also counts, right?

howardtw (approved):
LGTM

ecosystem/sep-0030.md
JakeUrban: I would add an explicit example of the account sponsorship situation you mention. Otherwise the generic explanation doesn't quite make it clear.

ecosystem/sep-0030.md
```diff
@@ -59,8 +59,13 @@ The signing flow is as follows:
 - The client, authenticated as the external authentication method (e.g. phone, email, etc) or another account, requests the server to sign a transaction.
 - The server verifies that the [JWT] is valid according to the authentication provider's specification.
 - The server verifies that the external authentication method (e.g. phone, email, etc) or account in the [JWT] is an alternative identity of the account.
-- The server verifies that the source account of the transaction is the account.
-- The server verifies that the source account of all operations on the transaction are not set or set to the account.
+- The server verifies that the source account of the transaction is one of:
+  - The account the client has authenticated for.
+  - In some unique cases an account the server knows definitively for the validity period of the transaction will not be authorized by the signing key.<sup>[1](#note1)</sup>
+- The server verifies that the source account of each operation in the transaction are one of:
```
JakeUrban, line 65:
```suggestion
- The server verifies that the source account of each operation in the transaction is:
```

ecosystem/sep-0030.md
```diff
@@ -59,8 +59,13 @@ The signing flow is as follows:
 - The client, authenticated as the external authentication method (e.g. phone, email, etc) or another account, requests the server to sign a transaction.
 - The server verifies that the [JWT] is valid according to the authentication provider's specification.
 - The server verifies that the external authentication method (e.g. phone, email, etc) or account in the [JWT] is an alternative identity of the account.
-- The server verifies that the source account of the transaction is the account.
-- The server verifies that the source account of all operations on the transaction are not set or set to the account.
+- The server verifies that the source account of the transaction is one of:
```
JakeUrban, line 62:
```suggestion
- The server verifies that the source account of the transaction is:
```

JakeUrban (approved):

ecosystem/sep-0030.md
leighmcculloch: I think we need the one of here otherwise it isn't explicit if the list is `and` or `or`. Is this incorrect to have one of here?

ecosystem/sep-0030.md
leighmcculloch: @howardtw If the signing keys of the server were also signing keys of the sponsoring account and the server allowed the signing of transactions for the sponsoring account then any registered account could request signatures for transactions that authorize operations on the sponsoring account, which is not desirable.

If by partially authorizing you mean the signer was a partial signer of the account, then yes, I think this is implied, but I can make this more explicit. Addressed in 50e1334.

@JakeUrban The discussion about a sponsoring account is the example. What did you have in mind that isn't here?

ecosystem/sep-0030.md
JakeUrban: You're right, and it is a small change anyway

ecosystem/sep-0030.md
```diff
@@ -78,6 +83,9 @@ A client can also use the endpoints in the specification to:
 - Find out if a new signing key is available due to signing key rotation.
 - Delete accounts.
 
+Notes:
+1. <a name="note1"></a>See [Endpoints `POST /accounts/<address>/sign/<signing-address>`](#post-accountsaddresssignsigning-address) for more details.
```
JakeUrban, line 87: After reading this endpoint description, it is still not completely clear to me how this section provides more details to line 64 and 68. This endpoint appears to sign transactions containing signer adjustments, but its unclear what the explicit connection is.

Is the concern that an account being authorized (made a signer) for another account shouldn't be used as the source account for any operations being described above?

ecosystem/sep-0030.md
```diff
@@ -467,9 +475,18 @@ available for signing. An account should check what signers are available
...
+that the server does not authorize operations for an account the client is not
+authorized. Generally to achieve this the transaction must only be signed if it
+contain operations for the authenticated account. The server can verify that
+the source account of the transaction and the source account of all operations
+in the transaction match the account in the request. In some cases servers may
+need to sign transactions that contain other accounts, like a sponsoring
+account used with CAP-33. In those cases a server may be implemented to allow
+transactions to contain that specific sponsoring account as a source account,
+but the server operator must ensure that the signing keys used by the server
+are never signers of the sponsoring account.
```
JakeUrban, line 489: I would split the generalized and concrete examples into two different sections. After re-reading this several times I'm starting to get it, but this paragraph leaves the implementor to interpret what other valid operations could have non-request source account.

Could we define exactly what operations could be permitted? Maybe thats too restrictive, but it feels like there are quasi-defined edge cases lingering here.

leighmcculloch: I agree, I think that is too restrictive. It is also harder to implement and the additional requirements/limitations that would add don't actually help to prevent the problem and the approach isn't applicable to other non-sponsorship use cases that might have the same problem.

I've broken up the large paragraph and reworded some of it to try and be more direct. See 9540af0.

JakeUrban: This is much better, thanks.

ecosystem/sep-0030.md
JakeUrban: This is resolve from my point of view.

ecosystem/sep-0030.md
leighmcculloch: The connection is the paragraph being added about what accounts a server must authorize.

JakeUrban (approved):
LGTM

howardtw (approved):
LGTM

PR #746, opened 2020-10-14
SEP-0010: Add back home domain verification with clarity (v3.0.0)

leighmcculloch:
### What
Add back the home domain verification, but with greater clarity about exactly what the home domain is.

### Why
In v2.1.0 the home domain verification that was added in v2.0.0 was removed because the exact contents of the home domain field could be interpreted in at least two different ways that led to inconsistent implementations. The home domain is still present in the SEP-10 challenge transaction and this change provides clarity to what it should contain and how clients should verify it. This change represents the intended understanding of the home domain field. See v2.0.0 and v2.1.0 for prior context.

### Notes
Servers should use this change to verify that they are placing the correct value for the home domain into the challenge transaction. Once clients implement v3.0.0 they will reject challenge transactions from servers who are providing home domain values that are incorrect.

This is an alternative proposal to #741.

### DIFF
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -25,8 +25,8 @@ It involves three components:
 
 The discovery flow is as follows:
 
-1. The client retrieves the [SEP-1 stellar.toml](sep-0001.md) from the home domain. The home domain hosts the stellar.toml.
-1. The client looks up the `WEB_AUTH_ENDPOINT` and `SIGNING_KEY` from the [SEP-1 stellar.toml](sep-0001.md).
+1. The client retrieves the `stellar.toml` from the home domain in accordance with [SEP-1 stellar.toml](sep-0001.md).
+1. The client looks up the `WEB_AUTH_ENDPOINT` and `SIGNING_KEY` from the `stellar.toml`.
 
 The authentication flow is as follows:
 1. The client obtains a unique [`challenge`](#challenge) from the `WEB_AUTH_ENDPOINT`, which is represented as specially formed Stellar transaction
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -81,7 +81,7 @@ In order for browsers-based wallets to validate the CORS headers, as [specified 
 
 This endpoint must respond with a Stellar transaction signed by the server that has an invalid sequence number (0) and thus cannot be executed on the Stellar network. The client can then sign the transaction using standard Stellar libraries and submit it to [`token`](#token) endpoint to prove that they control their account. This approach is compatible with hardware wallets such as Ledger. The client must also verify the server's signature to be sure the challenge is signed by the `SIGNING_KEY` from the server's [`stellar.toml`](sep-0001.md), and the home domain in the challenge is the domain of the domain hosting the [`stellar.toml`](sep-0001.md).
 
-Client and server arrangements where a client connects directly to a hardcoded server without interacting with a home domain hosting a stellar.toml may verify the signing key and home domain using other methods not defined here. The client should still perform verification of these values, and the home domain should be unique in a way similar to if a home domain was in use. For example, the home domain could be defined as the domain of the SEP-10 server.
+If a client is interacting directly with a SEP-10 server without using a stellar.toml, it must still verify the home domain to ensure it only signs challenges issued by the server that the client intends to authenticate.
 
 #### Request
 
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -32,7 +32,10 @@ The authentication flow is as follows:
 1. The client obtains a unique [`challenge`](#challenge) from the `WEB_AUTH_ENDPOINT`, which is represented as specially formed Stellar transaction
 1. The client verifies that the transaction has an invalid sequence number 0.  This is extremely important to ensure the transaction isn't malicious.
 1. The client verifies that the transaction is signed by the `SIGNING_KEY` obtained through discovery flow.
-1. The client verifies that the transaction's first operation is a Manage Data operation that has its source account set to the user's account and value set to a nonce value. The client ignores the home domain included.
+1. The client verifies that the transaction's first operation is a Manage Data operation that has its:
+   1. Source account set to the user's account.
+   1. Key set to `<home domain> auth` where the home domain is the home domain from the discovery flow.
+   1. Value set to a nonce value.
 1. The client verifies that if the transaction has other operations they are Manage Data operations that all have their source accounts set to the the server's account.
 1. The client signs the transaction using the secret key(s) of signers for the user's Stellar account
 1. The client submits the signed challenge back to the server using [`token`](#token) endpoint
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -6,8 +6,8 @@ Title: Stellar Web Authentication
 Author: Sergey Nebolsin <@nebolsin>, Tom Quisel <tom.quisel@gmail.com>, Leigh McCulloch <@leighmcculloch>, Jake Urban <jake@stellar.org>
 Status: Active
 Created: 2018-07-31
-Updated: 2020-10-07
-Version 2.1.0
+Updated: 2020-10-28
+Version 3.0.0
 ```
 
 ## Simple Summary
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -78,7 +78,7 @@ In order for browsers-based wallets to validate the CORS headers, as [specified 
 
 ### Challenge
 
-This endpoint must respond with a Stellar transaction signed by the server that has an invalid sequence number (0) and thus cannot be executed on the Stellar network. The client can then sign the transaction using standard Stellar libraries and submit it to [`token`](#token) endpoint to prove that they control their account. This approach is compatible with hardware wallets such as Ledger. The client must also verify the server's signature to be sure the challenge is signed by the `SIGNING_KEY` from the server's [`stellar.toml`](sep-0001.md), and the home domain in the challenge is the domain of the domain hosting the [`stellar.toml`](sep-0001.md).
+This endpoint must respond with a Stellar transaction signed by the server that has an invalid sequence number (0) and thus cannot be executed on the Stellar network. The client can then sign the transaction using standard Stellar libraries and submit it to [`token`](#token) endpoint to prove that they control their account. This approach is compatible with hardware wallets such as Ledger. The client must also verify the server's signature to be sure the challenge is signed by the `SIGNING_KEY` from the server's [`stellar.toml`](sep-0001.md), and the home domain in the challenge is the domain from the discovery flow.
 
 
 #### Request
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -27,6 +27,7 @@ The discovery flow is as follows:
 
 1. The client retrieves the [SEP-1 stellar.toml](sep-0001.md) from the home domain. The home domain hosts the stellar.toml.
 1. The client looks up the `WEB_AUTH_ENDPOINT` value from the [SEP-1 stellar.toml](sep-0001.md).
+1. The client looks up the `SIGNING_KEY` value from the [SEP-1 stellar.toml](sep-0001.md).
 
 The authentication flow is as follows:
 1. The client obtains a unique [`challenge`](#challenge) from the `WEB_AUTH_ENDPOINT`, which is represented as specially formed Stellar transaction
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -32,7 +32,7 @@ The authentication flow is as follows:
 1. The client obtains a unique [`challenge`](#challenge) from the `WEB_AUTH_ENDPOINT`, which is represented as specially formed Stellar transaction
 1. The client verifies that the transaction has an invalid sequence number 0.  This is extremely important to ensure the transaction isn't malicious.
 1. The client verifies that the transaction is signed by the `SIGNING_KEY` specified by the requested service's [SEP-1 stellar.toml](sep-0001.md).
-1. The client verifies that the transaction has a single Manage Data operation with its source account set to the user's account and value set to a nonce value. The client verifies the home domain included is the domain from the first step of the discovery flow.
+1. The client verifies that the transaction has a single Manage Data operation with its source account set to the user's account and value set to a nonce value. The client verifies the home domain included is the domain from the first step.
 1. The client verifies that if the transaction has other Manage Data operations they all have their source accounts set to the the server's account.
 1. The client signs the transaction using the secret key(s) of signers for the user's Stellar account
 1. The client submits the signed challenge back to the server using [`token`](#token) endpoint
@@ -92,7 +92,7 @@ Request Parameters:
 Name      | Type          | Description
 ----------|---------------|------------
 `account` | `G...` string | The stellar account that the wallet wishes to authenticate with the server
-`home_domain` | string | (optional) The home domain the client would like to authenticate for. Servers that generate tokens for multiple home domains can use this parameter to create valid tokens for each home domain. If not provided by the client, the server should assume a default for backwards compatibility with older clients.
+`home_domain` | string | (optional) The home domain the client would like to authenticate for. Servers that generate tokens for multiple home domains can use this parameter to identify which home domain the client hopes to authenticate for. If not provided by the client, the server should assume a default for backwards compatibility with older clients.
 
 Example:
 
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -18,10 +18,10 @@ This SEP defines the standard way for clients such as wallets or exchanges to cr
 
 This protocol is a variation of mutual challenge-response, which uses Stellar transactions to encode challenges and responses.
 
-It involves three parties:
-- A home domain hosting a [SEP-1 stellar.toml](sep-0001.md) containing a `WEB_AUTH_ENDPOINT` value.
-- A web auth endpoint providing the GET and POST operations discussed in this document.
-- A client who will authenticate with a Stellar account.
+It involves three components:
+- A domain hosting a [SEP-1 stellar.toml](sep-0001.md) containing a `WEB_AUTH_ENDPOINT` and `SIGNING_KEY`, referred to as the home domain.
+- An endpoint providing the GET and POST operations discussed in this document.
+- A client who will authenticate using a Stellar account.
 
 The discovery flow is as follows:
 
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -30,7 +30,7 @@ The authentication flow is as follows:
 1. The client obtains a unique [`challenge`](#challenge) from the `WEB_AUTH_ENDPOINT`, which is represented as specially formed Stellar transaction
 1. The client verifies that the transaction has an invalid sequence number 0.  This is extremely important to ensure the transaction isn't malicious.
 1. The client verifies that the transaction is signed by the `SIGNING_KEY` specified by the requested service's [SEP-1 stellar.toml](sep-0001.md).
-1. The client verifies that the transaction has a single Manage Data operation with its source account set to the user's account and value set to a nonce value. The client verifies the home domain included is the domain from the first step.
+1. The client verifies that the transaction has a single Manage Data operation with its source account set to the user's account and value set to a nonce value. The client verifies the home domain included is the domain from the first step.<sup>[1](#note1)</sup>
 1. The client verifies that if the transaction has other Manage Data operations they all have their source accounts set to the the server's account.
 1. The client signs the transaction using the secret key(s) of signers for the user's Stellar account
 1. The client submits the signed challenge back to the server using [`token`](#token) endpoint
@@ -58,7 +58,7 @@ The flow achieves several things:
 * The server can choose to authenticate accounts that do not yet exist
 
 Notes:
-1. <a name="note1"></a>Some use cases of SEP-10 may not involve a home domain. See the relevant notes throughout.
+1. <a name="note1"></a>Some use cases of SEP-10 may not involve a home domain hosting a stellar.toml. If a client is connecting directly to a SEP-10 serverit must still verify the home domain to ensure it only signs challenges issued by the server that the client intends to authenticate.
 
 ## Authentication Endpoint
 
@@ -81,7 +81,6 @@ In order for browsers-based wallets to validate the CORS headers, as [specified 
 
 This endpoint must respond with a Stellar transaction signed by the server that has an invalid sequence number (0) and thus cannot be executed on the Stellar network. The client can then sign the transaction using standard Stellar libraries and submit it to [`token`](#token) endpoint to prove that they control their account. This approach is compatible with hardware wallets such as Ledger. The client must also verify the server's signature to be sure the challenge is signed by the `SIGNING_KEY` from the server's [`stellar.toml`](sep-0001.md), and the home domain in the challenge is the domain of the domain hosting the [`stellar.toml`](sep-0001.md).
 
-If a client is interacting directly with a SEP-10 server without using a stellar.toml, it must still verify the home domain to ensure it only signs challenges issued by the server that the client intends to authenticate.
 
 #### Request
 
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0010.md
```diff
@@ -18,12 +18,18 @@ This SEP defines the standard way for clients such as wallets or exchanges to cr
...
+1. The client obtains the `WEB_AUTH_ENDPOINT` for the service by getting the [SEP-1 stellar.toml](sep-0001.md) from the services domain, referred to as the home domain. The home domain hosts the stellar.toml.
+1. The client obtains a unique [`challenge`](#challenge) from the `WEB_AUTH_ENDPOINT`, which is represented as specially formed Stellar transaction
 1. The client verifies that the transaction has an invalid sequence number 0.  This is extremely important to ensure the transaction isn't malicious.
 1. The client verifies that the transaction is signed by the `SIGNING_KEY` specified by the requested service's [SEP-1 stellar.toml](sep-0001.md).
-1. The client verifies that the transaction has a single Manage Data operation with its source account set to the user's account and value set to a nonce value. The client ignores the home domain included.
+1. The client verifies that the transaction has a single Manage Data operation with its source account set to the user's account and value set to a nonce value. The client verifies the home domain included is the domain from the first step.
```
JakeUrban, line 32: > the first step

are you referring to this?

> 1. The client obtains a unique [`challenge`](#challenge), which is represented as specially formed Stellar transaction

This makes it seem like the domain should be the WEB_AUTH_ENDPOINT, not the SEP-1 stellar.toml URL's hostname.

JakeUrban (commented):
One comment, otherwise LGTM. Well worded.

ecosystem/sep-0010.md
leighmcculloch: The first step was changed to be: 
>1. The client obtains the `WEB_AUTH_ENDPOINT` for the service by getting the [SEP-1 stellar.toml](sep-0001.md) from **the services domain, referred to as the home domain**. The home domain hosts the stellar.toml.

Maybe I can make this clearer. Suggestions?

JakeUrban: Oh no I was mistaken and thought the line you were removing was still present. This looks good.

leighmcculloch: I think it could be clearer. It's confusing the first step talks about the endpoint first rather than the stellar.toml. I've made a change in 0c4f2697d50b84e5a5a577d67e46ac5b5959b097 to make it clearer.

ecosystem/sep-0010.md
```diff
@@ -18,12 +18,21 @@ This SEP defines the standard way for clients such as wallets or exchanges to cr
...
+1. The client looks up the `WEB_AUTH_ENDPOINT` value from the [SEP-1 stellar.toml](sep-0001.md).
 
-1. The client obtains a unique [`challenge`](#challenge), which is represented as specially formed Stellar transaction
+The authentication flow is as follows:
+1. The client obtains a unique [`challenge`](#challenge) from the `WEB_AUTH_ENDPOINT`, which is represented as specially formed Stellar transaction
 1. The client verifies that the transaction has an invalid sequence number 0.  This is extremely important to ensure the transaction isn't malicious.
 1. The client verifies that the transaction is signed by the `SIGNING_KEY` specified by the requested service's [SEP-1 stellar.toml](sep-0001.md).
-1. The client verifies that the transaction has a single Manage Data operation with its source account set to the user's account and value set to a nonce value. The client ignores the home domain included.
+1. The client verifies that the transaction has a single Manage Data operation with its source account set to the user's account and value set to a nonce value. The client verifies the home domain included is the domain from the first step.
```
accordeiro, line 35: I think this should be:

```suggestion
1. The client verifies that the transaction has a single Manage Data operation with its source account set to the user's account and value set to a nonce value. The client verifies the home domain included is the domain from the first step of the discovery flow.
```

accordeiro (approved):
Looks great! Thanks for the quick iterations offline :)

ecosystem/sep-0010.md
JakeUrban: Now that there are multiple lists I want to reiterate that `the first step` should be replaced with a better reference.

ecosystem/sep-0010.md
```diff
@@ -82,7 +92,7 @@ Request Parameters:
 Name      | Type          | Description
 ----------|---------------|------------
 `account` | `G...` string | The stellar account that the wallet wishes to authenticate with the server
-`home_domain` | string | (optional) The [fully qualified domain name](https://en.wikipedia.org/wiki/Fully_qualified_domain_name) of the service requiring authentication. SEP-10 signing servers used to generate tokens for multiple different web services can use this parameter to create valid tokens for each service. If not provided by the client, a default should be used.
+`home_domain` | string | (optional) The home domain the client would like to authenticate for. Servers that generate tokens for multiple home domains can use this parameter to identify which home domain the client hopes to authenticate for. If not provided by the client, the server should assume a default for backwards compatibility with older clients.
```
JakeUrban, line 95: The only issue I have with this phrasing is that clients don't authenticate _for_ a home domain, I know what you're trying to say but it is a bit confusing because you're authenticating for a service on the TOML, not the home domain.

ecosystem/sep-0010.md
```diff
@@ -100,7 +110,7 @@ On success the endpoint must return `200 OK` HTTP status code and a JSON object
   * time bounds: `{min: now(), max: now() + 900 }` (we recommend expiration of 15 minutes to give user time to sign transaction)
   * operations:
     * `manage_data(source: client_account, key: '<home domain> auth', value: random_nonce())`
-      * The value of key is the fully qualified domain name of the web service requiring authentication followed by `auth`. It can be at most 64 characters.
+      * The value of key is the home domain the challenge is providing authentication for, followed by `auth`. It can be at most 64 characters.
```
JakeUrban, line 113: Same phrasing issue here.

JakeUrban (changes requested):
Theres some ambiguity in the language added around the relationship between the home domain and authentication.

ecosystem/sep-0010.md
leighmcculloch: I think both can be true. The client authenticates for the home domain and that authentication can be used with services referenced on the home domain. Is that accurate?

JakeUrban: That latter statement is accurate, but we shouldn't try to shorthand that full explanation by saying you authenticate for a home domain. The shorthand used doesn't imply the full explanation easily.

JakeUrban: When I read that phrase I think, "wait, I'm authenticating for a JWT to use with a Stellar service, what does the TOML home domain have to do with authentication?"

ecosystem/sep-0010.md
leighmcculloch: 💯 How is this? eb00ba8

ecosystem/sep-0010.md
leighmcculloch: I think the answer to the question about what odes the home domain have to do with it is answered further above in the abstract section where we now clearly call out the discovery flow.

How does this subtle change in the language sound to you? b22afa9

JakeUrban: I like that distinction, since the SDK's are in fact verifying the home domain, which is authenticating in a way.

JakeUrban (approved):
LGTM

ecosystem/sep-0010.md
```diff
@@ -18,12 +18,21 @@ This SEP defines the standard way for clients such as wallets or exchanges to cr
...
+The discovery flow is as follows:
+
+1. The client retrieves the [SEP-1 stellar.toml](sep-0001.md) from the home domain. The home domain hosts the stellar.toml.
+1. The client looks up the `WEB_AUTH_ENDPOINT` value from the [SEP-1 stellar.toml](sep-0001.md).
 
-1. The client obtains a unique [`challenge`](#challenge), which is represented as specially formed Stellar transaction
+The authentication flow is as follows:
+1. The client obtains a unique [`challenge`](#challenge) from the `WEB_AUTH_ENDPOINT`, which is represented as specially formed Stellar transaction
 1. The client verifies that the transaction has an invalid sequence number 0.  This is extremely important to ensure the transaction isn't malicious.
 1. The client verifies that the transaction is signed by the `SIGNING_KEY` specified by the requested service's [SEP-1 stellar.toml](sep-0001.md).
```
nebolsin, line 34:
```suggestion
1. The client verifies that the transaction is signed by the `SIGNING_KEY` obtained through discovery flow.
  * If this verification fails, client should re-execute the discovery flow and attempts to resume the authentication flow with the same challenge and a freshly retrieved `SIGNING_KEY`. This is necessary to support seamless key rotation for web services. If verification fails the second time, client must not continue the authentication and should present to the user (or write into the log, in case of machine-to-machine authentication) a warning about potential spoofing attack.
```

ecosystem/sep-0010.md
```diff
@@ -18,12 +18,21 @@ This SEP defines the standard way for clients such as wallets or exchanges to cr
...
-The authentication flow is as follows:
+It involves three parties:
+- A home domain hosting a [SEP-1 stellar.toml](sep-0001.md) containing a `WEB_AUTH_ENDPOINT` value.
+- A web auth endpoint providing the GET and POST operations discussed in this document.
+- A client who will authenticate with a Stellar account.
+
+The discovery flow is as follows:
+
+1. The client retrieves the [SEP-1 stellar.toml](sep-0001.md) from the home domain. The home domain hosts the stellar.toml.
+1. The client looks up the `WEB_AUTH_ENDPOINT` value from the [SEP-1 stellar.toml](sep-0001.md).
```
nebolsin, line 29:
```suggestion
1. The client looks up and notes the `WEB_AUTH_ENDPOINT` and `SIGNING_KEY` values from the [SEP-1 stellar.toml](sep-0001.md). If either of those values is missing, client must not continue with the authentication.
```

nebolsin (changes requested):
Overall it looks good :+1: Added a few comments and suggestions to eliminate ambiguity and also to allow future protocol extension by adding more operations to the challenge tx without breaking compatibility.

ecosystem/sep-0010.md
```diff
@@ -18,12 +18,21 @@ This SEP defines the standard way for clients such as wallets or exchanges to cr
...
 
-1. The client obtains a unique [`challenge`](#challenge), which is represented as specially formed Stellar transaction
+The authentication flow is as follows:
+1. The client obtains a unique [`challenge`](#challenge) from the `WEB_AUTH_ENDPOINT`, which is represented as specially formed Stellar transaction
 1. The client verifies that the transaction has an invalid sequence number 0.  This is extremely important to ensure the transaction isn't malicious.
 1. The client verifies that the transaction is signed by the `SIGNING_KEY` specified by the requested service's [SEP-1 stellar.toml](sep-0001.md).
-1. The client verifies that the transaction has a single Manage Data operation with its source account set to the user's account and value set to a nonce value. The client ignores the home domain included.
+1. The client verifies that the transaction has a single Manage Data operation with its source account set to the user's account and value set to a nonce value. The client verifies the home domain included is the home domain from the discovery flow.
```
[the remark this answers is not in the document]
leighmcculloch: In SEP-10 v2.1.0 the operations were relaxed to allow any number of operations greater than one, so I think this is already achieved. See https://github.com/stellar/stellar-protocol/commit/5acf11be9837969ec883c2f9fd7cd3201728ff07.

Let me know what I'm missing if this isn't already addressed sufficiently in v2.1.0.

ecosystem/sep-0010.md
leighmcculloch: I think adding this requirement is probably not going to change clients/servers. If a `WEB_AUTH_ENDPOINT` is not present the following steps are not possible so we do not need to call it out. If a `SIGNING_KEY` is not present and a client chose to continue the rest of the flow is unlikely to fail. I think the signing key verification being a hard requirement is something we need to revisit and so I'm hesitant to write it into more areas of the SEP unnecessarily.

ecosystem/sep-0010.md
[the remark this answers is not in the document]
nebolsin: @leighmcculloch Apparently I got confused but that might indicate that there's a room for improvement here. The way I read it was like two separate predicates on transaction and operation: 

> The client verifies that the transaction has a single Manage Data operation. This operation has a source account set to... => `tx.operations.size == 1 & valid_auth_op?(tx.operations[0])`.

And I guess what it suppose to mean is actually a single predicate

> The client verifies that `exactly one operation in the transaction is a Manage Data operation with its source account set to...`" => `tx.operations.find {|op| valid_auth_op?(op)}.size == 1`

Still I don't think it is clear how the client should proceed with the verification, it looks like it is expected to perform a search through transaction operations (which by itself is an extra complication, I think we can safely require that the auth operation is the first one), but the search condition a bit is misleading. Reading the current text exactly, it appears the client has to find operations with `source_account == account && value == nonce`, ensure that exactly one operation is found, and then verify that this operation contains correct home domain. But client doesn't know anything about the nonce and should only care that it exists, and the correct predicate is in fact `source_account == account && name == "#{home_domain} auth"`. Overall, it doesn't seem like the spec in the current edition can be directly translated to the code.

Does that makes sense, or am I missing something?

PS: Just realised that in the change suggested above I forgot about ` auth` suffix, so I updated my comment.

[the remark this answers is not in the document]
nebolsin: Another point I forgot to mention is about the next step in the flow, which requires other operations in the transaction to have a) Manage Data type; and b) have the signer set to the server acc. It is now clear that this condition is necessary because of the "find the only operation which has user as source account" semantics introduced in 2.1.0, but if we switch to the approach I'm suggesting "first operation in the transaction must be the auth op", then both those restrictions on the extra operations can be lifted, which I think is great from future extensibility standpoint. Existing clients would only care about the first op in the transactions, and just ignore the rest, if there're any.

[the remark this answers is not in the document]
leighmcculloch: I agree, we can do some things to improve the steps as described in the abstract, although this behavior is not something changing in v3 and is something described by v2.1.0 so we should improve it independently of the v2 vs v3. All SDKs I've reviewed implement the auth op as the first op as that is how the operation is represented further down in the SEP where the first op is always the auth op and zero or more ops are allowed to follow:
https://github.com/stellar/stellar-protocol/blob/a235c6366e0eef9c3f807222b1d3a8abad10405e/ecosystem/sep-0010.md#L112-L116
I think we could resolve this gap in a separate PR making it clear in the abstract that the auth op is the first operation.

ecosystem/sep-0010.md
nebolsin: This change was primarily to localise in a single place all the data that client has to obtain about the server and emphasise that all this happens during discovery flow, so that we can easily reference it from key rotation support condition below (re-run the discover on failed signing key validation). I added the obvious halt condition just for the sake of completeness. I'm not sure I understand your concerns regarding hard requirement verification of the signing key, lets discuss it on keybase.

ecosystem/sep-0010.md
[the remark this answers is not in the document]
leighmcculloch: I am addressing this issue separately in https://github.com/stellar/stellar-protocol/pull/754.

ecosystem/sep-0010.md
```diff
@@ -18,12 +18,21 @@ This SEP defines the standard way for clients such as wallets or exchanges to cr
...
 
-The authentication flow is as follows:
+It involves three parties:
+- A home domain hosting a [SEP-1 stellar.toml](sep-0001.md) containing a `WEB_AUTH_ENDPOINT` value.
+- A web auth endpoint providing the GET and POST operations discussed in this document.
+- A client who will authenticate with a Stellar account.
+
+The discovery flow is as follows:
+
+1. The client retrieves the [SEP-1 stellar.toml](sep-0001.md) from the home domain. The home domain hosts the stellar.toml.
```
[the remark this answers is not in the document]
leighmcculloch: I took a look at applying this change here, however I believe this is unnecessary because to do so is to respecify SEP-1 which is linked here. SEP-1 describes how a client should retrieve a stellar.toml from a domain. https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0001.md#specification

ecosystem/sep-0010.md
leighmcculloch: I applied this change but missed the change included adding the additional discussion about key rotation. This PR is focused on adding home domain back and so I've removed the comments about how to do key rotation in bb7700e0bfbbe930f328eafe27f2390875efadf6. I think how to do key rotation is something we should discuss independently.

ecosystem/sep-0010.md
leighmcculloch: fdcfa0c b37dac3

ecosystem/sep-0010.md
```diff
@@ -18,12 +18,21 @@ This SEP defines the standard way for clients such as wallets or exchanges to cr
 
 This protocol is a variation of mutual challenge-response, which uses Stellar transactions to encode challenges and responses.
 
-The authentication flow is as follows:
+It involves three parties:
+- A home domain hosting a [SEP-1 stellar.toml](sep-0001.md) containing a `WEB_AUTH_ENDPOINT` value.
+- A web auth endpoint providing the GET and POST operations discussed in this document.
+- A client who will authenticate with a Stellar account.
```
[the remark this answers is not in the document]
leighmcculloch: I think to add the terms web service here is more prescriptive than we need to be. An authentication service or simply an endpoint on some other service is an implementation detail and a decision the SEP doesn't need to make. The intention of this three point list is to call out upfront what the three things are that are involved: You need a domain hosting a stellar.toml, an endpoint to provide the challenge and token, and a client who is going to talk to the first two.

I applied some changes to the sentences to make it clearer and removed the term parties.

Let me know if I'm misunderstanding the distinction you're seeing that needs to be here.

leighmcculloch:
@JakeUrban @accordeiro @nebolsin Thank you for diligently reviewing this change. I've made what I think are the final changes to get this version ready for merging. Please let me know if there is anything else, or anything important I have missed, otherwise I believe this will be merged next week.

ecosystem/sep-0010.md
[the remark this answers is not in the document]
nebolsin: I don't think SEP-1 defines what we need here, specifically there's no reference to RFC3986 => no resolution for `host:port` vs `host` ambiguity. SEP-1 starts with `Given the domain DOMAIN`, which is one step ahead of  where we are in this spec, and there's a good reason for that.  SEP-1 originated from a need to link Stellar account to some metadata which is stored and accessed off-chain. It is not explicitly mentioned in the specification section, but account de facto provides the entrypoint and root context for SEP-1 flow. As far as SEP-1 is concerned, DOMAIN is whatever we find in account's `home_domain` field.

On the other hand, with SEP-10 we want to adopt Stellar keypairs as an authentication method for traditional web, which means that we need to carefully define how to plug Stellar concepts into the well-established framework of existing standards, in particular [HTTP/1.1: Message Syntax and Routing [RFC7230]](https://tools.ietf.org/html/rfc7230) and [HTTP/1.1: Authentication [RFC7235]](https://tools.ietf.org/html/rfc7235). The disposition for SEP-10 is user agent (client) wants to obtain authentication credentials in order to request some representation of the protected resource from the origin server. Resources in HTTP are identified by URI, so one thing we know for sure is that at the start of SEP-10 client is aware of the target URI, but there's no notion of "home domain" or "domain". Without specification every client implementing SEP-10 has to figure out how to convert the URI of the protected resource into the `DOMAIN` input expected by SEP-1, and that's not a good situation from the compatibility angle.

There's also a security angle, because any loosely specified term like home domain in security boundary definition has a potential for all sorts of vulnerabilities and attacks. For example, RFC3986 goes into great details defining the calculation of URIs equality, which if implemented incorrectly may result in compatibility and/or security issues down the road. Anchoring this SEP to existing well-established internet standards increases the odds that developers will rely on the proper library implementations, instead of going with simple but potentially dangerous approach (like comparing URIs as strings).

Maybe I put too much emphasis on this and this SEP will probably never see the level of adoption where things like that will become important. But if we can eliminate potential ambiguity and make the spec a little more formal, why not take this opportunity?

ecosystem/sep-0010.md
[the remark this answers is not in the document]
JakeUrban: I agree with the intent of the suggestion, just two recommendations:
- swap `wallet` for `client` and I approve of this change. Wallets are a subset of SEP-10 clients.
- We need a clear distinction between the SEP-10 service and the home domain that has the TOML. When saying `web service which requires authentication`, it is not clear which one we're talking about. I'd change it to `web service hosting a [SEP-1 stellar.toml](sep-0001.md) file containing a `WEB_AUTH_ENDPOINT` at the standardized `/.well-known/stellar.toml` URL path.`

ecosystem/sep-0010.md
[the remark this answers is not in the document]
JakeUrban: +1000 this change, we need to clarify the home domains format so there aren't mismatches of the same service represented by slightly different home domain representations.

A couple things I want to note:
- Add `the` before `client`
- Are we should RFC-3986 is the standard we want to cite? We need to be very careful with this

JakeUrban:
Yes, this needs to be merged by Wednesday. I made a few comments on @nebolsin 's original comments.

leighmcculloch:
I'll revisit those comments on Tuesday afternoon, then let's re-evaluate merging on Wednesday. If we need a couple more days I think it is okay to move forward with SDK changes on Wednesday given that we're really just iterating on wording that has no functional impact on the interactions between a client and a server.

ecosystem/sep-0010.md
```diff
@@ -18,12 +18,21 @@ This SEP defines the standard way for clients such as wallets or exchanges to cr
...
 
-1. The client obtains a unique [`challenge`](#challenge), which is represented as specially formed Stellar transaction
+The authentication flow is as follows:
+1. The client obtains a unique [`challenge`](#challenge) from the `WEB_AUTH_ENDPOINT`, which is represented as specially formed Stellar transaction
 1. The client verifies that the transaction has an invalid sequence number 0.  This is extremely important to ensure the transaction isn't malicious.
-1. The client verifies that the transaction is signed by the `SIGNING_KEY` specified by the requested service's [SEP-1 stellar.toml](sep-0001.md).
-1. The client verifies that the transaction has a single Manage Data operation with its source account set to the user's account and value set to a nonce value. The client ignores the home domain included.
+1. The client verifies that the transaction is signed by the `SIGNING_KEY` obtained through discovery flow.
+1. The client verifies that the transaction has a single Manage Data operation with its source account set to the user's account and value set to a nonce value. The client verifies the home domain included is the home domain from the discovery flow.
```
JakeUrban, line 35: I think this second sentence should be it's own bullet for readability.

ecosystem/sep-0010.md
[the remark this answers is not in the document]
leighmcculloch: >...one thing we know for sure is that at the start of SEP-10 client is aware of the target URI...
>...every client implementing SEP-10 has to figure out how to convert the URI of the protected resource into the `DOMAIN` input expected by SEP-1...

I'm not sure this is consistent with what I know of some wallets and their implementations. From the limited wallets I've seen if there's a fixed list of anchors within a wallet it's a domain, not a URL. Our stellar.toml clients in some SDKs take a domain as input and build the URL themselves. Examples of this are:
- Go SDK: https://github.com/stellar/go/blob/38aec89/clients/stellartoml/client.go#L14
- JS SDK: https://github.com/stellar/js-stellar-sdk/blob/6a508cb/src/stellar_toml_resolver.ts#L34-L37

I agree there is some room for improvement here, but I don't think it is detrimental to this changes success, and I still think the improvements you're specifying are improvements to SEP-1. SEP-1 is not limited to account-to-toml lookups, it also intended to generally specify how to go from a domain to a stellar.toml. You're correct that the domain is not sufficiently specified in regards to whether it includes a port or not. Practically this is unlikely to matter because production systems consistently use port 443 and HTTPS, or at least not enough for us to address urgently. None of the Stellar TOML clients in SDKs specify ports.

I think we can address this issue independent of SEP-10 v3.0.0.

I have changed the language at this line slightly in e732188 to more explicitly reference SEP-1, rather than casually reference it.

ecosystem/sep-0010.md
leighmcculloch: 👍  1d066da

leighmcculloch (commented):
@JakeUrban @Nebolsin Could you take a look at my comments and changes since your last 👀 ?

JakeUrban (approved):

JakeUrban:
I'm not particularly worried about SDK users figuring out how SEP-1 works and how to use the SDK for it. The logic is much simpler than some of the basic Stellar concepts like building transactions and submitting them to Horizon.

JakeUrban:
@leighmcculloch are we good to merge this?

PR #750, opened 2020-10-20
SEP-0036: Use sandwich transactions and simplify approval server responses

howardtw:
### What
This PR replaces the `Account Setup` section with the `Transaction Composition`. It also simplifies the response of an approval server so that it either accepts or rejects a transaction.

### Why
There are several drawbacks of configuring an asset issuer as a signer as mentioned in [CAP-18](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0018.md#motivation).
```
1. The owner of an account cannot unilaterally utilize assets unrelated to the issuer without the signature of that issuer
2. If multiple issuers want to become signers on a single account, then many signatures are potentially required for simple operations
3. Issuers cannot easily rotate keys, because their keys are on every account holding their asset
```

Having the approval server return a revised transaction is not straightforward as the approval server will have to understand the intention of the received transaction and propose another one. Revising the transaction also means that the original signatures are removed and the client will have to examine the transaction before signing it again. This will complicate the flows on the client side and make them inconsistent.

Finally the original `pending` and `action required` states can be combined with the `rejected` state as those states mean that the transaction was not accepted at the time of the submission. The clients can check the error message to see if the issuer needs more time or what actions to take in order to make the transaction in compliance.

### Notes
There are a few questions in the discussion section which will impact the design of the approval server. Any feedback will be welcome!

### DIFF
--- a/ecosystem/README.md
+++ b/ecosystem/README.md
@@ -66,7 +66,7 @@
 | [SEP-0033](sep-0033.md) | Identicons for Stellar Accounts | Lobstr.co, Gleb Pitsevich | Standard | Draft |
 | [SEP-0034](sep-0034.md) | Wallet Attribution for Anchors | Jake Urban and Leigh McCulloch | Standard | Final Comment Period |
 | [SEP-0035](sep-0035.md) | Operation IDs | Isaiah Turner, Debnil Sur, Scott Fleckenstein | Standard | Draft |
-
+| [SEP-0036](sep-0036.md) | Regulated Assets Transacting | Howard Chen | Standard | Draft |
 
 
 ### Rejected and Deprecated Proposals
@@ -74,6 +74,7 @@
 | Number | Title | Author | Track | Status |
 | --- | --- | --- | --- | --- |
 | [SEP-0003](sep-0003.md) | Compliance Protocol | SDF | Standard | Deprecated |
+| [SEP-0008](sep-0008.md) | Regulated Assets | Interstellar | Standard | Deprecated |
 | [SEP-0013](sep-0013.md) | DEPOSIT_SERVER proposal | @no, @ant, @manran, @pacngfar | Informational | Rejected |
 
 
--- a/ecosystem/README.md
+++ b/ecosystem/README.md
@@ -66,7 +66,7 @@
 | [SEP-0033](sep-0033.md) | Identicons for Stellar Accounts | Lobstr.co, Gleb Pitsevich | Standard | Draft |
 | [SEP-0034](sep-0034.md) | Wallet Attribution for Anchors | Jake Urban and Leigh McCulloch | Standard | Final Comment Period |
 | [SEP-0035](sep-0035.md) | Operation IDs | Isaiah Turner, Debnil Sur, Scott Fleckenstein | Standard | Draft |
-| [SEP-0036](sep-0036.md) | Regulated Assets Transacting | Howard Chen | Standard | Draft |
+| [SEP-0036](sep-0036.md) | Transaction Approval for Regulated Assets | Howard Chen | Standard | Draft |
 
 
 ### Rejected and Deprecated Proposals
@@ -74,7 +74,6 @@
 | Number | Title | Author | Track | Status |
 | --- | --- | --- | --- | --- |
 | [SEP-0003](sep-0003.md) | Compliance Protocol | SDF | Standard | Deprecated |
-| [SEP-0008](sep-0008.md) | Regulated Assets | Interstellar | Standard | Deprecated |
 | [SEP-0013](sep-0013.md) | DEPOSIT_SERVER proposal | @no, @ant, @manran, @pacngfar | Informational | Rejected |
 
 
--- a/ecosystem/sep-0008.md
+++ b/ecosystem/sep-0008.md
@@ -4,15 +4,10 @@
 SEP: 0008
 Title: Regulated Assets
 Author: Interstellar.com
-Status: Deprecated
+Status: Final
 Created: 2018-08-22
-Updated: 2020-10-26
-Version 1.0.1
+Version 1.0.0
 ```
-
-## Deprecation Notice
-
-For most use cases, we recommend the workflow specified in [SEP-0036](sep-0036.md) instead of SEP-0008.
 
 ## Simple Summary
 
--- a/ecosystem/sep-0008.md
+++ b/ecosystem/sep-0008.md
@@ -136,8 +136,7 @@ Ideally, no. Implementing Regulated Assets should only be used when absolutely n
 
 - We should define the request and the response format in this case.
 
-### How should the rule server obtain the KYC information in order to make a decision?
-
+### How should issuers obtain the KYC information from the wallet in order to make a decision?
 
 [Stellar.toml]: #stellartoml
 [Approval Server]: #approval-server
--- a/ecosystem/sep-0008.md
+++ b/ecosystem/sep-0008.md
@@ -3,12 +3,16 @@
 ```
 SEP: 0008
 Title: Regulated Assets
-Author: Interstellar.com, Howard Chen <@howardtw>
-Status: Active
+Author: Interstellar.com
+Status: Deprecated
 Created: 2018-08-22
-Updated: 2020-10-20
-Version 2.0.0
+Updated: 2020-10-26
+Version 1.0.1
 ```
+
+## Deprecation Notice
+
+For most use cases, we recommend the workflow specified in [SEP-0036](sep-0036.md) instead of SEP-0008.
 
 ## Simple Summary
 
@@ -23,22 +27,25 @@ Stellar aims to be an ideal platform for issuing securities. As such, it must pr
 ## Overview
 
 Implementing a Regulated Asset consists of these parts:
-- [Stellar.toml]: Issuers will advertise the existence of an Approval Service and the approval criteria via their stellar.toml file.
-- [Approval Server]: HTTP protocol for transaction signing.
-- [Transaction Composition]: Wallets will have to put operations that fully authorize the accounts for the asset, transact the asset, and deauthorize the accounts in a single transaction.
+- **Stellar.toml**: Issuers will advertise the existence of an Approval Service and the approval criteria via their stellar.toml file.
+- **Approval Server**: HTTP protocol for transaction signing.
+- **Account Setup**: Wallets will have to work with accounts that are controlled or at least partially controlled (via multisig) by asset issuers, rather than the wallet end user.<br/>
+**Note**: this step may not be required once the proposed [protocol change](https://github.com/stellar/stellar-protocol/issues/146) allowing for protocol-level per-transaction approval is implemented.
 
 ## Regulated Assets Approval Flow
 
 1. User creates and signs a transaction.
 2. Wallet resolves asset information and detects that it's a regulated asset.
 3. Wallet sends the transaction to the approval server.
-4. Approval server determines whether the transaction is compliant based on the compliance ruleset and looks up the current state of the ledger if necessary.
+4. Approval server determines whether the transaction is compliant based on the current state of the ledger, known pending transactions, and its compliance ruleset.
 5. Wallet handles approval response:
-    1. *Success?* Transaction has been approved and signed by issuer.
-    2. *Rejected?* Wallet should display given error message to the user.
+    1. *Success?* Transaction has been approved and signed by issuer. Submit to Horizon.
+    2. *Revised?* Transaction has been revised to be made compliant, and signed by the issuer. Wallet will show the changes to the user and ask them to sign the new transaction. Submit to Horizon once signed.
+    3. *Pending?* The issuer will need to asynchronously approve this transaction. Wallet should send the same transaction to the approval service after a specified timeout (Return to 3).
+    4. *Action Required?* Transaction requires a user action to be completed. Wallet will present the attached action url as a clickable link, along with the attached message and an option to resubmit once the action has been taken.
+    5. *Rejected?* Wallet should display given error message to the user.
 
 ## Stellar.toml
-
 Issuers will advertise the existence of an Approval Service through their stellar.toml file. This is done in the [[CURRENCIES]] section as different assets can have different requirements.
 
 ### Fields:
@@ -59,20 +66,23 @@ approval_criteria="The goat approval server will ensure that transactions are co
 ```
 
 ## Approval Server
-
-The transaction approval server receives a signed transaction and checks for compliance and signs it on success.
+The transaction approval server receives a signed transaction, checks for compliance and signs if possible.
 
 ### Possible results
 - Success: Transaction was found compliant and signed by service.
+- Revised: Transaction was modified to be compliant and signed by service. It should be re-signed by the client.
+- Pending: The issuer will asynchronously validate the transaction and respond later.
+- Action Required: The user must complete an action in order to have the transaction approved.
 - Rejected: Transaction was rejected.
 
 ### Request
 
-Transaction approval requests are executed by submitting an `HTTP POST` to `approval_server` with `Content-Type=application/x-www-form-urlencoded` and a single `tx` parameter. `tx` value is a base64 encoded transaction envelope XDR signed by the user. This is the transaction that will be tested for compliance and signed on success.
+Transaction approval requests are executed by submitting an `HTTP POST` to `approval_server` with `Content-Type=application/x-www-form-urlencoded` and a single `tx` parameter. `tx` value is a base64 encoded Transaction Envelope XDR signed by the user. This is the transaction that will be tested for compliance and signed if possible.
 
 ### Responses
 
 HTTP responses have an `application/json` encoded body. All responses will contain, at least, a top level `status` parameter that indicates the type of the result.
+
 
 #### Success Response
 
@@ -83,62 +93,79 @@ Parameters:
 Name | Data Type | Description
 -----|-----------|------------
 status|string|`"success"`
-tx|string|Transaction envelope XDR, base64 encoded. This transaction will have the original signature(s) from the request as well as the asset issuer's signature.
+tx|string|Transaction Envelope XDR, Base64 encoded. This transaction will have both the original signature(s) from the request, as well as an additional issuer signature.
 message|string|A human readable string containing information to pass on to the user (optional).
+
+#### Revised Response
+
+A Revised response will have a `200` HTTP status code and `revised` as the `status` value. It means that the transaction was revised to be made compliant. The user should examine and resign the transaction.
+
+Parameters:
+
+Name | Data Type | Description
+-----|-----------|------------
+status|string|`"revised"`
+tx|string|Transaction Envelope XDR, Base64 encoded. This transaction is a revised compliant version of the original request transation, signed by the issuer.
+message|string|A human readable string explaining the modifications made to the transaction to make it compliant.
+
+#### Pending Response
+
+A Pending response will have a `200` HTTP status code and `pending` as the `status` value. It means that the issuer needs to asynchronously validate the transaction. The user should resubmit the transaction after a given timeout.
+
+Parameters:
+
+Name | Data Type | Description
+-----|-----------|------------
+status|string|`"pending"`
+timeout|integer|Number of milliseconds to wait before submitting the same transaction again.
+message|string|A human readable string containing information to pass on to the user (optional).
+
+#### Action Required Response
+
+An Action Required response will have a `200` HTTP status code and `action_required` as the `status` value. It means that the user must complete an action before this transaction can be approved. The approval service will provide a URL that facilitates the action. Upon completion, the user will resubmit the transaction.
+
+Parameters:
+
+Name | Data Type | Description
+-----|-----------|------------
+status|string|`"action_required"`
+message|string|A human readable string containing information regarding the action required.
+action_url|string|A URL that allows the user to complete the actions required to have the transaction approved.
 
 #### Rejected Response
 
-A Rejected response will have a `400` HTTP status code and `rejected` as the `status` value. It means that the transaction is not compliant and thus not signed by the server.
-
-The error message should explain why the transaction is not compliant and advice further actions for the client if possible. If the issuer needs to asynchronously validate the transaction, the adviced wait time should be included in the error message if possible so the client would know when to retry.
+A Rejected response will have a `400` HTTP status code and `rejected` as the `status` value. It means that the transaction is not compliant and could not be revised to be made compliant.
 
 Parameters:
 
 Name | Data Type | Description
 -----|-----------|------------
 status|string|`"rejected"`
-error|string|A human readable string explaining why the transaction is not compliant.
+error|string|A human readable string explaining why the transaction is not compliant and could not be made compliant.
 
-## Transaction Composition
+### Best practices
 
-Wallets will have to put operations that fully authorize the accounts for the asset, transact the asset, and deauthorize the accounts in a single transaction. In other words, wallets can build transactions where a single operation, which has been approved by the issuer, can be run. If wallets compose transactions this way, there will be no point where accounts have open, authorized trustlines to run unapproved operations because of transaction atomocity.
+- If a transaction was revised, ALWAYS explain the changes that were made to a transaction through the `message` parameter.
+- Core operations shouldn't be modified as that can be confusing and misleading. For example, if the user wishes to put an offer for 1000 GOATS but due to velocity limits they can only put an offer for 500 GOATS, it is better to error with a message than to change the amount.
+- Adding an upper timebound to a transaction can help the issuer ensure that their view of the world does not get out of sync.
+- Issuers can enforce additional fees by adding additional operations. For example, any transaction involving goats, will also send 0.1 GOAT to Boris’ account.
+- Once a transaction has been signed, the issuer should mark it as pending and take it into account, as long as it hasn’t been timed out, so that they can have a accurate view of the world.
 
-Depending on whether issuers want to allow the accounts to maintain offers, issuers can leave the accounts in the `AUTHORIZE_TO_MAINTAIN_LIABILITIES_FLAG` state or completely deauthorize the accounts when completing the operation. See [CAP-18] for the fine-grained control of authorization.
+## Account Setup
 
-Here is an example of the transaction:
+Implementing Regulated Assets requires the participating accounts to be “managed” by the issuer. This is achieved by having the issuer be a co-signer on the account, such that a medium threshold operation cannot be submitted without the issuer’s approval.
 
-```
-Operation 1: Issuer uses AllowTrust to fully authorize account A, asset X
-Operation 2: Issuer uses AllowTrust to fully authorize account B, asset X
-Operation 3: Payment from A to B
-Operation 4: Issuer uses AllowTrust to set account B, asset X to AUTHORIZED_TO_MAINTAIN_LIABILITIES_FLAG state
-Operation 5: Issuer uses AllowTrust to set account A, asset X to AUTHORIZED_TO_MAINTAIN_LIABILITIES_FLAG state
-```
+In the future, this requirement can be replaced by introducing protocol level [CoSigned assets](https://github.com/stellar/stellar-protocol/issues/146).
 
 ## Discussion
 
 ### Should my asset be a regulated asset?
 
 Ideally, no. Implementing Regulated Assets should only be used when absolutely necessary, such as for regulatory reasons. It comes with a substantial operational overhead, added complexity, and burden for users. Issuers should only go down this route if it is absolutely required.
+Alternatively, some other available options are to utilize [SEP006](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0006.md) in order to perform KYC on deposit and withdraw, and/or use [AUTHORIZATION_REQUIRED](https://www.stellar.org/developers/guides/issuing-assets.html#requiring-or-revoking-authorization) which allows issuers to authorize specific accounts to hold their issued assets.
 
-### Why doesn’t the approval server submit transactions to the network?
+### Why doesn’t the approval service submit transactions to the network?
 
 - Separation of concerns between signing transactions and submitting them.
 - Transactions might require more signatures from further regulated asset issuers.
-- Wallets can handle the failure from transaction submission based on the type of errors and reconstruct the transaction.
 
-### Should the ruleset simply sit in the approval server?
-
-- Issuers can configure the ruleset with server configs.
-- The server can read the ruleset from a file and we can format the file so that it has a section about ledger state under Horizon.
-
-### Should the approval server make an API call to the issuer's server for rule checking if any?
-
-- We should define the request and the response format in this case.
-
-### How should issuers obtain the KYC information from the wallet in order to make a decision?
-
-[Stellar.toml]: #stellartoml
-[Approval Server]: #approval-server
-[Transaction Composition]: #transaction-composition
-[CAP-18]: https://github.com/stellar/stellar-protocol/blob/master/core/cap-0018.md
--- a/ecosystem/sep-0008.md
+++ b/ecosystem/sep-0008.md
@@ -32,7 +32,7 @@ Implementing a Regulated Asset consists of these parts:
 1. User creates and signs a transaction.
 2. Wallet resolves asset information and detects that it's a regulated asset.
 3. Wallet sends the transaction to the approval server.
-4. Approval server determines whether the transaction is compliant based on the compliance ruleset and look up the current state of the ledger if necessary.
+4. Approval server determines whether the transaction is compliant based on the compliance ruleset and looks up the current state of the ledger if necessary.
 5. Wallet handles approval response:
     1. *Success?* Transaction has been approved and signed by issuer.
     2. *Rejected?* Wallet should display given error message to the user.
@@ -68,7 +68,7 @@ The transaction approval server receives a signed transaction and checks for com
 
 ### Request
 
-Transaction approval requests are executed by submitting an `HTTP POST` to `approval_server` with `Content-Type=application/x-www-form-urlencoded` and a single `tx` parameter. `tx` value is a base64 encoded Transaction Envelope XDR signed by the user. This is the transaction that will be tested for compliance and signed on success.
+Transaction approval requests are executed by submitting an `HTTP POST` to `approval_server` with `Content-Type=application/x-www-form-urlencoded` and a single `tx` parameter. `tx` value is a base64 encoded transaction envelope XDR signed by the user. This is the transaction that will be tested for compliance and signed on success.
 
 ### Responses
 
@@ -83,12 +83,14 @@ Parameters:
 Name | Data Type | Description
 -----|-----------|------------
 status|string|`"success"`
-tx|string|Transaction Envelope XDR, base64 encoded. This transaction will have the original signature(s) from the request as well as the asset issuer's signature.
+tx|string|Transaction envelope XDR, base64 encoded. This transaction will have the original signature(s) from the request as well as the asset issuer's signature.
 message|string|A human readable string containing information to pass on to the user (optional).
 
 #### Rejected Response
 
-A Rejected response will have a `400` HTTP status code and `rejected` as the `status` value. It means that the transaction is not compliant and thus not signed by the server. The client should act accordingly based on the error message or retry if the issuer needs more time to validate the transaction.
+A Rejected response will have a `400` HTTP status code and `rejected` as the `status` value. It means that the transaction is not compliant and thus not signed by the server.
+
+The error message should explain why the transaction is not compliant and advice further actions for the client if possible. If the issuer needs to asynchronously validate the transaction, the adviced wait time should be included in the error message if possible so the client would know when to retry.
 
 Parameters:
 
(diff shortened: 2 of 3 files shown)


ecosystem/sep-0008.md
```diff
@@ -88,79 +83,62 @@ Parameters:
...
-Name | Data Type | Description
------|-----------|------------
-status|string|`"action_required"`
-message|string|A human readable string containing information regarding the action required.
-action_url|string|A URL that allows the user to complete the actions required to have the transaction approved.
-
 #### Rejected Response
 
-A Rejected response will have a `400` HTTP status code and `rejected` as the `status` value. It means that the transaction is not compliant and could not be revised to be made compliant.
+A Rejected response will have a `400` HTTP status code and `rejected` as the `status` value. It means that the transaction is not compliant and thus not signed by the server.
```
leighmcculloch, line 91: 💡 The status code used with the `rejected` status should probably be one of:
- `422` The request was well-formed but was unable to be followed due to semantic errors.
- `403` The request contained valid data and was understood by the server, but the server is refusing action.

`400` is inappropriate because the request was understood, and just not allowed or not appropriate.

ecosystem/sep-0008.md
```diff
@@ -88,79 +83,62 @@ Parameters:
...
+- Wallets can handle the failure from transaction submission based on the type of errors and reconstruct the transaction.
+
+### Should the ruleset simply sit in the approval server?
+
+- Issuers can configure the ruleset with server configs.
+- The server can read the ruleset from a file and we can format the file so that it has a section about ledger state under Horizon.
+
+### Should the approval server make an API call to the issuer's server for rule checking if any?
+
+- We should define the request and the response format in this case.
```
leighmcculloch, line 137: 💡 These discussion comments appear to be out-of-scope of the SEP, which I think should be focused on the interaction between the client and the service and the responsibilities of each. What goes on behind the scenes or how the service implements the servers responsibilities is unnecessary to constrain.

ecosystem/sep-0008.md
```diff
@@ -61,24 +59,21 @@ approval_criteria="The goat approval server will ensure that transactions are co
...
 - Success: Transaction was found compliant and signed by service.
-- Revised: Transaction was modified to be compliant and signed by service. It should be re-signed by the client.
-- Pending: The issuer will asynchronously validate the transaction and respond later.
-- Action Required: The user must complete an action in order to have the transaction approved.
 - Rejected: Transaction was rejected.
 
 ### Request
 
-Transaction approval requests are executed by submitting an `HTTP POST` to `approval_server` with `Content-Type=application/x-www-form-urlencoded` and a single `tx` parameter. `tx` value is a base64 encoded Transaction Envelope XDR signed by the user. This is the transaction that will be tested for compliance and signed if possible.
+Transaction approval requests are executed by submitting an `HTTP POST` to `approval_server` with `Content-Type=application/x-www-form-urlencoded` and a single `tx` parameter. `tx` value is a base64 encoded transaction envelope XDR signed by the user. This is the transaction that will be tested for compliance and signed on success.
```
leighmcculloch, line 71: 💡 I think we should make the request allowed to be either form urlencoded or JSON encoded. We do this in some other SEPs, e.g. SEP-10, SEP-30. The viewpoint from some people outside SDF that led to this pattern on SEP-10 was that it makes it easier to integrate with the APIs in some cases.

ecosystem/sep-0008.md
```diff
@@ -88,79 +83,62 @@ Parameters:
 Name | Data Type | Description
 -----|-----------|------------
 status|string|`"success"`
-tx|string|Transaction Envelope XDR, Base64 encoded. This transaction will have both the original signature(s) from the request, as well as an additional issuer signature.
+tx|string|Transaction envelope XDR, base64 encoded. This transaction will have the original signature(s) from the request as well as the asset issuer's signature.
```
leighmcculloch, line 86: 💡 If we are replacing this SEP I think we should rename `tx` to `transaction` because we don't use the tx abbreviation in other SEPs that surface APIs and consistency in naming can help reduce cognitive load and downstream consistency in SDKs and code. e.g. SEP-10, SEP-30, SEP-24.

ecosystem/sep-0008.md
```diff
@@ -88,79 +83,62 @@ Parameters:
...
-
-An Action Required response will have a `200` HTTP status code and `action_required` as the `status` value. It means that the user must complete an action before this transaction can be approved. The approval service will provide a URL that facilitates the action. Upon completion, the user will resubmit the transaction.
-
-Parameters:
-
-Name | Data Type | Description
------|-----------|------------
-status|string|`"action_required"`
-message|string|A human readable string containing information regarding the action required.
-action_url|string|A URL that allows the user to complete the actions required to have the transaction approved.
```
leighmcculloch, line 128: ❓ Are you sure an issuer would not need action required? This change appears to be focused on replacing the cosigning attribute of SEP-8's approach with the use of the allow trust sandwich, and removing the action required is independent of that.

leighmcculloch (commented):
Changing the SEP from Final to Active and making changes to the SEP is problematic because final communicates that no further changes will be made to the SEP.

There's also enough core fundamental concepts changing that I think this new proposal is worthy of being a new SEP.

I think you should create a new SEP. If SEP-8 is not-in-use and never took off, we could deprecate it.

---

Ignoring the issue of logistics above this new proposal looks very nice. 🎉 I made some suggestions inline and I'd like to take another look once you have a chance to address them. There is also one question.

JakeUrban:
I second @leighmcculloch 's idea to create a new SEP, but the changes look good.

ecosystem/sep-0008.md
howardtw: That's a good point! I think I will go with `403` because it suits it the best. The approval server understands the request and refuses to sign.

ecosystem/sep-0008.md
howardtw: I feel this can be an important part of the process for the approval server to determine whether the transaction is valid. However, I also think your point makes sense. This SEP-8 should only focus on the interaction between the client and the approval server. Do you think it's worth having another SEP for the interaction between the approval server and the issuer's rule checking server if we decide that this is the way to go compared to baking all the rules in the approval server?

ecosystem/sep-0008.md
howardtw: I'm not sure as I haven't been able to gather information about having which status will facilitate the issuers the most. I was simply thinking that this can be combined with the `rejected` status by putting the `action required` message in the error message.

Having said that, do you think changing the response would be a problem if this were to be done in a new SEP?

ecosystem/sep-0008.md
JakeUrban: I would actually strongly recommend `422`, the SDF's has overloaded the use of `403` is our SEPs. `403` --> `Permission Denied`, implying the client doesn't have the privilege of making the request.

Here, the client is submitting a request with two possible error cases:
- The HTTP request was invalid in some way, like missing request parameter or providing data that is invalid. If this is the case, we should return `400`
- The HTTP request was valid, but the server cannot honor the request due to it's internal state. This sounds like what we're doing here, and it should be `422`

howardtw:
@tomerweller what do you feel about creating a new SEP for this now? I know you were leaning toward just updating SEP-8.

ecosystem/sep-0008.md
leighmcculloch: If the approval server and issuer rule checking are hosted by the same entity, then I don't think that needs to be prescribed as there's no interoperability concerns and the issuer can build whatever they need is relevant to their use case, region, and concerns.

ecosystem/sep-0008.md
leighmcculloch: If we're doing this in a new SEP then I agree, don't add anything you don't know of needing. Just the fact we had it here suggests someone thought it was needed so it might be worth asking around. If there's no evidence though, yeah I'd leave it out in a new SEP.

ecosystem/sep-0008.md
howardtw: > If the approval server and issuer rule checking are hosted by the same entity, then I don't think that needs to be prescribed as there's no interoperability concerns.

It's true that we probably don't have to prescribe if they build everything on their own.

We are trying to figure out the way that issuers might it useful to host the approval server and use it for rule checking. If they find it more useful that the approval server can make an API call, then it's where the complication can come in because we might build a generalized approval server in our mono repo similar to `webauth` and `recoverysigner`. Without knowing the interface, it's hard to add that functionality.

howardtw: Thinking about it more, it might be very difficult to turn every single rule into implementation in the prototype server. I guess the intention of providing a prototype of the approval server might just be using it as an example.

tomerweller:
I suggested updating this SEP as it's not used in production right now so it's safe to introduce backward incompatible changes. With that said, I'm not offended by the idea of a new SEP.

tomerweller:
I would consider separating out the sandwich addition and the "simplification" into two separate PRs as the scope is different. 

Regarding simplification, you've removed a set of machine readable response codes in favor of undefined, non-deterministic error messages. This reduces the ability of a wallet to provide an optimized user experience. 

Examples:
- a `pending` response allows the wallet to mark the transaction as pending and auto-resubmit after the given timeout.
- an `action_required` response allows the wallet to redirect the user to a website to complete the action required. 
- a `revised` response allows the wallet to compare the transactions and show the user a diff or in a simple case just resign the new tx (example: if the anchor added a near future timebound)

Instead, you now suggest that the user reads a message and act accordingly. 

I'm not religious about these response codes, especially if key partners signal that they're not necessary. But I don't think we should slash features in the name of perceived simplicity.

leighmcculloch:
>this SEP as it's not used in production right now

If this is the case, and since the proposal is a broad change to the concepts and approach, I think SEP-8 should be deprecated and a new SEP created with this proposal. When anyone talks about this new proposal or SEP-8 it should not be confusing which approach is being discussed and to reuse the SEP number for a new proposal would do that.

+1 to the concerns on simplifying the responses. How do we validate if these responses are necessary?

howardtw:
> you've removed a set of machine readable response codes in favor of undefined, non-deterministic error messages.

That's a good point. I neglect the fact that there are `timeout integer` and `action_url string` in the responses of `pending` and `action_required` respectively where the wallet can act upon without the user's interpretation.

That main reason I made that change is that I think transactions with a status `pending` and `action_required`, can be considered as a `rejected` transaction. I'm not a fan of having a `revised` transaction returned given the reason I stated in the description.

Let me try to see what Securrency thinks about it. Without knowing what's really useful on the issuer's standpoint, it's hard to make a decision. 

For the time being, I think I will move the changes to a new SEP and simplify the response as it is in the PR. I can add the original statuses back if Securrency finds it useful or anyone in the organization thinks it will be useful. In fact, I can also add the `timeout` and `action_url` fields in the `rejected` status response too. 💡

shanzzam:
With a new SEP do you intend to pick what is needed from SEP-08 and add onto it for a new SEP, before deprecating SEP-08? 
Related: How do you deprecate a SEP that's Final as defined: "Final - A finalized SEP will not be changed aside from minor errata. For a proposal to be a candidate to be made Final it must be being used in live products."

leighmcculloch:
The definitions of Active and Final have not historically been followed, so SEPs created a little while ago may have progressed quicker to Final than newer SEPs have. In this case since @tomerweller confirmed this SEP is not used in production we can just change its status to deprecated. We could leave the PR open for a couple weeks if we wanted to leave it open for feedback or ask on stellar-dev mailing list, but that might not be necessary with this one as there is no discussion link referenced at the top of the SEP.

ecosystem/sep-0008.md
howardtw: https://github.com/stellar/stellar-protocol/pull/750/commits/92ce0df718a0ff9b483e441eac5fcd95b836d1d5

ecosystem/sep-0008.md
howardtw: https://github.com/stellar/stellar-protocol/pull/750/commits/a8d3bf64b67694086d4b7a2042e08a6b2ed5f265

ecosystem/sep-0008.md
howardtw: https://github.com/stellar/stellar-protocol/pull/750/commits/2254a830f9bb5fe107e4c166c21cb1f5d46ca42f

ecosystem/sep-0008.md
howardtw: https://github.com/stellar/stellar-protocol/pull/750/commits/0fe94d7a79cb6ddf04ae024fcd95e1f4890a4745

ecosystem/sep-0008.md
howardtw: Combine it with `rejected`: https://github.com/stellar/stellar-protocol/pull/750/commits/051a787a8c7cdb0204c7cf6683001e4bfba54e57

howardtw:
> With a new SEP do you intend to pick what is needed from SEP-08 and add onto it for a new SEP, before deprecating SEP-08?

@shanzzam That's right except I intend to do them in a single [step](https://github.com/stellar/stellar-protocol/pull/750/commits/b21f6a88ab43205386e8a896c9b87d667058407d). However, since the new SEP will be in `draft` state, I'm not sure whether it's a good practice to deprecate SEP-0008 before the successor becomes `active`. On the flip side, I can see that deprecating a soon-to-be-deprecated SEP earlier can prevent people from adopting an obsolete approach, and we can save ourselves the efforts to "update" those ones later on. Additionally, I'm not sure whether having a `final` SEP and a `draft` SEP on the same topic at the same time is a good idea because people can get confused by which one to follow. Finally I can see that we "actively" help our partner to adopt a `draft` SEP like SEP-0030. Does the status of a new SEP really matter when it comes to deprecating the predecessor? Thoughts @tomerweller @JakeUrban @leighmcculloch ? I don't think we have this process (replacing an existing SEP with a new SEP) documented in `README.md`. It will help if we can add a new section to address this.

howardtw:
@leighmcculloch @JakeUrban I have addressed your feedback and move the changes into a new SEP. Can you take another look?

ecosystem/sep-0036.md
```diff
@@ -0,0 +1,137 @@
...
+## Motivation
+
+Stellar aims to be an ideal platform for issuing securities. As such, it must provide tools to comply with various regulatory requirements. Regulation often requires asset issuers to monitor and approve every transaction involving the issued asset and to enforce an assortment of constraints. This SEP provides support for these requirements.
+
+## Overview
+
+Implementing a Regulated Asset consists of these parts:
+- [Stellar.toml]: Issuers will advertise the existence of an Approval Service and the approval criteria via their stellar.toml file.
+- [Approval Server]: HTTP protocol for transaction signing.
+- [Transaction Composition]: Wallets will have to put operations that fully authorize the accounts for the asset, transact the asset, and deauthorize the accounts in a single transaction.
```
JakeUrban, line 29: I think this is the place where we disambiguate between the operations the client actually wants and the operations that are necessary to allow it to happen. The client wants an operation to be executed and it needs to be preceded and followed up with operations that authorize/deauthorize the transaction.

 Also, its not clear if the Wallet/Client or Approval Server is the one to add the auth/deauth operations to the set of transaction that actually do what the client wants.

ecosystem/sep-0036.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Regulated Assets Approval Flow
+
+1. User creates and signs a transaction.
```
JakeUrban, line 33: The user doesn't do this, the wallet does for the user right?

ecosystem/sep-0036.md
```diff
@@ -0,0 +1,137 @@
...
+2. Wallet resolves asset information and detects that it's a regulated asset.
+3. Wallet sends the transaction to the approval server.
+4. Approval server determines whether the transaction is compliant based on the compliance ruleset and looks up the current state of the ledger if necessary.
+5. Wallet handles approval response:
+    1. *Success?* Transaction has been approved and signed by issuer.
+    2. *Rejected?* Wallet should display given error message to the user.
+
+## Stellar.toml
```
JakeUrban, line 41: This seems to be redundant, the same definition is used earlier

ecosystem/sep-0036.md
```diff
@@ -0,0 +1,137 @@
...
+
+Issuers will advertise the existence of an Approval Service through their stellar.toml file. This is done in the [[CURRENCIES]] section as different assets can have different requirements.
+
+### Fields:
```
JakeUrban, line 45: I would leave these definitions to SEP-1. The example is sufficient in my opinion.

ecosystem/sep-0036.md
```diff
@@ -0,0 +1,137 @@
...
+```toml
+[[CURRENCIES]]
+code="GOAT"
+issuer="GD5T6IPRNCKFOHQWT264YPKOZAWUMMZOLZBJ6BNQMUGPWGRLBK3U7ZNP"
+regulated=true
+approval_server="https://goat.io/tx_approve"
+approval_criteria="The goat approval server will ensure that transactions are compliant with NFO regulation"
+```
+
+## Approval Server
```
JakeUrban, line 62: Redundant

ecosystem/sep-0036.md
```diff
@@ -0,0 +1,137 @@
...
+
+The transaction approval server receives a signed transaction and checks for compliance and signs it on success.
+
+### Possible results
```
JakeUrban, line 66: Redundant

ecosystem/sep-0036.md
```diff
@@ -0,0 +1,137 @@
...
+- Success: Transaction was found compliant and signed by service.
+- Rejected: Transaction was rejected.
+
+### Request
```
JakeUrban, line 70: I would take a look at other SEPs and how they define endpoints. Most have a bullet list with links to each endpoint request/response definition. The specific HTTP method and URI are also often used in the title.

ecosystem/sep-0036.md
```diff
@@ -0,0 +1,137 @@
...
+-----|-----------|------------
+status|string|`"success"`
+transaction|string|Transaction envelope XDR, base64 encoded. This transaction will have the original signature(s) from the request as well as the asset issuer's signature.
+message|string|A human readable string containing information to pass on to the user (optional).
+
+#### Rejected Response
+
+A Rejected response will have a `422` HTTP status code and `rejected` as the `status` value. It means that the transaction is not compliant and thus not signed by the server.
+
+The error message should explain why the transaction is not compliant and advice further actions for the client if possible. If there are further actions that the user must complete before this transaction can be approved, the issuer may provide a URL via `action_url` to faciliate the actions. If the issuer needs to asynchronously validate the transaction, the adviced wait time should be included in the error message if possible so the client would know when to retry.
```
JakeUrban, line 94: I'm not a fan of a single human readable error message in this situation, especially if important pieces of info like a wait time is going to be included.

I don't quite understand the `if the issuer needs to asynchronously validate the transaction` situation, can you explain what they would be validating and how it's async?

Finally, it sounds like the `action_url` is similar to the `interactive_url` returned by SEP-24 endpoints. If it is, I would explicitly mention that this URL must provide an intuitive UX for the user to navigate. In SEP-24, I think we require the `Content-Type` to be `text/html`.

ecosystem/sep-0036.md
```diff
@@ -0,0 +1,137 @@
...
+
+Parameters:
+
+Name | Data Type | Description
+-----|-----------|------------
+status|string|`"rejected"`
+error|string|A human readable string explaining why the transaction is not compliant.
+action_url|string|A URL that allows the user to complete the actions required to have the transaction approved (optional).
+
+## Transaction Composition
```
JakeUrban, line 104: I think this section should be much higher, since it explains how this actually works. Maybe after `Example`?

ecosystem/sep-0036.md
```diff
@@ -0,0 +1,137 @@
...
+
+Wallets will have to put operations that fully authorize the accounts for the asset, transact the asset, and deauthorize the accounts in a single transaction. In other words, wallets can build transactions where a single operation, which has been approved by the issuer, can be run. If wallets compose transactions this way, there will be no point where accounts have open, authorized trustlines to run unapproved operations because of transaction atomocity.
+
+Depending on whether issuers want to allow the accounts to maintain offers, issuers can leave the accounts in the `AUTHORIZE_TO_MAINTAIN_LIABILITIES_FLAG` state or completely deauthorize the accounts when completing the operation. See [CAP-18] for the fine-grained control of authorization.
+
+Here is an example of the transaction:
```
JakeUrban, line 110: This is a bit confusing, you mention that the Wallet is the one to create the transaction but the descriptions of the operations make it seem like the issuer is creating it.

ecosystem/sep-0036.md
```diff
@@ -0,0 +1,137 @@
...
+Operation 5: Issuer uses AllowTrust to set account A, asset X to AUTHORIZED_TO_MAINTAIN_LIABILITIES_FLAG state
+```
+
+## Discussion
+
+### Should my asset be a regulated asset?
+
+Ideally, no. Implementing Regulated Assets should only be used when absolutely necessary, such as for regulatory reasons. It comes with a substantial operational overhead, added complexity, and burden for users. Issuers should only go down this route if it is absolutely required.
+
+### Why doesn’t the approval server submit transactions to the network?
```
JakeUrban, line 126: Eh, I feel like this was an internal question during design, but I'm not sure Wallets are going to be asking this question. Wallets already submit transactions to Stellar (for example, SEP-24 withdrawal payments), so I don't think it requires an explanation.

ecosystem/sep-0036.md
```diff
@@ -0,0 +1,137 @@
...
+
+- Separation of concerns between signing transactions and submitting them.
+- Transactions might require more signatures from further regulated asset issuers.
+- Wallets can handle the failure from transaction submission based on the type of errors and reconstruct the transaction.
+
+### How should issuers obtain the KYC information from the wallet in order to make a decision?
```
JakeUrban, line 132: The title asks a question the content doesn't answer.

JakeUrban (changes requested):
At a high level, I approve of the overall flow. My comments are mostly about SEP clarity/readability.

ecosystem/sep-0036.md
```diff
@@ -0,0 +1,137 @@
...
+Status: Draft
+Created: 2020-10-26
+Version 0.1.0
+```
+
+## Simple Summary
+
+Regulated Assets provide ecosystem support for assets that require an issuer’s approval (or a delegated third party’s approval, such as a licensed securities exchange) on a per-transaction basis. It standardizes the identification of such assets as well as defines the protocol for performing compliance checks and requesting issuer approval.
+
+This SEP is based on [SEP-0008](sep-0008.md), but the issuer no long has to a co-signer on the account holding the regulated asset. It also updates the API of the approval server to better align with other active SEPs.
```
leighmcculloch, line 16: I think you should drop this sentence. You can mention this in the PR and/or commit comment and beyond that this SEP should standalone.

ecosystem/sep-0036.md
```diff
@@ -0,0 +1,137 @@
...
+
+**Target Audience**: Asset issuers, issuance platforms, wallet developers, and exchanges
```
leighmcculloch, line 18: Can you make this a section `## Target Audience`?

ecosystem/sep-0036.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Motivation
+
+Stellar aims to be an ideal platform for issuing securities. As such, it must provide tools to comply with various regulatory requirements. Regulation often requires asset issuers to monitor and approve every transaction involving the issued asset and to enforce an assortment of constraints. This SEP provides support for these requirements.
+
+## Overview
+
+Implementing a Regulated Asset consists of these parts:
+- [Stellar.toml]: Issuers will advertise the existence of an Approval Service and the approval criteria via their stellar.toml file.
+- [Approval Server]: HTTP protocol for transaction signing.
```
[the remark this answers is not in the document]
leighmcculloch: HTTP Protocol is confusing because within the HTTP ecosystem protocols are a specific thing, they are the leading part of a URL. Instead I think it would be clearer to be specific:
```suggestion
- [Approval Server]: Endpoints defined in this SEP that a client use to sign a transaction that temporarily authorizes the client to operate with the asset.
```
+1 to the comments about phrasing consistency. I'd stick to each sentence describing what is on the left. It's simpler to understand.

ecosystem/sep-0036.md
leighmcculloch: I think these definitions belong here too. At the very least we need to list them, and if we're going to list them we may as well define them. IMO the fields and how they are used are owned by this SEP and the list in SEP-1 should just be a registry.

ecosystem/sep-0036.md
leighmcculloch: I think we also need examples for both the JSON and formurlencoded requests and a table for the request parameters like the response ones.

e.g. https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0030.md#endpoints

ecosystem/sep-0036.md
leighmcculloch: I think this needs to be called out because people have asked this question, who should submit the transaction, and it hasn't always been the wallets.

ecosystem/sep-0036.md
```diff
@@ -0,0 +1,137 @@
...
+
+### Should my asset be a regulated asset?
+
+Ideally, no. Implementing Regulated Assets should only be used when absolutely necessary, such as for regulatory reasons. It comes with a substantial operational overhead, added complexity, and burden for users. Issuers should only go down this route if it is absolutely required.
+
+### Why doesn’t the approval server submit transactions to the network?
+
+- Separation of concerns between signing transactions and submitting them.
+- Transactions might require more signatures from further regulated asset issuers.
+- Wallets can handle the failure from transaction submission based on the type of errors and reconstruct the transaction.
```
leighmcculloch, line 130: Also it is a scaling concern, it would hold up the server and require more infrastructure. The server has one job, to approve.

ecosystem/sep-0036.md
leighmcculloch: It's a good question though! It's not clear to me how KYC should take place, if an approver required it.

ecosystem/sep-0036.md
```diff
@@ -0,0 +1,137 @@
+## Preamble
+
+```
+SEP: 0036
+Title: Regulated Assets Transacting
```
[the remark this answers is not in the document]
leighmcculloch: I think this is a case where we are building a solution or one use case, but it may be relevant to others and we do not need to bind the definitions to terms like regulated assets. We are building a thin/small interface between wallets to get issuer approval.
```suggestion
Title: Transaction Approval for Assets Requiring Issuer Approval
```

ecosystem/sep-0036.md
howardtw: > Also, its not clear if the Wallet/Client or Approval Server is the one to add the auth/deauth operations to the set of transaction that actually do what the client wants.

How is it not clear which party is the one building the transaction? I thought this sentence states that `Wallets will have to put operations that...`. Got suggestions?

How about this?
```
Wallets will have to build and sign a transaction containing operations that fully authorize the accounts for the asset, transact the asset, and deauthorize the accounts.
```
By mentioning the **sign**, it should be clear that no party should modify the transaction.

ecosystem/sep-0036.md
howardtw: This section not only contains a short description but subsections like `Fields` and `Example`. The description is a little more descriptive than what's in the overview.

ecosystem/sep-0036.md
howardtw: This section contains subsections like `Request` and `Responses`.

ecosystem/sep-0036.md
howardtw: > I don't quite understand the if the issuer needs to asynchronously validate the transaction situation, can you explain what they would be validating and how it's async?

This is carried over from SEP-8. I don't know exactly how an issuer might do the validation to be honest.

Coupled with another comment
> I'm not a fan of a single human readable error message in this situation, especially if important pieces of info like a wait time is going to be included.

I could remove the mention about async validation and timeout or add a machine readable field `timeout` in the response. Which one do you prefer?

> In SEP-24, I think we require the Content-Type to be text/html.

Do you mean when an `action_url` is present, we want to use `text/html` instead of `application/json` as the content-type for the response? I can't find the requirement in SEP-24.

ecosystem/sep-0036.md
howardtw: Ah agree. I will rephrase them

ecosystem/sep-0036.md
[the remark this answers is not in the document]
howardtw: Because of what we say in `Simple Summary`, I feel it makes more sense to include `Regulated Assets` in the title. Hence, I came up with something that is the mix of your suggestions - `Transaction Approval for Regulated Assets` in https://github.com/stellar/stellar-protocol/pull/750/commits/8ec2d6d6dc55d5996a838ce4b2bbca1874f1c8d9

ecosystem/sep-0036.md
```diff
@@ -0,0 +1,137 @@
...
+Author: Howard Chen <@howardtw>
+Status: Draft
+Created: 2020-10-26
+Version 0.1.0
+```
+
+## Simple Summary
+
+Regulated Assets provide ecosystem support for assets that require an issuer’s approval (or a delegated third party’s approval, such as a licensed securities exchange) on a per-transaction basis. It standardizes the identification of such assets as well as defines the protocol for performing compliance checks and requesting issuer approval.
```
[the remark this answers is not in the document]
howardtw: https://github.com/stellar/stellar-protocol/pull/750/commits/117cbb9b5a0097fa792032b9200bb755ecd16135

ecosystem/sep-0036.md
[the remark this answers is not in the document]
howardtw: https://github.com/stellar/stellar-protocol/pull/750/commits/5145866e3a3e7e21d30670c2d55d0d9559911c71

ecosystem/sep-0036.md
howardtw: Good catch! https://github.com/stellar/stellar-protocol/pull/750/commits/954bde53bb2683f344c92b01fcdbf10e6df62a09

ecosystem/sep-0036.md
```diff
@@ -0,0 +1,145 @@
...
+
+This SEP is based on [SEP-0008](sep-0008.md), but the issuer no long has to a co-signer on the account holding the regulated asset. It also updates the API of the approval server to better align with other active SEPs.
+
+**Target Audience**: Asset issuers, issuance platforms, wallet developers, and exchanges
+
+## Motivation
+
+Stellar aims to be an ideal platform for issuing securities. As such, it must provide tools to comply with various regulatory requirements. Regulation often requires asset issuers to monitor and approve every transaction involving the issued asset and to enforce an assortment of constraints. This SEP provides support for these requirements.
```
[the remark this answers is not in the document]
howardtw: Nice one, especially that `Stellar is an ideal platform...` 💪

ecosystem/README.md
```diff
@@ -66,14 +66,15 @@
...
-
+| [SEP-0036](sep-0036.md) | Regulated Assets Transacting | Howard Chen | Standard | Draft |
 
 
 ### Rejected and Deprecated Proposals
 
 | Number | Title | Author | Track | Status |
 | --- | --- | --- | --- | --- |
 | [SEP-0003](sep-0003.md) | Compliance Protocol | SDF | Standard | Deprecated |
+| [SEP-0008](sep-0008.md) | Regulated Assets | Interstellar | Standard | Deprecated |
```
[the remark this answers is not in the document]
howardtw: https://github.com/stellar/stellar-protocol/pull/750/commits/002c05e9ebed15721b79c767874156ec5897bbfd

ecosystem/sep-0008.md
```diff
@@ -4,11 +4,16 @@
...
+Status: Deprecated
 Created: 2018-08-22
-Version 1.0.0
+Updated: 2020-10-26
+Version 1.0.1
 ```
 
+## Deprecation Notice
+
+For most use cases, we recommend the workflow specified in [SEP-0036](sep-0036.md) instead of SEP-0008.
```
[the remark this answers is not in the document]
howardtw: https://github.com/stellar/stellar-protocol/pull/750/commits/9b8132fc36874b550b2ecdca37a80235bf3178b2

ecosystem/sep-0036.md
```diff
@@ -0,0 +1,145 @@
...
+
+## Overview
+
+Implementing a Regulated Asset consists of these parts:
+- [Stellar.toml]: Issuers will advertise the existence of an Approval Service and the approval criteria via their stellar.toml file.
```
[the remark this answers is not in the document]
howardtw: https://github.com/stellar/stellar-protocol/pull/750/commits/6f8a8a847f1c0c2989e3f31dbb1f3b7fb3ce0b5d

ecosystem/sep-0036.md
```diff
@@ -0,0 +1,145 @@
...
+2. Wallet resolves asset information and detects that it's a regulated asset.
+3. Wallet sends the transaction to the approval server.
+4. Approval server determines whether the transaction is compliant based on the compliance ruleset and looks up the current state of the ledger if necessary.
+5. Wallet handles approval response:
+    1. *Success?* Transaction has been approved and signed by issuer.
+    2. *Rejected?* Wallet should display given error message to the user.
+
+## Stellar.toml
+
+Issuers will advertise the existence of an Approval Service through their stellar.toml file. This is done in the [[CURRENCIES]] section as different assets can have different requirements.
```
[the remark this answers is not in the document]
howardtw: https://github.com/stellar/stellar-protocol/pull/750/commits/6f8a8a847f1c0c2989e3f31dbb1f3b7fb3ce0b5d

ecosystem/sep-0036.md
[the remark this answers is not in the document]
howardtw: https://github.com/stellar/stellar-protocol/pull/750/commits/3473b69aa3f70d9216fbfcbcc506b04c6f9899ed

ecosystem/sep-0036.md
```diff
@@ -0,0 +1,145 @@
...
+- [Stellar.toml]: Issuers will advertise the existence of an Approval Service and the approval criteria via their stellar.toml file.
+- [Approval Server]: HTTP protocol for transaction signing.
+- [Transaction Composition]: Wallets will have to put operations that fully authorize the accounts for the asset, transact the asset, and deauthorize the accounts in a single transaction.
+
+## Regulated Assets Approval Flow
+
+1. User creates and signs a transaction.
+2. Wallet resolves asset information and detects that it's a regulated asset.
+3. Wallet sends the transaction to the approval server.
+4. Approval server determines whether the transaction is compliant based on the compliance ruleset and looks up the current state of the ledger if necessary.
```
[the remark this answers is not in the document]
howardtw: https://github.com/stellar/stellar-protocol/pull/750/commits/e77ae23b3420d6dd2bac3fe8bc602590e027a2c8

ecosystem/sep-0036.md
```diff
@@ -0,0 +1,145 @@
...
+
+Implementing a Regulated Asset consists of these parts:
+- [Stellar.toml]: Issuers will advertise the existence of an Approval Service and the approval criteria via their stellar.toml file.
+- [Approval Server]: HTTP protocol for transaction signing.
+- [Transaction Composition]: Wallets will have to put operations that fully authorize the accounts for the asset, transact the asset, and deauthorize the accounts in a single transaction.
+
+## Regulated Assets Approval Flow
+
+1. User creates and signs a transaction.
+2. Wallet resolves asset information and detects that it's a regulated asset.
```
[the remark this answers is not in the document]
howardtw: https://github.com/stellar/stellar-protocol/pull/750/commits/95c4461336b370d491d681ed3503044490abed60

ecosystem/sep-0036.md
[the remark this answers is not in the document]
howardtw: Added a section called `Identification` to consolidate the two. https://github.com/stellar/stellar-protocol/pull/750/commits/95c4461336b370d491d681ed3503044490abed60

The term `identification` is used in the simple summary section:
```
It standardizes the identification of such assets as well as defines the protocol for performing compliance checks and requesting issuer approval.
```

ecosystem/sep-0036.md
```diff
@@ -0,0 +1,145 @@
...
+    1. *Success?* Transaction has been approved and signed by issuer.
+    2. *Rejected?* Wallet should display given error message to the user.
+
+## Stellar.toml
+
+Issuers will advertise the existence of an Approval Service through their stellar.toml file. This is done in the [[CURRENCIES]] section as different assets can have different requirements.
+
+### Fields:
+
+- `regulated` is a boolean indicating whether or not this is a regulated asset. If missing, `false` is assumed.
```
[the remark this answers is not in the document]
howardtw: It's also addressed in the same commit: https://github.com/stellar/stellar-protocol/pull/750/commits/95c4461336b370d491d681ed3503044490abed60

ecosystem/sep-0036.md
howardtw: https://github.com/stellar/stellar-protocol/pull/750/commits/d0eb0d9f071e4ebe6c5d6f4aada5389b749dca58

ecosystem/sep-0036.md
howardtw: https://github.com/stellar/stellar-protocol/pull/750/commits/8969cf3495abf615b25834dfb4cdbe8367cafed0

ecosystem/sep-0036.md
howardtw: @JakeUrban it was intended to be a question for the reviewers as I don't know the answers.

However, based on the conversation in [this](https://github.com/stellar/stellar-protocol/pull/758#issuecomment-719511453) PR, I think part of the solution can be having the wallet send POST requests to the `action_url` in the response.

However, it doesn't feel like this should be the main solution as having the wallet send the approval server a sandwich transaction in the first place in order to start the KYC process sounds a bit odd. What do you guys think? @leighmcculloch @JakeUrban

tomerweller:
I would put this SEP on the back burner. 

1. Account setup was never an integral part of the technical approval server definition. So we can fairly easily backport sandwiched txs into SEP8 without changing anything technically. 

2. Changing request/response format is just not important enough without concrete ecosystem feedback. While SEP8 is not in production there are multiple SEP8 sandbox environments in the wild and we shouldn't disrupt these because of perceived simplicity/elegance. 

Would love to hear @dandoney's opinion.

howardtw:
@tomerweller I feel like a similar question was asked before, but what is the procedure to update a SEP with a `Final` status? I don't thinking replacing the `Account Setup` section with `Transaction Composition` is considered as a minor erratum.

From our last meeting, we found out that SEP-8 is extensively used, so we thought it's better to not create a new SEP to address the same topic at this point to reduce noise. Once I sort out the process of updating a `Final` SEP, I plan on closing this PR to start over again with the feedback in the PR.

Thoughts? @JakeUrban @leighmcculloch

tomerweller:
@howardtw good point. Instead of removing information we can append the new information. So modify the account setup section with a description of both approaches. wdyt?

howardtw:
@tomerweller if the SEP team is fine with appending information to a final SEP, I'm good with your suggestion.

By appending new information, if my understanding is correct, the status of this SEP will still be `Final` but the `Version` will be incremented with a `Updated` timestamp?

leighmcculloch:
If there aren't API changes being made, then it's probably possibly to make this change to SEP-8 in a way that extends it, but let's evaluate the change once it's been proposed for whether we all think it's safe to make to SEP-8.

---

Also, I think our use of Final was a little liberal previously. I guess we could back status SEP-8 to Active if the only implementations are outside production and limited. This wouldnt change what were doing here at all, but would communicate the actual state of the SEP a little better. What do you think @tomerweller?

tomerweller:
agreed

Manny27nyc (approved):

Manny27nyc (approved):

howardtw:
Closing due to https://github.com/stellar/stellar-protocol/pull/781

PR #754, opened 2020-10-23
SEP-0010: Clarify the first manage data operation is the operation containing the client account, home domain, etc

leighmcculloch:
### What
Clarify that the first manage data operation is the operation containing the client account, nonce, and home domain.

### Why
The first manage data operation has always been the first operation. This is stated explicitly further down in the SEP when the challenge transaction structure is described. All SDKs I have reviewed also follow this pattern. The abstract is somewhat ambiguous though because it states that there is a single manage data operation with those values without stating it is the first.

This change is in response to the discussion here:
https://github.com/stellar/stellar-protocol/pull/746#discussion_r509686786

### Note
This change is only improving the language and is not adding or changing the functionality of the SEP. This change is not urgent and may end up being merged before or after the v3 changes being proposed. If it is merged after the PR needs updating with an appropriate version and timestamp.

cc @nebolsin @JakeUrban @accordeiro

### DIFF
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -23,8 +23,8 @@ The authentication flow is as follows:
 1. The client obtains a unique [`challenge`](#challenge), which is represented as specially formed Stellar transaction
 1. The client verifies that the transaction has an invalid sequence number 0.  This is extremely important to ensure the transaction isn't malicious.
 1. The client verifies that the transaction is signed by the `SIGNING_KEY` specified by the requested service's [SEP-1 stellar.toml](sep-0001.md).
-1. The client verifies that the transaction's first Manage Data operation has its source account set to the user's account and value set to a nonce value. The client ignores the home domain included.
-1. The client verifies that if the transaction has other Manage Data operations they all have their source accounts set to the the server's account.
+1. The client verifies that the transaction's first operation is a Manage Data operation that has its source account set to the user's account and value set to a nonce value. The client ignores the home domain included.
+1. The client verifies that if the transaction has other operations they are Manage Data operations that all have their source accounts set to the the server's account.
 1. The client signs the transaction using the secret key(s) of signers for the user's Stellar account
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0010.md
```diff
@@ -23,7 +23,7 @@ The authentication flow is as follows:
 1. The client obtains a unique [`challenge`](#challenge), which is represented as specially formed Stellar transaction
 1. The client verifies that the transaction has an invalid sequence number 0.  This is extremely important to ensure the transaction isn't malicious.
 1. The client verifies that the transaction is signed by the `SIGNING_KEY` specified by the requested service's [SEP-1 stellar.toml](sep-0001.md).
-1. The client verifies that the transaction has a single Manage Data operation with its source account set to the user's account and value set to a nonce value. The client ignores the home domain included.
+1. The client verifies that the transaction's first Manage Data operation has its source account set to the user's account and value set to a nonce value. The client ignores the home domain included.
```
nebolsin, line 26: I don't want to be nitpicking, but it seems that this wording would allow, for example, a challenge with some `payment` as its first operation, followed by the correct `manage_data` op, and I don't think we should allow that. I guess we need something like "the transaction's first operation is Manage Data..." instead of "transaction's first Manage Data operation".

leighmcculloch: Good point. Addressed in 2efd3b960d7e66b53c7555cd9dd910bb095c73d3.

JakeUrban (approved):
LGTM. I don't think we ever resolved it, but are we expecting the operations in the challenge to be ordered or unordered? With this change, I think we need to go the ordered route.

leighmcculloch:
@JakeUrban The first operation must be the auth op so it must be ordered. Any other ops the order is not defined.

JakeUrban:
This should be merged prior to 3.0, can I merge it? @leighmcculloch

PR #757, opened 2020-10-28
Update SEP-0024 claimable balance op support

Zagan202:
## What 

Update SEP-0024 to include support for claimable balances.

## Why 

Currently Anchors are blocked issuing payments (deposits) if an account doesn't have an established trust line.
This update will encourage anchors and wallets to adopt Protocol 14/15 claimable balance operations.
Such that, when an account is newly created and/or a trust line isn't established the anchor will submit the transaction as a claimable balance to stellar for the user to claim in their own time.
Wallets should periodically query stellar for pending claimable balances for said user account.


ecosystem/sep-0024.md
```diff
@@ -190,6 +191,24 @@ The deposit flow can only be fulfilled if the Stellar `Account` has established
 5. `Anchor` detects the trust line creation in the `Account`. If the asset is `AUTH_REQUIRED`, `Anchor` approves the new trust line.
 6. `Anchor` proceeds with the deposit flow.
 
+##### Claimable Balances
```
JakeUrban, line 194: There are a lot of weird line breaks in this section

ecosystem/sep-0024.md
```diff
@@ -190,6 +191,24 @@ The deposit flow can only be fulfilled if the Stellar `Account` has established
...
+
+[Protocol
+14/15](https://www.stellar.org/blog/protocol-14-improvements?locale=en) now allows anchors to
+complete payments using the [ClaimableBalance
+functionality](../core/cap-0023.md) and the user can collect the funds at any
+time.
+
+This is an optional, yet recommended, feature that will no longer need to wait
+for the `Wallet` or the user to establish a
+trust line for the `Account` for the given asset.
```
JakeUrban, line 204:
```suggestion
[Claimable Balances](https://www.stellar.org/blog/protocol-14-improvements) allow for payments to accounts that do not have a trustline for the asset sent. 

Anchors can optionally use this functionality to send SEP-24 deposit transactions to Stellar accounts that don't yet have a trustline to the asset they have requested.

Wallets should poll for their users' received Claimable Balances so that users can "claim" them. Claiming a Claimable Balance adds a trustline to the account for the asset claimed and adds the amount transacted.
```

ecosystem/sep-0024.md
```diff
@@ -190,6 +191,24 @@ The deposit flow can only be fulfilled if the Stellar `Account` has established
...
+
+1. If the given Stellar `account` does not exist, on receipt of the deposit, the
```
JakeUrban, line 206: I don't think these 2 points are necessary. The logic is simple, no trustline --> claimable balance, trustline --> payment. I wouldn't introduce the "if the stellar account doesn't exist" case.

JakeUrban (commented):
I made a suggestion, I think its important to always remember who you're talking to when writing these SEPs.

ecosystem/sep-0024.md
```diff
@@ -190,6 +191,72 @@ The deposit flow can only be fulfilled if the Stellar `Account` has established
...
+1. The given Stellar `account` does not have a trust-line to the asset's issuer account.
+2. The `Anchor` proceeds with the deposit flow via the [CAP-0023](../core/cap-0023.md) Create Claimable Balance operation.
+
+**Anchor Claimable Balance Field Changes**
+
+### Fields for deposit transactions
+
+Name | Type | Description
+-----|------|------------
+`claimable_balance_id` | string | (optional) ClaimableBalanceID is a required parameter when claiming a Claimable Balance operation.
```
JakeUrban, line 242: Same thing here, this needs to go in the actual response from the `POST /interactive` endpoints and referenced here.

ecosystem/sep-0024.md
```diff
@@ -190,6 +191,72 @@ The deposit flow can only be fulfilled if the Stellar `Account` has established
...
+
+The `claimable_balance_id` can be added to the [response](#deposit-and-withdraw-shared-responses)
+This can be helpful for wallets to communicate to the user that their
+transaction was submitted as a claimable balance and can be claimed at the relevant location (`claimable_balance_id`).
+
+In order to claim it the recipient must either get access to the account that the anchor created and or establish a trust-line (for the asset code of the claimableBalance) with their existing account.
```
JakeUrban, line 248: `get access to the account that the anchor created` is a bit confusing. The wallet will already have access. The anchor creates the account but the wallet is the only one with the secret key.

I would remove this sentence, the flow is the same regardless of who created the account.

ecosystem/sep-0024.md
```diff
@@ -190,6 +191,72 @@ The deposit flow can only be fulfilled if the Stellar `Account` has established
...
+
+**Anchor Claimable Balance Recommended Options**
+
+#### Claimable Balance Predicates
+
+[https://developers.stellar.org/docs/glossary/claimable-balance/#parameters](https://developers.stellar.org/docs/glossary/claimable-balance/#parameters)
```
JakeUrban, line 254: Lets not do raw links in the SEP. We should also precede the link with context so they know what the link is before reading the URL or clicking it.

ecosystem/sep-0024.md
```diff
@@ -190,6 +191,72 @@ The deposit flow can only be fulfilled if the Stellar `Account` has established
...
+
+We suggest if you decide to set predicates to claim the balance in 30 days. `BEFORE_RELATIVE_TIME(X)`
```
JakeUrban, line 256: Did we decide 30 days was a good suggestion? Whats the rationale?

ecosystem/sep-0024.md
```diff
@@ -190,6 +191,72 @@ The deposit flow can only be fulfilled if the Stellar `Account` has established
...
+
+However if implemented, its up to the anchors to handle refunds (of off network funds) to the the waller/user.
```
JakeUrban, line 258: I think we need to expand on this. _why_ does this introduce the need to handle refunds? Why would an anchor want to have an expiration? Why do we recommend an expiration? This SEP is the introduction to this functionality for most people reading this.

ecosystem/sep-0024.md
```diff
@@ -190,6 +191,72 @@ The deposit flow can only be fulfilled if the Stellar `Account` has established
 5. `Anchor` detects the trust line creation in the `Account`. If the asset is `AUTH_REQUIRED`, `Anchor` approves the new trust line.
 6. `Anchor` proceeds with the deposit flow.
 
+##### Claimable Balances
+
+[Claimable Balances](https://www.stellar.org/blog/protocol-14-improvements) allows payments to accounts that do not have a trust-line for the asset sent.
+
+**Wallets** poll for their users' received Claimable Balances so that users can "claim" them.
+
+**Anchors** can optionally use this functionality to send SEP-24 deposit transactions to Stellar accounts that don't yet have a trust-line to the asset they have requested.
```
JakeUrban, line 200: I think we should emphasize that supporting claimable balances is currently _optional_ for anchors but _mandatory_ for wallets. Wallets reading this need to infer that some of their partner anchors may implement claimable balances, and therefore need to support it as well, but it's not super clear.

ecosystem/sep-0024.md
JakeUrban: Now that I'm thinking about this more, I'm realizing it's impossible for anchors to return a `claimable_balance_id` for a payment that hasn't been submitted to the network at this point. So I don't think we can have this response attribute at all.

JakeUrban (changes requested):
Aside from my PR comments, I'd like to outline why the `claimable_balance` request parameter and `claimable_balance_id` are hardly ever useful to either the wallet or anchor. The request parameter may even confuse wallets into thinking an anchor supports claimable balances when they don't.

We're starting off with neither wallets or anchors having CB support. This change is *requiring* wallets to add support while making it *optional* for anchors.

Now let's take an example Wallet and Anchor. When the Wallet adds CB support, it will begin including the `claimable_balance: true` parameter in deposit requests. At this point, the Anchor does not implement CBs and simply ignores the `claimable_balance: true` parameter and starts waiting for a trust-line to be established.

The Wallet _must_ realize that the transaction is `pending_trust` and that their `claimable_balance: true` parameter was ignored. If they don't, a trust-line will never be established and the deposit will never take place.

Furthermore, the response parameter `claimable_balance_id` can't be returned because the claimable balance op hasn't been submitted to Stellar when the anchor responds to the request containing the `claimable_balance: true` parameter.

So, there is exactly one situation where these request parameter is helpful: when the anchor **has** implemented CBs and the Wallet **has not**. The wallet can tell the anchor that they don't support CBs, so the anchor should _not_ send a CB and instead should wait until the account has a trust-line.

Due to the fact that this PR requires Wallets to support CBs while making it optional for anchors, the one case where the `claimable_balance: true` parameter will likely not occur IRL. 

If we remove the added request and response parameter from this change, it's also much simpler to implement for both wallets and anchors. We can also talk more about how wallets can figure out whether to expect a CB or payment.

ecosystem/sep-0024.md
```diff
@@ -190,6 +191,72 @@ The deposit flow can only be fulfilled if the Stellar `Account` has established
...
+
+New Request Parameter:
+
+Name | Type | Description
+-----|------|------------
+`claimable_balance` | boolean | True if submitting transactions as a claimable balance is accepted, false otherwise.
+
+**Wallet Claimable Balance Polling**
+
+The wallet must either listen for a callback from the anchor with the
```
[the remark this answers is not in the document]
Zagan202: I was suggesting that anchors append `claimable_balance_id` to the deposit response.

ecosystem/sep-0024.md
```diff
@@ -190,6 +191,72 @@ The deposit flow can only be fulfilled if the Stellar `Account` has established
...
+`claimable_balance_id` or poll Horizon's
+[`/claimable_balances/`](https://developers.stellar.org/api/resources/claimablebalances/list/)
+endpoint to pull any available claimable balances for the claimant.
+
+*Recommended request*
+`GET /claimable_balances?claimant={:claimant}&order={asc,desc}&limit={1-200}`
```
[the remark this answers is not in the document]
Zagan202: > Does this request change depending on the architecture of the client's system? 

I don't anticipate that the GET request to horizon would change. This is an example of how the client can discover that there's CB available for the account.

>Could a client extract the claimable balance ID from the `results_xdr` and just use it in the GET request?

Either way is possible.
The GET request surfaces _all_ of the account's available CB; from there they could communicate what those id(s) are to the account holder.
or
They can decode the `results_xdr` of the transaction submitted, discover that it was submitted as a claimable balance, then pull the balance id from the xdr.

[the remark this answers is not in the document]
Zagan202: > I wouldn't try to anticipate how the client will want to poll, but I think outlining the 1-3 common approaches would be helpful.

I think thats a great idea.

ecosystem/sep-0024.md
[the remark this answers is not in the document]
JakeUrban: Oh, we have several real callbacks in this SEP so lets not use that word.

ecosystem/sep-0024.md
```diff
@@ -190,6 +191,72 @@ The deposit flow can only be fulfilled if the Stellar `Account` has established
...
+
+This allows the wallet to update the claimant that their is a balance available
+to claim and they may need to establish a [trust-line](#stellar-account-doesnt-trust-asset) in order to do so.
```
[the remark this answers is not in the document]
Zagan202: Apologies my editor tries to keep line length under 80.
I extended it so it shouldn't be a problem in the future.

ecosystem/sep-0024.md
```diff
@@ -190,6 +191,72 @@ The deposit flow can only be fulfilled if the Stellar `Account` has established
...
+
+**Anchor Claimable Balance Flow**
+
+1. The given Stellar `account` does not have a trust-line to the asset's issuer account.
```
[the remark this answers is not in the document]
Zagan202: That's correct.
The sender can make as many payments (of an asset that the recipient doesn't trusts) that they want as long as their CBs.
The recipient has to trust the asset in order to add that balance to their account.

ecosystem/sep-0024.md
Zagan202: We discussed that 30 days might be a good sweet spot to not keep balances piling up on the dex without them being claimed.

I'm still open to discussion on this one.

On one hand I think its totally fine to keep in unconditional and let the recipient claim on their own time whether it be a week a month or a year.
On the other hand, I think that if we help push recipients to claim their balance with a bit of haste it could lead them to do more activity (payments and whatnot) on the network. Instead of having balances eternally rot.

tomerweller:
@JakeUrban I hear you and you're making a good argument. I do think you're overestimating the cost of implementing this parameter and downplaying the cost of *not* implementing it. 

> The Wallet must realize that the transaction is pending_trust and that their claimable_balance: true parameter was ignored. If they don't, a trust-line will never be established and the deposit will never take place.

That is true. And is true regardless of whether or not this parameter flag is implemented. 

> Due to the fact that this PR requires Wallets to support CBs while making it optional for anchors, the one case where the claimable_balance: true parameter is helpful will likely not occur IRL.

You're treating wallet adoption as a binary but that's not the case. We'll consider wallet adoption complete when a critical mass of wallets have adopted - ideally this is all of them but there may be outliers that we just accept have not changed. For these edge cases wallets it is crucial that anchors *explicitly fail* rather than sending a claiming balance to a wallet that doesn't support it. Without this parameter anchors are in the dark. 

If there's one thing that we learned in our history of SEP changes is that adoption is not as simple as we think it is and theoretical edge cases have tendency to .. not be so theoretical.

ecosystem/sep-0024.md
```diff
@@ -190,6 +191,72 @@ The deposit flow can only be fulfilled if the Stellar `Account` has established
...
+```
+POST TRANSFER_SERVER_SEP0024/transactions/deposit/interactive
+Content-Type: multipart/form-data
+```
+
+New Request Parameter:
+
+Name | Type | Description
+-----|------|------------
+`claimable_balance` | boolean | True if submitting transactions as a claimable balance is accepted, false otherwise.
```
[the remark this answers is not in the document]
tomerweller: `claimable_balance` is not really self descriptive - it doesn't convey that cap23 is supported, not required. while a bit cumbersome I would go with something like 
`is_calimable_balance_supported` or `is_claimble_balance_accepted`. You can drop the `is` if it's over the top and go with something like `claimable_balance_accepted`.

ecosystem/sep-0024.md
Zagan202: Correct me if I'm off the mark on this theoretical flow if the anchor sets the 30 day predicate.

1. Alice deposits off chain funds (US dollar)
2. Alice doesn't have a trust-line established for (USD) such that the anchor makes a claimable balance for Alice to claim (30 days to claim)
3. Alice doesnt claim the balance before 30 days and the balance is locked.
4. The anchor should refund Alice of her off chain funds

Zagan202: Honestly after writing that and demoing what that behavior would be I think its probably for the best if we ensure that they set the predicates as unconditional.

JakeUrban: I don't think we should recommend going either way, but outline why an anchor would take one approach over the other. 

Anchors that really really care about UX will probably want to handle refunds so users don't think they stole funds only to realize it's on them to "claim" the balance. 

For other anchors, handling refunds won't be worth it because there is no real benefit to them if they refund. The have to do 2X the logistical effort for no profit (fees are refunded).

JakeUrban:
> For these edge cases wallets it is crucial that anchors explicitly fail

So you're saying Wallets that don't support CBs will still send the `is_claimable_balance_supported: false` key-value pair? If a wallet hasn't build CB support, how likely do you think they are to implement this key-value pair? I think we're overestimating the number of wallets who won't implement this AND will make use of `is_claimable_balance_supported`

tomerweller:
> So you're saying Wallets that don't support CBs will still send the is_claimable_balance_supported: false key-value pair? If a wallet hasn't build CB support, how likely do you think they are to implement this key-value pair? I think we're overestimating the number of wallets who won't implement this and will make use of is_claimable_balance_supported

No :) 

I might have forgot to mention a very important piece of information that we should add to the SEP: Not having the `is_claimable_balance_supported` parameter in the payload is equivalent to having it set to `false`. So if wallet didn't migrate and an anchor did, and the anchor doesn't see the parameter they will error out.

JakeUrban:
Ohh thats important and helpful to hear. I was imagining anchors who supported both types, but most anchors will migrate to CBs exclusively and will hence require `is_claimable_balance_supported: true`.

I agree then we should keep the request parameter but we unfortunately can't have a `claimable_balance_id` response attribute as I explained in my PR comments.

Zagan202:
@tomerweller 
> I might have forgot to mention a very important piece of information that we should add to the SEP: Not having the `is_claimable_balance_supported` parameter in the payload is equivalent to having it set to `false`. So if a wallet didn't migrate, the anchor did , and it only support claimable balances, the anchor should error.

Should I write in the SEP that if a wallets do not implement this change and attempt to deposit through an anchor that supports only claimable balances the the deposit will fail?

@JakeUrban 

>I agree then we should keep the request parameter but we unfortunately can't have a claimable_balance_id response attribute as I explained in my PR comments.

Would it be best to not only remove this from the SEP but remove claimable_balance_id from the transaction model in polaris?
Since the anchors are not going to utilize claimable_balance_id in any way.

Zagan202:
After some thought I think we should we also add to the SEP that wallets should implement claiming balances from CBs.
Thoughts?

JakeUrban:
> Should I write in the SEP that if a wallets do not implement this change and attempt to deposit through an anchor that supports only claimable balances the the deposit will fail?

Yes. Overall I think we need to be a lot more explicit about the possible flows related to `is_claimable_balance_supported`. The situation Tomer outlines needs to described in the SEP, as well as what should happen in other situations, like when the anchor doesn't support CBs.

JakeUrban:
> Would it be best to not only remove this from the SEP but remove claimable_balance_id from the transaction model in polaris? Since the anchors are not going to utilize claimable_balance_id in any way.

I think `Transaction.claimable_balance_id` should be populated when the claimable balance is detected by Polaris. It will have to be `None` until the transaction is submitted to the network, streamed, and matched. But once Polaris actually has the claimable balance tx, it can extract the ID and save it to the column.

Basically, lets not remove the column because it's still useful when the TX is matched, even if we don't get an ID in the initial `POST /deposit` response body.

ecosystem/sep-0024.md
Zagan202: @tomerweller 
Would you be opposed to strongly suggesting unconditional claiming conditions?

ecosystem/sep-0024.md
```diff
@@ -150,6 +150,7 @@ Name | Type | Description
 `wallet_name` | string | (optional) In communications / pages about the deposit, anchor should display the wallet name to the user to explain where funds are going.
 `wallet_url` | string | (optional) Anchor should link to this when notifying the user that the transaction has completed.
 `lang` | string | (optional) Defaults to `en`. Language code specified using [ISO 639-1](https://en.wikipedia.org/wiki/ISO_639-1). `error` fields in the response, as well as the interactive flow UI and any other user-facing strings returned for this transaction should be in this language.
+`claimable_balance_supported` | boolean | (optional) True if submitting transactions as a claimable balance is accepted, false otherwise.
```
leighmcculloch, line 153: >`claimable_balance_supported`...a claimable balance is accepted...

Does this field indicate to the server that the client supports claimable balances?

ecosystem/sep-0024.md
```diff
@@ -190,6 +192,73 @@ The deposit flow can only be fulfilled if the Stellar `Account` has established
 5. `Anchor` detects the trust line creation in the `Account`. If the asset is `AUTH_REQUIRED`, `Anchor` approves the new trust line.
 6. `Anchor` proceeds with the deposit flow.
 
+##### Claimable Balances
+
+[Claimable Balances](https://www.stellar.org/blog/protocol-14-improvements) allows payments to accounts that do not have a trust-line for the asset sent.
+
+**Wallets**: It's mandatory for wallets to support Claimable Balances and should poll for their users' received Claimable Balances so that users can "claim" them.
+
+**Anchors**: It's optional for anchors to use submit Claimable Balances as deposit transactions to Stellar accounts that don't yet have a trust-line to the asset they have requested.
```
leighmcculloch, line 201:
```suggestion
**Anchors**: It's optional for anchors to submit Claimable Balances as deposit transactions to Stellar accounts that don't yet have a trust-line to the asset they have requested.
```

ecosystem/sep-0024.md
```diff
@@ -190,6 +192,73 @@ The deposit flow can only be fulfilled if the Stellar `Account` has established
...
+
+**Wallet Claimable Balance Flow**
+
+Wallets will still follow the [typical deposit flow](#Basic-Wallet-Implementation).
+The POST request however will extend to include a `claimable_balance_supported` parameter
+
+**Wallet Claimable Balance Polling**
+
+Claimable Balance Polling allows the wallet to update the claimant that their is a balance available to claim and they may need to establish[trust-line](#stellar-account-doesnt-trust-asset) in order to do so.
+There are a few ways a wallet can support .
```
leighmcculloch, line 211: >There are a few ways a wallet can support **?????**.

ecosystem/sep-0024.md
```diff
@@ -190,6 +192,73 @@ The deposit flow can only be fulfilled if the Stellar `Account` has established
...
+**Anchors**: It's optional for anchors to use submit Claimable Balances as deposit transactions to Stellar accounts that don't yet have a trust-line to the asset they have requested.
+
+**Wallet Claimable Balance Flow**
+
+Wallets will still follow the [typical deposit flow](#Basic-Wallet-Implementation).
+The POST request however will extend to include a `claimable_balance_supported` parameter
+
+**Wallet Claimable Balance Polling**
+
+Claimable Balance Polling allows the wallet to update the claimant that their is a balance available to claim and they may need to establish[trust-line](#stellar-account-doesnt-trust-asset) in order to do so.
```
leighmcculloch, line 210: Wallets should probably poll claimable balances as a general feature, but for SEP-24 I think the flow as an isolated experience shouldn't require it. The anchor should provide a claimable balance ID and the wallet should try and claim only that ID. Otherwise the UI/UX could become quite confusing. If for example a claimable balance comes in unrelated to the anchor at the same time for a different amount, that would be very confusing for the user and lead to a support ticket being opened about the other amount. From a security point-of-view we should also avoid arbitrary transactions impacting a well defined user flow as well.

ecosystem/sep-0024.md
```diff
@@ -344,13 +413,13 @@ When a successful response comes back (either from polling or via callback), the
 The wallet must use the response fields in the following way to complete the withdrawal:
 
 - `status`: `pending_user_transfer_start` means the user has given all necessary info to the anchor, and it is up to the wallet to send the actual stellar assets.
-- `withdraw_anchor_account`: send the withdrawal payment to this Stellar account. 
+- `withdraw_anchor_account`: send the withdrawal payment to this Stellar account.
```
leighmcculloch, line 416: Seems like there's a view formatting issues in this PR that could be fixed independent of a major change to the SEP. Could they be moved to a separate PR we can merge quickly?

ecosystem/sep-0024.md
```diff
@@ -569,6 +638,7 @@ Name | Type | Description
 `deposit_memo_type` | string | (optional) Type for the `deposit_memo`.
 `from` | string | Sent from address, perhaps BTC, IBAN, or bank account.
 `to` | string | Stellar address the deposited assets were sent to.
+`claimable_balance_id` | string | (optional) ClaimableBalanceID is a required parameter when claiming a Claimable Balance operation.
```
leighmcculloch, line 641: I found this description confusing. I'd avoid the term parameter since we use the term field in the title. The use of the term required is also confusing in the context of an optional field. And its not clear to me in this context what it means for it to be required when claiming the operation, since the operation isn't claimed. I think we should say what the field contains and when it is used.
```suggestion
`claimable_balance_id` | string | (optional) ID of the Claimable Balance containing the asset, included when an anchor is issuing the asset via a Claimable Balance.
```

leighmcculloch (commented):
I shared some thoughts below, none are intended to block this work.

ecosystem/sep-0024.md
Zagan202: That is correct.

ecosystem/sep-0024.md
Zagan202: >Wallets should probably poll claimable balances as a general feature, but for SEP-24 I think the flow as an isolated experience shouldn't require it. 

I agree that is should be a general feature however it should be a required feature for wallets who intend to support claimable balances as a fallback to communicate with Wallet users that they have balances available.
It's currently implemented [here](https://github.com/stellar/stellar-demo-wallet/pull/49). 
I want to provide wallets as little friction as possible when interacting with anchors who may be still in the process of supporting claimable balances.

> The anchor should provide a claimable balance ID and the wallet should try and claim only that ID. 

We discussed this and we would like to have this feature when anchor adoption of claimable balances are relatively high.
Then we can require anchors send claimable balance IDs in their deposit response that Wallets can anticipate.
So we're attempting to strategize our phases for wallets and anchors not to overwhelm them with too many changes immediately.

ecosystem/sep-0024.md
```diff
@@ -190,6 +192,73 @@ The deposit flow can only be fulfilled if the Stellar `Account` has established
 5. `Anchor` detects the trust line creation in the `Account`. If the asset is `AUTH_REQUIRED`, `Anchor` approves the new trust line.
 6. `Anchor` proceeds with the deposit flow.
 
+##### Claimable Balances
+
+[Claimable Balances](https://www.stellar.org/blog/protocol-14-improvements) allows payments to accounts that do not have a trust-line for the asset sent.
```
JakeUrban, line 197:
```suggestion
[Claimable Balances](https://www.stellar.org/blog/protocol-14-improvements) enable payments to accounts that do not have a trust-line for the asset being sent.
```

ecosystem/sep-0024.md
```diff
@@ -190,6 +192,73 @@ The deposit flow can only be fulfilled if the Stellar `Account` has established
...
+
+**Wallets**: It's mandatory for wallets to support Claimable Balances and should poll for their users' received Claimable Balances so that users can "claim" them.
```
JakeUrban, line 199:
```suggestion
**Wallets**: It's mandatory for wallets to support Claimable Balances. The methods for supporting Claimable Balances are outlined below.

File: ecosystem/sep-0024.md [hunk 23]
```diff
@@ -190,6 +192,73 @@ The deposit flow can only be fulfilled if the Stellar `Account` has established
...
+
+**Anchors**: It's optional for anchors to use submit Claimable Balances as deposit transactions to Stellar accounts that don't yet have a trust-line to the asset they have requested.
+
+**Wallet Claimable Balance Flow**
+
+Wallets will still follow the [typical deposit flow](#Basic-Wallet-Implementation).
+The POST request however will extend to include a `claimable_balance_supported` parameter
+
+**Wallet Claimable Balance Polling**
```
JakeUrban, line 208: Leigh also commented on this section, but the way I see it there are two methods for Wallets to detect claimable balances:

1. Checking Horizon
2. Checking the anchor's `GET /transaction(s)` endpoint

With the Horizon option, they could poll on demand or stream transactions to the users Stellar account.

ecosystem/sep-0024.md
```diff
@@ -190,6 +192,73 @@ The deposit flow can only be fulfilled if the Stellar `Account` has established
...
+
+Claimable Balance Polling allows the wallet to update the claimant that their is a balance available to claim and they may need to establish[trust-line](#stellar-account-doesnt-trust-asset) in order to do so.
+There are a few ways a wallet can support .
+
+1. The wallet can poll Horizon's [`/claimable_balances/`](https://developers.stellar.org/api/resources/claimablebalances/list/) list endpoint to pull any available claimable balances for the claimant.
+
+   * *Recommended request* `GET /claimable_balances?claimant={:claimant}&order={asc,desc}&limit={1-200}` or [See List Claimable Balances](https://developers.stellar.org/api/resources/claimablebalances/list/)
+
+1. Wallets could decode the `results_xdr` of the transaction submitted
```
JakeUrban, line 217: I think we should leave this option out. Its complicated and the `/claimable_balances` endpoint polling is easy.

ecosystem/sep-0024.md
```diff
@@ -190,6 +192,73 @@ The deposit flow can only be fulfilled if the Stellar `Account` has established
...
+    const operationIndex = 0; // if you have multiple operations then you the index to correspond to he claimable balance operation
+    console.log(tr.result().results()[operationIndex].value().createClaimableBalanceResult().balanceId().toXDR('hex'));
+     ```
+
+**Wallet Claim Account Claimable Balance**
+
+It's recommended that Wallets have an feature that allows the account to claim the balances that were polled from Horizon.
+[Example on how to claim a balance](https://developers.stellar.org/docs/glossary/claimable-balance/#claim-claimable-balance)
+
+**Anchor Claimable Balance Flow**
```
JakeUrban, line 243: instead of **bolding** all these sections, I would use the markdown `####` strategy so readers can clearly see where the `Wallet` section ends and `Anchor` section starts.

ecosystem/sep-0024.md
```diff
@@ -190,6 +192,73 @@ The deposit flow can only be fulfilled if the Stellar `Account` has established
...
+
+1. If a user's Stellar account does not have a trust-line to the asset's issuer account, one must be added via the [Change Trust](https://developers.stellar.org/docs/start/list-of-operations/#claim-claimable-balance) operation.
+2. The `Anchor` proceeds with the deposit flow via the [CAP-0023](../core/cap-0023.md) Create Claimable Balance operation.
```
JakeUrban, line 246: Is this correct? Why would an anchor have to add a change_trust op to the transaction? In fact, I don't think this _can_ happen because the user's SK is required to change the user account's trustlines

ecosystem/sep-0024.md
```diff
@@ -190,6 +192,127 @@ The deposit flow can only be fulfilled if the Stellar `Account` has established
 5. `Anchor` detects the trust line creation in the `Account`. If the asset is `AUTH_REQUIRED`, `Anchor` approves the new trust line.
 6. `Anchor` proceeds with the deposit flow.
 
+##### Claimable Balances
+
+[Claimable Balances](https://www.stellar.org/blog/protocol-14-improvements) enables payments to accounts that do not have a trust line for the asset being sent.
```
JakeUrban, line 197:
```suggestion
[Claimable Balances](https://www.stellar.org/blog/protocol-14-improvements) enable payments to accounts that do not have a trust line for the asset being sent.
```

ecosystem/sep-0024.md
```diff
@@ -190,6 +192,127 @@ The deposit flow can only be fulfilled if the Stellar `Account` has established
...
+This feature provides less friction for Stellar accounts who wish to make an offchain deposit to anchors without making a trust line before the deposit is complete.
```
JakeUrban, line 198: This doesn't make sense. Stellar accounts can't make off-chain deposits.

ecosystem/sep-0024.md
```diff
@@ -190,6 +192,127 @@ The deposit flow can only be fulfilled if the Stellar `Account` has established
...
+Anchors will no longer have to wait during the time when the account doesn't have a trust line. Instead, anchors can make the payment using ClaimableBalance operation and the stellar account (wallet user) can collect the funds at their own convenience.
```
JakeUrban, line 199:
```suggestion
Anchors will no longer have to wait until the user's Stellar account has trust line to the sending asset before sending funds. Instead, anchors can make the payment using a ClaimableBalance operation and the user's Stellar account can claim the funds at their own convenience.
```

ecosystem/sep-0024.md
```diff
@@ -190,6 +192,127 @@ The deposit flow can only be fulfilled if the Stellar `Account` has established
...
+
+**NOTE**: The Stellar Account will still need to establish trust line to the asset before claiming the outstanding balance.
```
JakeUrban, line 201: maybe this belongs in the `Wallet Feature: Claim Account Claimable Balance` section? It seems a bit early to say this here.

ecosystem/sep-0024.md
```diff
@@ -190,6 +192,127 @@ The deposit flow can only be fulfilled if the Stellar `Account` has established
...
+
+**Wallets**: To support claimable balances wallets must
+
+* Send `claimable_balance_supported=true` in their POST /transactions/deposit/interactive request.
```
JakeUrban, line 205: I wouldn't specify the `true` value like you are right now. It looks like a URL-encoded parameter, but client will be making a post request. Instead I'd provide a JSON example:
```json
{
    "claimable_balance": true
}
```

ecosystem/sep-0024.md
```diff
@@ -190,6 +192,127 @@ The deposit flow can only be fulfilled if the Stellar `Account` has established
...
+* Periodically poll for account's available claimable balances.
+* Provide a UI that allows users to claim a claimable balance. (The wallet UI should allow users to trust the asset of the claimable balance in order to claim)
```
JakeUrban, line 207: I would leave out the trust-line/TX-construction details until you get to the `Claim Account Claimable Balance` section
```suggestion
* Provide a UI that allows users to claim a claimable balance.
```

ecosystem/sep-0024.md
```diff
@@ -190,6 +192,127 @@ The deposit flow can only be fulfilled if the Stellar `Account` has established
 5. `Anchor` detects the trust line creation in the `Account`. If the asset is `AUTH_REQUIRED`, `Anchor` approves the new trust line.
 6. `Anchor` proceeds with the deposit flow.
 
+##### Claimable Balances
```
JakeUrban, line 195: Can you start with stating that claimable balances are currently optional by highly recommended? Also, note that in the future, claimable balances will be required.

Eventually, the hope is that all anchors will support CB's, and therefore we no longer need to support the 'waiting for trust-line' flow.

ecosystem/sep-0024.md
```diff
@@ -190,6 +192,127 @@ The deposit flow can only be fulfilled if the Stellar `Account` has established
...
+
+**Wallets**: To support claimable balances wallets must
+
+* Send `claimable_balance_supported=true` in their POST /transactions/deposit/interactive request.
+* Periodically poll for account's available claimable balances.
+* Provide a UI that allows users to claim a claimable balance. (The wallet UI should allow users to trust the asset of the claimable balance in order to claim)
+
+**Anchors**: To support claimable balances anchors must
+
+* Accept the `claimable_balance_supported` field in their `POST /transactions/deposit/interactive` request and submit the deposit transaction as a Claimable Balance operation to Stellar account that don't yet have a trust-line to the asset they have requested.
```
JakeUrban, line 211: Making the singular v. plural consistent
```suggestion
* Accept the `claimable_balance_supported` field in `POST /transactions/deposit/interactive` requests and submit deposit transactions using Claimable Balance operations to Stellar accounts that don't yet have a trust-line to the asset requested.
```

ecosystem/sep-0024.md
```diff
@@ -190,6 +192,127 @@ The deposit flow can only be fulfilled if the Stellar `Account` has established
...
+* Add `claimable_balance_id` to their [deposit transactions](####Fields-for-deposit-transactions)
```
JakeUrban, line 212:
```suggestion
* Add `claimable_balance_id` to their [deposit transactions responses](####Fields-for-deposit-transactions)
```

ecosystem/sep-0024.md
```diff
@@ -190,6 +192,127 @@ The deposit flow can only be fulfilled if the Stellar `Account` has established
...
+
+**Anchors and Wallets**: Both anchors and wallets must also support the aforementioned ["stellar account doen't have a trust line for that asset."](#stellar-account-doesnt-trust-asset).
```
JakeUrban, line 214:
```suggestion
**Anchors and Wallets**: Both anchors and wallets must also support the aforementioned ["stellar account doen't have a trust line for that asset."](#stellar-account-doesnt-trust-asset) flow.
```

ecosystem/sep-0024.md
```diff
@@ -190,6 +192,127 @@ The deposit flow can only be fulfilled if the Stellar `Account` has established
...
+Wallets need to be interoperable with anchors that have not adopted the [claimable balance feature](#Claimable-Balances) and vice versa.
+
+##### Wallet Claimable Balance Flow
+
+Wallets will still follow the [typical deposit flow](#Basic-Wallet-Implementation).
+Wallets will provide `claimable_balance_supported=true` field and value in their POST parameters for `/transactions/deposit/interactive`.
```
JakeUrban, line 220: Same here with the URL-coded example

ecosystem/sep-0024.md
```diff
@@ -190,6 +192,127 @@ The deposit flow can only be fulfilled if the Stellar `Account` has established
...
+
+If the anchor we're depositing to **supports** claimable balances.
```
JakeUrban, line 222: Wallets unfortunately won't know if claimable balances are supported unless they hard-code it per-anchor. So i think we need to rewrite this **supports**/**doesn't support** section from the perspective that the wallet doesn't know.

That would involve sending `claimable_balance_supported: true`, then polling the service for `pending_trust`. If the status goes to `completed` without ever moving to `pending_trust`, the wallet knows to check for claimable balances, and the `claimable_balance_id` should be populated.

ecosystem/sep-0024.md
```diff
@@ -190,6 +192,127 @@ The deposit flow can only be fulfilled if the Stellar `Account` has established
...
+If the anchor we're depositing to **doesn't support** claimable balances.
+
+1. Make a request to `/transactions/deposit/interactive` and provide a value for the `claimable_balance_supported=true` field.
+1. When the transaction status becomes `pending_user_transfer_start` send the required payment as described in the interactive webapp callback or the `/transaction` endpoint.
+1. The anchor will ignore the `claimable_balance_supported` field, discover the account does not have a trust line, and report transaction status as `pending_trust`.
+1. At this point both the wallet and anchor are waiting on the trust line to be established in order to continue the deposit payment. This is described here [steps 4-6](#stellar-account-doesnt-trust-asset)
+
+##### Wallet Feature: Claimable Balance Polling
+
+Claimable Balance Polling allows the wallet to update the claimant that their is a balance available to claim and they may need to establish a [trust line](#stellar-account-doesnt-trust-asset) in order to do so.
```
JakeUrban, line 238:
```suggestion
Polling allows the wallet to detect claimable balances and notify their users that their is a balance available to claim. See the [Claim Account Claimable Balance](#Wallet Feature: Claim Account Claimable Balance) section to learn more about how to claim a balance.
```

ecosystem/sep-0024.md
```diff
@@ -190,6 +192,127 @@ The deposit flow can only be fulfilled if the Stellar `Account` has established
...
+There are a few options as to how a wallet can poll and surface claimable balance information for their users.
+
+1. The wallet can poll Horizon's [`/claimable_balances/`](https://developers.stellar.org/api/resources/claimablebalances/list/) list endpoint to pull any available claimable balances for the claimant.
+
+   * *Recommended request* `GET /claimable_balances?claimant={:claimant}&order={asc,desc}&limit={1-200}` or [See List Claimable Balances](https://developers.stellar.org/api/resources/claimablebalances/list/)
+
+2. The wallet can poll Horizon's [`/claimable_balances/`](https://developers.stellar.org/api/resources/claimablebalances/single/) single endpoint to pull the submitted transactions' available claimable balance(s) for the claimant(user) using the claimable_balance_id.
```
JakeUrban, line 245: I don't think this option will be valuable for wallets, because they won't know the balance ID until they stream the transaction or poll the anchor for the `claimable_balance_id`. In either scenario, they won't need to make an additional Horizon request to get info on the balance, they can just immediately submit the claim op.

Not to mention that decoding the result XDR and getting the balance ID is complex, as evident by how much you had to write in order to explain it. I vote to remove this method.

ecosystem/sep-0024.md
```diff
@@ -190,6 +192,127 @@ The deposit flow can only be fulfilled if the Stellar `Account` has established
...
+      const { xdr } = StellarBase;
+
+      const b64 = "AAAAAAAAAGQAAAAAAAAAAQAAAAAAAAAOAAAAAAAAAAA9/w3pAlg0fEoZp+eC9g0slbYtvoySN3i5y+F4WJY8sgAAAAA=";
+      const tr = xdr.TransactionResult.fromXDR(b64, 'base64');
+
+      const operationIndex = 0; // if you have multiple operations then you the index to correspond to he claimable balance operation
+      console.log(tr.result().results()[operationIndex].value().createClaimableBalanceResult().balanceId().toXDR('hex'));
+      ```
+
+##### Wallet Feature: Claim Account Claimable Balance
```
JakeUrban, line 271: I think it would be valuable to outline the operations needed right here. You can keep the docs reference if they want to check it out, but ideally they don't have to leave the SEP to know what the TX should look like. See SEP-10's description of challenge transactions for inspiration.

ecosystem/sep-0024.md
```diff
@@ -190,6 +192,127 @@ The deposit flow can only be fulfilled if the Stellar `Account` has established
...
+
+It's recommended that Wallets have an feature that allows the account to claim the balances that were polled from Horizon.
+[Example on how to claim a balance](https://developers.stellar.org/docs/glossary/claimable-balance/#claim-claimable-balance)
+
+##### Anchor Claimable Balance Flow
+
+1. Wallets make a request to `/transactions/deposit/interactive` and provide a value for the `claimable_balance_supported=true` field.
+1. Anchors mark the transaction as [`claimable_balance_supported=true`](#deposit)
```
JakeUrban, line 279: URL-encoding thing

ecosystem/sep-0024.md
```diff
@@ -190,6 +192,127 @@ The deposit flow can only be fulfilled if the Stellar `Account` has established
...
+1. Anchors discover that the stellar account that their sending the asset to doesnt have a trust line.
```
JakeUrban, line 280:
```suggestion
1. Anchors detect that the user's Stellar account doesn't have a trust-line.
```

ecosystem/sep-0024.md
```diff
@@ -190,6 +192,127 @@ The deposit flow can only be fulfilled if the Stellar `Account` has established
...
+1. The anchor should then report transaction status as `pending_anchor` and make the payment as a claimable balance operation.
```
JakeUrban, line 281:
```suggestion
1. The anchor should then report transaction status as `pending_anchor` and submit a Stellar transaction containing a create claimable balance operation
```

ecosystem/sep-0024.md
```diff
@@ -190,6 +192,127 @@ The deposit flow can only be fulfilled if the Stellar `Account` has established
...
+1. From there the anchor will continue the [deposit flow](#basic-anchor-implementation)
+
+##### Anchor Feature: Extend Single Historical Transaction endpoint For Claimable Balances
```
JakeUrban, line 284: They need to extend both `/transaction` and `/transactions`

ecosystem/sep-0024.md
```diff
@@ -569,6 +692,7 @@ Name | Type | Description
 `deposit_memo_type` | string | (optional) Type for the `deposit_memo`.
 `from` | string | Sent from address, perhaps BTC, IBAN, or bank account.
 `to` | string | Stellar address the deposited assets were sent to.
+`claimable_balance_id` | string | (optional) ID of the Claimable Balance containing the asset, included when an anchor is issuing the asset via a Claimable Balance.
```
JakeUrban, line 695: I'm not sure what that second part means
```suggestion
`claimable_balance_id` | string | (optional) ID of the Claimable Balance used to send the asset initially requested. 
```

JakeUrban (changes requested):
Lots of copy-editting changes, but overall I like the structure. Just make sure we're not missing any places in the rest of the doc where things should be adjusted.

ecosystem/sep-0024.md
Zagan202: I'll change it to something more appropriate
```suggestion
This feature provides less friction for Stellar accounts who wish to make a deposit external assets with an anchor without making a trust line before the deposit is complete.
```

ecosystem/sep-0024.md
Zagan202: Yeah thats a cleaner way to split the options.

ecosystem/sep-0024.md
Zagan202: I'm more in favor of leaving it in order to cover our bases. 
It is kinda verbose but this was info I had to figure out so wouldn't be better to share it in the SEP?
Or is there a better place to communicate this info?

Zagan202: Im going to take it out and shorten it significantly.
The more I look at it the more I see how much of a mouthful it is.

Zagan202:
Is it `trust line` or `trust-line`?
I want to do a find and replace in this doc to find a standard and stick to it.

JakeUrban:
> Is it `trust line` or `trust-line`?
> I want to do a find and replace in this doc to find a standard and stick to it.

Its`trustline`, one word.  I kinda like `trust-line` but I'd just stick to the original usage.

ecosystem/sep-0024.md
Zagan202: The way I intended to frame `If the anchor we're depositing to **supports** claimable balances.` and `If the anchor we're depositing to ** doesn't support** claimable balances.` is from the wallet's perspective they do not know if the anchor their issuing a deposit to is `claimable_balance_supported` or not, they're just shooting their shot. 
Let me explain using the demo wallet implementation as an example as far as I understand it.

1. Wallet issues a deposit for a user (no trustline to depositing asset) and sends the `claimable_balance_supported` POST request param.
2. The wallet listens for the anchor's callback for the transaction status.
3. Until the transaction is completed we communicate its pending.
From the Wallet logs
```
To collect the interactive information we launch the interactive URL in a frame or webview, and await payment details from a postMessage callback
Transaction status pending...
Transaction status complete
```
If the anchor supports claimable balances it will do all the things it needs to do to process the claimable balance transaction.
Then mark it as complete

If the anchor doesnt support claimable balances (lets say it ignores the param) 
The wallet will keep listening to the transaction status from the callback.
```
To collect the interactive information we launch the interactive URL in a frame or webview, and await payment details from a postMessage callback
Transaction status pending...
Transaction status pending...
Transaction status pending...
```
On the interactive window from `sep24/transaction/more_info` we'll see the status `waiting for trust-line to be established`
from there the wallet is busy waiting (like before) but this time its because it `pending_trust`  until the user makes that trustline.
I reproduced this by sending `claimable_balance_supported: false` from my wallet and letting polaris(`316-claimable-balance-attempt2`) do it's trustline flow.

This is the wallet's callback loop 
```
window.onmessage = ({ data: { transaction } }) => {
      if (transaction.status === 'completed') {
        this.updateAccount()
        this.logger.instruction('Transaction status complete')
        finish()
      } else {
        this.logger.instruction('Transaction status pending...')
        setTimeout(() => {
          const urlBuilder = new URL(transaction.more_info_url)
          urlBuilder.searchParams.set('callback', 'postMessage')

          popup.location.href = urlBuilder.toString()
        }, 1000)
      }
    }
```

ecosystem/sep-0024.md
Zagan202: I think I prefer to link both the documentation and showcase example code for example.
Similar to [SEP 20](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0020.md#example-code-for-linking)
Thoughts?

```suggestion
##### Wallet Feature: Claim Account Claimable Balance

It's recommended that Wallets provides a UI to enable accounts to claim their balances [polled](#wallet-feature-claimable-balance-polling)

[Example on how to claim a balance](https://developers.stellar.org/docs/glossary/claimable-balance/#claim-claimable-balance)
```javascript
const transaction = new TransactionBuilder(account, {
      fee: 100,
      networkPassphrase: this.network_passphrase,
    })
      .addOperation(
        Operation.claimClaimableBalance({
          balanceId,
        })
      )
      .setTimeout(0)
      .build()
    transaction.sign(keypair)
    this.logger.request(
      'Submitting claimClaimableBalance transaction',
      transaction
    )
    const result = await this.server.submitTransaction(transaction)
```
```

File: ecosystem/sep-0024.md (hunk 33 above)
Comment (Zagan202) [reply ecosystem/sep-0024.md:195]: How does the description read now?

File: ecosystem/sep-0024.md [hunk 47]
```diff
@@ -170,26 +171,183 @@ There are several possible kinds of response, depending on whether the anchor ne
...
+2. When the transaction status becomes `pending_user_transfer_start` send the required payment as described in the interactive webapp callback or the `/transaction` endpoint.
+3. The anchor will ignore the `claimable_balance_supported` field, discover the account does not have a trustline, and report transaction status as `pending_trust`.
+4. At this point both the wallet and anchor are waiting on the trustline to be established in order to continue the deposit payment. This is described here [steps 4-6](#stellar-account-doesnt-trust-asset)
+
+##### Wallet Feature: Claimable Balance Polling
+
+Polling allows the wallet to detect claimable balances and provide a UI to users that there is a balance available to claim. See the ["Claim Account Claimable Balance"](#wallet-feature-claim-account-claimable-balance) section to learn more about how to claim a balance.
+There are a few options as to how a wallet can poll and surface claimable balance information for their users.
+1. The wallet can poll the anchor's [transaction history endpoints](#anchor-feature-extend-transaction-history-endpoints-for-claimable-balances)
+
```
JakeUrban, line 256:
```suggestion
```

ecosystem/sep-0024.md
```diff
@@ -170,26 +171,183 @@ There are several possible kinds of response, depending on whether the anchor ne
...
+2. The wallet can poll Horizon's [`/claimable_balances/`](https://developers.stellar.org/api/resources/claimablebalances/list/) endpoint (or Horizon's [`/claimable_balances/`](https://developers.stellar.org/api/resources/claimablebalances/single/) single endpoint) to pull any available claimable balances for the claimant.
+
```
JakeUrban, line 258:
```suggestion
```

ecosystem/sep-0024.md
JakeUrban: That works, but note your current code is missing a change trust op. In order to accept a claimable balance the user needs to have a trustline to the asset. The tricky piece is that the change trust op only needs to be included once per asset. You should explain this nuance in this section before showing the code.

ecosystem/sep-0024.md
```diff
@@ -170,26 +171,183 @@ There are several possible kinds of response, depending on whether the anchor ne
...
+   * The single endpoint needs the `claimable_balance_id` in order to retrieve the claimable balance info.
+     * This can be obtained by extracting the `claimable_balance_id` from `results_xdr` of the deposit transaction submitted if the anchor does not extend their [transaction history endpoints](#anchor-feature-extend-transaction-history-endpoints-for-claimable-balances)
+
+##### Wallet Feature: Claim Account Claimable Balance
+
+It's recommended that Wallets provides a UI to enable accounts to claim their balances [polled](#wallet-feature-claimable-balance-polling)
+
+[Example on how to claim a balance](https://developers.stellar.org/docs/glossary/claimable-balance/#claim-claimable-balance)
+Accounts will still need to establish trustline to the asset before claiming the outstanding balance.
```
JakeUrban, line 267: Make sure you remove this code if you commit your suggestion:
```suggestion
```

ecosystem/sep-0024.md
```diff
@@ -170,26 +171,183 @@ There are several possible kinds of response, depending on whether the anchor ne
...
+
+##### Anchor Claimable Balance Flow
+
+1. Wallets make a request to `/transactions/deposit/interactive` and provide the `claimable_balance_supported` request parameter.
+2. Anchors mark the transaction as [`claimable_balance_supported=true`](#deposit)
```
JakeUrban, line 272: I think this needs to be more explicit.
```suggestion
2. Anchors update their internal database record of the transaction to indicate the wallet supports receiving a claimable balance
```

ecosystem/sep-0024.md
```diff
@@ -170,26 +171,183 @@ There are several possible kinds of response, depending on whether the anchor ne
...
+3. Anchors detect that the user's Stellar account doesn't have a trustline.
+4. The anchor should then report transaction status as `pending_anchor` and submit a Stellar transaction containing a create claimable balance operation
```
JakeUrban, line 274: I think we can remove this line because:
1. The transaction status should already be in `pending_anchor` while the anchor checks for a trustline, and
2. The correct status once the anchor begins submitting the transaction is `pending_stellar`.

ecosystem/sep-0024.md
```diff
@@ -170,26 +171,183 @@ There are several possible kinds of response, depending on whether the anchor ne
...
+It's recommended that Wallets provides a UI to enable accounts to claim their balances [polled](#wallet-feature-claimable-balance-polling)
+
+[Example on how to claim a balance](https://developers.stellar.org/docs/glossary/claimable-balance/#claim-claimable-balance)
+Accounts will still need to establish trustline to the asset before claiming the outstanding balance.
+
+##### Anchor Claimable Balance Flow
+
+1. Wallets make a request to `/transactions/deposit/interactive` and provide the `claimable_balance_supported` request parameter.
+2. Anchors mark the transaction as [`claimable_balance_supported=true`](#deposit)
+3. Anchors detect that the user's Stellar account doesn't have a trustline.
```
JakeUrban, line 273: Can we add a step before this one like `User sends funds to the anchor's off-chain account`?

ecosystem/sep-0024.md
```diff
@@ -170,26 +171,183 @@ There are several possible kinds of response, depending on whether the anchor ne
...
+
+##### Anchor Feature: Extend Transaction history endpoints For Claimable Balances
+
+Wallets query the [transaction endpoint](#transaction-history) to see and display the status of deposits.
+Anchors should extend their transaction(s) endpoint response to include the `claimable_balance_id` that the anchor added to the [transaction field](####Fields-for-deposit-transactions).
+This will help wallets to provide UI to showcase that their user's deposit was successful and theres an [available claimable balance for them.](#wallet-feature-claimable-balance-polling)
+
+```
+GET TRANSFER_SERVER_SEP0024/transactions
+```
```
JakeUrban, line 285: I don't think these examples are necessary, we already give json examples in the transaction section. Just add `null` and the non-null value to those examples. Anchors who don't implement claimable balances will either not add the attribute or keep it `null`.

ecosystem/sep-0024.md
```diff
@@ -170,26 +171,183 @@ There are several possible kinds of response, depending on whether the anchor ne
...
+      "claimable_balance_id": "00000000c2d8c89264288dbde8488364fd3fd30850fd4e7fbf6d1e9809702558afa4fdea",
+    }
+}
+```
+
+##### Anchor Claimable Balance Claimants and Predicates
+
+ Predicates are one of the [Claimable balance parameters](https://developers.stellar.org/docs/glossary/claimable-balance/#parameters) used when craftting a Claimable Balance transaction.
+They're used as conditions to meet in order for the recipient to claim their balance.
+      It's recommended to use the `UNCONDITIONAL` predicate to allow the claimant to claim the balance without any barriers.
```
JakeUrban, line 349: Remove the indent. I've noticed you like putting 1 sentence per line too, which is fine I guess but it may not be necessary. There's nothing wrong with multi-sentence lines.

JakeUrban (changes requested):
The content is much improved! Most of my comments are about formatting, copy editing, or simply removing unneeded content.

ecosystem/sep-0024.md
```diff
@@ -170,26 +171,183 @@ There are several possible kinds of response, depending on whether the anchor ne
...
+
+**Wallets**: To support claimable balances wallets must
+
+* Send `claimable_balance_supported` request parameter in their POST /transactions/deposit/interactive request.
+
+```json
+{
+"claimable_balance": true
+}
+```
```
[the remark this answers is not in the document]
Zagan202: `commit suggestions` is sus sometimes haha

ecosystem/sep-0024.md
Zagan202: Remove only this correct?
```
[Example on how to claim a balance](https://developers.stellar.org/docs/glossary/claimable-balance/#claim-claimable-balance)
```

JakeUrban: No, I think you need to introduce the "the account needs to have a trustline prior to claiming a balance" before you show the code.

ecosystem/sep-0024.md
```diff
@@ -170,26 +171,128 @@ There are several possible kinds of response, depending on whether the anchor ne
...
+**NOTE** in the *future* the following claimable balance implementations will be mandatory in order to be SEP-0024 interoperable.
+
+**Wallets**: To support claimable balances wallets must
+
+* Send the additional `claimable_balance_supported` request parameter in the POST /transactions/deposit/interactive request body.
+
+```json
+{
+  ...
+  "claimable_balance": true
```
tomerweller, line 210: Shouldn't this be `claimable_balance_supported` ?

ecosystem/sep-0024.md
```diff
@@ -170,26 +171,128 @@ There are several possible kinds of response, depending on whether the anchor ne
...
+Wallets need to be interoperable with anchors that have not adopted the claimable balance feature and vice versa.
+
+##### Wallet Claimable Balance Flow
+
+Wallets follow the [typical deposit flow](#Basic-Wallet-Implementation) but will provide the `claimable_balance_supported` request parameter in their POST parameters for `/transactions/deposit/interactive`.
+
+```json
+{
+  ...
+  "claimable_balance": true
```
tomerweller, line 232: Shouldn't this be `claimable_balance_supported` ?

ecosystem/sep-0024.md
```diff
@@ -170,26 +171,128 @@ There are several possible kinds of response, depending on whether the anchor ne
...
+
+Wallets query the [transaction endpoint](#transaction-history) to see and display the status of deposits.
+Anchors should extend their transaction(s) endpoint response to include the `claimable_balance_id` that the anchor added to the [transaction field.](####Fields-for-deposit-transactions)
+This will help wallets detect that the user's deposit was successful and theres an [available claimable balance for them.](#wallet-feature-claimable-balance-polling)
+
+##### Anchor Claimable Balance Claimants and Predicates
+
+Predicates are one of the [claimable balance parameters](https://developers.stellar.org/docs/glossary/claimable-balance/#parameters) used to craft a Claimable Balance transaction. They're conditions to meet in order for the recipient to claim their balance.
+
+It's recommended to use the `UNCONDITIONAL` predicate to allow the claimant to claim the balance without any barriers.
```
tomerweller, line 294: Why is putting some sane predicates (collect within 30 days, for example) not recommended?

tomerweller (approved):
LGTM. just minor nits. 

This PR contains a lot of formatting changes outside the context of claimable balances. In the future these types of changes should probably be in a separate PR.

ecosystem/sep-0024.md
Zagan202: If the account doesn't claim the balance in 30 days the balance unfortunately doesn't refund back into the sender's account. It's lost for both parties.
In order to provide a friendlier experience for the claimant/receiver its best not to restrict them from when they want to establish their trustline and claim.

If the claimant cannot claim the balance due to time constraints or whatnot then we would have to recommend the following for Wallets and Anchors

- Wallets need to provide more aggressive messaging to their users that there's a deadline to claim their balance
- Wallets and Anchors need to provide ways to refund deposits in the scenario where a claimablebalance is dead on the DEX
- (or) Anchors need refund themselves by issuing more of their asset on the network to send another claimable balance to the accounts that forget to claim in 30 days

JakeUrban: I think the simplest option would to have the sponsor list themselves as a claimant, and claim the balance if it's outstanding for too long.

JakeUrban (approved):
LGTM!

ecosystem/sep-0024.md
Zagan202: I think thats an interesting idea.
Maybe we should see what the community thinks over time.

tomerweller: I prefer not having a recommendation rather than recommending something that we're not sure is the best practice

tomerweller: we can leave it up for the anchor for now

Zagan202: @tomerweller  what do you think of this?
```suggestion
Anchors are free to set whichever predicates they feel are necessary in order to claim the balance.
If there are no predicate preferences, `UNCONDITIONAL` allows accounts to claim balances at anytime.
```

tomerweller: LGTM!

PR #758, opened 2020-10-29
SEP-0008: Add additional optional action params for the action required response

shredding:
We have implemented SEP-8 at DSTOQ, however for regulatory reasons (e.g. we need to inform users on actions such as stock splits or a change in terms) we require the email and have build a webflow that is a bit cumbersome to the user, as they have to verify an email (meaning that the webflow is already an app in app that in turn requires them to open a third app [mail client] that opens a webpage, after which they can go back to the webflow).

This backward compatible change would allow us to set this up behind the scenes if the client service does already have an email of the user. 

I added firstname/lastname for convenience. The public key may be required if the user sends a tx with multiple source accounts.

Finally, the synchronous action is interesting to give better feedback to the user, as (in our case) the required action is a KYC process that may take some time for us to complete.

Here's a sample of our flow if you want to try it:

```python
from stellar_sdk import TransactionBuilder, Server, Network
import requests
import json 

STELLAR_TOML = 'https://dsq.technology/.well-known/stellar.toml'
APPROVAL_SERVER = 'https://dsq.technology/sep-8/tx-approve/'
PUBLIC_KEY = 'GDIMI5MOM5KKB7RGAW3SB5KJOEDOCNHDGS3ANS3BL55MTJ6OTUSJGVLE'

server = Server(horizon_url='https://horizon.stellar.org')
builder = TransactionBuilder(server.load_account(PUBLIC_KEY), Network.PUBLIC_NETWORK_PASSPHRASE)

tx_xdr = builder.append_manage_buy_offer_op(
    selling_code='USD',
    selling_issuer='GDUKMGUGDZQK6YHYA5Z6AY2G4XDSZPSZ3SW5UN3ARVMO6QSRDWP5YLEX',
    buying_code='TSLA',
    buying_issuer='GBRDHSZL4ZKOI2PTUMM53N3NICZXC5OX3KPCD4WD4NG4XGCBC2ZA3KAG',
    amount='1.00',
    price='402.9721521'
).build().to_xdr()

print(json.dumps(requests.post(APPROVAL_SERVER, data={'tx': tx_xdr}).json(), indent=2))
```

Prints:

```json
{
  "status": "action_required",
  "message": "Register your account now to start trading securities. It's easy!",
  "action_url": "https://api.dstoq.com/accounts/webflow?publicKey=GDIMI5MOM5KKB7RGAW3SB5KJOEDOCNHDGS3ANS3BL55MTJ6OTUSJGVLE"
}
```


ecosystem/sep-0008.md
```diff
@@ -126,6 +126,8 @@ Name | Data Type | Description
 status|string|`"action_required"`
 message|string|A human readable string containing information regarding the action required.
 action_url|string|A URL that allows the user to complete the actions required to have the transaction approved.
+action_params|string|A comma separated list of additional parameters that the client may optionally provide to the approval service; options are `first_name`, `last_name`, `email` and `public_key`
```
leighmcculloch, line 129: Are these params query parameters or fields to be POSTed to the URL?

Given this list of params is likely to expand, and we might need to specify descriptions for each, can we start a table that lists the params and reference that table here?

Or, if this list is likely to overlap significantly with [SEP-9](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0009.md), I think we should reference that SEP and support any of those fields, using the field names there for consistency.

We could also probably transition towards using [SEP-12](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0012.md) since that SEP provides greater control for a client to maintain and delete personal information, although that might be a breaking change and something we'd need to work towards.

ecosystem/sep-0008.md
```diff
@@ -117,7 +117,7 @@ message|string|A human readable string containing information to pass on to the
 
 #### Action Required Response
 
-An Action Required response will have a `200` HTTP status code and `action_required` as the `status` value. It means that the user must complete an action before this transaction can be approved. The approval service will provide a URL that facilitates the action. Upon completion, the user will resubmit the transaction.
+An Action Required response will have a `200` HTTP status code and `action_required` as the `status` value. It means that the user must complete an action before this transaction can be approved. The approval service will provide a URL that facilitates the action. Upon completion, the user will resubmit the transaction if it's synchronous (the default) or wait for a confirmation from the approval service to continue.
```
leighmcculloch, line 120: What does the waiting process look like? Does a client wait for some signal? What is that signal?

ecosystem/sep-0008.md
```diff
@@ -126,6 +126,8 @@ Name | Data Type | Description
 status|string|`"action_required"`
 message|string|A human readable string containing information regarding the action required.
 action_url|string|A URL that allows the user to complete the actions required to have the transaction approved.
```
leighmcculloch, line 128: Today it's assumed the client will open the URL in a browser, but providing these options might need to be POSTed to keep sensitive information out of the URL query params. We probably need to specify what request format this URL should be in the case actions params are provided.

leighmcculloch (commented):
This change makes sense to me, if a client already has this information this would support prefilling the data. In your mind would the client still display the actional URL in a browser so the user would still interact or is the goal for this to support a non-interactive flow?

ecosystem/sep-0008.md
shredding: Yeah, i thought that the client would add those to the query params, as it's currently exactly like that (e.g. as you can see from our sample we are already adding the public key, taken from the tx.

ecosystem/sep-0008.md
shredding: In our case, the user get's an email that she has passed kyc and then she can trade. The client is not involved here. It basically means that afterwards the client does nothing.

ecosystem/sep-0008.md
shredding: I thought about that as well, in our case it would be overhead. Yeah, the naming should be consistent with SEP-9; for us SEP-12 does not make sense as we have our own KYC Flow.

shredding:
> In your mind would the client still display the actional URL in a browser so the user would still interact or is the goal for this to support a non-interactive flow?

For us it has to be interactive. We have integrated a KYC Provider (https://passbase.com/) and have to pipe users through their Flow. I want to strip down the Webflow to this, essentally.

ecosystem/sep-0008.md
leighmcculloch: Got it, so the app if it is synchronous would return to resending the same request again and retrying after the action required step, where-as if it is not synchronous the app would stop. Is that an accurate understanding?

shredding: That's up to the implementation, I'm not sure how current clients handle that, but if it's synchronous, then yes, the client could do that. 

Maybe we can leave that out for now as well, I'm not aware of other SEP-8 implementations so I'm not sure if that is actually a thing.

leighmcculloch:
>pipe users through their Flow. I want to strip down the Webflow to this, essentally.

Got it, so the goal with this is to optimize the experience for the user so that the data you already possess is pre-filled?

shredding:
I'm even aiming to not only prefill, but getting rid of that entire flow and just give the user KYC. 

Technically someone could fake an Email by using SEP-8 without a client that has the Email verified; but that's not really a usecase, as we do a full blown KYC Process with Idenity Check, Selfie and Passport Verification afterwards.

ecosystem/sep-0008.md
leighmcculloch: I don't see a problem with adding the public key to the URL but the name and email could be leaked in other places by including it there. URLs are not always secure as they get easily propagated to logs, ad tracking, analytics systems, etc, so it would be ideal to keep these details in a POST either by submitting them in the request body as a POST separately and getting back a URL with an ID referencing that data, or by using a POST at this point although that might make it harder to implement in some cases such as mobile.

SEP-12 would help provide that parallel delivery of data but it sounds like that would be significantly different to what you're doing today. I think we should probably explore that. Return of a list of data fields needed could be delivered separately as the custom, and then the action URL is loaded for the user. Prior to SEP-12 the flow would probably need SEP-10 to authenticate as the account first which could be seamless.

shredding: Those are valid points, especially with the POST request.

SEP-12, as I understand it, has only `NEEDS_INFO`, which isn't the case here. We can work without E-Mail by collecting it for ourself (not all clients may have that info at hand), so it's optional.

leighmcculloch: Got it, so what you're saying is if you don't provide this information then it would be okay, the interactive browser page would then ask for it so it isn't the same thing because this is prefilling not requiring data, and if all the data is prefilled it can be skipped, and if there are fields still required the page opened can collect it. Is that accurate?

dmgmyza:
@leighmcculloch @shredding 
Hi, from the wallet side - we also support the idea of adding email/name as a POST parameters. That works for us.

ecosystem/sep-0008.md
leighmcculloch: Okay got it. If it works for you, let's change this to reference SEP-9 and say any parameters there may be requested as params. Also, let's refer to them as fields because that's what SEP-9 refers to them as.
```suggestion
action_fields|string| (Optional) A comma separated list of additional fields defined by [SEP-9](sep-0009.md) that the client may optionally provide to the approval service.
```

ecosystem/sep-0008.md
leighmcculloch: @shredding @dmgmyza and I discussed this offline and an alternative approach where:

Add a new optional response field, `action_method`. The field has three valid states:
- Not provided, interpreted the same as if it is given the value `GET`
- `GET` – the action URL expects to receive a GET request, any fields requested may be passed as query parameters
- `POST` – the action URL expects to receive a POST request, any fields requested may be passed url-form-encoded or as JSON

Action URLs providing a `POST` method should response to the POST with a HTTP response with status code 301 and a `Location` header that is a URL where the user can complete the required actions with all of the POST data pre-filled or already accepted.

This approach supports products entirely in a web browser because the browser will follow the 301 redirect and will prevent odd UI where a user for some reason refreshes the page, they won't see warnings about they are refreshing a page that is accepting post data.

This approach supports products that are mobile apps loading the URL in a system browser because the app can make the POST request internally, read the location out of the response and provide it to a browser to load.

This approach is additive, any existing implementation would see no change. Any existing implementations that are providing optional fields/params as query parameters would be compatible. Any new implementations that plan to provide personal data or sensitive data could choose to use the POST method.

This approach does put some burden on the client to support different approaches though which is not great. We could encourage the POST approach for new implementations.

leighmcculloch (commented):
@shredding @dmgmyza Does the changes proposed below align with what you have in mind?

leighmcculloch:
@shredding Would you like to update this proposal to match? (I'm also happy to open the proposal when I'm back in the office next week if you'd prefer.)

> can we add `action_auth_required`, defaulting to false and if true, the client must authenticate with sep-10 and by client I mean the wallet provider. E.g. we could add Lobstr as a registered wallet to our backend.

I think we should take that change to a separate PR. SEP-10 does not currently support identifying the wallet, although it may soon, or there may be some other SEP that specifies that. I'd suggest for the sake of simplicity returning a HTTP status code 401/403 in the case that the API requires auth from a specific wallet. A wallet like Lobstr could for the moment always auth with your server. Would that work for the moment? or is there a case a wallet won't know it needs to auth until after first making the request to your server?

JakeUrban:
Hey y'all, I haven't been involved in this PR but @shredding I'm curious to know if you agree with the problem outlined in  #773. Its about clarifying the structure of transactions submitted to approval servers.

shredding:
You are right. For the moment and in the light of getting things done, we can just leave it as you proposed and we'll do this in a separate PR. We are working on a second integration right now and I'll check how we'll do it there and come up with a combined solution.

shredding:
(@JakeUrban thanks for pointing me to that, i answered in the thread)

shredding:
Are we set here? I would like to implement that in the next week.

leighmcculloch:
> Are we set here? I would like to implement that in the next week.

@shredding I've opened a PR against this PR with the changes I proposed already above (https://github.com/stellar/stellar-protocol/pull/758#discussion_r518319840): https://github.com/DSTOQ/stellar-protocol/pull/1. If you're happy with that change can you merge it into this PR?

leighmcculloch:
>We currently plan to give back a magic token and therefore auto-login the user afterwards.

>I would be fine with either SEP-10 or (preferred) Authentication Header via HTTP.

@shredding I think we should consider SEP-10 in your case so that the client is authing independent of the regulated transaction, with the send of the transaction to the approval service being authenticated. There's a PR that's still a WIP for SEP-10 that in a non-breaking way adds verifying the client application's identity independent of the account, that's #760. Could you take a look at it and let me know what you think?

---

Independent of this issue does this change still work: https://github.com/stellar/stellar-protocol/pull/758#issuecomment-738527177? or does it need adapting?

howardtw:
@shredding I just merged https://github.com/stellar/stellar-protocol/pull/813 that improves the readability of SEP-8 without changing any of the current functionalities. The conflict should be fairly straightforward to resolve. I can help doing it if you'd like to.

To surface @leighmcculloch 's question in [this](https://github.com/stellar/stellar-protocol/pull/758#issuecomment-743387783) comment:
> Independent of this issue does this change still work: #758 (comment)? or does it need adapting?

Do you see this PR being valuable if you plan on pursuing SEP-10?

leighmcculloch:
>Do you see this PR being valuable if you plan on pursuing SEP-10?

@howardtw I can't find where this was discussed now, but there's no need to authenticate the user any longer, or at least for this change.

ecosystem/sep-0008.md
```diff
@@ -225,17 +225,44 @@ Name | Type | Description
...
+   -----|------|------------
+   `next_url` | string | A URL where the user can complete the required actions with all the parameters included in the original POST pre-filled or already accepted.
+
+   Example:
+
+   ```json
+   {
+     "next_url": "https://..."
+   }
+   ```
```
leighmcculloch, line 264: This section provides clear instruction for what the action_url's response message should contain, but is not clear about how the request should be formatted or constructed. There are a couple different approaches that an implementer could take, either using SEP-12 for inspiration, or using each SEP-9 field as a key in a flat object, or using each SEP-9 field as a hierarchy.

@shredding @dmgmyza Could you share how the request is formatted in your implementation?

I'll update this PR with the additional details once I hear from you.

leighmcculloch (commented):
This PR is almost in a place where it can be merged.

After doing a final review I realize it is missing some details we need to add to make sure that all clients and servers implementing this new functionality, go about it the same way, and that interoperability is achieved. See comment inline below.

ecosystem/sep-0008.md
leighmcculloch: I've opened PR https://github.com/DSTOQ/stellar-protocol/pull/3 that contains additional details.

ecosystem/sep-0008.md
```diff
@@ -225,17 +225,56 @@ Name | Type | Description
...
+
+   Any fields defined by [SEP-9 Standard KYC / AML fields].
+
+   Request Example:
+
+   ```json
+   {
+     "email_address": "user@example.com"
+   }
+   ```
```
leighmcculloch, line 262: Note: This portion discussing the request format does not discuss how to send SEP-9 fields that have sub-fields. I'm intentionally leaving that undefined until I get an understanding of how SEP-12 works, so that I can make SEP-8 consistent with SEP-12. Related discussion: https://github.com/stellar/stellar-protocol/issues/832

ecosystem/sep-0008.md
```diff
@@ -225,17 +225,56 @@ Name | Type | Description
 `status` | string | `"action_required"`
 `message` | string | A human readable string containing information regarding the action required.
 `action_url` | string | A URL that allows the user to complete the actions required to have the transaction approved.
+`action_method` | string | (Optional) `GET` or `POST`, indicating the type of request that should be made to the `action_url`. If not provided, `GET` is assumed.
```
howardtw, line 228: nit: I think in other places of this doc, we use `(optional)` instead of `(Optional)`. Maybe it's worth making it consistent.

ecosystem/sep-0008.md
```diff
@@ -225,17 +225,56 @@ Name | Type | Description
...
   "message": "Please sign the terms of service via the provided URL.",
-  "action_url": "https://..."
+  "action_url": "https://...",
+  "action_method": "POST",
+  "action_fields": ["email_address", "mobile_number"]
 }
 ```
 
+###### Action URL
+If the `action_method` is provided the client uses the type of request indicated when sending a request to the `action_url`. If `action_fields` are to be provided and the fields contain sensitive or personal information the `POST` case is preferred to ensure sensitive information is not transmitted in the URL.
```
howardtw, line 244: Adding a comma helps me understand it better.

```suggestion
If the `action_method` is provided, the client uses the type of request indicated when sending a request to the `action_url`. If `action_fields` are to be provided and the fields contain sensitive or personal information, the `POST` case is preferred to ensure sensitive information is not transmitted in the URL.
```

ecosystem/sep-0008.md
```diff
@@ -225,17 +225,56 @@ Name | Type | Description
...
+
+- In the case that `action_method` is `GET` any fields requested may be passed as query parameters in the `action_url`.
```
howardtw, line 246:
```suggestion
- In the case that `action_method` is `GET`, any fields requested may be passed as query parameters in the `action_url`.
```

ecosystem/sep-0008.md
```diff
@@ -225,17 +225,56 @@ Name | Type | Description
...
+
+- In the case that `action_method` is `POST` any fields requested may be passed as JSON in the request body, using content type `application/json`. The server should respond to the `POST` with a HTTP response with status code `200`, content-type `application/json` and a body containing the next URL that the user should be taken to.
```
howardtw, line 248:
```suggestion
- In the case that `action_method` is `POST`, any fields requested may be passed as JSON in the request body using content type `application/json`. The server should respond to the `POST` with a HTTP response with status code `200`, content-type `application/json` and a body containing the next URL that the user should be taken to.
```

howardtw (approved):

ecosystem/sep-0008.md
leighmcculloch: https://github.com/DSTOQ/stellar-protocol/pull/4

ecosystem/sep-0008.md
leighmcculloch: https://github.com/DSTOQ/stellar-protocol/pull/4

ecosystem/sep-0008.md
leighmcculloch: https://github.com/DSTOQ/stellar-protocol/pull/4

ecosystem/sep-0008.md
leighmcculloch: https://github.com/DSTOQ/stellar-protocol/pull/4

leighmcculloch (commented):
@howardtw I have addressed your comments in https://github.com/DSTOQ/stellar-protocol/pull/4.

@accordeiro Also shared with me feedback about the action_url response and highlighted that if the `POST` happens to provide all the details required and the server doesn't need to display a terms of service or anything else to the user then there is no need to share a next_url and there should be a way for the action_url to indicate an early exit. I've added that case. I've made the change in https://github.com/DSTOQ/stellar-protocol/pull/4.

@shredding @dmgmyza Is https://github.com/DSTOQ/stellar-protocol/pull/4 acceptable to you?

@shredding I don't think this needs to change your implementation much other than adding in the `"result": "follow_next_url"` to your action_url response.

@dmgmyza For an existing integration with @shredding that never returns the terminal case you technically don't need to change anything for it to work, but it would be great to support the terminal case at some point if it makes sense to you.

leighmcculloch:
@dmgmyza confirmed by DM that this is also good from the perspective of a client who is calling the API.

shredding:
Thanks for your efforts @leighmcculloch

dmgmyza:
Looks good for us, works perfect in LOBSTR.
Thanks @leighmcculloch and @shredding

PR #759, opened 2020-10-29
SEP-0012: Make less about SEP-6

leighmcculloch:
### What
Remove mentions of SEP-6 from the summary and mention other SEPs that use SEP-12 instead of only SEP-6.

### Why
On a first read of it's easy for someone without context to think that SEP-12 is only intended for use with SEP-6, but it is intended to be used for any data collection that other SEPs might need to do with an authenticated user.


JakeUrban (approved):
This is great, thanks for clarifying this.

PR #760, opened 2020-10-29
SEP-10: Client Attribution

JakeUrban:
resolves stellar/stellar-protocol#685

Clients can pass a `client_domain` parameter to SEP-10 servers who can verify the client's identity by ensuring the challenge POSTed is signed by the `SIGNING_KEY`  from the client's SEP-1 TOML file.


JakeUrban:
@pitsevich @andywer this is the current WIP solution

JakeUrban:
I think I'm going to change the way the client indicates that it wants to be attributed to future requests to the anchor. The client should do this at the beginning of the flow with a `GET <WEB_AUTH_ENDPOINT>?client_domain=` request.

The anchor could then add the client domain to the challenge using a ManageData operation (now that we have v2.1). This makes things way nicer because the client domain is included in the cryptographically signed data instead of an additional POST request parameter.

The client can take the challenge that includes the domain passed to its backend service, which validates the client domain and signs the challenge (or generates the signature), returning it to the client. The client can verify the signature, add it to the challenge, and post it back to the server without any API schema changes. The server can then store the client domain value from the posted challenge in the JWT returned if the challenge is valid. The client then uses the JWT on each request and the anchor attributes it to the client by reading the `client_domain` JWT claim.

leighmcculloch:
I think the change you suggest sounds good.

I'm also wondering if a domain is necessary or if it is enough for the client to request an arbitrary claim be added for another account, and that other account signs, and the server sees that the other account confirms that claim. My understanding reading issue #685 is that this would support the second point:
>the ability to process users from specific wallet applications differently

It would satisfy that point because the anchor could distinguish the wallet from the account, domains aren't really necessary. I'm not convinced about this, just sharing some thoughts out loud.

andywer:
Looks good to me 👍

@leighmcculloch I didn't fully think it through, so I am not sure whether the domain is strictly needed either, but to me it feels like a reasonable choice to go by the domain, anyway. Might be easier to debug than handling the public keys directly. Another very minor argument: Domains lead to shorter JWTs 😅

JakeUrban:
@leighmcculloch unless each public key address had a `home_domain` set on the account, there would be no way to know who these public keys represent. The anchor would need to match public keys to client organizations, and this would only be possible if they're hard coded or if the account has a `home_domain` set.

Home domains are the de-facto way of identifying an organization (TOML file), and this is a natural extension of that principal. I think not using home domains would end up being a barrier to understanding what we're doing here.

JakeUrban:
This is ready for review now. No rush though

ecosystem/sep-0010.md
```diff
@@ -250,6 +257,25 @@ A server may choose to issue JWTs for less than all thresholds and based on any
 
 A server that needs to support validating accounts that do not exist can require a signature of the master key of the account address for accounts that do not exist.
 
+### Verifying Client Application Identity
+
+A server may want to attribute each HTTP request to a specific client application. For example, a server and client could make a deal to elimate fees for the users of that client. SEP-10 offers an optional variation to the flow outlined in the [Abstract](#abstract) to enable this use case.
```
leighmcculloch, line 262:
```suggestion
A server may want to attribute each HTTP request to a specific client application. For example, a server and client may agree to change fees for the users of that client. SEP-10 offers an optional variation to the flow outlined in the [Abstract](#abstract) to enable this use case. 
```

ecosystem/sep-0010.md
```diff
@@ -250,6 +257,25 @@ A server may choose to issue JWTs for less than all thresholds and based on any
 
 A server that needs to support validating accounts that do not exist can require a signature of the master key of the account address for accounts that do not exist.
 
+### Verifying Client Application Identity
+
+A server may want to attribute each HTTP request to a specific client application. For example, a server and client could make a deal to elimate fees for the users of that client. SEP-10 offers an optional variation to the flow outlined in the [Abstract](#abstract) to enable this use case. 
+
+In order to use this optional feature, the organization that owns the client application must host a [SEP-1 stellar.toml](sep-0001.md) file containing a `SIGNING_KEY` attribute. This allows the SEP-10 server to verify that the challenge sent by a client is signed by the `SIGNING_KEY`, proving that the organization takes ownership of the client application. The SEP-10 server can now attribute requests made with the resulting JWT to the organization that signed the challenge.
```
leighmcculloch, line 264:
```suggestion
In order to use this optional feature, the organization that owns the client application must host a [SEP-1 stellar.toml](sep-0001.md) file containing a `SIGNING_KEY` attribute. This allows the SEP-10 server to verify that the challenge returned by the client is signed with the `SIGNING_KEY`. This proves that the client is associated with the client domain. The SEP-10 server can now attribute requests made with the resulting JWT to the client domain that signed the challenge.
```

ecosystem/sep-0010.md
```diff
@@ -250,6 +257,25 @@ A server may choose to issue JWTs for less than all thresholds and based on any
...
+
+1. The client makes the initial `GET <WEB_AUTH_ENDPOINT>` request with an added `?client_domain=example.com` parameter and value
+1. The SEP-10 server generates the challenge transaction with an additional Manage Data operation containing the `<client_domain> attribution>` passed
```
leighmcculloch, line 267: Can you use the same pattern as defined in SEP-10 v3.1 where you use the key and value fields of the manage data operation instead of trying to bundle everything into the key? Make the key `client_domain`, and the value `exampe.com`.

ecosystem/sep-0010.md
```diff
@@ -250,6 +257,25 @@ A server may choose to issue JWTs for less than all thresholds and based on any
...
+
+1. The client makes the initial `GET <WEB_AUTH_ENDPOINT>` request with an added `?client_domain=example.com` parameter and value
+1. The SEP-10 server generates the challenge transaction with an additional Manage Data operation containing the `<client_domain> attribution>` passed 
+1. The client obtains a signature generated using its organization's `SIGNING_KEY` and adds it to the challenge transaction. How this happens is up to the client. However:
+  1. It is highly recommended that the client-side application does not hold the secret key for the `SIGNING_KEY`
+  1. Instead, an internal service should accept a request from the organization's client application, verify the challenge sent, and generate the signature if valid
+1. The client verifies the signature and adds the it to the challenge transaction if valid
```
leighmcculloch, line 271: I think you can drop this step, this is internal behavior of the client and there's nothing particularly dangerous about not verifying it here.
```suggestion
```

ecosystem/sep-0010.md
```diff
@@ -250,6 +257,25 @@ A server may choose to issue JWTs for less than all thresholds and based on any
...
+1. The client signs the challenge as it normally would with the user's Stellar account signer secret key(s)
+1. The client makes the `POST <WEB_AUTH_ENDPOINT>` request including a `client_domain` parameter containing the home domain of the client organization's `stellar.toml` file
```
leighmcculloch, line 273: Why do we need to include the client_domain in the POST request's URL? It should be inside the manage data operation. Or is that what you're meaning?
```suggestion
1. The client makes the `POST <WEB_AUTH_ENDPOINT>` request with the signed transaction
```

ecosystem/sep-0010.md
```diff
@@ -82,7 +82,8 @@ Request Parameters:
...
 ----------|---------------|------------
 `account` | `G...` string | The stellar account that the wallet wishes to authenticate with the server
-`home_domain` | string | (optional) The [fully qualified domain name](https://en.wikipedia.org/wiki/Fully_qualified_domain_name) of the service requiring authentication. SEP-10 signing servers used to generate tokens for multiple different web services can use this parameter to create valid tokens for each service. If not provided by the client, a default should be used.
+`home_domain` | string | (optional) The [fully qualified domain name](https://en.wikipedia.org/wiki/Fully_qualified_domain_name) of the service requiring authentication. SEP-10 signing servers used to generate tokens for multiple different web services can use this parameter to create valid tokens for each service. If not provided by the client, a default should be used
+`client_domain` | string | (optional) The [fully qualified domain name](https://en.wikipedia.org/wiki/Fully_qualified_domain_name) of the service that hosts the client organizations `stellar.toml` file. Only used for [Verifying Client Application Identity](#verifying-client-application-identity)
```
leighmcculloch, line 86: We should state here or somewhere that if the client_domain parameter is provided, but not supported by the server, the server should ignore the value but continue as per normal. It's important clients can start including the field without servers returning an error.

ecosystem/sep-0010.md
```diff
@@ -103,6 +104,8 @@ On success the endpoint must return `200 OK` HTTP status code and a JSON object
       * The value of key is the fully qualified domain name of the web service requiring authentication followed by `auth`. It can be at most 64 characters.
       * The value must be 64 bytes long. It contains a 48 byte cryptographic-quality random string encoded using base64 (for a total of 64 bytes after encoding).
     * zero or more `manage_data(source: server_account, ...)` reserved for future use
+      * If the client passed a `client_domain` parameter in the request, the challenge should include an operation like so:
```
leighmcculloch, line 107:
```suggestion
      * If the client passed a `client_domain` parameter in the request and the server supports [Verifying Client Application Identity](#verifying-client-application-identity), the challenge should include an operation like so:
```

ecosystem/sep-0010.md
leighmcculloch, line 267: We should add a note here that this behavior continues if the server supports Verifying Client Application Identity. If the server doesn't support it, it should continue with the ordinary flow.

leighmcculloch, line 267: We also should specify here what the source account should be.

ecosystem/sep-0010.md
```diff
@@ -138,18 +141,21 @@ To validate the challenge transaction the following steps are performed by the s
...
+  * verify that client signatures on the transaction are signers of the first operation's source account;
   * verify that client signatures are correct;
   * verify that client signatures provide weight that meets the required threshold(s), if any;
-* if the operation's source account does not exist:
+* if the first operation's source account does not exist:
   * verify that client signature count is one;
   * verify that client signature is correct for the master key of the account address;
 * verify that transaction containing additional Manage Data operations have their source account as the server signing key;
+* if the transaction contains a Manage Data operation with a key containing `attribution`:
+  * verify that the client domain included in the key has a `stellar.toml` file with a `SIGNING_KEY` used to sign the challenge. An internal mapping from client domain to SIGNING_KEY may be used.
```
leighmcculloch, line 158: I think we need to go into more detail here, can we model this off the similar change in SEP-10 v3.1 and be explicit about what each field of the operation needs to contain, especially the source account field?

leighmcculloch (changes requested):
👏🏻 I think this change does a great job of adding client domain as a claim to the SEP-10 process. There's a few things we need to tweak and make more clear, see below.

I think it would also be helpful if you merge the SEP-10 v3.1 change into this PR so that it is based on that change. That change adds a new manage data operation and sets a pattern we should use for others.

ecosystem/sep-0010.md
JakeUrban: I stated this in the parameter definition, but I don't think it needs to be specified in this section, since it's specifically for cases where the parameter is passed.

JakeUrban: Why is the source account important? The SEP currently just requests that source accounts not be null.

ecosystem/sep-0010.md
JakeUrban: I adjusted the Manage Data op definition, let me know if I need to make more adjustments

ecosystem/sep-0010.md
leighmcculloch: This transaction should follow the basic rules of Stellar transactions. The signature for an operation needs to be the signer of the source account of the operation. If the source account is the server account then the server should sign for it.

The rules about additional operations signed by the server is just to allow a server to include any additional information itself.

JakeUrban: Hmm ok I see where you're coming from, but I don't think we should try to apply these rules to challenge transactions, and I would argue that we're already breaking these rules with SEP-10 3.1's changes.

In a payment operation from account A to account B, I can have a transaction source account C, as long as A is the source account on the payment operation. Account A needs to sign the transaction along with account C because A's funds are being used and C's sequence number is being used. That all makes sense.

But in challenge transactions, how to go about applying these rules is not as clear. In the first Manage Data operation, the client's account is the source account, but the data is the home domain of the server's TOML file. You could argue that the client needs to verify the home domain included and signs the transaction to "approve" of the home domain -- this is why we updated to SEP-10 3.0.

But before SEP-10 3.0, the client's signature's only use was proving to the server the client controlled the Stellar account specified. My point here is that unlike in payment operations, the Manage Data operation was not the reason the client needed to sign the transaction.

My point is further reinforced by SEP-10 3.1's changes. The added operation uses the server's account as the source account. Why is the server the source account for the `web_auth_domain` but not the `home_domain`? Shouldn't the client account be the source account on this new op, since the reason we're introducing it is for the client to verify that the challenge originated from that domain?

All this is to say that instead trying to apply the signing rules of normal "submittable" transactions to challenge transactions, we should acknowledge that challenge transaction signatures prove a specific claim.
- the server's signature proves possession of the secret for the `SIGNING_KEY`'s on the `home_domain`'s TOML
- the user's signature proves the client is in possession of the secret for the Stellar account being authenticated
- the client's signature proves the client is in possession of the secret for the `SIGNING_KEY` on the `client_domain`'s TOML, verifying that the client wants to be attributed to the authentication request

leighmcculloch: The first operation needs to be an exception. It is how it is and future operations we add don't need to conform to it. I think these rules should be applied to how we define the subsequent operations.

I think the source account of an operation should otherwise indicate who is signing off on that claim to make it clear if the client is signing it, a third party, or simply the server. This is why SEP-2.1 included a statement that "if the transaction has other operations they are Manage Data operations that all have their source accounts set to the the server's account." The reason it makes that statement is if there are operations included that the client doesn't recognize, and they're not the servers, the client shouldn't sign in case the server thinks the client is signing off on them when it doesn't intend to.

Another issue is if the source account isn't included then the server is required to either store the information about who else is supposed to sign, or look that information up fresh when receiving the challenge.

Sticking it in the source account field makes it really clear who is signing for that operation.

What's the motivation for going a different route with this?

JakeUrban: I think the notion of _signing for an operation_ doesn't cleanly apply to special non-submittable transactions like challenges. For challenge transactions, both client and server sign to prove each party is who they say they are, not because a particular operation requires either party's signature.

But in this instance we agree that the source account is the server account for the attribution Manage Data op.

leighmcculloch: I think the source account for the attribution op should be the party signing to prove that claim, which is the third party. The SEP-10 server isn't proving that attribute.

JakeUrban: Oh I see, this requires the SEP-10 server to make request to `<client_domain>/.well-known/stellar.toml` so it can extract the `SIGNING_KEY`. But maybe this isn't a concern because Anchors could only attribute challenges to known `client_domains` for which they have an internal mapping to the client server's `SIGNING_KEY`.

leighmcculloch: Yup. The server needs the signing key of the third party anyway, to verify the signature on the challenge tx.

JakeUrban: I'm going to add a point about the above ^

JakeUrban:
Not stale

leighmcculloch (commented):
@JakeUrban This is looking great. I've left some comments throughout.
❗ – Things I think need addressing.
💡 – Suggestions.
❓ – Questions that may change the course of things.
💭 – Thoughts that do not need addressing in this change.

Also:
❗ Meta information needs updating, version, etc.

ecosystem/sep-0010.md
```diff
@@ -122,7 +123,10 @@ On success the endpoint must return `200 OK` HTTP status code and a JSON object
     * subsequent operations, order unimportant:
       * `manage_data(source: server_account, key: 'web_auth_domain', value: web_auth_domain)`
         * The value is the web auth domain, the domain hosting the web auth endpoint and that the client connected to when requesting the challenge transaction. It can be at most 64 characters.
-      * zero or more `manage_data(source: server_account, ...)` reserved for future use
+      * (optional) `manage_data(source: client_signing_key, key: 'client_domain', value: client_domain)`
+        * Only add this operation if the client passed a `client_domain` parameter in the request and the server supports [Verifying Client Application Identity](#verifying-client-application-identity)
+        * The `client_signing_key` is the `SIGNING_KEY` from the `client_domain`'s [SEP-1 stellar.toml](sep-0001.md)
```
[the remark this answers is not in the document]
JakeUrban: I replaced `client_signing_key` with `client_domain_signing_key`, but I think we should keep `client_domain` as-is. SEP-10 defines 'home domain' to be the domain of the service accepting JWT tokens in the abstract.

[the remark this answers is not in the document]
JakeUrban: This is where my [Clarifying Participants PR](#851) would be helpful. Currently, 'client' refers to both the user who controls the account as well as the application the user is using.

> - A client who will authenticate using a Stellar account.

This is how the user/client is defined in the abstract. Its a bit confusing to add the term `client domain` or `client domain signing key` when the user obviously doesn't have a domain.

ecosystem/sep-0010.md
```diff
@@ -162,15 +166,20 @@ To validate the challenge transaction the following steps are performed by the s
   * has a non-null source account
   * has the home domain as the key of the operation
 * verify that transaction envelope has a correct signature by server's signing key;
-* if the operation's source account exists:
+* verify that transaction operations:
+  * are all [Manage Data](https://www.stellar.org/developers/guides/concepts/list-of-operations.html#manage-data) operations
+  * have non-null source accounts
```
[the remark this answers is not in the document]
JakeUrban: This comment must have been made before #868 was merged? Resolving

ecosystem/sep-0010.md
```diff
@@ -271,6 +281,29 @@ A server may choose to issue JWTs for less than all thresholds and based on any
 
 A server that needs to support validating accounts that do not exist can require a signature of the master key of the account address for accounts that do not exist.
 
+### Verifying Client Application Identity
+
+A server may want to attribute each HTTP request to a specific client application. For example, a server and client may agree to change fees for the users of that client. SEP-10 offers an optional variation to the flow outlined in the [Abstract](#abstract) to enable this use case. 
+
+In order to use this optional feature, the organization that owns the client application must host a [SEP-1 stellar.toml](sep-0001.md) file containing a `SIGNING_KEY` attribute. This allows the SEP-10 server to verify that the challenge returned by the client is signed with the `SIGNING_KEY`. This proves that the client is associated with the client domain. The SEP-10 server can now attribute requests made with the resulting JWT to the client domain that signed the challenge.
```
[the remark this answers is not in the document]
JakeUrban: Committing suggestion but keeping `client domain` instead of using `client home domain`

ecosystem/sep-0010.md
```diff
@@ -271,6 +281,29 @@ A server may choose to issue JWTs for less than all thresholds and based on any
 
 A server that needs to support validating accounts that do not exist can require a signature of the master key of the account address for accounts that do not exist.
 
+### Verifying Client Application Identity
+
+A server may want to attribute each HTTP request to a specific client application. For example, a server and client may agree to change fees for the users of that client. SEP-10 offers an optional variation to the flow outlined in the [Abstract](#abstract) to enable this use case.
```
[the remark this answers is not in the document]
JakeUrban: From a readers perspective, isn't it better to start with 'why' so they don't have to read the 'how' before they determine if its relevant to them?

ecosystem/sep-0010.md
```diff
@@ -271,6 +281,29 @@ A server may choose to issue JWTs for less than all thresholds and based on any
...
+  * If it did not, the client aborts the flow outlined here and continues the authentication process as if the `client_domain` parameter was not passed in the request
+* The client obtains a signature generated using its organization's `SIGNING_KEY` and adds it to the challenge transaction. How this happens is up to the client. However:
+  * It is highly recommended that the client-side application does not hold the secret key for the `SIGNING_KEY`
+  * Instead, an internal service should accept a request from the organization's client application, verify the challenge sent, and generate the signature if valid
+* The client signs the challenge as it normally would with the user's Stellar account signer secret key(s)
+* The client makes the `POST <WEB_AUTH_ENDPOINT>` request with the signed transaction
+* In addition to the usual validation checks described in [Token](#token), the SEP-10 server verifies that the source account of the `client_domain` Manage Data operation signed the transaction
+* The SEP-10 server responds with a [token](#token) including a `client_domain` claim. Services that accept this token can attribute requests made with it to `client_domain`
```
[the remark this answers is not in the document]
JakeUrban: I agree, will update 👍

ecosystem/sep-0010.md
```diff
@@ -271,6 +281,29 @@ A server may choose to issue JWTs for less than all thresholds and based on any
...
+
+While SEP-10 servers may chose to verify all client applications who request attribution, it is highly recommended to only verify a specific set of client applications and ignore attribution requests from other clients. **Making requests to unknown domains may slow down or halt the SEP-10 server's ability to respond to other requests.**
+
+Outlined below is the flow between client applications and SEP-10 servers when attribution is requested:
+
+* The client makes the initial `GET <WEB_AUTH_ENDPOINT>` request with an added `?client_domain=example.com` parameter and value
+* If the SEP-10 server supports attribution requests for the specified `client_domain`, it fetches the `SIGNING_KEY` from the `client_domain`'s [SEP-1 stellar.toml](sep-0001.md) and generates the challenge transaction with an additional Manage Data operation as described in the [Response](#response) section, referred to here as the `client_domain` Manage Data operation
+  * If the SEP-10 server is unable to fetch the `SIGNING_KEY`, return a `424 Failed Dependency` HTTP status code
```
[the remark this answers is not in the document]
JakeUrban: I think it should fail, because there is a difference between the server choosing not to support attribution for the provided client domain and being unable to support attribution for the provided client domain. One is the server's prerogative, and the other is a failure of the client to make attribution possible.

I'm not tied to `424`, but it did seem to perfectly fit this situation. Another choice may be `422`, but that also seems to be defined within the WebDAV context. However, 422 is definitely used outside the WebDAV context in the wild. 

`400 Bad Request` is not a good fit here because the request can be well formed and the server would still return an error if the `SIGNING_KEY` couldn't be obtained.

The best option remaining may be `409 Conflict`

ecosystem/sep-0010.md
```diff
@@ -271,6 +281,29 @@ A server may choose to issue JWTs for less than all thresholds and based on any
 
 A server that needs to support validating accounts that do not exist can require a signature of the master key of the account address for accounts that do not exist.
 
+### Verifying Client Application Identity
```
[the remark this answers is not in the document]
JakeUrban: I have to adjust the links to this section but yes I'll adjust it 👍

JakeUrban:
@leighmcculloch I've rebased this PR on #851 and updated the language to match the changes introduced there.

ecosystem/sep-0010.md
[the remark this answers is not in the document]
leighmcculloch: If this is going to error, I think a `400` is fine. [RFC7231](https://tools.ietf.org/html/rfc7231#section-6.5.1) superseded [RFC2616](https://tools.ietf.org/html/rfc2616#section-10.4.1) and broadened the scope of the status code to apply to any request where the server won't process the request because it perceives the client made a mistake. There could be multiple reasons for why a `SIGNING_KEY` is not fetchable. Maybe the client specified the wrong domain, maybe the domain is currently down, but the domain is provided by the client organization so I think we can say that it is the clients fault if the domain they wish to verify is not setup appropriately.

ecosystem/sep-0010.md
```diff
@@ -124,6 +131,9 @@ On success the endpoint must return `200 OK` HTTP status code and a JSON object
       * `manage_data(source: server account, key: 'web_auth_domain', value: web_auth_domain)`
         * The source account is the **Server Account**
         * The value is the **Server**'s domain. It can be at most 64 characters.
+      * (optional) `manage_data(source: client domain account, key: 'client_domain', value: client_domain)`
+        * The source account is the **Client Domain Account**
+        * Only add this operation if the client passed a `client_domain` parameter in the request and the server supports [Verifying the Client Domain](#verifying-the-client-domain)
```
leighmcculloch, line 136: 💡 I think it is worth reordering these to first gate including the operation on what the server supports. I'm a little concerned in a quick read someone might think they should include it only based on what is in the request.
```suggestion
        * Add this operation if the server supports [Verifying the Client Domain](#verifying-the-client-domain) and the client provided a `client_domain` parameter in the request.
```

ecosystem/sep-0010.md
```diff
@@ -100,6 +106,7 @@ Name      | Type          | Description
 ----------|---------------|------------
 `account` | `G...` string | The stellar account that the wallet wishes to authenticate with the **Server** 
 `home_domain` | string | (optional) a **Home Domain**. Servers that generate tokens for multiple **Home Domain**s can use this parameter to identify which home domain the **Client** hopes to authenticate with. If not provided by the **Client**, the **Server** should assume a default for backwards compatibility with older **Clients**.
+`client_domain` | string | (optional) a **Client Domain**. Supplied by **Clients** that intend to verify their domain in addition to the **Client Account**. See [Verifying the Client Domain](#verifying-the-client-domain). **Servers** should ignore this parameter if the **Server** does not support **Client Domain** attribution, or the **Server** does not support attribution for the specific **Client Domain** included in the request.
```
leighmcculloch, line 109: 💡 I suggest changing attribution to verification so that we use a single term throughout the document to describe the feature.
```suggestion
`client_domain` | string | (optional) a **Client Domain**. Supplied by **Clients** that intend to verify their domain in addition to the **Client Account**. See [Verifying the Client Domain](#verifying-the-client-domain). **Servers** should ignore this parameter if the **Server** does not support **Client Domain** verification, or the **Server** does not support verification for the specific **Client Domain** included in the request.
```

ecosystem/sep-0010.md
```diff
@@ -43,6 +47,7 @@ The authentication flow is as follows:
    1. Source account set to the **Server Account**.
    1. Value set to the **Server**'s domain that the client requested the challenge from.
 1. The **Client** verifies that if the transaction has other operations they are Manage Data operations that all have their source accounts set to the **Server Account**
+1. If the transaction has a Manage Data operation with key `client_domain`, the **Client** obtains a signature from the **Client Domain Account** and adds it to the challenge transaction
```
leighmcculloch, line 50: ❗ The additional `if` if worthwhile here to distinguish the client should only do this if it requested client verification, and not otherwise.
```suggestion
1. If the client included a client domain in the request, and the transaction has a Manage Data operation with key `client_domain`, the **Client** obtains a signature from the **Client Domain Account** and adds it to the challenge transaction
```

ecosystem/sep-0010.md
```diff
@@ -273,6 +286,14 @@ A **Server** may choose to issue JWTs for less than all thresholds and based on
 
 A **Server** that needs to support validating accounts that do not exist can require a signature of the master key of the account address for accounts that do not exist.
 
+### Verifying the Client Domain
+
+A web service requirng SEP-10 authentication may want to attribute each HTTP request made to it to a specific **Client**. For example, a web service may want to offer reduced fees for the users of a specific **Client**.
```
leighmcculloch, line 291: 💡 One typo, and I also think adding the word software in this sentence helps clarify that when the document says client it don't mean a single instance of a client, but the software used by all instances of the client. Technically the abstract already communicates this, but when reading this specific sentence that meaning didn't feel evident.
```suggestion
A web service requiring SEP-10 authentication may want to attribute each HTTP request made to it to a specific **Client** software. For example, a web service may want to offer reduced fees for the users of a specific **Client**.
```

ecosystem/sep-0010.md
```diff
@@ -273,6 +286,14 @@ A **Server** may choose to issue JWTs for less than all thresholds and based on
...
+
+In order to use this optional feature, the organization that owns the **Client** must host a [SEP-1 stellar.toml](sep-0001.md) file containing a `SIGNING_KEY` attribute (i.e. the **Client Domain Account** on the **Client Domain**. This allows the **Server** to verify that the challenge returned by the **Client** is also signed with the **Client Domain Account**, proving that the **Client** is associated with the **Client Domain**. Web services requiring SEP-10 authentication can now attribute requests made with the resulting JWT to the **Client Domain** that signed the challenge.
```
leighmcculloch, line 293: ❗ I think we need to say something about the importance of protecting the key.
```suggestion
In order to use this optional feature, the organization that provides the **Client** must host a [SEP-1 stellar.toml](sep-0001.md) file containing a `SIGNING_KEY` attribute (i.e. the **Client Domain Account** on the **Client Domain**). The `SIGNING_KEY` attribute must be a Stellar public key in the form of a `G` address. The secret key paired with the `SIGNING_KEY` should be protected as anyone in possession of the secret can verify the **Client Domain**.

This setup allows the **Server** to verify that the challenge returned by the **Client** is also signed with the **Client Domain Account**, proving that the **Client** is associated with the **Client Domain**. Web services requiring SEP-10 authentication can now attribute requests made with the resulting JWT to the **Client Domain** that signed the challenge.
```

ecosystem/sep-0010.md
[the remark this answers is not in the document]
JakeUrban: Ok, I'm alright with that 👍

leighmcculloch (approved):
I think this looks great 👏🏻. I've left two small changes I'd like included before merging (❗) and some other small suggestions that I defer to you on (💡).

@tomerweller @accordeiro Could you also review this change?

ecosystem/sep-0010.md
```diff
@@ -165,14 +175,16 @@ To validate the challenge transaction the following steps are performed by the *
...
-  * verify that remaining signatures on the transaction are signers of the operation's source account;
+  * verify that remaining signatures on the transaction are signers of the **Client Account**
   * verify that remaining signatures are correct;
   * verify that remaining signatures provide weight that meets the required threshold(s), if any;
-* if the operation's source account does not exist:
+* if the first operation's source account does not exist:
   * verify that remaining signature count is one;
   * verify that remaining signature is correct for the master key of the **Client Account**
+* if the transaction contains a Manage Data operation with the key `client_domain`:
+  * verify that the transaction was signed by the source account of the Manage Data operation
```
accordeiro, line 187: I know this is further described in the doc (in the "Verifying the Client Domain" section), but I'd still add these extra steps here for clarity:

```suggestion
  * verify that the `value` provided in this Manage Data operation is a domain that serves a valid [SEP-1 stellar.toml](sep-0001.md) file containing a `SIGNING_KEY` attribute (see [Verifying the Client Domain](#verifying-the-client-domain))
  * verify that the source account of this Manage Data operation is the same as the `SIGNING_KEY` retrieved from the `stellar.toml`
  * verify that the transaction was signed by the source account of this Manage Data operation
```

accordeiro (approved):
This looks good to me.

I've added one inline suggestion to clarify the validation process from the server perspective.

A few general comments / questions:
1. We don't provide any guidance as to how clients obtain signatures from their client domains. I imagine that's intentional, since it's particular to their implementation?
2. I'm torn on the `400` status when there is a failure in the server-side client domain verification, but I can't think of a better status. It seems like this has [been previously discussed](https://github.com/stellar/stellar-protocol/pull/760#discussion_r574091444), so this is probably the best solution for now.

ecosystem/sep-0010.md
JakeUrban: The server doesn't need to check that the `SIGNING_KEY` matches the source account in the `POST` request logic because the server assigns `SIGNING_KEY` as the source account in the `GET` request logic and signs the transaction.

The source account cannot be altered without invalidating the server's signature, so the server can trust it's past work by verifying it's own signature.

accordeiro: Oh yeah you're right, the server already sends a signed transaction on `GET`, so verifying its own signature should be enough in that case.

PR #762, opened 2020-10-31
SEP-0032: Allow arbitrary extensions to asset addresses using colon delimiters

leighmcculloch:
### What
Experiment with what changes would be required to SEP-32 to allow asset addresses to be extended with namespaces that conceptually exist off-network within the asset.

### Why
@dandoney [shared](https://groups.google.com/g/stellar-dev/c/rWThI_BIs0Y/m/CrZY9gXBBgAJ) some insights into a more complex asset address scheme that they used and there could be value in allowing more complex asset addresses to exist and be parseable by other applications, encouraging interoperability, without specifying the namespaces and therefore without limiting what they could be.

### Why not
It could be argued that any application supporting an extension to the asset address format could do so on its own terms, and any application not supporting that custom extension could accept data entry of the standard asset address and that a user could know to enter only that portion of the address. It could be argued similarly that a developer connecting two systems where one uses the standard format and one extends the format could easily truncate the address when moving from one system to the next. In both of these situations the protocol doesn't need to do anything and therefore is simpler to implement for the many of applications.

### Draft
This is a draft to illustrate an idea, and isn't planned for merging yet.

### DIFF
--- a/ecosystem/sep-0032.md
+++ b/ecosystem/sep-0032.md
@@ -67,18 +67,30 @@ The format is based on [SEP-11 Txrep]'s definition for rendering assets as
 a string with the parts separated by a colon:
 
 ```
-{asset-code}:{asset-domain}
+{asset-code}:{asset-domain}[:]
 ```
 
 Where:
 - `{asset-code}` is the 1-12 alphanumeric Stellar asset code.
 - `{asset-domain}` the domain name that is hosting a `stellar.toml` file that
   contains a [`CURRENCIES`] section for the asset, as defined by [SEP-1].
+- `[:]` optional, terminates the asset address allowing for extension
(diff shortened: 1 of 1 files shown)


JakeUrban (commented):
Maybe its not necessary, but I would suggest a format that allows for key-value extensions. `:` can serve as a separator, so we just need another character to disambiguate key from value.

This format change would increase readability at no real cost (unless there are potential length limits?). Additionally, we may begin to see common key-value patterns emerge, which could be standardized here in the future.

leighmcculloch:
I'm going to close this idea in light of @orbitlens comments [here](https://groups.google.com/g/stellar-dev/c/rWThI_BIs0Y/m/F45DGmagBwAJ) and consider them more holistically and possibly propose different changes in the new year.

PR #763, opened 2020-11-03
SEP-0030: Change title to Account Recovery

leighmcculloch:
### What
Change the title to `Account Recovery` and remove mention of key management in title.

### Why
The SEPs purpose is to provide a multi-party approach to recovering Stellar accounts, either accounts a user created or owns, or accounts that another user has created and shared with the user. It is not technically key management because it is an interoperability layer that sits in the middle of an approach to key management.

The title `Recoverysigner` has also not been very useful. It was added to the beginning of the title to provide a single word keyword that could be used to reference SEP-30, but it turns out the term is odd, and so isn't being used to reference SEP-30.

### DIFF
--- a/ecosystem/sep-0030.md
+++ b/ecosystem/sep-0030.md
@@ -2,7 +2,7 @@
 
 ```
 SEP: 0030
-Title: Account Recovery: multi-party account recovery of Stellar accounts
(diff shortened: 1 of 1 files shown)


accordeiro (approved):
I love that change!

rice2000:
I think simplifying the title is a great move.

PR #764, opened 2020-11-05
SEP-0035: explicitly credit nullstyle

debnil:
This PR explicitly credits nullstyle with the design and implementation of the ID approach that this SEP is derived from. As the engineering work was done by him, we want to make it as explicit as possible in the text of the SEP.


accordeiro:
Looks great! :)

accordeiro (approved):

Issue #573, opened 2020-03-17
SEP-0030: External authentication is not sufficiently specified


leighmcculloch:
### Problem
[SEP-30 v0.2.0] references the phone and email authentication vaguely and without any specifics.
https://github.com/stellar/stellar-protocol/blob/63dde446500f9b156306505c5eddb4b013764558/ecosystem/sep-0030.md#L136-L137

In the SEP it is referred to as the `External` authentication provider, but beyond issuing some authentication token it doesn't specify the details. The lack of specification here will get in the way of two independently hosted recoverysigner servers being interoperable.

In the [reference implementation] we just happen to have hardcoded Firebase as the external authentication provider, but that is just a temporary hold over till we define this better.

[SEP-30 v0.2.0]: https://github.com/stellar/stellar-protocol/blob/63dde446500f9b156306505c5eddb4b013764558/ecosystem/sep-0030.md
[reference implementation]: https://github.com/stellar/go/tree/14a482c1286875cee8caef5ec1fa2fdb27e923c3/exp/services/recoverysigner

leighmcculloch:
TODO:
- [ ] Look at existing standards to utilize like OpenID, Oauth 2

leighmcculloch:
I think [OpenID Connect](https://openid.net/specs/openid-connect-core-1_0.html) which is layered on-top of Oauth 2 will work pretty well. This [FAQ](https://openid.net/connect/faq/) is a good starting point about it.

These features of OpenID Connect will support SEP-30's external auth well:

- OpenID Connect supports discovery, so a client can discover how to use an identity provider based on files it publishes to a .well-known address. A recoverysigner can indicate a list of identity providers it supports. It's unclear if folks would use this since at least initially a wallet is likely to integrate with specific identity providers and specific recoverysigners but longer term there could be significant value in supporting discovery. We're unlikely to use the resolution of a user identifier to identity provider in Section 2 of the spec and only the publishing of the server details in Section 4.
- [OpenID Connect Client Initiated Backchannel Authentication Flow](https://openid.net/specs/openid-client-initiated-backchannel-authentication-core-1_0.html) which is still in draft supports authentication being completed on a separate device without redirects. E.g. Initiate auth via email on a phone but click the email link on a computer, or Initiate auth via phone number on a computer and reply to an SMS on the phone to approve login.
[...]

PR #765, opened 2020-11-07
SEP-0030: Add clearer instruction that other authentication types are supported

leighmcculloch:
### What
Add clearer instruction that other authentication types are supported and recommended formats for their use.

### Why
SEP-30 has always stated that implementers may choose any authentication methods, and there are no limitations within SEP-30 that require an authentication method to be synchronous and immediate. However this has not been clear as the SEP only discusses phone and email authentication in any real detail.

This change does not yet improve interoperability when other authentication methods are used because the formats for those authentication methods are not yet defined. I'll follow up with that in separate PRs.

Note: SEP-30 has other issues regarding interoperability in that external authentication is not defined in anyway within this SEP, however this change does not address that and that issue remains to be address by https://github.com/stellar/stellar-protocol/issues/573.

### DIFF
--- a/ecosystem/sep-0030.md
+++ b/ecosystem/sep-0030.md
@@ -147,11 +147,11 @@ prove they meet the high threshold on the account.
 - `External`: Defined by the implementer. Proves possession of an identity that
   is external to the Stellar network such as a phone number, email address, or
   any other electronic or physical identity. Any forms of identity may be
-  supported. Authentication may be immediate and synchronous, as would be the
-  case with phone, email, and many OpenID Connect providers. Authentication may
-  be delayed and asynchronous, as would be the case with verifying a physical
-  address, national identity document, passport, or any other non-electronic
-  proof of identity.
+  supported. Authentication may be immediate and synchronous, as is the case with
+  electronic identities such as phone, email, and many OpenID Connect
+  providers. Authentication may be delayed and asynchronous, as might be the
+  case with physical identities such as verifying a physical address, national
+  identity document, passport, etc.
 
 ### Content Type
 
--- a/ecosystem/sep-0030.md
+++ b/ecosystem/sep-0030.md
@@ -8,8 +8,8 @@ Track: Standard
 Status: Draft
 Discussion: https://groups.google.com/forum/#!topic/stellar-dev/SFr2dHBZlsY
 Created: 2019-12-20
-Updated: 2020-11-02
-Version: 0.6.1
+Updated: 2020-11-16
+Version: 0.6.2
 ```
 
 ## Summary
--- a/ecosystem/sep-0030.md
+++ b/ecosystem/sep-0030.md
@@ -14,7 +14,7 @@ Version: 0.6.1
 
 ## Summary
 
-This protocol defines an API that enables an individual (e.g., a user or wallet) to regain access to a Stellar account that it owns after the individual has lost its private key without providing any third party control of the account. Using this protocol, the user or wallet will preregister the account and a phone number or email with two or more servers implementing the protocol and add those servers as signers of the account. No individual server will have control of the account, but collectively, they may help the individual recover access to the account. The protocol also enables individuals to pass control of a Stellar account to another individual.
+This protocol defines an API that enables an individual (e.g., a user or wallet) to regain access to a Stellar account that it owns after the individual has lost its private key without providing any third party control of the account. Using this protocol, the user or wallet will preregister the account and a phone number, email, or other form of authentication with two or more servers implementing the protocol and add those servers as signers of the account. No individual server will have control of the account, but collectively, they may help the individual recover access to the account. The protocol also enables individuals to pass control of a Stellar account to another individual.
 
 ## Motivation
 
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0030.md
```diff
@@ -199,8 +205,22 @@ Name | Type | Description
...
+`identities[].auth_methods[].type` | string | The type of identity. See [Common Authentication Methods].
+`identities[].auth_methods[].value` | string | The identifier or value that uniquely identifies the identity that if authenticated will be given full permissions of the account. For example if the type of authentication is an email address, the value is the email address.
+
+##### Common Authentication Methods
+
+Implementers define the external authentication methods supported by the
+implementation. If an implementation supports any of the following common
+authentication methods the type values below should be used along with the
+specifications for how to format the identifiers to improve interopability
+between clients and servers that may have multiple integraters.
```
accordeiro, line 217: 💡  I like the idea of listing the common authentication methods here. If we're aiming at maximizing interoperability, do you think it's worth suggesting to readers that, if they decide to add a new custom method, they are also highly encouraged to update this SEP with a new entry on the list, so other implementers will also adhere to the standard?

ecosystem/sep-0030.md
```diff
@@ -144,8 +144,14 @@ either. The two issuers are:
...
 account. The [SEP-10] server should be configured to only issue [JWT]s if signers
 prove they meet the high threshold on the account.
-- `External`: Defined by the implementer. Proves control of a phone number,
-email address, or other identity.
+- `External`: Defined by the implementer. Proves possession of an identity that
+  is external to the Stellar network such as a phone number, email address, or
+  any other electronic or physical identity. Any forms of identity may be
+  supported. Authentication may be immediate and synchronous, as is the case with
+  electronic identities such as phone, email, and many OpenID Connect
+  providers. Authentication may be delayed and asynchronous, as might be the
```
accordeiro, line 152: ❓ Do we want to give any sort of recommendation here? Immediate and synchronous are probably better from the user experience perspective, albeit not applicable to all authentication cases.

accordeiro (approved):
The clarifications looks great! Left one idea and one question, both inline, both nonblocking.

ecosystem/sep-0030.md
leighmcculloch: >Do we want to give any sort of recommendation here?

I don't think so, mostly because better user experience is subjective depending on the use case. In the right context slower authentication can actually create a better user experience. Also the SEP should be mostly unopinionated about what methods an implementer picks. Wallets/implementers need to make their own business decisions about that.

ecosystem/sep-0030.md
leighmcculloch: I agree with the sentiment, but I don't think that needs to live in the SEP. That's what the discussion thread and elsewhere is for. The SEP probably shouldn't be meta referencing itself and SEP process? Anyone can propose changes to any SEP.

ecosystem/sep-0030.md
accordeiro: Yep, good point. Agreed.

ecosystem/sep-0030.md
accordeiro: Got it. So you think people should update the list as I suggested, but we shouldn't mention this on the SEP text as it would be too meta?

ecosystem/sep-0030.md
leighmcculloch: That's what I was thinking, but I'm not hard set on that if you think that level of meta is worth it? I was planning to send a message to stellar-dev asking for suggestions for what other formats to include.

ecosystem/sep-0030.md
leighmcculloch: That's what I was thinking, but I'm not hard set on that if you think that level of meta is worth it? I was planning to send a message to stellar-dev asking for suggestions for what other formats to include.

accordeiro: I think what you're saying makes sense, thinking more about it. It's the same case as for any other interoperability updates: you send a PR, ask for community input, and update the SEP at the end of the process.

ecosystem/sep-0030.md
JakeUrban, line 152: What makes email confirmation or SMS verification synchronous? In fact, these are known to be asynchronous forms of communication compared to a phone call or F2F conversation. The user can wait as long as he/she pleases to confirm control of a phone number or email address unless the sender of the confirmation request limits the window in which the code/link sent is valid.

I think the difference you're pointing out is how long it takes the authentication provider to verify the identity based on the method used. So verifying that a user lives at a certain address or verifying that a user's passport is authentic may take more time than SMS/email verification _starting from when the user provides the code/clicks the link/sends the document_.

I think it would be more clear to remove the distinction and just list the potential forms of identify verification. An authentication provider reading this SEP will likely just pick the one easiest for them given their circumstances. They won't be asking themselves "should I do this synchronously or asynchronously"?

JakeUrban (commented):
Only one comment. I think the engineer in you is thinking sync-async but authentication providers will use whatever method makes sense for them regardless of how long it takes them to verify. How long it takes them to verify identity is just one factor when they choose which route to take.

JakeUrban:
Ah looks like I missed it. No worries, its not a big deal

ecosystem/sep-0030.md
leighmcculloch: Ah great distinction, thank you. I'm addressing this feedback in https://github.com/stellar/stellar-protocol/pull/770.

leighmcculloch:
>Ah looks like I missed it. No worries, its not a big deal

It's rarely too late to make a change. 😄  I'm addressing the feedback in #770.

PR #767, opened 2020-11-09
SEP-0010: Add web auth domain manage data operation (v3.1.0)

leighmcculloch:
### What
Add a new manage data operation containing the server domain of the SEP-10 server that the server includes in the challenge transaction.

### Why
The challenge transaction includes the home domain in the challenge transaction which provides some certainty to the client that the challenge transaction they are signing is for the home domain they intend to sign for. In some situations where a home domain is not involved in the SEP-10 process it would be useful to have an explicit field containing the domain of the issuer so there is something the client can verify to ensure it is signing a change that was generated by the issuer and not just for the home domain.

In some cases the server domain and home domain are the same and this is somewhat redundant. However, the server domain are not always the same domain as the home domain because it is common to deploy SEP-10 as a separate service, and deploying separate services is usually easier implemented on different subdomains or entirely separate domains. This is reasonably common. One example of this is [this](https://stablecoin.anchorusd.com/.well-known/stellar.toml):
   - Home domain is `stablecoin.anchorusd.com`
   - SEP-10 domain is `api.anchorusd.com`

This change is not breaking for any server and client pair who are using [SEP-10 v2.1] or greater. SEP-10 v2.1 changes clients to allow there to be additional manage data operations as long as they followed some basic rules. It is not breaking because if servers include the new operation and clients have not upgraded, clients will ignore it. If clients upgrade and servers do not, clients will continue to verify challenge transactions because the new operation is only verified if the server includes it. Servers and clients opt-in to the benefit of an improved challenge-response handshake. SDKs that implement this should probably make the new field mandatory in their APIs, but the optionality supports this change being backwards compatible and allows us to roll it out with zero coordination between clients and servers. 

[SEP-10 v2.1]: https://github.com/stellar/stellar-protocol/commit/5acf11be9837969ec883c2f9fd7cd3201728ff07


ecosystem/sep-0010.md
```diff
@@ -36,6 +36,10 @@ The authentication flow is as follows:
    1. Source account set to the user's account.
    1. Key set to `<home domain> auth` where the home domain is the home domain from the discovery flow.
    1. Value set to a nonce value.
+1. The client verifies that if the transaction has a Manage Data operation with key `server_domain` that it has:
```
JakeUrban, line 39: I'm a bit concerned that `server_domain` is too vague. What about `sep10_server_domain`?

ecosystem/sep-0010.md
```diff
@@ -115,7 +119,10 @@ On success the endpoint must return `200 OK` HTTP status code and a JSON object
     * `manage_data(source: client_account, key: '<home domain> auth', value: random_nonce())`
       * The value of key is the home domain the challenge is authenticating, followed by `auth`. It can be at most 64 characters.
       * The value must be 64 bytes long. It contains a 48 byte cryptographic-quality random string encoded using base64 (for a total of 64 bytes after encoding).
-    * zero or more `manage_data(source: server_account, ...)` reserved for future use
+    * subsequent operations, order unimportant:
```
JakeUrban, line 122:
```suggestion
    * subsequent operations must conform to one of the following patterns listed below. The order of these `manage_data` does not matter.
```

ecosystem/sep-0010.md
leighmcculloch: I think we should keep the term `SEP-10` out of the code interface, SEPs do not normally include that. How about `issuer_domain`? Using this name will do one of two things:
- Clear up misconception because the issuer of the SEP-10 challenge is the SEP-10 server.
- Any ambiguity should be more apparent since the term `issuer` may more likely lead to the reader digging in to find out what it means rather than glossing over on a generic term such as `server`.

ecosystem/sep-0010.md
leighmcculloch: The statement `must conform to one of the following patterns` reads to me as if the server must choose one dot point and only implement that, which isn't the intention. The server does not need to choose only one of the listed patterns, it can include any of the data that is listed, and the order with which it includes them in the transaction is not important.

It's unclear from your comment what is unclear about the current text that you're wanting to make clear. Could you elaborate so I understand the problem you're solving with the suggestion?

JakeUrban: It was really a matter of verbosity, we can keep it as-is but I felt the current language was a bit too terse.

ecosystem/sep-0010.md
JakeUrban: I think `server_domain` is a better than `issuer_domain` due to the fact that the term `issuer` is already a keyword within the Stellar context and could confuse the reader. What about `web_auth_server`, which would be the domain used in SEP-1's `WEB_AUTH_ENDPOINT`?

leighmcculloch: Oh I like `web_auth_domain`. It correlates directly to this SEP. I like it. Updated. 01a77b4

leighmcculloch: One more thought on this though, we use the term `server_account`  to mean the account of the SEP-10 server, so if we continued to use the term `server_domain` it correlates.

ecosystem/sep-0010.md
leighmcculloch: Agreed, it's a bit terse. I'm struggling to come up with more words that don't raise other questions so I'm going to leave it for now.

ecosystem/sep-0010.md
JakeUrban: Ah thats true, that definitely makes it more plausible. I still think `web_auth_domain` is better though.

JakeUrban (approved):

accordeiro (approved):
This looks good, and I like that this is a small change. I have a few points potentially worth discussing, but none of them should block this from being merged.

### a) How do we deal with apex/subdomains?

Should a client trust a subdomain if the response returns with the apex domain? As in:

1. **Client** asks `subdomain.domain.com` (aka **Server**) for a challenge transaction
2. **Server** responds with a signed challenge transaction, which contains the operation `manage_data(..., key: 'web_auth_domain', value: 'domain.com')`?

Should the client accept that challenge? I feel like it shouldn't, and that we should require the FQDN of the SEP-10 server. If that makes sense, should we make that explicit? It might sound obvious, but I'd rather not leave margin for interpretation.

### b) Standardization of manage_data fields:

1. For the `home_domain`, we use the following structure: `manage_data(source: client_account, key: '<home domain> auth', value: random_nonce())` – i.e. all important data is on the `key`, and value is just a nonce
2. For the `web_auth_domain`, we use a different structure: `key` is an identifier and `value` is the actual value.

I know (2) is probably more in line with what any developer would expect and (1) is a bit of an anti-pattern – but maybe it's worth standardizing to (1) since it's already what SDKs have implemented, so there isn't need for yet another parsing logic? I don't feel strongly about this, but it did feel a bit strange to me that we're using a different format.

leighmcculloch:
@accordeiro Thanks for reviewing, will respond to each discussion points separately:

a) I don't think fully-qualified domain name is the appropriate term to use here. Really a fully-qualified domain is required to have a dot at the end. Normally we all use relative domains relative to that root dot. In some documentation the FQDN term is used to be a specific computer or host, which again isn't really what we want here, because we want this value to refer to the domain name that the client has made a request to, a service in general and not a particular machine or hardware.

I think this ambiguity is worth addressing though if it exists. How we addressed this with the home domain was to be more specific about exactly what the home domain references. I think we could do that with the web auth domain by specifying that the value should match the [`host` portion of the URL](https://tools.ietf.org/html/rfc1738#section-5) that the client requested the challenge. The reason I didn't add this originally was because I am concerned a developer could read that statement and think the server should blindly include the host in the request, which could be dangerous.

Giving it another look, I think we can add these additional statements to clear up the ambiguity: https://github.com/stellar/stellar-protocol/pull/767/commits/5c0ca6aca49b8c1d7f6cf7260851a5bb6b1e4d54.

How does that look to you?

b) The new parsing logic for key and values is simpler than what already exists and so even though it is different I think the effort to add the new is not a net cost over the other. The existing operation has some problems because it leans on a whitespace separator in the key to separate the key and the value, and it may be difficult to continue to use that for all future operations so I think we should make the jump on this one.

ecosystem/sep-0010.md
```diff
@@ -19,8 +19,8 @@ This SEP defines the standard way for clients such as wallets or exchanges to cr
 This protocol is a variation of mutual challenge-response, which uses Stellar transactions to encode challenges and responses.
 
 It involves three components:
-- A domain hosting a [SEP-1 stellar.toml](sep-0001.md) containing a `WEB_AUTH_ENDPOINT` and `SIGNING_KEY`, referred to as the home domain.
-- An endpoint providing the GET and POST operations discussed in this document.
+- A domain hosting a [SEP-1 stellar.toml](sep-0001.md) containing a `WEB_AUTH_ENDPOINT` (URL) and `SIGNING_KEY` (`G...`), referred to as the home domain.
+- An endpoint providing the GET and POST operations discussed in this document. The endpoint may be hosted on the home domain, a sub-domain of the home domain, or any other domain. The domain that it is hosted is referred to as the web auth domain.
```
JakeUrban, line 23: This last sentence doesn't read well, `The domain that it is hosted`?

leighmcculloch:
```suggestion
- An endpoint providing the GET and POST operations discussed in this document. The endpoint may be hosted on the home domain, a sub-domain of the home domain, or any other domain. The domain that the endpoint is hosted is referred to as the web auth domain. 
```

JakeUrban: Added `on`, LGTM now

accordeiro (approved):
Looks great!

> b) The new parsing logic for key and values is simpler than what already exists and so even though it is different I think the effort to add the new is not a net cost over the other. The existing operation has some problems because it leans on a whitespace separator in the key to separate the key and the value, and it may be difficult to continue to use that for all future operations so I think we should make the jump on this one.

Agreed with your rationale here :)

ecosystem/sep-0010.md
```diff
@@ -115,7 +119,10 @@ On success the endpoint must return `200 OK` HTTP status code and a JSON object
     * `manage_data(source: client_account, key: '<home domain> auth', value: random_nonce())`
       * The value of key is the home domain the challenge is authenticating, followed by `auth`. It can be at most 64 characters.
```
JakeUrban, line 120: `the home domain the challenge is authenticating`

The challenge does not authenticate a home domain. I think we should change this to either mention the discovery flow like in prior statements, or we should be explicit about what domain we're talking about.

```suggestion
      * The value of key is the domain hosting the service's [SEP-1 stellar.toml](sep-0001.md), followed by `auth`. It can be at most 64 characters.
```

leighmcculloch: The home domain is defined at the top of the document, so that we do not need to redefine it at each mention point throughout the document.

>The challenge does not authenticate a home domain.

I agree, and I think there's a word missing here. It should say `the home domain the challenge is authenticating [with]...`.

I think this is out-of-scope of this PR. If you agree, can you open a new PR with the improvements you propose?

JakeUrban (approved):

PR #768, opened 2020-11-09
SEP-0010: Require the server hostname manage data operation (v4.0.0)

leighmcculloch:
### What
Require the server domain be specified in challenge transactions.

### Why
I'm proposing we add the server domain in #767, and after sometime if all servers are providing this field it might be feasible to remove that feature from being optional in clients to verify.

### Why not
If all servers are including the server domain it is not likely to make any difference if we make this change or not since all clients that verify will verify if the field is present. Also, as I think about the trends we want to set with this SEP it's unclear to me if we should do two-step releases where we add a feature as optional and then make it required. We have the capability for many features to make them additive and reach approximately the same goals. This might be one of them. I'd like to hear other thoughts.

### Reviewing
This change includes changes from #767 which is intended to merge first. Look at the final commit on this PR to see the meaningful change: https://github.com/stellar/stellar-protocol/pull/768/commits/f9d51237cd19651f11d38fb3ede3e9969084fa1f

### Merging
There may be other changes we want to introduce into SEP-10 in a new major version. We maybe should consider merging this in conjunction with any other changes we're considering when the time comes.

### DIFF
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -36,7 +36,7 @@ The authentication flow is as follows:
    1. Source account set to the user's account.
    1. Key set to `<home domain> auth` where the home domain is the home domain from the discovery flow.
    1. Value set to a nonce value.
-1. The client verifies that if the transaction has a Manage Data operation with key `server_domain` that has its:
+1. The client verifies that if the transaction has a Manage Data operation with key `server_domain` that it has:
    1. Source account set to the server's account.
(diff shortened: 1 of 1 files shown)


leighmcculloch:
I'm closing this change because I don't think there are practical reasons to do it, and it perceivably appears to add breaking changes but practically once servers and clients upgrade to v3.1 (#767) they'll be getting all the benefits of the new fields added in that release, and can update in any order over time without any dependency on each other.

PR #769, opened 2020-11-17
SEP-0030: Be unopinionated about number of servers required by protocol

leighmcculloch:
### What
Be unopinionated about number of servers required by protocol and do not exclude use of the protocol with a single server, but include note that the single server configuration means control of the account.

### Why
The protocol specifies an interface a client and server can use to facilitate account recovery through transaction signing and authentication with providers outside of the Stellar network. It supports a configuration where a client uses multiple servers so that no individual server has control of the account on its own, but that is one use case and the protocol can support being used with a client and server configuration where there is only one server. There is little reason to limit the SEP to only multiple servers as it is up to the client how many servers they integrate with and utilize and are valid use cases where a client may wish to use a single server.

cc @capybaraz

### DIFF
--- a/ecosystem/sep-0030.md
+++ b/ecosystem/sep-0030.md
@@ -14,7 +14,7 @@ Version: 0.6.2
 
 ## Summary
 
-This protocol defines an API that enables an individual (e.g., a user or wallet) to regain access to a Stellar account that it owns after the individual has lost its private key without providing any third party control of the account. Using this protocol, the user or wallet will preregister the account and a phone number, email, or other form of authentication with one or more servers implementing the protocol and add those servers as signers of the account. If two or more servers are used no individual server will have control of the account, but collectively, they may help the individual recover access to the account. The protocol also enables individuals to pass control of a Stellar account to another individual.
+This protocol defines an API that enables an individual (e.g., a user or wallet) to regain access to a Stellar account that it owns after the individual has lost its private key without providing any third party control of the account. Using this protocol, the user or wallet will preregister the account and a phone number, email, or other form of authentication with one or more servers implementing the protocol and add those servers as signers of the account. If two or more servers are used with appropriate signer configuration no individual server will have control of the account, but collectively, they may help the individual recover access to the account. The protocol also enables individuals to pass control of a Stellar account to another individual.
 
 ## Motivation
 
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0030.md
```diff
@@ -14,7 +14,7 @@ Version: 0.6.1
...
+This protocol defines an API that enables an individual (e.g., a user or wallet) to regain access to a Stellar account that it owns after the individual has lost its private key without providing any third party control of the account. Using this protocol, the user or wallet will preregister the account and a phone number or email or other form of authentication with one or more servers implementing the protocol and add those servers as signers of the account. If two or more servers are used no individual server will have control of the account, but collectively, they may help the individual recover access to the account. The protocol also enables individuals to pass control of a Stellar account to another individual.
```
accordeiro, line 17: I think this statement: `If two or more servers are used no individual server will have control of the account` seems inaccurate, or at least incomplete. It is true only if the account that is being registered is correctly configured, as you later describe in this PR.

So maybe we could change the wording to say `If two ore more servers are used and the account has the right signer configuration, no individual server will have control of the account`. What do you think?

accordeiro (approved):
Looks great! Left a suggestion to make the summary more complete / accurate.

ecosystem/sep-0030.md
leighmcculloch: That's a great point. Actually that raises another point, you can use a single server and still not give the server control because you could use a SEP-5 mnemonic phrase as the other party to a multi-party recovery. I think I need to revisit the phrasing here and in the paragraphs further down too, so that it is less opinionated but not so general as to be meaningless.

ecosystem/sep-0030.md
```diff
@@ -51,7 +51,11 @@ The registration flow is as follows:
 - The server responds to the client with the signing public key.
 - The client adds the signing public key as a signer of the account with an appropriate weight.
 
-A client can register with multiple servers and give each signing key a weight that is appropriate for the use case the client and servers are targeting. As an example, if a client is intending to use two servers that should individually have no control of the account but together are able to approve transactions, then the client can configure [account thresholds] to high 20, med 20, low 20, and assign weights 10 to both server signing keys.
+A client can register with multiple servers and give each signing key a weight that is appropriate for the use case the client and servers are targeting.
+
+As examples:
+- If a client intends to use two servers that should individually have no control of the account but together are able to approve transactions, then the client could configure [account thresholds] to high 2, med 2, low 2, and assign weights 1 to both server signing keys.
```
JakeUrban, line 57: I wouldn't mention medium or low threshold values here. This SEP only cares about the high threshold right? Theres no reason each threshold should be equal right? And in fact, they really shouldn't or wouldn't be IRL.

JakeUrban (commented):
This is an improvement for sure. Just one comment :+1

ecosystem/sep-0030.md
leighmcculloch: Made this change in 64e8382.

ecosystem/sep-0030.md
leighmcculloch: I see your point that high should be irrelevant because for the SEP-10 proof only the high threshold is relevant, but this specific paragraph is discussing as an example how to configured account thresholds such that neither server has control of the account on their own. It is important to configure the medium and low thresholds greater than the servers signing weight otherwise the server would have some independent control. This is a minor point I think and so there isn't much discussion in the document on this. Ultimately the SEP provides an interface to support a wallet/server doing whatever they need for their product and doesn't go deep into prescribing the signer setups and only provides examples.

marcelosalloum (approved):
LGTM!

leighmcculloch:
I got @capybaraz blessing offline for this.

PR #770, opened 2020-11-17
SEP-0030: Remove mention of async/sync authentication methods

leighmcculloch:
### What
Remove mention of async/sync authentication methods and replace it with a more concise list of example authentication methods.

### Why
@JakeUrban pointed out in https://github.com/stellar/stellar-protocol/pull/765#pullrequestreview-531948828 that the section we have here that is being changed focuses on the authentication methods being synchronous or asynchronous and that is not particularly helpful to an integrator. The change in this change refocus the paragraph on what the external authentication can be which is more meaningful.


JakeUrban (approved):
LGTM!

PR #772, opened 2020-11-23
Update definitions to SEP-12 GET request

JakeUrban:
Makes the SEP-12 request parameters more clear. 

Also changes mentions of `user` to `customer` to avoid confusion (in the context of SEP-12, they are synonymous).

Context:
SEP-12 was originally created for SEP-6, but was later adapted to work with SEP-31 as well. This is why you see the removal of `deposit/withdrawal`, which is a SEP-6 construct.


ecosystem/sep-0012.md
```diff
@@ -54,22 +54,22 @@ Alternatively, if the client cannot add the authorization header. The JWT should
 
 Checks the status of a customers information. This allows clients to check whether the customers information was accepted, rejected, or still needs more info.  Lookups are done via the stellar account ID and memo that were used for the initial PUT request.
 
-This also provides the information needed from a KYC provider to KYC a user.  If a new account that has never been KYC'd is fetched,the server should respond with `NEEDS_INFO` and a list of all the fields the client will need to provide.
+This also provides the information needed from a KYC provider to KYC a customer.  If a new account that has never been KYC'd is fetched,the server should respond with `NEEDS_INFO` and a list of all the fields the client will need to provide.
 
 ### Request
```
Zagan202, line 59: More of a general SEP-12 question, since all the request fields are optional what is the expected response if the GET was just `GET [KYC_SERVER]/customer`?

JakeUrban: Thats a good question because it points out that the SEP isn't clear.

To answer your question: `GET /customer` would return the fields required for the default `type` of customer. If no identifying parameters are sent by the client, the endpoint should assume its a request for a new customer that doesn't yet exist.

There are 2 purposes to the `GET /customer` endpoint:
- Fetch the status of an existing customer
- Fetch the fields required for a new customer

If the sending anchor is registering a customer that is _new to the receiving anchor_, then there's no reason to pass parameters, because the receiving anchor doesn't have the customer record in their DB yet.

If the sending anchor is checking on the status of an existing customer, the (`id`) or (`memo`, `memo_type`) parameters should be used. Both can identify the customer. But if the sender doesn't have an `id` returned previously from the receiving anchor, it can only use `memo` and `memo_type`.

JakeUrban: I'll add more to the SEP to clarify that.

JakeUrban:
@Zagan202 can you take another look?

Zagan202 (approved):
This much more clear.
LGTM

PR #774, opened 2020-11-28
SEP-0037: Add address directory API

orbitlens:


ecosystem/sep-address-directory-api.md
```diff
@@ -0,0 +1,256 @@
...
+    "description": "Reserved, custodian account, or cold wallet"
+  },
+  {
+    "name": "malicious",
+    "description": "Account involved in theft/scam/spam/phishing"
+  },
+  {
+    "name": "unsafe",
+    "description": "Obsolete or potentially dangerous account"
+  },
```
leighmcculloch, line 111: ❓ It's unclear to me what the distinction between malicious and unsafe is practically for a user.

leighmcculloch (approved):
I think this looks good. Couple minor logistical things (✏️), one question (❓), and one suggestion (💡). Could you fix up the ✏️ items, then let's merge this.

ecosystem/sep-address-directory-api.md
```diff
@@ -0,0 +1,256 @@
...
+  ]
+}
+```
+
+### CORS, Rate Limiting and Caching Considerations
+
+All described API endpoints have [CORS headers](https://enable-cors.org/index.html)
+enabled, so they can be used directly from any website.
+
+The effective API request rate is limited to 5 requests per second for each
```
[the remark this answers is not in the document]
orbitlens: Agree, this definitely makes sense

ecosystem/sep-address-directory-api.md
orbitlens: "Malicious" are the addresses reported for illicit on-chain activity (scam, spam, funds theft). It implies that wallets should block deposits/withdrawals to such addresses. At the same time, "unsafe" tag refers to organization's accounts that may malfunction for some reason, like no longer used custodial deposit addresses, nonfunctioning anchors, potentially shady ICOs, etc. In this case the wallet can notify the user about potential problem while still allowing transfers to such an address.

orbitlens:
@leighmcculloch Thanks for the quick review :+1:

leighmcculloch (approved):
The changes look good to me.

PR #779, opened 2020-12-09
SEP-24: Remove the pending_stellar status step in deposits flow diagram

JakeUrban:
Our status diagram inaccurately showed `pending_stellar` as used in the deposit flow, but in reality this status is never used. `pending_stellar` is used to indicate the Stellar transaction is in-flight, but since the anchor won't know exactly when a withdrawal payment is coming, they won't ever update to this status. 

Instead, they'll simply detect the Stellar transaction was made to their distribution account and update the status from `pending_user_transfer_start` --> `pending_anchor`.


JakeUrban:
Ah I'm realizing the resolution of the new diagram is much smaller, let me try to get a better resolution image

Zagan202:
Are the `too_small` or `too_large` not relevant or are they captured during another status/stage of the flow?

JakeUrban:
I guess it would be relevant when the anchor receives the withdrawal payment. Polaris doesn't use these statuses and just reports `error` though, since the anchor had to collect the amount originally via interactive flow.

JakeUrban:
Its not like SEP-6 where the anchor doesn't collect the amount prior to the wallet sending it.

JakeUrban:
So what I mean is, the wallet will send whatever funds the anchor requests via the `amount_in` field after the interactive flow completes. Its not like deposit where the user has the freedom to send whatever amount he/she wants from his/her bank.

Zagan202 (approved):
LGTM

PR #782, opened 2020-12-11
SEP-0001: Improve language on organization licensing fields

rice2000:
Adding a bit of language in the 3 license fields at legal's suggestion.


leighmcculloch (approved):
👏🏻 This seems fine and safe change to make. Can you also update the verison field to `2.1.1`, incrementing the patch version since this expands on existing language and is not adding any new functionality? Can you also update the updated date at the same time? Thanks!

rice2000:
Good call re: the metadata.  Done.  Thanks, @leighmcculloch!

PR #783, opened 2020-12-14
SEP-9: Add beneficial owner organization fields

gaclaudiu:
new organization fields

[Here ](https://docs.google.com/document/d/1ZXuEmDAVhka5PHL6dIget-0LsitOFLC184Ysy_5mkmM/edit?usp=sharing) is a more detailed document with business requirements.

### DIFF
--- a/ecosystem/sep-0009.md
+++ b/ecosystem/sep-0009.md
@@ -94,8 +94,6 @@ Name | Type | Description
 `organization.individual_beneficial_owners.{index}.ownership_percentage` | string | Holds the percentage owned by the beneficial owner. Must be a number in the range 0 - 100 with at most 2 decimal places
 `organization.entity_beneficial_owners.{index}.ownership_percentage` | string | Holds the percentage owned by the beneficial owner. Must be a number in the range 0 - 100 with at most 2 decimal places
 
-
-
 ## Explanation
 
 Where possible we use field names from [schema.org](https://schema.org/Person). Words are separated with underlines as that convention has previously been established in Stellar protocols.
--- a/ecosystem/sep-0009.md
+++ b/ecosystem/sep-0009.md
@@ -7,7 +7,7 @@ Author: stellar.org
 Status: Active
 Created: 2018-07-27
 Updated: 2020-12-14
-Version 1.2.1
+Version 1.3.0
 ```
 
 ## Simple Summary
--- a/ecosystem/sep-0009.md
+++ b/ecosystem/sep-0009.md
@@ -90,7 +90,7 @@ Name | Type | Description
 `organization.individual_beneficial_owners.{index}.{SEP-0009 person field}` | array | Organization individual beneficial owners and control person
 `organization.entity_beneficial_owners.{index}.{SEP-0009 organization field}`| array | Organization entity beneficial owners
 `organization.individual_beneficial_owners.{index}.is_control_person` | boolean | Tells if an individual beneficial owner is also a control person
-`organization.individual_beneficial_owners.{index}.title` | string | If an individual benefiacial owner is control person then this field will store the control person title. Field is optional
+`organization.individual_beneficial_owners.{index}.title` | string | If an individual beneficial owner is control person then this field will store the control person title
 `organization.individual_beneficial_owners.{index}.ownership_percentage` | string | Holds the percentage owned by the beneficial owner. Must be a number in the range 0 - 100 with at most 2 decimal places
 `organization.entity_beneficial_owners.{index}.ownership_percentage` | string | Holds the percentage owned by the beneficial owner. Must be a number in the range 0 - 100 with at most 2 decimal places
 
--- a/ecosystem/sep-0009.md
+++ b/ecosystem/sep-0009.md
@@ -87,12 +87,12 @@ Name | Type | Description
 `organization.website` | string | Organization website
 `organization.email` | string | Organization contact email
 `organization.phone` | string | Organization contact phone
-`organization.beneficial_owners.person.{uniqueIdentifier}.{SEP-0009 person field}` | array | Organization person beneficial owners and control person
-`organization.beneficial_owners.organization.{uniqueIdentifier}.{SEP-0009 organization field}`| array | Organization beneficial owners which are organizations
-`organization.beneficial_owners.person.{uniqueIdentifier}.is_control_person` | boolean | Tells if a person beneficial owner is also a control person
-`organization.beneficial_owners.person.{uniqueIdentifier}.title` | string | If an person beneficial owner is control person then this field will store the control person title
-`organization.beneficial_owners.person.{uniqueIdentifier}.ownership_percentage` | string | Holds the percentage owned by the beneficial owner. Must be a number in the range 0 - 100 with at most 2 decimal places
-`organization.beneficial_owners.organization.{uniqueIdentifier}.ownership_percentage` | string | Holds the percentage owned by the beneficial owner. Must be a number in the range 0 - 100 with at most 2 decimal places
+`organization.beneficial_owners.person.{id}.{SEP-0009 person field}` | array | Organization person beneficial owners and control person
+`organization.beneficial_owners.organization.{id}.{SEP-0009 organization field}`| array | Organization beneficial owners which are organizations
+`organization.beneficial_owners.person.{id}.is_control_person` | boolean | Tells if a person beneficial owner is also a control person
+`organization.beneficial_owners.person.{id}.title` | string | If an person beneficial owner is control person then this field will store the control person title
+`organization.beneficial_owners.person.{id}.ownership_percentage` | string | Holds the percentage owned by the beneficial owner. Must be a number in the range 0 - 100 with at most 2 decimal places
+`organization.beneficial_owners.organization.{id}.ownership_percentage` | string | Holds the percentage owned by the beneficial owner. Must be a number in the range 0 - 100 with at most 2 decimal places
 
 ## Explanation
 
--- a/ecosystem/sep-0009.md
+++ b/ecosystem/sep-0009.md
@@ -6,8 +6,8 @@ Title: Standard KYC / AML fields
 Author: stellar.org
 Status: Active
 Created: 2018-07-27
-Updated: 2020-04-03
-Version 1.2.0
+Updated: 2020-12-14
+Version 1.2.1
 ```
 
 ## Simple Summary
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0009.md
```diff
@@ -87,6 +87,14 @@ Name | Type | Description
...
 `organization.email` | string | Organization contact email
 `organization.phone` | string | Organization contact phone
+`organization.individual_beneficial_owners.{index}.{SEP-0009 person field}` | array | Organization individual beneficial owners and control person
+`organization.entity_beneficial_owners.{index}.{SEP-0009 organization field}`| array | Organization entity beneficial owners
+`organization.individual_beneficial_owners.{index}.is_control_person` | boolean | Tells if an individual beneficial owner is also a control person
+`organization.individual_beneficial_owners.{index}.title` | string | If an individual benefiacial owner is control person then this field will store the control person title. Field is optional
+`organization.individual_beneficial_owners.{index}.ownership_percentage` | decimal |  Optional field , holds the percentage owned by the beneficial owner
+`organization.entity_beneficial_owners.{index}.ownership_percentage` | decimal |  Optional field , holds the percentage owned by the beneficial owner
+
+
```
JakeUrban, line 97: Lets remove the added whitespace
```suggestion
```

ecosystem/sep-0009.md
```diff
@@ -95,3 +103,53 @@ Where possible we use field names from [schema.org](https://schema.org/Person).
...
+### Beneficial Owners
+* Individual Beneficial Owners - individual (person)  which holds ownership.
+* Entity Beneficial Owner - entity (organization)  which holds ownership. 
+* Control Person - an important person in a corporation. Control persons include senior managers, members of the board of directors, and officers such as the CEO and CFO. Control persons are able to use both their authority and their influence to make decisions on the corporation's activities. A control person is also called an affiliated person.
+
+### Control Person Title List
```
JakeUrban, line 112: Would a request containing a `title` value not given here be rejected? If not, we should remove this list.

ecosystem/sep-0009.md
```diff
@@ -95,3 +103,53 @@ Where possible we use field names from [schema.org](https://schema.org/Person).
...
+
+### Beneficial Owners
+* Individual Beneficial Owners - individual (person)  which holds ownership.
+* Entity Beneficial Owner - entity (organization)  which holds ownership. 
+* Control Person - an important person in a corporation. Control persons include senior managers, members of the board of directors, and officers such as the CEO and CFO. Control persons are able to use both their authority and their influence to make decisions on the corporation's activities. A control person is also called an affiliated person.
```
JakeUrban, line 110:
```suggestion
* Control Person - Control persons are able to use their authority or influence to make decisions on the corporation's activities. Control persons include but are not limited to senior managers, members of the board of directors, and officers such as the CEO and CFO. A control person is also called an affiliated person.
```

JakeUrban (changes requested):
Thanks for submitting this @gaclaudiu, I added some suggestions but the overall change is good to have 👍 . Once these comments are addressed I can approve.

ecosystem/sep-0009.md
JakeUrban: This is more of a suggestion, not a big deal

ecosystem/sep-0009.md
```diff
@@ -6,8 +6,8 @@ Title: Standard KYC / AML fields
 Author: stellar.org
 Status: Active
 Created: 2018-07-27
-Updated: 2020-04-03
-Version 1.2.0
+Updated: 2020-12-14
+Version 1.2.1
```
leighmcculloch, line 10: This adds new fields and needs to be a minor version update rather than a patch version.
```suggestion
Version 1.3.0
```

ecosystem/sep-0009.md
```diff
@@ -87,6 +87,14 @@ Name | Type | Description
 `organization.website` | string | Organization website
 `organization.email` | string | Organization contact email
 `organization.phone` | string | Organization contact phone
+`organization.individual_beneficial_owners.{index}.{SEP-0009 person field}` | array | Organization individual beneficial owners and control person
```
leighmcculloch, line 90: Because the term `person` is used instead of `individual` in this SEP I would suggest using that term here instead too so that we maintain consistent ubiquitous language for the concept of a single person.

ecosystem/sep-0009.md
```diff
@@ -87,6 +87,14 @@ Name | Type | Description
...
+`organization.entity_beneficial_owners.{index}.{SEP-0009 organization field}`| array | Organization entity beneficial owners
```
leighmcculloch, line 91: Because the term `organization` is used instead of entity in this SEP I would suggest using that term here instead too so that we maintain consistent ubiquitous language for the concept of a non-person entity. Also, in many jurisdictions a person can be a subtype of an entity, so it is somewhat confusing to use the term entity to distinguish from a person.

ecosystem/sep-0009.md
```diff
@@ -87,6 +87,14 @@ Name | Type | Description
...
+`organization.individual_beneficial_owners.{index}.is_control_person` | boolean | Tells if an individual beneficial owner is also a control person
```
leighmcculloch, line 92: I think there's some value in grouping all beneficial owners under a single group, then grouping persons and organizations under it. In the JSON it keeps them scoped to the one place and appears elegant, I think, although this is definitely personal preference bleeding into this request.
```suggestion
`organization.beneficial_owners.organizations.{index}.{SEP-0009 organization field}`| array | Organization entity beneficial owners
`organization.beneficial_owners.persons.{index}.is_control_person` | boolean | Tells if an individual beneficial owner is also a control person
```

leighmcculloch (changes requested):
I have four requests below, in addition to the requests that @JakeUrban made. Otherwise this looks great, thanks @gaclaudiu for expanding the usefulness of SEP-9.

ecosystem/sep-0009.md
leighmcculloch, line 90: This is the first case where a SEP-9 field contains an array with an index rather than sub-fields.

The SEP probably needs to specify some rules about how indexes and arrays work. Questions I have so far, which aren't exhaustive:
- Is zero or one the first index?
- Must indexes be sequential?
- Is an index forever? For example, if the list of beneficial owners changes from 4 persons to 3 persons, where of the 4, 2 are removed and 1 is added, how is that communicated in SEPs that use SEP-9? Do the removed persons indexes get reused? How do SEPs know that the person has changed vs a detail about a person has changed? Depending on the answers to these questions a unique identifier for a person may be preferred over an index.

Another thing we need to consider is that recently we had that case in another SEP where it became clear that encoding arrays in url-form-encoded format was ambiguous. We couldn't find any common way to encode the array where the index was preserved. URL encoding arrays using the only standard form does not traditionally preserve the order because there is no index encoded.

Other SEPs that use SEP-9 probably don't currently have instructions for how to use array elements, and we should include changes to those SEPs with this change otherwise it will be ambiguous for how to use and interpret then.

cc @JakeUrban

leighmcculloch: SEPs we need to consider are: SEP-12, SEP-6, SEP-24, and soon SEP-8 once #758 merges.

JakeUrban: I think the best way to solve this problem is to avoid it entirely. That means using a unique identifier instead of array index as you suggest.

There are two alternatives:
- define our own standard for encoding arrays for key-value based content types: `application/x-www-form-urlencoded` and `multipart/form-data`
- use `application/json`. The blocker here is binary data, as JSON doesn't natively support it. However, we _could_ provide a standardized approach to encoding binary data in JSON: base64-encoding it.

To me, asking clients to base64-encode binary data seems easier than asking them to conform to a non-standard array-encoding scheme. I would also argue that there is precedent for using base64-encodings on binary data types within Stellar's ecosystem: `memo`.

leighmcculloch: I like both approaches you discuss, and I think we could use both. Unique identifiers to make this problem mute, and supporting binary where-ever we use JSON via base64 encoding works. For the base64 encoding to work the field will also need to communicate the content-type of the contents. This could be in a sibling field with the same name suffixed with _content_type, or we could make the field in JSON format a JSON object with subfields `.content_type`, and `.data`.

JakeUrban: Don't we have the content type regardless of the base64 encoding? If an anchor requires `photo_id_front`, there's no way for the client to specify the image format used (png, jpeg, etc.). Why does encoding this data change that?

leighmcculloch: We don't have the problem with multipart form encoding because each party can specify a `Content-Type` header, along with a `name` and even a `filename` field. This isn't shown in the examples of SEP-12, but it is available.

JakeUrban: Ah I see. I like the latter option more as well. I'll write up some PR's to add to each SEP's `Content-Type` section.

gaclaudiu: used unique identifier to identify a beneficial owner. 
Do I need to specify what data type is allowed as unique identifier ? like is a Guid or a number or a string ? 
What normally you guys use for unique identifiers ?

leighmcculloch: I think the unique identifier can be left undefined.

ecosystem/sep-0009.md
```diff
@@ -87,6 +87,12 @@ Name | Type | Description
 `organization.website` | string | Organization website
 `organization.email` | string | Organization contact email
 `organization.phone` | string | Organization contact phone
+`organization.beneficial_owners.person.{uniqueIdentifier}.{SEP-0009 person field}` | array | Organization person beneficial owners and control person
```
leighmcculloch, line 90: I'd suggest using the term `id` for brevity.
```suggestion
`organization.beneficial_owners.person.{id}.{SEP-0009 person field}` | array | Organization person beneficial owners and control person
```

mojoflower:
I've been wondering why it is preferred to have BO information in a hosted .toml file rather than making a dedicated manageData entry to the ledger. 

The reasoning for using manageData is that way the information can be queried from the same source and changes recorded with changes in the asset itself rather than hosted separately. I can also see it as being a nice way to verify KYC'd persons, for instance, authenticate them as signers for other applications. It makes sense that this type of information stays with the asset, and keeps an immutable record of changes as well. 

Example: 
Entry Name: CEO
Entry Value: GCRS...........................................MJXRRV

Is this a scaling issue or a cost issue, or is there anything else that I'm missing? 

*Note - this information is not updated frequently and I don't think that cost would be an issue.

// 

I´m thinking it can make compliance simpler and reduce the need for duplicate recordkeeping, self-reporting issues and especially in those cases where the .toml is self-hosted by the company.

leighmcculloch:
@JakeUrban @gaclaudiu I suggest the following edits to this change: https://github.com/gaclaudiu/stellar-protocol-1/pull/1.

PR #784, opened 2020-12-14
Replace ambiguous home domain parameter definition

JakeUrban:
Changes the `home_domain` parameter in the SEP-10 `GET` endpoint to reference the definition of "the home domain" in the abstract.

### DIFF
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -113,7 +113,7 @@ On success the endpoint must return `200 OK` HTTP status code and a JSON object 
   * time bounds: `{min: now(), max: now() + 900 }` (we recommend expiration of 15 minutes to give user time to sign transaction)
   * operations:
     * `manage_data(source: client_account, key: '<home domain> auth', value: random_nonce())`
-      * The value of key is the home domain the client would like to authenticating with, followed by `auth`. It can be at most 64 characters.
+      * The value of key is the home domain the client is authenticating with, followed by `auth`. It can be at most 64 characters.
       * The value must be 64 bytes long. It contains a 48 byte cryptographic-quality random string encoded using base64 (for a total of 64 bytes after encoding).
     * zero or more `manage_data(source: server_account, ...)` reserved for future use
   * signature by the service's stellar.toml `SIGNING_KEY`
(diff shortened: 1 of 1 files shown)


JakeUrban:
As discussed in https://github.com/stellar/stellar-protocol/pull/767#discussion_r542595681

ecosystem/sep-0010.md
```diff
@@ -95,7 +95,7 @@ Request Parameters:
 Name      | Type          | Description
 ----------|---------------|------------
 `account` | `G...` string | The stellar account that the wallet wishes to authenticate with the server
-`home_domain` | string | (optional) The home domain the client would like to authenticate. Servers that generate tokens for multiple home domains can use this parameter to identify which home domain the client hopes to authenticate. If not provided by the client, the server should assume a default for backwards compatibility with older clients.
+`home_domain` | string | (optional) The home domain as defined in the [Abstract](#abstract). Servers that generate tokens for multiple home domains can use this parameter to identify which home domain the client hopes to authenticate. If not provided by the client, the server should assume a default for backwards compatibility with older clients.
```
leighmcculloch, line 98: >As discussed in https://github.com/stellar/stellar-protocol/pull/767#discussion_r542595681

We were discussing the discussion about the home domain being the key in the Response discussion, but this PR looks like it is changing the parameter description.

I think we still need to specify that the client provides a value that they want to authenticate with. The abstract doesn't communicate that.

I think to address your original concern (https://github.com/stellar/stellar-protocol/pull/767#discussion_r542129321) that this makes it sounds like the client is authenticating the home domain rather than with the home domain, we need to add the word with here:

```suggestion
`home_domain` | string | (optional) The home domain the client would like to authenticate with. Servers that generate tokens for multiple home domains can use this parameter to identify which home domain the client hopes to authenticate. If not provided by the client, the server should assume a default for backwards compatibility with older clients.
```

If it's not clear enough that the term home domain has been previously defined, let's add an anchor tag around the definition, and link to it from here?

ecosystem/sep-0010.md
```diff
@@ -113,7 +113,7 @@ On success the endpoint must return `200 OK` HTTP status code and a JSON object
   * time bounds: `{min: now(), max: now() + 900 }` (we recommend expiration of 15 minutes to give user time to sign transaction)
   * operations:
     * `manage_data(source: client_account, key: '<home domain> auth', value: random_nonce())`
-      * The value of key is the home domain the challenge is authenticating, followed by `auth`. It can be at most 64 characters.
+      * The value of key is the home domain the client would like to authenticating with, followed by `auth`. It can be at most 64 characters.
```
leighmcculloch, line 116: Can we keep this sentence present tense? I don't think that needs to change to address the issue.
```suggestion
      * The value of key is the home domain the client is authenticating with, followed by `auth`. It can be at most 64 characters.
```

leighmcculloch (commented):
One change requested below. Can you also update the meta info, and increment the patch version since there is no functional change?

PR #789, opened 2020-12-15
SEP-12: Added Content-Type policy section

JakeUrban:
relates to #788 

Adds a `Content-Type` section to outline what encodings are accepted for requests and responses. SEP-12 differs from most standards by supporting `multipart/form-data` since requests could contain file uploads.

### DIFF
--- a/ecosystem/sep-0012.md
+++ b/ecosystem/sep-0012.md
@@ -64,6 +64,7 @@ All endpoints respond with content type:
 
 - `application/json`
 
+
 ## Customer GET
 
 This endpoint allows clients to:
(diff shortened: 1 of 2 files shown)


ecosystem/sep-0012.md
```diff
@@ -50,6 +50,20 @@ Alternatively, if the client cannot add the authorization header. The JWT should
...
+
+- `multipart/form-data`
+- `application/x-www-form-urlencoded`
+- `application/json`
+
+All endpoints respond with content type:
+
+- `application/json`
+
+`multipart/form-data` should be used instead of `application/x-www-form-urlencoded` for requests transmitting binary data type values from [SEP-9](sep-0009.md).
```
leighmcculloch, line 65: I think we should reword this a little to make it clear this is backwards compatible and you can continue to use the multipart content-type without binary data.
```suggestion
`multipart/form-data` should be used for requests including binary data type values from [SEP-9](sep-0009.md). `multipart/form-data` or `application/x-www-form-urlencoded` may be used when no binary fields are included.
```

I think this should be moved up to the request paragraph too because it's a little confusing for this to run after the comments about the response format.

leighmcculloch (approved):
💯  To round this change out can you:
- Update the meta incrementing the minor version since there is new functionality.
- Add an example of a form url encoded request. Example of this in another SEP: 
   https://github.com/stellar/stellar-protocol/blob/12f8af237b220032fa8d9332ae7378229769a825/ecosystem/sep-0030.md#L245-L249

JakeUrban:
Added examples in the `PUT` section. 

Do you know if this line is still needed? It doesn't seem particularly helpful since most clients will not be manually constructing these request bodies:
> When uploading data for fields specificed in [SEP-9](./sep-0009.md), `binary` type fields (typically files) should be submitted **after all other fields**. The reason for this is that some web servers require `binary` fields at the end so that they know when they can begin processing the request as a stream.

leighmcculloch:
That statement is probably still required.

ecosystem/sep-0012.md
```diff
@@ -50,6 +50,21 @@ Alternatively, if the client cannot add the authorization header. The JWT should
...
+- `application/x-www-form-urlencoded`
+- `application/json`
+
+`multipart/form-data` should be used for requests including binary data type values from [SEP-9](sep-0009.md). `multipart/form-data` or `application/x-www-form-urlencoded` may be used when no binary fields are included.
+
+All endpoints respond with content type:
+
+- `application/json`
+
+
```
leighmcculloch, line 67: Extra new line here that isn't typical in this file or other SEPs afaik.
```suggestion
```

leighmcculloch (approved):

PR #790, opened 2020-12-15
SEP-31: added Content-Type policy section

JakeUrban:
relates to #788

### DIFF
--- a/ecosystem/sep-0031.md
+++ b/ecosystem/sep-0031.md
@@ -51,7 +51,6 @@ This protocol involves the transfer of value, and so HTTPS is required for all e
 
 All endpoints accept in requests the following `Content-Type`s:
 
-- `application/x-www-form-urlencoded`
 - `application/json`
 
 All endpoints respond with content type:
--- a/ecosystem/sep-0031.md
+++ b/ecosystem/sep-0031.md
@@ -47,6 +47,17 @@ Any API request that fails to meet proper authentication should return a 403 For
 
 This protocol involves the transfer of value, and so HTTPS is required for all endpoints for security.  Anchors should refuse to interact with any insecure HTTP endpoints.
 
+## Content Type
+
+All endpoints accept in requests the following `Content-Type`s:
+
+- `application/x-www-form-urlencoded`
+- `application/json`
+
+All endpoints respond with content type:
+
+- `application/json`
+
 ## API Endpoints
 
 * [`GET /info`](#info)
--- a/ecosystem/sep-0031.md
+++ b/ecosystem/sep-0031.md
@@ -7,8 +7,8 @@ Title: Direct Payments
 Author: SDF
 Status: Active
 Created: 2020-04-07
-Updated: 2020-05-06
-Version 1.0.0
+Updated: 2020-12-15
+Version 1.1.0
 ```
 
 ## Simple Summary


leighmcculloch (commented):
💯  To round this change out can you:
- Update the meta incrementing the minor version since there is new functionality.
- Add an example of a form url encoded request. Example of this in another SEP: 
   https://github.com/stellar/stellar-protocol/blob/12f8af237b220032fa8d9332ae7378229769a825/ecosystem/sep-0030.md#L245-L249

I'd like to take a quick look once the example is added to understand what format is being used given the request is more complex.

JakeUrban:
Ah we have an issue, thanks for asking me to create an example. 

I took the JSON request example:
```
{
  "amount": 100,
  "asset_code": "USD",
  "asset_issuer": "GDRHDSTZ4PK6VI3WL224XBJFEB6CUXQESTQPXYIB3KGITRLL7XVE4NWV",
  "sender_id": "d2bd1412-e2f6-4047-ad70-a1a2f133b25c",
  "receiver_id": "137938d4-43a7-4252-a452-842adcee474c",
  "fields": {
    "transaction": {
      "receiver_routing_number": "442928834",
      "receiver_account_number": "0029483242",
      "type": "SEPA"
    }
  }
}
```

And I tried to make a request using Python's `requests` package (the go-to package for sending requests). I assigned the JSON (in python, a dict) data to the `data` variable, and did this:
```
>>> r = requests.Request("POST", "https://testanchor.stellar.org/sep31/transactions", data=data)
>>> p = r.prepare()
>>> pretty_print_POST(p)
-----------START-----------
POST https://testanchor.stellar.org/sep31/transactions
Content-Length: 210
Content-Type: application/x-www-form-urlencoded

amount=100&asset_code=USD&asset_issuer=GDRHDSTZ4PK6VI3WL224XBJFEB6CUXQESTQPXYIB3KGITRLL7XVE4NWV&sender_id=d2bd1412-e2f6-4047-ad70-a1a2f133b25c&receiver_id=137938d4-43a7-4252-a452-842adcee474c&fields=transaction
```
Note the `fields=transaction` piece at the end. We lost the data under the `transaction` key. We could stringify the `fields` object, but the server would likely receive the data as a string and have to decode it separately.

However, the same problem is present in `multipart/form-data` requests as well. If I try to use this encoding with the same JSON payload, I have to encode it in bytes in order for it to work.

JakeUrban:
Given that in this SEP we never specified a content-type policy and all examples were in JSON, maybe our policy should only allow JSON in requests.

leighmcculloch:
We use urlformencoded content in complicated forms with arrays and sub objects in this app and it works:
https://github.com/stellar/go/tree/master/exp/services/recoverysigner

As an example htese two requests are the same:
```json
{
  "identities": [
    {
      "role": "sender",
      "auth_methods": [
        { "type": "stellar_address", "value": "GDUA..." },
        { "type": "phone_number", "value": "+10000000001" },
        { "type": "email", "value": "person1@example.com" }
      ]
    },
    {
      "role": "receiver",
      "auth_methods": [
        { "type": "stellar_address", "value": "GBEA..." },
        { "type": "phone_number", "value": "+10000000002" },
        { "type": "email", "value": "person2@example.com" }
      ]
    },
  ]
}
```
```
identities.0.role=sender&
identities.0.auth_methods.account=GDUA...&
identities.0.auth_methods.email=person1%40example.com&
identities.0.auth_methods.phone_number=%2B10000000001&
identities.1.role=receiver&
identities.1.auth_methods.account=GBEA...&
identities.1.auth_methods.email=person2%40example.com&
identities.1.auth_methods.phone_number=%2B10000000002
```
(I added new lines in the form encoding to make it easier to read.)

We use a library in Go projects to do decoding of forms like this, and there may not be something in the Python library for it.

leighmcculloch:
I think the real question is not what Python supports, but if you create a basic HTML form with this type of data, how does it get submitted? I'm pretty sure it is relatively simple to build a HTML form that uses this format.

JakeUrban:
URL-encoding arrays is done in several ways:
- `foo=bar1&foo=bar2`
- `foo=bar1,bar2`
- `foo[]=bar1&foo[]=bar2`

In the last example, `foo[]` is url-encoded to `foo%5B%5D`. See https://willhaley.com/blog/url-array-go/

I've never seen and cannot find any use of the array syntax you're using in the example (`key.{index}.key`).

Looking at your example a bit closer, I see there's some discrepancies. Take the `phone_number` case. In JSON, you display it as so:
```
{ "type": "phone_number", "value": "+10000000002" }
``` 
But in the urlformencoded version, there is no `type` and `value` keys. You just have:
```
phone_number=%2B10000000002
```

So I'm not convinced the example you gave is actually a standard.

leighmcculloch:
Oh good point, the request format changed at one point and wasn't updated. I'll update it and reshare here the format we've been using.

leighmcculloch:
This is an example that should be the same that I took from our tests in recoverysigner (SEP-30 reference implementation):

JSON ([test here](https://github.com/stellar/go/blob/013bf861cb0f060be1e2c1b0fdf4cb12a00d6b9d/exp/services/recoverysigner/internal/serve/account_post_test.go#L173-L264)):
```json
{
	"identities": [
		{
			"role": "sender",
			"auth_methods": [
				{ "type": "stellar_address", "value": "GBF3XFXGBGNQDN3HOSZ7NVRF6TJ2JOD5U6ELIWJOOEI6T5WKMQT2YSXQ" },
				{ "type": "phone_number", "value": "+10000000000" },
				{ "type": "email", "value": "user1@example.com" }
			]
		},
		{
			"role": "receiver",
			"auth_methods": [
				{ "type": "stellar_address", "value": "GB5VOTKJ3IPGIYQBJ6GVJMUVEAIYGXZUJE4WYLPBJSHOTKLZTETBYOBI" },
				{ "type": "phone_number", "value": "+20000000000" },
				{ "type": "email", "value": "user2@example.com" }
			]
		}
	]
}
```

Form ([test here](https://github.com/stellar/go/blob/013bf861cb0f060be1e2c1b0fdf4cb12a00d6b9d/exp/services/recoverysigner/internal/serve/account_post_test.go#L266-L355))
```
identities.0.auth_methods.0.type=stellar_address&identities.0.auth_methods.0.value=GBF3XFXGBGNQDN3HOSZ7NVRF6TJ2JOD5U6ELIWJOOEI6T5WKMQT2YSXQ&identities.0.auth_methods.1.type=phone_number&identities.0.auth_methods.1.value=%2B10000000000&identities.0.auth_methods.2.type=email&identities.0.auth_methods.2.value=user1%40example.com&identities.0.role=sender&identities.1.auth_methods.0.type=stellar_address&identities.1.auth_methods.0.value=GB5VOTKJ3IPGIYQBJ6GVJMUVEAIYGXZUJE4WYLPBJSHOTKLZTETBYOBI&identities.1.auth_methods.1.type=phone_number&identities.1.auth_methods.1.value=%2B20000000000&identities.1.auth_methods.2.type=email&identities.1.auth_methods.2.value=user2%40example.com&identities.1.role=receiver
```

You're correct that this isn't a specific standard. There is no standard I know of that preserves index order in arrays. One of the advantages we get this from specific design that is from the [schema](https://github.com/gorilla/schema) package is that indexes are encoded into the values which ensures query parameter reordering doesn't change the value encoded.

If there are better ways of encoding form data though, let's put them on the table here before we merge this and add complex form encoded data to an Active SEP.

leighmcculloch:
>Given that in this SEP we never specified a content-type policy and all examples were in JSON, maybe our policy should only allow JSON in requests.

I agree. The argument to support forms which goes back a couple years when SEP-10 was written is not compelling in the face of no consistent decoding/encoding.

leighmcculloch:
I think you should still make this change making the content type explicit, just drop the form encoding on requests as an option.

ecosystem/sep-0031.md
```diff
@@ -47,6 +47,17 @@ Any API request that fails to meet proper authentication should return a 403 For
 
 This protocol involves the transfer of value, and so HTTPS is required for all endpoints for security.  Anchors should refuse to interact with any insecure HTTP endpoints.
 
+## Content Type
+
+All endpoints accept in requests the following `Content-Type`s:
+
+- `application/x-www-form-urlencoded`
```
leighmcculloch, line 54:
```suggestion
```

leighmcculloch (approved):
👍🏻 with form encoding removed.

JakeUrban:
Thanks @leighmcculloch !

leighmcculloch:
As a side-effect of this, I will remove form encoding from SEP-30 while it is still in draft state.

PR #792, opened 2020-12-15
Add semantic section to CAP Template

leighmcculloch:
### What
Add semantic section to CAP template within the specification section.

### Why
I've noticed that in many of the CAPs that are authored by @jonjove the CAP separates the XDR definitions from the semantics of how those new or changed XDR types function. This template was recently updated. It's not clear to me if the semantics section was intentionally left out with the intention to discuss the semantics within the XDR section, or if it was accidental. I suspect it was accidental because otherwise there would be little reason to have the XDR sub-section of specification.

### DIFF
--- a/cap-template.md
+++ b/cap-template.md
@@ -80,6 +80,11 @@ on top of a different CAP).
 For large changes, it may be beneficial to link to actual XDR files copied
 in the relevant "contents" folder.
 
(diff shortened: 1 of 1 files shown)


MonsieurNicolas:
Yes this is good default stance for a CAP. I am merging it

PR #794, opened 2020-12-15
CAP-0035: Move XDRs into a single section and move discussion to a semantics section

leighmcculloch:
### What
Move XDRs into a single section and move discussion to a semantics section.

### Why
To adhere to the CAP template. No functional changes to the CAP.

### DIFF
--- a/core/cap-0035.md
+++ b/core/cap-0035.md
@@ -81,10 +81,6 @@ in the event of a fraudulent or mistaken transaction.
 
 ### XDR changes
 
-#### Account
-This proposal requires the addition of `CLAWBACK_ENABLED` to the issuing account 
-AccountFlags so the XDR becomes
- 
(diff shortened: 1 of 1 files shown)


core/cap-0035.md
```diff
@@ -159,7 +142,22 @@ struct Operation
...
     body;
 };
-``` 
+```
+
+### Semantics
+
+#### Account
+This proposal requires the addition of `CLAWBACK_ENABLED` to the issuing account 
+AccountFlags so the XDR becomes
```
accordeiro, line 151: Is the intentional to add the XDR here? Or should we just remove this line?

leighmcculloch: Oops: e049165

accordeiro (approved):
LGTM!

PR #795, opened 2020-12-16
CAP-0035: Address concerns raised by Jon and Nicolas on the mailing list

leighmcculloch:
### What
Address concerns raised by Jon and Nicolas on the [mailing list](https://groups.google.com/g/stellar-dev/c/hPhkXhrl5-Y/m/ZF6eJcqKAgAJ?pli=1).

Specifically:
- Remove `AUTH_CLAWBACKWABLE` and use `AUTH_REVOCABLE`.
- Change what happens when the `CallbackOp` is used on an account that has selling liabilities such that it functions like other operations and fails.
- Include an example of how to use `AllowTrustOp` to temporarily revoke authorization of an account to free up selling liabilities.
- Include discussion highlighting how the limitations of clawback match the limitations of revoking authorization, and how they cannot affect claimable balances.
- Include discussion about backwards compatibility.
- Include discussion about thresholds.

### Why
Addressing feedback from the [discussion thread](https://groups.google.com/g/stellar-dev/c/hPhkXhrl5-Y/m/ZF6eJcqKAgAJ?pli=1) from @jonjove and @MonsieurNicolas and @dandoney. 

It was agreed in the discussion thread that revoking all offers is simpler and acceptable, and this can be achieved with existing logic assuming the issuer account can revoke authorization (has its `AUTH_REVOCABLE` flag set). This simplification also lends to us fitting this functionality into the existing `AUTH_REVOCABLE` feature set. Clawback is really a fine grain version of revocation which has more flexibility than the existing revocation functionality provides.

It seems difficult to disconnect this new flag from `AUTH_IMMUTABLE` because it is reasonable for an asset holder to assume any asset already issued with `AUTH_IMMUTABLE` cannot be clawed back. Even though we could argue clawback has nothing or little to do with authorization, it seems to be tied to it conceptually. For this reason the functionality lends itself to being considered a form of auth revocation.

Claimable balances are already not revokable, so it seems consistent to make them not clawbackable. The changes to the proposal include discussion about this. I've opened a separate proposal, CAP-36 (#797) to discuss an approach we could take similar to this proposal for claimable balances.

### DIFF
--- a/core/cap-0035.md
+++ b/core/cap-0035.md
@@ -82,14 +82,6 @@ in the event of a fraudulent or mistaken transaction.
 ### XDR changes
 
 ```c
-enum AccountFlags
-{
-    AUTH_IMMUTABLE = 1,                                      
-    AUTH_REQUIRED = 2,                                  
-    AUTH_REVOCABLE = 4,                                  
-    AUTH_CLAWBACKABLE = 8
-};
-
 struct ClawbackOp 
 {
     union switch (AssetType type)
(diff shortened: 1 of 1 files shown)


core/cap-0035.md
```diff
@@ -182,15 +237,41 @@ in many cases should be reserved for licensed entities (like a transfer agent)
 holding the issuer credentials and aware of responsibilities under the law of 
 the jurisdiction of the affected party and asset. 
  
+### Reusing the AUTH_REVOCABLE flag
+The account `AUTH_REVOCABLE` flag allows the issuer to indicate that it has
+control over the use of the asset on the network. By including the
+`AUTH_REVOCABLE` flag in account flags, account owners may review the
+revocability of an asset issued by the issuer and have the choice to avoid this
+type of asset if they object to the implied trust in the issuer. Clawback is
+another form of an issuer revoking use of an asset with fine control over the
```
accordeiro, line 246: I agree with the similarities described here, but I think it's worth making it clearer that `ClawbackOp` effectively reduces the total number of an asset that exists in the network, whereas `AllowTrustOp` doesn't. So in some ways 
 ClawbackOp` might seem more definitive in theory, although not necessarily in practice.

core/cap-0035.md
```diff
@@ -63,115 +61,175 @@ exchange of assets, throughout the globe, enabling users to make payments betwee
...
+The `ClawbackOp` operation reduces the balance of the asset in the account by
+the specified amount of the specific `asset` from the `from` account,
+effectively returning it to the issuer account, burning it.
+
+Similar to other operations the clawback operation will fail if the account
+balance is less than the amount specified when account for selling liabilities.
+If clawback is required of asset amounts locked up with selling liabilities then
+the issuer may use the `AllowTrustOp` operation to revoke authorization of the
+trustline, which will cancel any existing ledger entries creating selling
+liabilities, such as offers, and issue the `ClawbackOp` in the same operation.
```
MonsieurNicolas, line 207: typo "operation" should be "transaction"

PR #797, opened 2020-12-16
CAP-0036: Add proposal for ClawbackClaimableBalanceOp as a sibling to CAP-35

leighmcculloch:
### What
Add proposal for ClawbackClaimableBalanceOp as a sibling to CAP-35 that proposes a method similar to CAP-35's clawback of an asset within an account that claws back an asset within a claimable balance.

### Why
The issue of claimable balances keeps occurring in relation to CAP-35 and this seems like the most explicit and simplest approach to also allowing issuers to clawback amounts in claimable balances. This should be considered in conjunction with CAP-35 but does not need to be tied to it and is otherwise independent.


core/cap-0036.md
```diff
@@ -0,0 +1,189 @@
...
+The change does make it possible for claimable balances to be destroyed, and
+claimable balances could form an important part of series of transactions that
+form a logical contract. The use of a clawback could break existing contracts
+that did not consider the possibility of clawback. Issuers today who use the
+`AUTH_REVOCABLE` flag get the same capability by freezing claimaint accounts.
+Auth revocation is reversable meaning an accidental auth revocation if reversed
+would have no impact on a planned series of transactions that form a logical
+contract. However, a clawed back claimable balance recreated would have a new
+claimable balance ID and would not be claimable with pre-signed or
+pre-authorized transactions.
```
leighmcculloch, line 178: @MonsieurNicolas This backwards incompatibility could be avoided if claimable balances were mutable and instead of destroying the claimable balance entry its amount was altered. I realize from the mailing list discussion thread you mentioned that mutability of this ledger entry is undesirable. This is the tradeoff.

core/README.md
```diff
@@ -65,6 +65,7 @@
 | [CAP-0029](cap-0029.md) | AllowTrust when not AUTH_REQUIRED | Tomer Weller | Draft |
 | [CAP-0032](cap-0032.md) | Trustline Preauthorization | Jonathan Jove | Draft |
 | [CAP-0035](cap-0035.md) | Asset Clawback | Dan Doney | Draft |
+| [CAP-0035](cap-0036.md) | Claimable Balance Clawback | Leigh McCulloch | Draft |
```
tomerweller, line 68:
```suggestion
| [CAP-0036](cap-0036.md) | Claimable Balance Clawback | Leigh McCulloch | Draft |
```

tomerweller (approved):
LGTM aside from that typo

core/cap-0036.md
leighmcculloch: Rather than this be a PR comment, I've placed this comment into the proposal in the design rationale section so it is visible when this is discussed on the mailing list. 94998f1cb8821fdbff1e54202d6636e6b89897fd

PR #798, opened 2020-12-16
Clarify SEP-24 account funding policy (special case)

### LINKED ISSUE #771: SEP-24: improve special cases handling
**Why**

In the [section](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0024.md#special-cases) of handling special cases, what we recommend for the two cases are inconsistent. In [stellar account does not exist](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0024.md#stellar-account-does-not-exist), we recommend anchors to convert part of the deposit into XLM while in [stellar account does not trust asset](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0024.md#stellar-account-doesnt-trust-asset), we recommend anchors to `send the needed amount of XLM to the Account for creating a trust line` without saying whether the needed amount should be taken from the deposit.

Regarding the section for `stellar account does not exist`:
1. We don't really know why 2.01 XLM was previously recommended.
2. The reserve should be in addition to the deposit amount. However, our current recommendation suggests that anchors should convert part of the deposit into XLM, but we don't have a guideline for anchors to communicate the fact that some of the deposit will be used as the account reserve.
3. If we end up having anchors send extra XLM to users' accounts as the account reserve and transaction fee, the same should be done in other cases where a user already has an account with or without a trustline.

Given that our goal is not to have users think about XLM reserves at all, we have rooms to improve for the above points. Finally, it would be great to learn which anchors have implemented the account creation flow or the trustline-reserve-sponsor flow so that we can know how they are doing it.

JakeUrban:
resolves #771 

Currently, the anchor is instructed to not list the minimum funding amount (2.01XLM recommended) as a fee because the user still receives that amount in their account. While true, the user cannot transact that minimum funding amount, due to it being the _minimum amount the account needs to exist_.

Its also in the anchors best interest to sponsor account creations, so deducting the minimum funding amount from the deposit given to the anchor likely brings more complexity and UX dissatisfaction than its worth.

### DIFF
--- a/ecosystem/sep-0024.md
+++ b/ecosystem/sep-0024.md
@@ -173,21 +173,21 @@ Responses are detailed in the [Deposit and Withdraw shared responses](#deposit-a
 ##### Special Cases
 ###### Stellar account does not exist
 
-If the given Stellar `account` does not exist, on receipt of the deposit, the anchor should use `create_account` to create the account with at least enough XLM for the minimum reserve and a trust line (2.01 XLM is recommended). The anchor should take some of the asset that is sent in to pay for the XLM. The anchor should not list this minimal funding as a fee because the user still receives the value of the XLM in their account. The anchor should detect if the account has been created before returning deposit information, and adjust the `min_amount` in the response to be at least the amount needed to create the account.
+If the given Stellar `account` does not exist on receipt of the deposit funds, the anchor should use the `CreateAccount` operation to create the account with at least enough XLM for the minimum reserve and a trust line to the requested asset (2.01 XLM is recommended). The anchor can add this minimal funding amount to the service fee, but this requires calculating the worth of the minimum funding amount in units of the requested asset.
 
-Since the anchor doesn't have the user account's secret key, the user must create a trust line to the anchor's asset before the anchor can send the remaining asset tokens to the user's account. The anchor should listen for the user to establish this trust line. Once the trust line is there, the anchor should send the remaining asset tokens to the account in Stellar to complete the deposit.
+Since the anchor doesn't have the user account's secret key, the user must create a trust line to the anchor's asset before the anchor can send the requested asset tokens to the user's account. The anchor should listen for the user to establish this trust line. Once the trust line is there, the anchor should send the requested asset tokens to the account on Stellar to complete the deposit.
 
-If the anchor does not support creating new accounts for users and `account` doesn't exist yet, the anchor should return a `400 Bad Request` error. The response body should be a JSON object containing an `error` field that explains why the request failed.
+If the anchor does not support creating new accounts for users and `account` doesn't exist yet, the anchor should return a `400 Bad Request` error in the `transactions/deposit/interactive` response. The response body should be a JSON object containing an `error` field that explains why the request failed.
 
 ###### Stellar account doesn't trust asset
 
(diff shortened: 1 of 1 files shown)


JakeUrban:
The equivalent changes will be made to SEP-6 once the changes here are finalized.

ecosystem/sep-0024.md
```diff
@@ -173,21 +173,21 @@ Responses are detailed in the [Deposit and Withdraw shared responses](#deposit-a
...
+Since the anchor doesn't have the user account's secret key, the user must create a trust line to the anchor's asset before the anchor can send the requested asset tokens to the user's account. The anchor should listen for the user to establish this trust line. Once the trust line is there, the anchor should send the requested asset tokens to the account on Stellar to complete the deposit.
 
-If the anchor does not support creating new accounts for users and `account` doesn't exist yet, the anchor should return a `400 Bad Request` error. The response body should be a JSON object containing an `error` field that explains why the request failed.
+If the anchor does not support creating new accounts for users and `account` doesn't exist yet, the anchor should return a `400 Bad Request` error in the `transactions/deposit/interactive` response. The response body should be a JSON object containing an `error` field that explains why the request failed.
```
howardtw, line 180: > in the `transactions/deposit/interactive` response

Can we use a link directing the user to the relevant section instead? Like #deposit or #response or #6-error?

ecosystem/sep-0024.md
```diff
@@ -173,21 +173,21 @@ Responses are detailed in the [Deposit and Withdraw shared responses](#deposit-a
...
-2. If `Account` doesn't have enough XLM, `Wallet` starts listening for transactions to the given `Account`, waiting for it to have enough XLM for a trust line.
-3. When asked for a deposit, `Anchor` detects if `Account` has enough XLM to create a trust line. If it doesn't, `Anchor` sends the needed amount of XLM to the `Account` for creating a trust line. `Anchor` then starts listening for trust line creations for that `Account`.
-4. `Wallet` detects the arrival of XLM in the `Account`, and establishes a trust line.
-5. `Anchor` detects the trust line creation in the `Account`. If the asset is `AUTH_REQUIRED`, `Anchor` approves the new trust line.
+1. `Wallet` checks if `account` has enough XLM to create a trust line. If it does, skip to step `4`.
+2. If `account` doesn't have enough XLM, `Wallet` starts listening for transactions to the given `account`, waiting for it to have enough XLM for a trust line.
+3. When asked for a deposit, `Anchor` detects if `account` has enough XLM to create a trust line. If it doesn't, `Anchor` sends the needed amount of XLM to the `Account` for creating a trust line. `Anchor` then starts listening for trust line creations for that `account`.
```
howardtw, line 188: Should we also clarify that the sponsored xlm should not be taken from the user deposit? It'd better to be added as the service fee if possible.

howardtw (commented):
LGTM overall. Think that we need to add some clarification for the sponsored reserve in the case of `Stellar account doesn't trust asset` for consistency.

JakeUrban:
@tomerweller here is the PR discussing XLM reserves and how the anchor should handle the cost.

howardtw (approved):
LGTM

tomerweller (approved):
LGTM. I agree that the previous version was too prescriptive and this looks strictly better.

PR #799, opened 2020-12-18
CAP-0035: Add note about how issuers can prevent creation of claimable balances

leighmcculloch:
### What
Add note about how issuers can prevent creation of claimable balances.

### Why
@MonsieurNicolas identified [on the mailing list](https://groups.google.com/g/stellar-dev/c/hPhkXhrl5-Y/m/l2OTWiKSAQAJ) that `AUTHORIZED_TO_MAINTAIN_LIABILITIES_FLAG` introduced in CAP-18 and SEP-8 could be used by a regulated asset issuer to restrict the creation of claimable balances.


tomerweller (commented):
LGTM. Thanks

PR #800, opened 2020-12-19
SEP-0030: Remove urlencoded form data as an accepted content-type in requests

leighmcculloch:
### What
Remove urlencoded form data as a accepted content-type in requests.

### Why
@JakeUrban pointed out how there is really no consistent standard for encoding complex nested data in urlencoded form, which means that integrations are highly likely to use it, or to be buggy. We started using dual content types in requests in SEP-10 and we applied that broadly here but the difficulty to use is compelling that this isn't going to have a positive return on investment to maintain and isn't a pattern we should propogate to other SEPs either.

Related conversation in https://github.com/stellar/stellar-protocol/pull/790.

This change is breaking but the proposal is in draft and that is acceptable. It is also unlikely that there are implementations using form encoding given the difficult to provide that feature.

The recoverysigner reference implementation can be updated to remove the urlencoded form data decoding, but it doesn't have to be to continue to meet the standard. When we have time we can update it. I have created https://github.com/stellar/stellar-protocol/issues/801 to track that work.

### DIFF
--- a/ecosystem/sep-0030.md
+++ b/ecosystem/sep-0030.md
@@ -9,7 +9,7 @@ Status: Draft
 Discussion: https://groups.google.com/forum/#!topic/stellar-dev/SFr2dHBZlsY
 Created: 2019-12-20
 Updated: 2020-11-17
-Version: 0.7.0
+Version: 0.8.0
 ```
 
 ## Summary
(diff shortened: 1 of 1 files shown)


JakeUrban (approved):
Love it 👍

leighmcculloch:
I have shared this change [here](https://groups.google.com/g/stellar-dev/c/SFr2dHBZlsY/m/CGzYzkDbAQAJ) and I'll merge this change on Dec 23rd if there are no objections.

PR #804, opened 2020-12-22
SEP-9: Add bank branch field

brunopedrazza:
Add `bank_branch` field to Natural Person KYC fields in order to identify customer's bank branch number wich is needed in Brazil, for example.

### DIFF
--- a/ecosystem/sep-0009.md
+++ b/ecosystem/sep-0009.md
@@ -6,8 +6,8 @@ Title: Standard KYC / AML fields
 Author: stellar.org
 Status: Active
 Created: 2018-07-27
-Updated: 2020-04-03
-Version 1.2.0
+Updated: 2020-12-22
+Version 1.2.1
 ```
 
 ## Simple Summary
@@ -48,6 +48,7 @@ Name | Type | Description
 `birth_place` | date | Place of birth (city, state, country; as on passport)
 `birth_country_code` | country code | ISO Code of country of birth
 `bank_account_number` | string | Number identifying bank account
+`bank_agency_number` | string | Number identifying bank agency
 `bank_number` | string | Number identifying bank in national banking system (routing number in US)
 `bank_phone_number` | string | Phone number with country code for bank
 `tax_id` | string | Tax identifier of user in their country (social security number in US)
--- a/ecosystem/sep-0009.md
+++ b/ecosystem/sep-0009.md
@@ -50,7 +50,7 @@ Name | Type | Description
 `bank_account_number` | string | Number identifying bank account
 `bank_number` | string | Number identifying bank in national banking system (routing number in US)
 `bank_phone_number` | string | Phone number with country code for bank
-`bank_branch` | string | Number identifying bank branch
+`bank_branch_number` | string | Number identifying bank branch
 `tax_id` | string | Tax identifier of user in their country (social security number in US)
 `tax_id_name` | string | Name of the tax ID (`SSN` or `ITIN` in the US)
 `occupation` | number | Occupation ISCO code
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0009.md
```diff
@@ -6,8 +6,8 @@ Title: Standard KYC / AML fields
 Author: stellar.org
 Status: Active
 Created: 2018-07-27
-Updated: 2020-04-03
-Version 1.2.0
+Updated: 2020-12-22
+Version 1.2.1
```
leighmcculloch, line 10: As this change adds a new field, it should increment the minor number instead of the patch number.

```suggestion
Version 1.3.0
```

ecosystem/sep-0009.md
```diff
@@ -50,6 +50,7 @@ Name | Type | Description
 `bank_account_number` | string | Number identifying bank account
 `bank_number` | string | Number identifying bank in national banking system (routing number in US)
 `bank_phone_number` | string | Phone number with country code for bank
+`bank_branch` | string | Number identifying bank branch
```
leighmcculloch, line 53: This field could be confused for a branch name, the fact it's a number is not clear from the name. Suggest using the format off some other fields here and using the `_number` suffix.

```suggestion
`bank_branch_number` | string | Number identifying bank branch
```

leighmcculloch (commented):
+1 This would be a good idea to add I think. Brazil isn't the only country which has a branch code/number.

At some point it might also make sense to have an IBAN field that could optionally used instead which is designed to make this easier across countries, but the localization is helpful so this is still good.

Some requested changes below.

I defer to @JakeUrban for a final look after those changes are applied.

brunopedrazza:
It makes a lot of sense. Changes were applied!

leighmcculloch (approved):
👍🏻 Looks great to me. 

@JakeUrban does this look fine to you?

ecosystem/sep-0009.md
```diff
@@ -50,6 +50,7 @@ Name | Type | Description
 `bank_account_number` | string | Number identifying bank account
 `bank_number` | string | Number identifying bank in national banking system (routing number in US)
 `bank_phone_number` | string | Phone number with country code for bank
+`bank_branch_number` | string | Number identifying bank branch
```
JakeUrban, line 53: Unless branch identifiers are _always_ numbers, I would make this field name more generic:
```suggestion
`bank_branch_identifier` | string | An identifier for the bank branch
```
If branch _numbers_ are the standard then this suggestion can be ignored.

JakeUrban (approved):
This LGTM with one suggestion

ecosystem/sep-0009.md
Bocoi5011: `bank_branch_identifier` Affin Bank 
An identifier for the. Mohd Faizal Bin Julung 
typically a branch
number. Bank 205070110343

ecosystem/sep-0009.md
```diff
@@ -50,6 +50,7 @@ Name | Type | Description
...
 `tax_id` | string | Tax identifier of user in their country (social security number in US)
 `tax_id_name` | string | Name of the tax ID (`SSN` or `ITIN` in the US)
```
Bocoi5011, line 55: Affin Bank
205070110343

JakeUrban:
@Bocoi5011 is this what `bank_branch_number` is supposed to refer to: https://en.wikipedia.org/wiki/ISO_9362 ?

leighmcculloch:
> @Bocoi5011 is this what `bank_branch_number` is supposed to refer to: https://en.wikipedia.org/wiki/ISO_9362 ?

ISO9362 defines the BIC/Swift code which is similar to an IBAN, both which include a branch code alongside the bank code and other details. The branch number being added in this pr is also commonly referred to as the branch code. Either term works. So the bank branch number is not a BIC, but a BIC will sometimes contain a branch code.

JakeUrban:
Ok, thanks @leighmcculloch I'll merge this then

PR #805, opened 2020-12-23
SEP-12 callback endpoint

morleyzhi:
Wallets need a way to check a user's KYC status even if the user isn't available to sign a SEP-10 request. This change adds a POST /customer endpoint that allows wallets to provide a callback URL. Anchors will POST whenever the status changes (in the same format as responses to GET /customer) until the user is accepted or rejected.

### DIFF
--- a/ecosystem/sep-0012.md
+++ b/ecosystem/sep-0012.md
@@ -339,7 +339,7 @@ PUT [KYC_SERVER || TRANSFER_SERVER]/customer/callback
 | `account`   | `G...` string | (optional) The Stellar account ID used to identify this customer. If many customers share the same Stellar account, the `memo` and `memo_type` parameters should be included as well.                                                                                                                 |
 `memo` | `string` | (optional) The memo used to create the customer record |
 `memo_type` | `string` | (optional) The type of memo used to create the customer record |
-| `url` | `string` | A valid URL. |
+| `url` | `string` | A callback URL that the SEP-12 server will POST to when the state of the account changes. |
 
 ### PUT Responses
 
--- a/ecosystem/sep-0012.md
+++ b/ecosystem/sep-0012.md
@@ -36,8 +36,8 @@ To support this protocol an anchor acts as a server and implements the specified
 
 * [`GET /customer`](#customer-get): Check the status of a customers info
 * [`PUT /customer`](#customer-put): Idempotent upload of customer info
-* [`DELETE /customer`](#customer-delete): Deletion of customer info
-* [`POST /customer`](#customer-post): Set callbacks for a wallet to receive status updates from the anchor
+* [`DELETE /customer/[account]`](#customer-delete): Deletion of customer info
+* [`POST /customer/callbacks`](#customer-post): Set a callback for a wallet to receive status updates from the anchor
 
 ## Authentication
 
@@ -317,7 +317,7 @@ Success | `200 OK`
 Client not authenticated properly | `401 Unauthorized`
 Anchor has no information on the customer | `404 Not Found`
 
-## Customer POST
+## Customer callback POST
 
 Allow the wallet to provide a callback URL to the anchor.
 
@@ -325,17 +325,19 @@ Whenever the user's
 `status` field changes, the anchor will issue a POST request to the callback
 URL. The payload of the POST request will be the same as [`GET /customer`](#customer-get).
 
-The anchor will POST updates to the callback URL until the user's status changes to ACCEPTED or REJECTED.
+If a wallet submits a callback URL that has already been set, the anchor shouldn't POST changes more than once. But a wallet might submit more than one unique callback URL.
 
 ### Request
 
 ```
-POST [KYC_SERVER || TRANSFER_SERVER]/customer/[account]
+POST [KYC_SERVER || TRANSFER_SERVER]/customer/callbacks
 ```
 
 | Name                 | Type     | Description             |
 | -------------------- | -------- | ----------------------- |
-| `on_change_callback` | `string` | (optional) A valid URL. |
+| `id`        | `string`      | (optional) The ID of the customer as returned in the response of a previous `PUT` request. If the customer has not been registered, they do not yet have an `id`.                                                                                                                                     |
+| `account`   | `G...` string | (optional) The Stellar account ID used to identify this customer. If many customers share the same Stellar account, the `memo` and `memo_type` parameters should be included as well.                                                                                                                 |
+| `url` | `string` | A valid URL. |
 
 ### POST Responses
 
--- a/ecosystem/sep-0012.md
+++ b/ecosystem/sep-0012.md
@@ -7,7 +7,7 @@ Author: Interstellar
 Status: Active
 Created: 2018-09-11
 Updated: 2020-12-23
-Version 1.2.0
+Version 1.2.1
 ```
 
 ## Abstract
@@ -16,39 +16,37 @@ This SEP defines a standard way for stellar clients to upload KYC (or other) inf
 
 This SEP was made with these goals in mind:
 
-- interoperability
-- Allow a customer to enter their KYC information to their wallet once and use it across many services without re-entering information manually
-- handle the most common 80% of use cases
-- handle image and binary data
-- support the set of fields defined in [SEP-9](sep-0009.md)
-- support authentication via [SEP-10](sep-0010.md)
-- support the provision of data for [SEP-6](sep-0006.md), [SEP-24](sep-0024.md), [SEP-31](sep-0031.md), and others
-- give customers control over their data by supporting complete data erasure
+* interoperability
+* Allow a customer to enter their KYC information to their wallet once and use it across many services without re-entering information manually
+* handle the most common 80% of use cases
+* handle image and binary data
+* support the set of fields defined in [SEP-9](sep-0009.md)
+* support authentication via [SEP-10](sep-0010.md)
+* support the provision of data for [SEP-6](sep-0006.md), [SEP-24](sep-0024.md), [SEP-31](sep-0031.md), and others
+* give customers control over their data by supporting complete data erasure
 
 To support this protocol an anchor acts as a server and implements the specified REST API endpoints, while a wallet implements a client that consumes the API. The goal is interoperability, so a wallet implements a single client according to the protocol, and will be able to interact with any compliant anchor. Similarly, an anchor that implements the API endpoints according to the protocol will work with any compliant wallet.
 
 ## Prerequisites
 
-- An anchor must define the location of their `KYC_SERVER` or `TRANSFER_SERVER` in their [`stellar.toml`](sep-0001.md). This is how a client app knows where to find the anchor's server. A client app will send KYC requests to the `KYC_SERVER` if it is specified, otherwise to the `TRANSFER_SERVER`.
-- Anchors and clients must support [SEP-10](sep-0010.md) web authentication and use it for all SEP-12 endpoints.
+* An anchor must define the location of their `KYC_SERVER` or `TRANSFER_SERVER` in their [`stellar.toml`](sep-0001.md). This is how a client app knows where to find the anchor's server. A client app will send KYC requests to the `KYC_SERVER` if it is specified, otherwise to the `TRANSFER_SERVER`.
+* Anchors and clients must support [SEP-10](sep-0010.md) web authentication and use it for all SEP-12 endpoints.
 
 ## API Endpoints
 
-- [`GET /customer`](#customer-get): Check the status of a customers info
-- [`PUT /customer`](#customer-put): Idempotent upload of customer info
-- [`DELETE /customer`](#customer-delete): Deletion of customer info
-- [`POST /customer`](#customer-post): Set callbacks for a wallet to receive status updates from the anchor
+* [`GET /customer`](#customer-get): Check the status of a customers info
+* [`PUT /customer`](#customer-put): Idempotent upload of customer info
+* [`DELETE /customer`](#customer-delete): Deletion of customer info
+* [`POST /customer`](#customer-post): Set callbacks for a wallet to receive status updates from the anchor
 
 ## Authentication
 
 Clients should submit the JWT previously obtained from the anchor via the [SEP-10](sep-0010.md) authentication flow. The JWT should be included in all requests as request header:
-
 ```
 Authorization: Bearer <JWT>
 ```
 
 Alternatively, if the client cannot add the authorization header. The JWT should be passed as a jwt query parameter:
-
 ```
 ?jwt=<token>
 ```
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0012.md
```diff
@@ -16,36 +16,39 @@ This SEP defines a standard way for stellar clients to upload KYC (or other) inf
...
 
-* interoperability
-* Allow a customer to enter their KYC information to their wallet once and use it across many services without re-entering information manually
-* handle the most common 80% of use cases
-* handle image and binary data
-* support the set of fields defined in [SEP-9](sep-0009.md)
-* support authentication via [SEP-10](sep-0010.md)
-* support the provision of data for [SEP-6](sep-0006.md), [SEP-24](sep-0024.md), [SEP-31](sep-0031.md), and others
-* give customers control over their data by supporting complete data erasure
+- interoperability
```
leighmcculloch, line 19: ❗This is an active SEP. It's helpful if we keep changes surgical, so it's ultra clear to implementers what has changed. We can refactor but like code it's helpful if the refactor commits make no functional change so implementers can ignore them. can you move the format changes to a separate PR? I'd also question if the format changes are necessary. Even with refactors we should weigh the return on investment vs churning diffs and making it harder to follow the diffs history.

ecosystem/sep-0012.md
```diff
@@ -6,7 +6,7 @@ Title: Anchor/Client customer info transfer
 Author: Interstellar
 Status: Active
 Created: 2018-09-11
-Updated: 2020-12-15
+Updated: 2020-12-23
 Version 1.2.0
```
leighmcculloch, line 10: ❗This change adds a new feature can you bump the minor version?

```suggestion
 Version 1.3.0
```

ecosystem/sep-0012.md
```diff
@@ -6,7 +6,7 @@ Title: Anchor/Client customer info transfer
 Author: Interstellar
```
leighmcculloch, line 6: 💡If you'd like you can add your name to the list of authors here.

ecosystem/sep-0012.md
```diff
@@ -16,36 +16,39 @@ This SEP defines a standard way for stellar clients to upload KYC (or other) inf
...
 
 ## API Endpoints
 
-* [`GET /customer`](#customer-get): Check the status of a customers info
-* [`PUT /customer`](#customer-put): Idempotent upload of customer info
-* [`DELETE /customer`](#customer-delete): Deletion of customer info
+- [`GET /customer`](#customer-get): Check the status of a customers info
+- [`PUT /customer`](#customer-put): Idempotent upload of customer info
+- [`DELETE /customer`](#customer-delete): Deletion of customer info
+- [`POST /customer`](#customer-post): Set callbacks for a wallet to receive status updates from the anchor
```
leighmcculloch, line 40: ❗I think this breaks the existing REST  pattern this SEP exposes. The description indicates that this sets a value which implies to me it can be called with an updated value. For consistency with the rest of this SEP it should use the `PUT`.

❓Is there a need to set the callback in a separate call? I think the new `on_change_callback` field should be added added to the existing `PUT /customer`. If there's a need to set the callback separately then we should make it a sub-path, e.g. `PUT /customer/callbacks`.

leighmcculloch (changes requested):
The idea makes sense to me, I like that the `on_change_callback` works the same way as with SEP-24. I've left some comments in line: critical asks (❗), questions (❓), and suggestions (💡). The diff is pretty large because of the reformatting included so I need to take another look after the reformatting is removed.

ecosystem/sep-0012.md
morleyzhi: Sorry, the reformatting happened automatically. just pushed it with the formatting removed.

ecosystem/sep-0012.md
morleyzhi: I think the POST verb applies better than the PUT verb. PUT implies idempotency, but a wallet may want to set more than one callback on a given KYC user.

I don't feel strongly about `/customer` vs. `/customer/callbacks`, and I think the distinction is a semantic / aesthetic one, so I'll just take your suggestion and change the URL.

morleyzhi:
@leighmcculloch Changes made!

ecosystem/sep-0012.md
leighmcculloch: Is the reason there would be multiple because an account holder might be using several different wallet applications, and each might register for callbacks?

I'd expect if it's just one application it'd be one callback url and if it a new url is generated updating would be sufficient, and so PUT would be more appropriate.

Could you elaborate on the multiple use case?

If we have multiple we need to think about deleting or unregistering callbacks. That probably means returning a callback id. Or define when they expire naturally. Exposing might be better so that old wallets don't get callbacks after a user switches wallets.

If we require the callback to be included in the main PUT call we could tie cannabis to that specific kyc run, as opposed to being registered with the account forever. Once that kyc run completes it could be considered expired.

leighmcculloch: Is the reason there would be multiple because an account holder might be using several different wallet applications, and each might register for callbacks?

I'd expect if it's just one application it'd be one callback url and if it a new url is generated updating would be sufficient, and so PUT would be more appropriate.

Could you elaborate on the multiple use case?

If we have multiple we need to think about deleting or unregistering callbacks. That probably means returning a callback id. Or define when they expire naturally. Exposing might be better so that old wallets don't get callbacks after a user switches wallets.

If we require the callback to be included in the main PUT call we could tie cannabis to that specific kyc run, as opposed to being registered with the account forever. Once that kyc run completes it could be considered expired.

ecosystem/sep-0012.md
```diff
@@ -315,3 +316,33 @@ Situation | Response
...
+### Request
+
+```
+POST [KYC_SERVER || TRANSFER_SERVER]/customer/callbacks
+```
+
+| Name                 | Type     | Description             |
+| -------------------- | -------- | ----------------------- |
+| `id`        | `string`      | (optional) The ID of the customer as returned in the response of a previous `PUT` request. If the customer has not been registered, they do not yet have an `id`.                                                                                                                                     |
+| `account`   | `G...` string | (optional) The Stellar account ID used to identify this customer. If many customers share the same Stellar account, the `memo` and `memo_type` parameters should be included as well.                                                                                                                 |
```
leighmcculloch, line 339: Can you explicitly list the memo fields?

leighmcculloch (commented):
One thing below, and I left a comment on a thread above too.

I think we need to consider and capture here how callbacks functions with multiple wallets, one anchor. It's not clear to me if multiple wallets would received all updates forever which could be a security concern. Typically wallets only gain insight into a users kyc status when authenticated with the accounts keys. Or do callbacks expire after that specific kyc run?

morleyzhi:
I'll just change the endpoint to a PUT to move things along.

morleyzhi:
This is now limited to one callback at a time that stops POSTing once accepted or rejected.

ecosystem/sep-0012.md
```diff
@@ -315,3 +316,38 @@ Situation | Response
...
+`memo` | `string` | (optional) The memo used to create the customer record |
+`memo_type` | `string` | (optional) The type of memo used to create the customer record |
+| `url` | `string` | A valid URL. |
+
+### PUT Responses
+
+| Situation                                 | Response           |
+| ----------------------------------------- | ------------------ |
+| Success                                   | `200 OK`           |
+| Client not authenticated properly         | `401 Unauthorized` |
```
JakeUrban, line 352: I think we should stick to using `403 Permission Denied` since thats used in SEP 6, 24, & 31. Maybe others as well.

JakeUrban: We should probably add a comment about using 403 in the authentication section since it applies to any request here.

ecosystem/sep-0012.md
```diff
@@ -315,3 +316,38 @@ Situation | Response
 Success | `200 OK`
 Client not authenticated properly | `401 Unauthorized`
 Anchor has no information on the customer | `404 Not Found`
+
+## Customer callback PUT
```
JakeUrban, line 320: Can we remove the newline characters in this section to match the style of the rest of the doc?

ecosystem/sep-0012.md
```diff
@@ -315,3 +316,38 @@ Situation | Response
...
+
+Whenever the user's
+`status` field changes, the anchor will issue a POST request to the callback
+URL. The payload of the POST request will be the same as [`GET /customer`](#customer-get).
+
+Anchors will submit POST requests until the user's status changes to `ACCEPTED`
+or `REJECTED`. If a wallet needs to watch a user's KYC status after that,
+it will need to set a callback again.
+
+### Request
```
JakeUrban, line 333: If a callback request to the provided URL fails, is the server obligated to retry? I don't think it is, but we should make that clear in this section.

JakeUrban (commented):
If we're committed to this approach (server's making callback requests) then this LGTM. We chose to take a similar approach with SEP-24's `on_change_callback` so I can see why this makes sense from a consistency perspective.

However, I'm not thrilled with server-side callback requests in general because it adds complexity to backend services. In order to implement server-side callbacks properly, the server needs to maintain a queue of callbacks to make and run a separate process for dequeueing and making the requests. This ensures the server's API remains responsive regardless of how many callbacks need to be made.

I would prefer to use an approach suggested earlier (can't find it now) that exposed the `id` and `status` of a customer to unauthenticated clients. This is much simpler to implement for both parties, and the information is still somewhat protected due to unguessable memos/ids.

If we agree this is a better approach, I'd like to entertain the idea of making a similar change to SEP-24: exposing `id` and `status` from `/transaction(s)` instead of using `on_change_callback`.

ecosystem/sep-0012.md
morleyzhi: The other, existing calls already return 401. Are you sure you want to change the error response if this SEP is already in use?

morleyzhi:
@JakeUrban I'm fine with either option (unauthorized fallback vs. callback). I opted against doing the unauthorized fallback option because KYC status might be privileged information, but we can write a separate PR for that if that's not the case. I'll let you decide if we should pursue that line of thinking.

ecosystem/sep-0012.md
JakeUrban: Ah you're right, darn. I don't want to make any breaking changes so lets just stick with 401.

JakeUrban:
@tomerweller do you think anyone would have objections to SEP-12 servers displaying a customer's `status` to unauthenticated requests? They would still have to know the `account` (and potentially `memo`) or the `id` already.

tomerweller:
@JakeUrban those are valid points. If we go down the route of unauthenticated resources then I think they should have different URLs from the authenticated ones. Getting different resources from the same url based on authentication status can lead to some weird edge cases in clients.

With that said, I'm leaning towards not leaking any type of information if it's not required. 

Given that SEP24 already has a callback feature, anchor backends should already be equipped to deal with callbacks. I say let's push this through.

ecosystem/sep-0012.md
```diff
@@ -315,3 +316,35 @@ Situation | Response
...
 Client not authenticated properly | `401 Unauthorized`
 Anchor has no information on the customer | `404 Not Found`
+
+## Customer callback PUT
+
+Allow the wallet to provide a callback URL to the anchor. The provided callback URL will replace (and supercede) any previously-set callback URL for this account.
+
+Whenever the user's `status` field changes, the anchor will issue a POST request to the callback URL. The payload of the POST request will be the same as [`GET /customer`](#customer-get).
+
+It's the wallet's responsibility to send a working callback URL to the anchor. If callback POST fails due to a problem on the callback URL's side, the anchor should not retry it.
```
leighmcculloch, line 326: >If callback POST fails due to a problem on the callback URL's side, the anchor should not retry it.

❗I think we should allow retrying, but leave the specifics undefined. Retrying callbacks is a common practice, but we can't be too prescriptive without burdening services that don't have the capabilities to support robust callback retry systems.

By disallowing retries here we encourage clients to build systems that can't handle duplicate requests, which could be dangerous. A callback URL should always handle cases like being called more than once, or other edge cases, and there's little reason to disallow retries when a callback endpoint needs to handle those other cases.

```suggestion
It's the wallet's responsibility to send a working callback URL to the anchor. If callback POST fails due to a problem on the callback URL's side, the anchor may retry it, although is not required to.
```

Alternatively, if we don't feel like we want to make a decision on this just yet, delete this statement and leave retries undefined until such time as they become defined.

```suggestion
It's the wallet's responsibility to send a working callback URL to the anchor.
```

ecosystem/sep-0012.md
```diff
@@ -315,3 +316,35 @@ Situation | Response
...
+PUT [KYC_SERVER || TRANSFER_SERVER]/customer/callback
+```
+
+| Name                 | Type     | Description             |
+| -------------------- | -------- | ----------------------- |
+| `id`        | `string`      | (optional) The ID of the customer as returned in the response of a previous `PUT` request. If the customer has not been registered, they do not yet have an `id`.                                                                                                                                     |
+| `account`   | `G...` string | (optional) The Stellar account ID used to identify this customer. If many customers share the same Stellar account, the `memo` and `memo_type` parameters should be included as well.                                                                                                                 |
+`memo` | `string` | (optional) The memo used to create the customer record |
+`memo_type` | `string` | (optional) The type of memo used to create the customer record |
+| `url` | `string` | A valid URL. |
```
leighmcculloch, line 342: 💡 Could we include some more detail here about what this URL is? Example:

```suggestion
| `url` | `string` | A callback URL that the SEP-12 server will POST to when the state of the account changes. |
```

leighmcculloch (commented):
@morleyzhi Thanks for addressing the other comments. Just one more ❗ thing to address, and one suggestion 💡 that I think will make the specification clearer, otherwise this looks great.

morleyzhi:
@leighmcculloch I accepted both your suggestions.

ecosystem/sep-0012.md
leighmcculloch: Returning to a question I asked that wasn't answered:
>Is the reason there would be multiple because an account holder might be using several different wallet applications, and each might register for callbacks?

I think this is a valid use case, and so it would have also been fine to support multiple callbacks but clean up the REST pattern in use to support registering a callback and deleting it. Example:
- `POST /customer/callbacks` returns a callback ID
- `DELETE /customer/callbacks/<id>` deletes the callback (optional given that callbacks stop once the account arrives at a terminal state).

If a single callback is all we need in the foreseeable future then we don't need to add the burden of multiple callbacks to SEP-12 implementations, and we can always add other endpoints when they're needed at a later date.

leighmcculloch (approved):
👍🏻 I think this works and is simple. One note here https://github.com/stellar/stellar-protocol/pull/805#discussion_r561278703 about a conversation I don't think we finished about the multiple callbacks, but it sounds like from our comment https://github.com/stellar/stellar-protocol/pull/805#issuecomment-750626011 that isn't actually a requirement today.

JakeUrban (approved):

Issue #780, opened 2020-12-10
SEP-6: Update the protocol to work well with the upgraded SEP-12


JakeUrban:
SEP-12 was originally built exclusively for SEP-6. Now that SEP-31 uses SEP-12, and SEP-12 was upgraded significantly to support that use case, we should update SEP-6 to use the new SEP-12 functionality in a non-breaking fashion.

Particularly, SEP-12 now supports different customer types. Anchors can specify different KYC requirements for different customer types.

Additionally, SEP-6 and SEP-24 alike have no way of communicating a KYC information error after the KYC information is originally passed and accepted. SEP-12's API's now offer a way to do this. For example, if a user provides a properly-formatted banking address originally, but it turns out the account at that address doesn't exist, SEP-12 can communicate that and expect the client to correct the mistake. SEP-6 should integrate with this functionality.

JakeUrban:
CC @leighmcculloch @Zagan202

JakeUrban:
CC @morleyzhi @nickgilbert

JakeUrban:
this isn't a stale issue, there are several related PR's outstanding related to it

PR #808, opened 2021-01-05
SEP-6: Add sep12_type to GET /deposit and /withdraw responses

JakeUrban:
This is the first of several PRs I'll be making to update SEP-6 to use SEP-12 expanded capabilities (it's new GET endpoint in particular).

Changes:
- Adds a `sep12_type` attribute to the `non_interactive_customer_info_needed` response
    - This allows wallets to fetch more detailed information on the fields requested by the server. For example, Wallets can determine which fields are required vs. optional, which fields have a fixed set of accepted values, etc. 
- The possible values for `sep12_type` are outlined in SEP-6's `/info` endpoint, similar to SEP-31's `/info`.

### DIFF
--- a/ecosystem/sep-0006.md
+++ b/ecosystem/sep-0006.md
@@ -492,16 +492,14 @@ The response should be a JSON object like:
         "bank_account": {
           "fields": {
             "dest": { 
-              "description": "your email address. Your cashout PIN will be sent here. If not provided, your account's default email will be used",
-              "optional": true
+              "description": "your email address. Your cashout PIN will be sent here. If not provided, your account's default email will be used"
             }
           }
         },
         "cash": {
           "fields": {
             "dest": { 
-              "description": "your email address. Your cashout PIN will be sent here. If not provided, your account's default email will be used",
-              "optional": true
+              "description": "your email address. Your cashout PIN will be sent here. If not provided, your account's default email will be used"
             }
           }
         }
--- a/ecosystem/sep-0006.md
+++ b/ecosystem/sep-0006.md
@@ -491,7 +491,7 @@ The response should be a JSON object like:
       "types": {
         "bank_account": {
           "fields": {
-            "email_address": { 
+            "dest": { 
               "description": "your email address. Your cashout PIN will be sent here. If not provided, your account's default email will be used",
               "optional": true
             }
@@ -499,7 +499,7 @@ The response should be a JSON object like:
         },
         "cash": {
           "fields": {
-            "email_address": { 
+            "dest": { 
               "description": "your email address. Your cashout PIN will be sent here. If not provided, your account's default email will be used",
               "optional": true
             }
--- a/ecosystem/sep-0006.md
+++ b/ecosystem/sep-0006.md
@@ -320,7 +320,16 @@ Example:
 ```json
 {
   "type": "non_interactive_customer_info_needed",
-  "fields" : ["family_name", "given_name", "address", "tax_id"],
+  "fields" : [
+    "family_name",
+    "given_name",
+    "address",
+    "mobile_number",
+    "tax_id",
+    "bank_account_number",
+    "bank_number",
+    "bank_branch_number"
+  ],
   "sep12_type": "default_deposit"
 }
 ```
--- a/ecosystem/sep-0006.md
+++ b/ecosystem/sep-0006.md
@@ -311,15 +311,17 @@ Name | Type | Description
 -----|------|------------
 `type` | string | Always set to `non_interactive_customer_info_needed`.
 `fields` | list of strings | A list of field names that need to be transmitted via SEP-12 for the deposit to proceed.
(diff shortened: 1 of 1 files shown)


JakeUrban:
Partially addresses #780

ecosystem/sep-0006.md
```diff
@@ -463,21 +473,33 @@ The response should be a JSON object like:
...
       "types": {
         "bank_account": {
           "fields": {
-              "dest": {"description": "your bank account number" },
-              "dest_extra": { "description": "your routing number" },
-              "bank_branch": { "description": "address of your bank branch" },
-              "phone_number": { "description": "your phone number in case there's an issue" }
+            "dest": { 
+              "description": "your bank account number"
+            }
```
JakeUrban, line 496: I removed the other attributes because SEP-12 accepts these parameters, and we want to encourage anchors to collect as much information via the standardized fields as opposed to the potentially custom fields described here. `dest` simply serves as the identifier for the bank information provided via SEP-12.

ecosystem/sep-0006.md
```diff
@@ -463,21 +473,33 @@ The response should be a JSON object like:
...
           }
         },
         "cash": {
           "fields": {
             "dest": { 
-              "description": "your email address. Your cashout PIN will be sent here. If not provided, your account's default email will be used",
-              "optional": true
+              "description": "your email address. Your cashout PIN will be sent here. If not provided, your account's default email will be used"
```
JakeUrban, line 502: `dest` is not optional

ecosystem/sep-0006.md
```diff
@@ -311,15 +311,26 @@ Name | Type | Description
 -----|------|------------
 `type` | string | Always set to `non_interactive_customer_info_needed`.
 `fields` | list of strings | A list of field names that need to be transmitted via SEP-12 for the deposit to proceed.
+`sep12_type` | string | (optional) The SEP-12 `type` value associated with `fields`. Wallets may use this value in [SEP-12 GET /customer](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0012.md#customer-get) requests for more information on the returned list.
```
leighmcculloch, line 314: 💡 Is `fields` essentially deprecated with the addition of type? If there's redundant data here I think we should say the fields field is deprecated. It should still be provided but new applications should use the type. Is that accurate?

leighmcculloch (approved):
One suggestion (💡). The version and updated fields need updating. You should increment the minor version.

ecosystem/sep-0006.md
JakeUrban: Thats accurate, I'll deprecate the field.

JakeUrban:
Actually, I need to think about this change more. Regardless of if we return `fields` or `sep12_type` in `/deposit` and `/withdraw` responses, there's an implicit assumption within this protocol that the anchor will know which fields need to be collected for a given deposit/withdraw request. 

However, there are no required `/deposit` or `/withdraw` request parameters that give the anchor information it could use to determine which fields are required / what `sep12_type` value to return.

In the current changes made in this PR, we're adding a `sep12_types` object to `/info`, which implies that the fields needed could differ, but we're not providing a mechanism for the anchor to determine what those fields are.

JakeUrban:
I can think of 4 pieces of information that could possibly change what KYC information needs to be collected:
- asset (ex. if the asset is regulated, maybe more information is needed)
- amount (ex. large amounts may require more KYC)
- country (ex. services operating in multiple countries need to collect KYC based on the user's country)
- rails type (ex. domestic bank transfer, international wire transfer, cash, crypto, mobile money, etc. all need different info)

Can you think of others? If clients were to specify all of the above in their `/deposit` or `/withdraw` requests, the anchor should have all the information it needs to know what KYC information should be collected.

If we look at the request parameters for both deposit and withdraw,
- `asset_code` is a required parameter
- `type` (meaning rails type) is an optional parameter
- there are no parameters for country or amount

I think the path of least resistance is to add parameters to `/deposit` and `/withdraw` for the transaction amount and user's country. Anchors who need any of these parameters (`asset_code`, `amount`, `type`, `country`) should be allowed to return 400 if not given. SEP-6's `/info` endpoint could be updated to indicate whether or not these fields are required.

If we do this, there's no need for SEP-6 to outline accepted SEP-12 types.

leighmcculloch:
This makes sense. Would the intention still be for SEP-6 to use SEP-9, it would just instruct the client exactly which type to use in response to the request to deposit/withdraw?

JakeUrban:
My original thought was to leave the SEP-6 `/deposit` and `/withdraw` response as is: returning a `fields` list of SEP-9 params. 

We could add a `sep12_type` response attribute as well, but as you pointed out, it is redundant, and `fields` may need to be deprecated if we go that route. I think it makes sense to do that, but that requires existing implementations to update.

JakeUrban:
I'm going to close this and start a new PR

PR #810, opened 2021-01-05
SEP-6: Added 'sep12_status_url' to 'customer_info_needed' response

JakeUrban:
This is the second of several PR's I'll be making to update SEP-6 to use SEP-12's new capabilities. First issue: #808

Changes:
- Adds a `sep12_status_url` attribute to SEP-6's `customer_info_status` response. 
    - This response and SEP-12's `GET /customer` response overlap significantly

### DIFF
--- a/ecosystem/sep-0006.md
+++ b/ecosystem/sep-0006.md
@@ -327,7 +327,7 @@ Example:
 
 Response code: `403 Forbidden`
 
-An anchor should use this response if customer information was submitted for the `account`, but the information is either still being processed or was not accepted.
+An anchor should use this response if customer information was submitted for the `account`, but the information is either still being processed or was not accepted. SEP-12's [GET /customer](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0012.md#customer-get) can also be used to check on the user's KYC status.
 
 Name | Type | Description
 -----|------|------------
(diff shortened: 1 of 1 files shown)


JakeUrban:
partially addresses #780

ecosystem/sep-0006.md
```diff
@@ -327,14 +327,15 @@ Example:
...
+An anchor should use this response if customer information was submitted for the `account`, but the information is either still being processed or was not accepted. SEP-12's [GET /customer](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0012.md#customer-get) can also be used to check on the user's KYC status.
 
 Name | Type | Description
 -----|------|------------
 `type` | string | Always set to `customer_info_status`.
 `status` | string | Status of customer information processing. One of: `pending`, `denied`.
 `more_info_url` | string | (optional) A URL the user can visit if they want more information about their account / status.
 `eta` | int | (optional) Estimated number of seconds until the customer information status will update.
+`sep12_status_url` | string | (optional) A [SEP-12 GET /customer](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0012.md#customer-get) URL that returns the status of the user. This response may have a human-readable message that the Wallet can display instead of opening `more_info_url`.
```
leighmcculloch, line 338: Is including this here a convenience or would there be no other way for the client to find out what the status URL is of the SEP-12 integration? I'm wondering what we get out of binding the two SEPs together and interweaving their functionality into each others specifications, vs a SEP-12 client just knowing it can find the status out via the SEP-12 API.

JakeUrban: It is included here for convenience. The two protocols are already implicitly bound though. Both `non_interactive_customer_info_needed` and `customer_info_status` responses imply the need for SEP-12.

leighmcculloch: I'm struggling to see the inconvenience this solves for. If we add a SEP-12 URL here, we'd probably also want SEP-6 to provide all the other SEP-12 URLs that the client will need to hit for symmetry and because otherwise the client will need to know the other URLs anyway, eroding any gain in convenience this adds.

leighmcculloch (commented):
I'm failing to see the convenience this adds or the problem this solves. I suspect that's because I'm missing some context. I've highlighted below why I think this doesn't significantly change the convenience factor. Could you elaborate if I'm missing anything?

JakeUrban:
I agree, your point about needing to know how to form the other URLs makes providing the GET URL not very valuable. Closing

Issue #809, opened 2021-01-05
SEP-6: Mirror account creation sponsorship policy from SEP-24


JakeUrban:
SEP-24 was recently updated via #798, and SEP-6 should use the same language

PR #811, opened 2021-01-05
SEP-6: mirror update statuses and endpoints from SEP-31

JakeUrban:
This is the third and final change related to #780. This change takes functionality originally introduced to SEP-31 and applies it within the SEP-6 context. See SEP-31's [update sections](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0031.md#pending-customer-info-update) for reference.

Changes:
- Adds two new statuses: `pending_customer_info_update` and `pending_transaction_info_update`
- Adds a new endpoint: `PATCH /transactions` that accepts updated transaction information 

This allows invalid information that was originally accepted via SEP-12 or via the `/deposit`or `/withdraw` endpoints to be reported as invalid and to accept updated information.

First & second issues for updating SEP-6 for SEP-12: #809 and #810


leighmcculloch (approved):
This looks good, assuming we still need it. Do we still need it?

JakeUrban:
@leighmcculloch yes we still need it!

ecosystem/sep-0006.md
```diff
@@ -768,6 +772,22 @@ Example response:
       "completed_at": "2017-03-20T17:09:58Z",
       "stellar_transaction_id": "17a670bc424ff5ce3b386dbfaae9990b66a2a37b4fbe51547e8794962a3f9e6a",
       "external_transaction_id": "2dd16cb409513026fbe7defc0c6f826c2d2c65c3da993f747d09bf7dafd31093"
+    },
+    {
+      "id": "52fys79f63dh3v1",
+      "kind": "withdrawal",
+      "status": "pending_transaction_info_update",
```
leighmcculloch, line 779: One quick question, why not change back to `incomplete`?

JakeUrban: > why not change back to `incomplete`?

That could work, but using a new status allows us to indicate that the transaction previously transitioned away from `incomplete`.

> What's the value in a separate status for these cases?

The client could deduce what is necessary by checking `required_info_updates` and if it were null/empty checking SEP-12's `GET /customer`, but this is more explicit.

If you see a strong reason to change the way this is designed we can do it, but since this uses the same pattern as SEP-31 I think there is value in being consistent across protocols.

leighmcculloch: If it's the same pattern as SEP-31 I think that's a strong reason to do this. 

I don't think there is an always best approach to less or more states. Reusing states when they essentially mean the same thing can make some APIs simpler while making others more complex. APIs that never revisit states can't be easier to debug and code for, even if it means understanding then required more effort.

I'm still for it 👍

leighmcculloch (approved):

PR #814, opened 2021-01-06
SEP-12: Allow GET /customer in PROCESSING to have 'message'

JakeUrban:
The `GET /customer` endpoint has an optional `message` response attribute that is currently implied to only be used for rejection responses. However, SEP-6/SEP-31 Wallets may want to display a human readable message to their users while KYC information is being processed instead of opening a `more_info_url` from the anchor.

### DIFF
--- a/ecosystem/sep-0012.md
+++ b/ecosystem/sep-0012.md
@@ -160,6 +160,14 @@ Name | Type | Description
 ```
 
 ```js
+{
+    "id": "46116754-695e-43f6-84c4-8c05e50a7b12",
+    "status": "PROCESSING",
+    "message": "Photo ID requires manual review. This process typically takes 1-2 business days."
+}
+```
+
+```js
 // The case when a customer has been rejected and cannot be KYC'd
 {
    "id": "d1ce2f48-3ff1-495d-9240-7a50d806cfed",
--- a/ecosystem/sep-0012.md
+++ b/ecosystem/sep-0012.md
@@ -107,8 +107,8 @@ Name | Type | Description
 -----|------|------------
 `id` | string | (optional) ID of the customer, if the customer has already been created via a `PUT /customer` request.
 `status` | string | Status of the customers KYC process.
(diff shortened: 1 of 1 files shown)


tomerweller:
@JakeUrban can you share a specific example?

JakeUrban:
Lets say a SEP-12 server accepts `photo_id_front` and `photo_id_back` which needs to be reviewed by a human.

Currently the response from `GET /customer` would look like this:
```
{
    "status": "PROCESSING",
    "id": "46116754-695e-43f6-84c4-8c05e50a7b12"
}
```

If the Wallet is making a SEP-6 deposit on behalf of it's user, then it probably got the following response back from `/deposit`:
```
{
    "type": "customer_info_status"
    "status": "pending",
    "more_info_url": "https://anchor.com/customer/status.html"
}
```

The wallet probably doesn't want to open a popup in their app for the user to know whats going on. Instead, the SEP-12 server could update it's `GET /customer response to include a `message`:
```
{
    "id": "46116754-695e-43f6-84c4-8c05e50a7b12",
    "status": "PROCESSING",
    "message": "Photo ID requires manual review. This process typically takes 1-2 business days."
}
```

Now the Wallet can just display `message` to the user instead of opening a third-party HTML page in their app.

JakeUrban:
@tomerweller

tomerweller (changes requested):
So the idea that the wallet will pull the sep12 customer info whenever it sees `"status": "pending"` on sep6? 

I think this is a simple step in the right direction.  

Can you add this `PROCESSING` example in the examples section?

ecosystem/sep-0012.md
```diff
@@ -159,6 +159,14 @@ Name | Type | Description
 }
 ```
 
+```js
```
tomerweller, line 162:
```suggestion
```js
// The case when the Anchor is processing KYC information  
```

tomerweller (approved):
One comment about adding a comment to the example and otherwise LGTM

PR #816, opened 2021-01-08
SEP-8: add application/json as request content type

howardtw:
**What**

Support JSON encoded request.

**Why**

We accept the `application/json` content type in some other SEPs, e.g. SEP-10 and SEP-30. The viewpoint from some people outside SDF that led to this pattern on SEP-10 was that it makes it easier to integrate with the APIs in some cases.

**Note**

This PR would have a version conflict with https://github.com/stellar/stellar-protocol/pull/817. One that gets merged first will have version 1.1.1; one that gets merged later will have version 1.2.1.


JakeUrban (approved):
Looks good to me!

PR #817, opened 2021-01-08
SEP-8: replace account setup with transaction composition

howardtw:
**What**

This PR replaces Account Setup with Transaction Composition.

**Why**

SEP-8 was introduced before CAP-18 was introduced. The Account Setup section in the current SEP-8 has a few drawbacks as mentioned in CAP-18.
```
1. The owner of an account cannot unilaterally utilize assets unrelated to the issuer without the signature of that issuer
2. If multiple issuers want to become signers on a single account, then many signatures are potentially required for simple operations
3. Issuers cannot easily rotate keys, because their keys are on every account holding their asset
```

We are seeing a few active developments on regulated assets in the ecosystem, and we want to make sure the associated SEP is up-to-date and is easy to use without introducing breaking changes.

**Note**

This PR would have a version conflict with https://github.com/stellar/stellar-protocol/pull/816. One that gets merged first will have version `1.1.1`; one that gets merged later will have version `1.2.1`.

### DIFF
--- a/ecosystem/sep-0008.md
+++ b/ecosystem/sep-0008.md
@@ -260,7 +260,7 @@ Name | Type | Description
 
 ### Best practices
 
-- Issuers using the `revised` status to make a transaction complaint is encouraged as it can take the burdens away from the wallets.
+- Issuers are encouraged to use the `revised` status to make a transaction complaint when possible, since Wallets may not know how to form a transaction that is compliant with the server's criteria.
 - If a transaction was revised, ALWAYS explain the changes that were made to a transaction through the `message` parameter.
 - Wallet should inspect revised transactions and alert the user if any of the original operations are changed.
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0008.md
```diff
@@ -227,16 +251,13 @@ Name | Type | Description
 
 ### Best practices
 
+- Issuers using the `revised` status to make a transaction complaint is encouraged as it can take the burdens away from the wallets.
```
JakeUrban, line 254:
```suggestion
- Issuers are encouraged to use the `revised` status to make a transaction complaint when possible, since Wallets may not know how to form a transaction that is compliant with the server's criteria.
```

JakeUrban (approved):
One phrasing suggestion, otherwise LGTM

ecosystem/sep-0008.md
howardtw: Thanks! Much better. https://github.com/stellar/stellar-protocol/pull/817/commits/0b464cafd42ad658e9b1281698abbedd61b4f380

PR #820, opened 2021-01-13
SEP-12: Added Binary Data Types section under Content Type section

JakeUrban:
originally discussed in [#783](https://github.com/stellar/stellar-protocol/pull/783#discussion_r556693629), `application/json` can be used to send binary data if that data is base64-encoded. 

Once the referenced PR is merged, a `Nested Data Types` subsection should be added to point out that `application/json` data should be used when SEP-9 attributes cannot be encoded in a standardized manner (i.e. sub-objects, arrays).

Once this PR is merged, I will create issues for SEP-6, 8, 24, and 31 containing the same information.

### DIFF
--- a/ecosystem/sep-0012.md
+++ b/ecosystem/sep-0012.md
@@ -58,11 +58,17 @@ All endpoints accept in requests the following `Content-Type`s:
 - `application/x-www-form-urlencoded`
 - `application/json`
 
-`multipart/form-data` should be used for requests including binary data type values from [SEP-9](sep-0009.md). Any of the above encoding schemes may be used when requests do not include binary data.
-
 All endpoints respond with content type:
 
 - `application/json`
+
+### Binary Data Types
+
+`multipart/form-data` or `application/json` should be used for requests including binary data type values from [SEP-9](sep-0009.md). 
+
+If `application/json` is used, values for the binary data type attributes must be objects containing `data` and `content_type` keys, where the `data` is base64-encoded. See the [PUT /customer](#customer-put) section for request examples. 
+
+Any of the above encoding schemes may be used when requests do not include binary data.
 
 ## Customer GET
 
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0012.md
```diff
@@ -255,7 +271,16 @@ Content-Type: application/json
 {
    "account":"GBORFR3GDNVZ5PLUTBDQHKGWVD26CQUHORO2T3SDQ2JPLGLUJCCA5GK6",
    "memo":"21bf91a4-7db1-401d-8108-fab7660a45d6",
-   "memo_type":"text"
+   "memo_type":"text",
+   "photo_id_front": {
+      "data": "VGhpcyB0ZXh0IHJlcHJlc2VudHMgYmluYXJ5IGltYWdlIGRhdGE=",
+      "content_type": "image/png"
+   },
```
leighmcculloch, line 278: This should be okay for smaller files, but this could be problematic for larger files because most JSON parsing libraries require the entire JSON object to be loaded into memory. It will be difficult for implementers to stream a file to some other destination, such as AWS S3.

What size files do you think we'll be handling?

This is more complicated, but it might be necessary to explore file upload as a separate API that returns a file identifier that is then provided here. This is a common pattern. e.g. https://stripe.com/docs/api/files https://developers.braintreepayments.com/reference/request/document-upload/create

leighmcculloch (commented):
I'd like to think about this a little more before we commit to this approach given the reasons I share below in comment.

ecosystem/sep-0012.md
tomerweller: I agree with @leighmcculloch  that sending binary blobs in json is less than ideal. 

However, I'm concerned about the complexity of adding a file upload API to the spec.

What if we have the senders simply send a URL to the resource? this decouples the sending and receiving and allows both parties to utilize infrastructure that they (probably) already have. If we go down this route it's important to empathsize that these urls should be considered ephemeral so the receiver should fetch and save for their records.

leighmcculloch: Using URLs would be harder to secure. These documents may be highly sensitive and submitting a URL that anyone could unauthenticated access isn't ideal. I think we need to define the method for the document to go from client to server without anyone else having access.

JakeUrban:
I'm closing this PR since we've decided to abandon the strategy of encoding binary data as base64. #904 has been opened in the draft state, adding a `customer/files` endpoint.

Issue #812, opened 2021-01-06
SEP-0012: property verification


tomerweller:
Some KYC properties such as phone number, email address and even home address can be verified using a verification code.

Verifying phone numbers using text messages is especially popular. 

SEP12 currently doesn't provide a mechanism to convey if a property is verified/unverified. 

More so, there's no clear way for a wallet to initiate verification of a property.

This issue is created to facilitate discussion of potential solutions.

morleyzhi:
This is how we've modeled phone validation in Vibrant:

- PUTting mobile_number alone will send a code + return an error (indicating that a code was sent)
- The anchor will only store a mobile_number when it's PUT with a valid mobile_number_verify
- Vibrant would PUT:/customer a mobile_number
- The anchor would send a code to the phone
- Vibrant would show a code input
- The user would enter the code
- Vibrant would PUT:/customer mobile_number_verify and mobile_number
- mobile_number_verify should otherwise not be returned by GET:/customer as a missing field or whatever

JakeUrban:
My issue with the current approach is the lack of separation between KYC information and data used to verify that information.

What I suggest is a new endpoint, `PUT /customer/verification` that accepts forms of verification the following fields:
```
{
    # ID of an existing customer, probably in PROCESSING status
    "id": "e793dc26-c8fc-4b95-bf3d-231f8126e8e1",
    # data used for verification
    "fields": {
        "mobile_number_verification": "185014",
        "email_address_verification": "205721",
        "address_verification": "457023"
    }
}
```
Email address verification is probably easier to do the typical way -- the service sends a confirmation link that the user has to click, but you get the point.

JakeUrban:
Actually, to match the format used in `PUT /customer` I would remove the `fields` attribute:
```
{
    # ID of an existing customer, probably in PROCESSING status
    "id": "e793dc26-c8fc-4b95-bf3d-231f8126e8e1",
    # data used for verification
    "mobile_number_verification": "185014",
    "email_address_verification": "205721",
    "address_verification": "457023"
}
```

tomerweller:
I'm with @JakeUrban on separating this out. Overloading the customer PUT can be somewhat confusing. 

In both of these suggestions there's a pre-requisite that the anchor knows that a field requires verification - should the server communicate this to the wallet? If so, how?

JakeUrban:
Good point @tomerweller, I have a solution

[...]

PR #821, opened 2021-01-13
SEP-12: Added new verification endpoint and expanded fields object usage

JakeUrban:
resolves stellar/stellar-protocol#812

Changes:
- Added a `PUT /customer/verification` endpoint that accepts a customer `id` along with SEP-9 attribute names concatenated with `_verification`. All verification values must be strings.
- Expanded `fields` object usage in `GET /customer`:
    - `fields` can now be used in any `GET /customer` request (that is, for any customer `status`)
    - Per-field `status` and `error` attributes were added. `status` values can be one of:
        - `NOT_PROVIDED`
        - `PROCESSING`
        - `VERIFICATION_REQUIRED`
        - `ACCEPTED`
        - `REJECTED`

There are some potential issues the following design raises:
- What happens if the confirmation code sent to the customer expires? This leads to the following issue:
- How would our API support requesting a new verification code?


tomerweller:
Not stale

ecosystem/sep-0012.md
```diff
@@ -107,14 +108,31 @@ Name | Type | Description
 -----|------|------------
 `id` | string | (optional) ID of the customer, if the customer has already been created via a `PUT /customer` request.
 `status` | string | Status of the customers KYC process.
-`fields` | object | (optional) An object containing the fields still needed to finish KYC Process for the given KYC type.
+`fields` | object | (optional) An object containing the fields for the given KYC type. See [fields](#fields) for more detailed information.
```
leighmcculloch, line 111: ❓ The fields used to contain the fields still needed, but now it sounds like it will contain all the fields. It sounds like the user could provide the fields in multiple calls before. Now must it be in one call? Or was it always intended to be a single call?

ecosystem/sep-0012.md
```diff
@@ -188,14 +217,16 @@ REJECTED | This customer's KYC has failed and will never succeed.  The `message`
...
-The fields entry is required for the `NEEDS_INFO` status.  Fields should be specified as an object with keys representing field names required, preferably from SEP-9.
+The fields entry is required for the `NEEDS_INFO` status but may be included with any status. Fields should be specified as an object with keys representing the [SEP-9](sep-0009.md) field names required.
 
 Property | Type | Description
 -----|------|------------
 `type` | enum | The data type of the field value. Can be `string`, `binary`, `number`, or `date`
 `description` | string | A human-readable description of this field, especially important if this is not a SEP-9 field.
 `choices` | array | (optional) An array of valid values for this field.
 `optional` | boolean | (optional) A boolean whether this field is required to proceed or not. Defaults to false.
+`status` | string | (optional) One of `NOT_PROVIDED`, `PROCESSING`, `UNVERIFIED`, `ACCEPTED`, `REJECTED`. See [PUT /customer/verification](#customer-put-verification) for more information on `UNVERIFIED` fields.
```
leighmcculloch, line 228: ❗   What is the `UNVERIFIED` state? It doesn't look like all the statuses are defined:

https://github.com/stellar/stellar-protocol/blob/6320ef8124c85805206045ec31448358aa4d656b/ecosystem/sep-0012.md#L209-L216

💡 It's quite difficult to ascertain the meaning of UNVERIFIED in the context of the other statuses. It sounds like it could be the same thing as PROCESSING. I think we should find another name for it, maybe `VERIFICATION_REQUIRED` which signals a little more explicitly that there is some action required by the user. We could also name it `ACTION_REQUIRED` to generalize it a little more and be more explicit, but `VERIFICATION_REQUIRED` I think is better because it uses the same terms as the new verification endpoint.

ecosystem/sep-0012.md
```diff
@@ -302,6 +340,41 @@ For example:
 }
 ```
 
+## Customer PUT Verification
+
+This endpoint allows servers to accept data values, usually confirmation codes, that verify a previously provided field via `PUT /customer`, such as `mobile_number` or `email_address`. Note that while fields such as `photo_proof_residence` or `notary_approval_of_photo_id` are verifications of other fields described in [SEP-9](#sep-0009.md), the server does not require the associated fields _before_ verification can be accomplished.
```
leighmcculloch, line 345: ~❓ Where are `photo_proof_residence` and `notary_approval_of_photo_id`, and `mobile_number_verification` used in the example, defined?~

💡 I think it is worth moving the paragraph about how the `_verification` fields are defined, that currently lives in the request section, ahead of any examples of fields here. When reading top-to-bottom seeing examples ahead of a definition creates confusion.

ecosystem/sep-0012.md
```diff
@@ -302,6 +340,41 @@ For example:
...
+
+### Request
+
+```
+PUT [KYC_SERVER || TRANSFER_SERVER]/customer/verification
+```
+
+Name | Type | Description
+-----|------|------------
+`id` | string | The ID of the customer as returned in the response of a previous `PUT` request.
```
leighmcculloch, line 355: ❗ Can you add a row here for the additional fields so all fields that can be sent are defined?
```suggestion
`id` | string | The ID of the customer as returned in the response of a previous `PUT` request.
`*_verification` | string | One or more fields that ...
```

ecosystem/sep-0012.md
```diff
@@ -302,6 +340,41 @@ For example:
...
+```json
+{
+   "id": "391fb415-c223-4608-b2f5-dd1e91e3a986",
+   "mobile_number_verification": "2735021"
+}
+```
+
+### Response
+
+Success responses should return a `202 Accepted` status, and the `GET /customer` `status` of the field(s) being verified should be changed to `PROCESSING` or `ACCEPTED`.
```
leighmcculloch, line 368: ❓ What should the response body contain in the success case?

ecosystem/sep-0012.md
leighmcculloch, line 345: ❓ The sentence about photo proof of resident and notary verifications isn't clear to me. What you want to communicate? How do these fields work?

leighmcculloch (commented):
👏🏻 I like the simplicity of verifications. I have some questions that probably will affect the change(❓), some things I think need addressing (❗), and some suggestions (💡).

ecosystem/sep-0012.md
JakeUrban: It was never clearly defined whether multiple calls were acceptable or single calls required. Polaris leaves this decision up to the anchor.

I don't see a reason to pick a policy here either. In fact, doing so would be a breaking change. Anchors can return `400` if they'd like all fields at once, and its likely less complex to implement that way.

However, I do see the possibility of the client wanting to verify one field at a time. From a UX perspective, the client may want to know if the value provided by the user is valid before moving on to collect other pieces of information.

If the client wants this UX, it can reach out to the anchor directly to make their implementation more flexible. The `NOT_PROVIDED` field status will be useful in this case too.

ecosystem/sep-0012.md
JakeUrban: I added a `Field Statuses` section, and renamed `UNVERIFIED` to `VERIFICATION_REQUIRED`.

ecosystem/sep-0012.md
JakeUrban: This is actually undefined for DELETE, and `PUT` used to be undefined until we added the `id` attribute. We could return the field `status`?

ecosystem/sep-0012.md
JakeUrban: `photo_proof_residence` is a verification of the address fields, like `address`, `state_or_province`, etc.

`notary_approval_of_photo_id` is a verification of `photo_id_front` and `photo_id_back`.

However, you wouldn't listed `photo_id_front` as `VERIFICATION_REQUIRED`. Instead you would just also list `notary_approval_of_photo_id` in your field set.

In general, a field only needs `VERIFICATION_REQUIRED` if that field cannot be verified until it is first provided to the server, because if that weren't the case, the server would just request another field that verifies the field needing verification.

JakeUrban: Another way to think about it is whether or not the server wants to verify a field themselves or _trust a third party_.

Lets say `photo_proof_residence` is required and the client submits a utility bill. The server is trusting the utility company to have verified that the customer lives at the address on the bill. 

`mobile_number` is different because the server doesn't have to trust a third party. Instead it can just verify that the customer has that number themselves using verification codes.

ecosystem/sep-0012.md
JakeUrban: As specified in the request description:
> Data values used for verification should be included in the request using the SEP-9 attribute name plus `_verification`.

If you want me to add a single entry for `*_verification`, I can do that. But I don't think we should define a list of fields where verification is valid. We can leave that to the anchor to decide.

JakeUrban:
Thanks for your review @leighmcculloch. I've updated the PR and responded to your comments where discussion is still needed. Don't respond until you're back!

ecosystem/sep-0012.md
leighmcculloch: I think we should add something, so adding something like `*_verification` at least makes it clear that's where the fields go.

ecosystem/sep-0012.md
leighmcculloch: I think that sounds good.

JakeUrban: Actually, the request format allows for multiple field verifications to be sent, so a single status wouldn't work. I think returning the same response sent in `GET /customer` would make sense though.

leighmcculloch: Sounds good to me.

ecosystem/sep-0012.md
JakeUrban: I found it difficult to move the verification text above the examples from a flow perspective, so I moved the relevant example down into the request section.

ecosystem/sep-0012.md
```diff
@@ -182,25 +196,38 @@ Name | Type | Description
...
-The fields entry is required for the `NEEDS_INFO` status.  Fields should be specified as an object with keys representing field names required, preferably from SEP-9.
+The fields entry is required for the `NEEDS_INFO` status but may be included with any status. Fields should be specified as an object with keys representing the [SEP-9](sep-0009.md) field names required.
 
 Property | Type | Description
 -----|------|------------
 `type` | enum | The data type of the field value. Can be `string`, `binary`, `number`, or `date`
 `description` | string | A human-readable description of this field, especially important if this is not a SEP-9 field.
 `choices` | array | (optional) An array of valid values for this field.
 `optional` | boolean | (optional) A boolean whether this field is required to proceed or not. Defaults to false.
+`status` | string | (optional) One of the values described in [Field Statuses](#field-statuses).
```
leighmcculloch, line 218: ❓ I like that this field is optional because in some jurisdictions the server may wish to state that the customer is accepted or rejected without indicating which fields caused the decision. Is it acceptable for a server to implement SEP-12 and not output the status? Or is the optionality only to accommodate implementations that existed prior to this change?

ecosystem/sep-0012.md
```diff
@@ -182,25 +196,38 @@ Name | Type | Description
...
+
+#### Field Statuses
+
+Status | Description
+-------|------------
+ACCEPTED | The field has been validated. When all required fields are accepted, the [Customer Status](#customer-statuses) should also be accepted.
+PROCESSING | The field is being validated. The client can make [GET /customer](#customer-get) requests to check on the result of this validation in the future.
+NOT_PROVIDED | The field has not been sent by the client. The field is required unless the `optional` property is `true`. The [Customer Status](#customer-statuses) should be `NEEDS_INFO` if a field is required but has not been provided.
+REJECTED | The field was in the `PROCESSING` status but did not pass validation. If the client may resubmit this field, the [Customer Status](#customer-statuses) should be `NEEDS_INFO`, otherwise it should be `REJECTED`.
+VERIFICATION_REQUIRED | The field must be verified using the [PUT /customer/verification](#customer-put-verification) endpoint. For example, the `mobile_number` field could be placed in this status until a confirmation code is sent to the customer and passed back to the this endpoint.
```
leighmcculloch, line 229: 💡 Not needed for this PR, but these statuses would benefit from a diagram showing the entry and exit statuses, and from which states they can transition.

ecosystem/sep-0012.md
```diff
@@ -309,6 +343,55 @@ For example:
...
+```json
+{
+   "id": "391fb415-c223-4608-b2f5-dd1e91e3a986",
+   "mobile_number_verification": "2735021"
+}
+```
+
+### Response
+
+Success responses should return a `202 Accepted` status as well as a body matching the [GET /customer](#customer-get) response schema. The field statuses for which verifications were sent must be updated to either `PROCESSING` or `ACCEPTED`.
```
leighmcculloch, line 385: 💡 If the status is Accepted the I think 200 OK would be more appropriate because there is nothing else to do, processing has already completed. 202 is intended for when processing still needs to take place, which worse for the case where it is still processing, although you could even use 200 there too without much contention I think.

leighmcculloch (approved):
I think this looks great. I have one question (❓), also a couple suggestions (💡), regardless I think this change is good.

ecosystem/sep-0012.md
JakeUrban: I made it optional to accommodate existing implementations, but you make a good point, the field could be omitted for that reason as well.

leighmcculloch: I think it could be worth making that more explicit, by adding a statement that servers may choose not to expose the status per-field status if they don't wish expose which fields were accepted or rejected. Does the protocol still work in this case do you think?

ecosystem/sep-0012.md
JakeUrban: Ah I didn't know that. I used 202 because thats what is return in `PUT /customer` responses. If `Accepted` should only be used when process is still necessary, do you think its worth adding 200 as an acceptable status for that endpoint as well, since 202 would only make sense if the customer was put into the `PROCESSING` status?

leighmcculloch: I think 200 OK would suffice for both endpoints, because a full response is available in all cases. Clients should focus on using the customer status as the indicator not the http status code, because we can add new statuses. I get to make it always 200 would be a breaking change though, so another option is to say any 2xx response is fine.

ecosystem/sep-0012.md
JakeUrban: I think it still works. Anchors who don't want to expose `accepted`/`rejected` information can omit the `status` attribute unless verification is needed. I'll add a statement.

leighmcculloch (approved):

leighmcculloch (approved):

PR #827, opened 2021-01-14
SEP-0008: support CORS header

### LINKED ISSUE #823: SEP-0008: Add info about requiring support for cross-domain requests
CORS requests should be supported by the endpoint so that web apps on other domains can hit them cross-domain.
See language in [SEP-10](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0010.md#cross-origin-headers) that may be used.

howardtw:
**What**

Add support for CORS header.

**Why**

CORS requests should be supported by the endpoint so that web apps on other domains can hit them cross-domain.

Closes #823 

cc @shredding @dmgmyza

### DIFF
--- a/ecosystem/sep-0008.md
+++ b/ecosystem/sep-0008.md
@@ -6,8 +6,8 @@ Title: Regulated Assets
 Author: Interstellar.com, Howard Chen <@howardtw>
 Status: Active
 Created: 2018-08-22
-Updated: 2021-01-11
-Version 1.3.1
+Updated: 2021-01-15
+Version 1.4.1
 ```
 
 ## Simple Summary
--- a/ecosystem/sep-0008.md
+++ b/ecosystem/sep-0008.md
@@ -101,6 +101,14 @@ Approval server is a single endpoint that receives a signed transaction, checks 
 
 The specification of an approval server is defined as follows:
 
+### Cross-Origin Headers
+
+Valid CORS headers are necessary to allow web clients from other sites to use the endpoints. The following HTTP header must be set for the endpoint on the approval server, including error responses.
+
+```
+Access-Control-Allow-Origin: *
+```
+
 ### Content Type
 
 Approval server accepts requests in the following `Content-Type`s:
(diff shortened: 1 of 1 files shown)


JakeUrban (approved):

ecosystem/sep-0008.md
```diff
@@ -101,6 +101,14 @@ Approval server is a single endpoint that receives a signed transaction, checks
 
 The specification of an approval server is defined as follows:
 
+### Cross-Origin Headers
+
+Valid CORS headers are necessary to allow web clients from other sites to use the endpoints. The following HTTP header must be set for the endpoint on the approval server, including error responses.
```
leighmcculloch, line 106: Let's use this opportunity to make this definition more accurate, and we can update SEP-10 at some point in the future so it is also more accurate, where I can see this was copied from.

The reason I'm suggesting this change is because the header displayed below is the minimum change needed with `application/x-www-form-urlencoded` requests, but it won't work for JSON requests unless the OPTIONS request is also supported. It's also more common to not use the wildcard and instead echo the origin in the request as that works in a greater number of cases where the wildcard is disallowed by most browsers (when cookies are present is one example). I don't think we need to teach people how to use CORS, or prescribe the approach they take, we just need to indicate cross origin requests must be supported.

```suggestion
Cross-origin requests to the endpoints in this specification must be supported by the server so that web clients from other sites can use use the endpoints. Valid CORS headers and requests that allow any domain origin to access the endpoints in this specification are necessary. The following HTTP header and support of preflight OPTIONS requests are the smallest change necessary. More complex headers configurations can be used.
```

leighmcculloch (commented):
I think we can make the discussion about what needs implementing more focused on the desired outcome, and broader. As it is written today if a developer just did what it said, it would fail in some cases. See below.

I see this is a copy of SEP-10's statements on this matter, we can update SEP-10 as well later.

ecosystem/sep-0008.md
howardtw: https://github.com/stellar/stellar-protocol/pull/827/commits/41475ca61e39d003e526aae422026782df51fdca

leighmcculloch (approved):

leighmcculloch:
FYI @shredding

PR #830, opened 2021-01-15
CAP-0035: mention sponsored reserve behavior

leighmcculloch:
### What
Mention what happens to sponsored reserves when a claimable balance is clawed back.

### Why
@bartekn asked what happens on the mailing list, and whilst there is nothing unique being introduced regarding this feature we can add a statement making it clear that is the case.


sisuresh (approved):
Looks good to me

tomerweller (commented):
lgtm

PR #833, opened 2021-01-16
Close stale issues and pull requests automatically after 35 days

leighmcculloch:
### What
Close stale issues and pull requests automatically after 35 days with a 5 day warning.

### Why
@MonsieurNicolas pointed out that we've got a build up of stale issues and pull requests. This should help clean things up. It is non-destructive so issues and pull requests can be reopened by their authors if they get closed. We can improve a selection of labels that prevent auto-close.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -2,6 +2,7 @@ name: 'Stale Issues'
 on:
   schedule:
     - cron: '30 1 * * *'
+  pull_request:
 
 jobs:
   stale:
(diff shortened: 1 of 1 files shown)


leighmcculloch:
@MonsieurNicolas This is ready to merge. Let's merge it with debug-only enabled. It'll run on this repo and we can see the log output. If that looks good in a separate PR we can disable debug-only.

MonsieurNicolas:
LGTM

PR #837, opened 2021-01-19
SEP-6 & SEP-24: make asset codes used consistent

JakeUrban:
88312b04e1334147b8dab8763dce0ed58dcd3ce3 clarified the `asset_code` parameter to `/deposit` to be the code of the asset being deposited to the anchor's off-chain account. For example, if an anchor issues a `MyBTC` token, the client should still pass `BTC` as the `asset_code` parameter value.

This PR changes that to require clients to use the code of the anchored asset (`MyBTC` in the example). Here is why:
1. SEP-24 specifically states that the `asset_code` parameter should be the asset code of the anchored asset, not the off-chain asset. So this PR makes SEP-6 consistent with SEP-24.
2. Using potentially different asset codes for `/deposit` makes what the `/info` response should be unclear. In the `MyBTC` example, should `MyBTC` or `BTC` be listed in the `/info` response? This change makes this question mute.

### DIFF
--- a/ecosystem/sep-0006.md
+++ b/ecosystem/sep-0006.md
@@ -6,7 +6,7 @@ Title: Anchor/Client interoperability
 Author: SDF
 Status: Active (Interactive components are deprecated in favor of SEP-24)
 Created: 2017-10-30
-Updated: 2020-04-01
+Updated: 2021-01-19
 Version 3.1.2
 ```
 
--- a/ecosystem/sep-0006.md
+++ b/ecosystem/sep-0006.md
@@ -7,7 +7,7 @@ Author: SDF
 Status: Active (Interactive components are deprecated in favor of SEP-24)
 Created: 2017-10-30
 Updated: 2020-04-01
-Version 3.1.1
+Version 3.1.2
 ```
 
 ## Simple Summary
--- a/ecosystem/sep-0006.md
+++ b/ecosystem/sep-0006.md
@@ -131,7 +131,7 @@ Request Parameters:
 
 Name | Type | Description
 -----|------|------------
-`asset_code` | string | The code of the asset the user wants to deposit with the anchor. Ex BTC, ETH, USD, INR, etc. This may be different from the asset code that the anchor issues. Ex if a user deposits BTC and receives MyBTC tokens, `asset_code` must be BTC.
+`asset_code` | string | The code of the asset the user wants to deposit with the anchor. The value passed must match one of the codes listed in the [/info](#info) response's deposit object.
 `account` | `G...` string | The stellar account ID of the user that wants to deposit. This is where the asset token will be sent.
 `memo_type` | string | (optional) Type of memo that anchor should attach to the Stellar payment transaction, one of `text`, `id` or `hash`.
 `memo` | string | (optional) Value of memo to attach to transaction, for `hash` this should be base64-encoded.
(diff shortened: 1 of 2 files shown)


ecosystem/sep-0006.md
```diff
@@ -131,7 +131,7 @@ Request Parameters:
 
 Name | Type | Description
 -----|------|------------
-`asset_code` | string | The code of the asset the user wants to deposit with the anchor. Ex BTC, ETH, USD, INR, etc. This may be different from the asset code that the anchor issues. Ex if a user deposits BTC and receives MyBTC tokens, `asset_code` must be BTC.
+`asset_code` | string | The code of the asset the user wants to deposit with the anchor. The value passed must match one of the codes listed in the [/info](#info) response's deposit object.
```
leighmcculloch, line 134: 💡 I think it is helpful to continue to include the last example, it at least makes it clearer to me:
```suggestion
`asset_code` | string | The code of the asset the user wants to deposit with the anchor. For example, if a user deposits BTC and receives MyBTC tokens, `asset_code` will be BTC. The value passed must match one of the codes listed in the [/info](#info) response's deposit object.
```

ecosystem/sep-0006.md
```diff
@@ -245,7 +245,7 @@ Request parameters:
 
 Name | Type | Description
 -----|------|------------
-`asset_code` | string | Code of the asset the user wants to withdraw. This must match the asset code issued by the anchor. Ex if a user withdraws MyBTC tokens and receives BTC, the `asset_code` must be MyBTC.
+`asset_code` | string | Code of the asset the user wants to withdraw. The value passed must match one of the codes listed in the [/info](#info) response's withdraw object.
```
leighmcculloch, line 248: 💡 
```suggestion
`asset_code` | string | Code of the asset the user wants to withdraw. For example, if a user deposits BTC and receives MyBTC tokens, `asset_code` will be BTC. The value passed must match one of the codes listed in the [/info](#info) response's withdraw object.
```

ecosystem/sep-0024.md
```diff
@@ -154,7 +154,7 @@ Request Parameters:
 
 Name | Type | Description
 -----|------|------------
-`asset_code` | string | The code of the stellar asset the user wants to receive for their deposit with the anchor.
+`asset_code` | string | The code of the stellar asset the user wants to receive for their deposit with the anchor. The value passed must match one of the codes listed in the [/info](#info) response's deposit object.
```
leighmcculloch, line 157: ❓ This seems to differ from SEP-6, and says it is the asset that the user will receive, not the asset they will deposit. Is that correct?

leighmcculloch (commented):
One blocking question (❓), and some optional suggestions (💡).

JakeUrban:
@leighmcculloch This change makes it so `MyBTC` will _always_ be used.

leighmcculloch:
The old examples suggest that this is a breaking change. Has anyone implemented it to behave that way? Are all implementers doing it the new way? What's the known impact of this change?

JakeUrban:
All known anchors whose tokens differ from the official acronym use it the way I'm suggesting:
- Cowrie (NGNT): https://api.cowrie.exchange/transfer/info
- Tempo (EURT):  https://k.tempocrypto.com/t1/info
- Settle (ARST & BRLT): https://transfer-server.stablex.cloud/info
- Saldo (SMX): https://smx.saldo.mx/api/v1/smx/transfer/info
- Bitbond (EURB): https://stablecoin-bankhaus.bitbond.biz/anchor/api/v1/info

Granted, only Cowrie is using SEP-6, which instructs anchors to use the asset code of the issued asset. I don't know of any other SEP-6 anchor that uses an asset code that differs from the official acronym (AnchorUSD uses USD, Apay uses BTC)

leighmcculloch:
Got it, thanks. Sounds like this is clearly fixing a bug in the specification.

ecosystem/sep-0024.md
leighmcculloch: Answered here: https://github.com/stellar/stellar-protocol/pull/837#issuecomment-763161098

leighmcculloch (approved):
Makes sense.

tomerweller (approved):
If this is just adjusting the spec to existing behavior SGTM.

PR #839, opened 2021-01-20
SEP-12: add 'type' parameter to PUT /customer

### LINKED ISSUE #803: Feature Request: SEP12 PUT - add type field(the type of action the customer is being KYCd for)
Server will know which fields are sent when doing a put for SEP12 customer. 
Right now there is no way to know for what type are the fields for : 
* Counter Party 
* individual (person) client
* or entity (organization) client

Add SEP type field in the SEP12 /PUT model .

JakeUrban:
resolves #803

Improves the phrasing of the "Type Specification" section and adds the `type` parameter to `PUT /customer`

@gurub109


ecosystem/sep-0012.md
```diff
@@ -90,16 +90,16 @@ Name | Type | Description
...
+Different types of customers may have different KYC requirements depending on the action that customer wants to take. The type parameter is used to specify the action a customer wants to make so the server can identify what information must be collected. For example, if a customer is being KYC'd as a SEP31 sender, they may only require full name and email, but a SEP31 receiver would require banking information. The `type` parameter could also be used to specify if an organization or person is being KYC'd, which would require different fields. The values accepted for `type` are up to the implementor in order to signify the types of transactions their services support, and is optional if only one type of customer is needed.
```
leighmcculloch, line 102: I think we need to explicitly state that the values of `type` are undefined by SEP-12 and may be defined by SEPs that utilize SEP-12.

leighmcculloch (approved):
This looks like it needs the patch version incremented, and updated field updated, and I think there could be a statement added (see below), otherwise looks good.

ecosystem/sep-0012.md
JakeUrban: Done. I took out the part where I said `the values accepted for type are up to the implementor` because that information can be communicated by the protocol defining the types.

Issue #802, opened 2020-12-21
Feature Request: More than one SEP12 type support in SEP31 info endpoint


gaclaudiu:
### What problem does your feature solve?
More SEP12 types  to cover the use case when a sender or receiver is a : 
* counter party 
* individual (person) client 
* entity (organization) client

### What would you like to see?
I would like to see SEP31 info point configuration per SEP12 type , or if a config can be reused then the sender_sep12_type or receiver_sep12_type to allow an array. 

For our use case the array implementation should be enough.
Configuration example : 
`{
   "receive":{
      "USD":{
         "enabled":true,
         "fee_fixed":5,
         "fee_percent":1,
         "min_amount":0.1,
         "max_amount":1000,
         "sender_sep12_type": "sep31-senderIndividualClient, sep-31EntityClient",
         "receiver_sep12_type": "sep31-receiverIndividualClient, sep-31EntityClient, sep-31CounterPaty",
         "fields":{
            "transaction":{
               "receiver_routing_number":{
                  "description": "routing number of the destination bank account"
               },
               "receiver_account_number":{
                  "description": "bank account number of the destination"
               },
               "type":{
                  "description": "type of deposit to make",
                  "choices":[
                     "SEPA",
                     "SWIFT"
                  ]
               }
            }
         }
      }
   }
}`

JakeUrban:
Hi @gaclaudiu, I agree we need to be able to specify multiple values for both sender & receiver:
```
{
   "receive":{
      "USD":{
         ...
         "sender_sep12_type": ["sep31-senderIndividualClient, sep-31EntityClient"],
         "receiver_sep12_type": ["sep31-receiverIndividualClient, sep-31EntityClient, sep-31CounterPaty"],
         ...
      }
   }
}
```

Regarding "SEP31 info point configuration per SEP12 type", its a bit more complicated. Some anchors would like to have different info configurations per transaction type (SEPA, SWIFT, etc.) or per-country of operation, see #731, #716, #743. So more discussion on this is needed.

Can you give more info on what fields would need to be different per SEP-12 type?

gaclaudiu:
Hi @JakeUrban , 
"SEP31 info point configuration per SEP12 type"  : 
Use case example : 
You can pass for receiver a client (person or organization ) or a counter party. 

For a client we need only his finClusive account ID to put the money in . 

For a counter party we need it's full bank details to do an ACH to it account . 

[...]

PR #840, opened 2021-01-20
SEP-31: Added 'sep12_types' object for all accepted 'type' values for SEP-12

JakeUrban:
resolves stellar/stellar-protocol#802

This change **deprecates** `sender_sep12_type` and `receiver_sep12_type` in favor of three new fields: `sender_sep12_required`, `receiver_sep12_required`, and `sep12_types`. The former 2 indicate if KYC is necessary for the sending and receiving client respectively, and the `sep12_types` attribute defines the various accepted values for the `type` parameter used in SEP-12 GET requests.

Note that `sender_sep12_required` and `receiver_sep12_required` are mandatory fields, and `sep12_types` is mandatory if either one of previously mentioned values are `true`.


JakeUrban:
@gurub109

ecosystem/sep-0031.md
```diff
@@ -153,8 +153,25 @@ The response should be a JSON object like:
          "fee_percent":1,
          "min_amount":0.1,
          "max_amount":1000,
-         "sender_sep12_type": "sep31-sender",
-         "receiver_sep12_type": "sep31-receiver",
+         "sender_sep12_required": true,
+         "receiver_sep12_required": true,
+         "sep12_types": {
+          "sep31-sender": {
```
leighmcculloch, line 159: 🤔 How does someone go from knowing that `sender_sep12_required` is `true`, to needing to send the `sep31-sender` type? How do they know not to use `sep31-receiver`? How does an implementer communicate that KYC is required for amounts over a certain amount, but not required for under? Or is that not relevant for this SEP?

In the cases that the `_required` fields are `false`, would there still be optional types defined for them, or there would be no types?

If there are two types of SEP-12 types for SEP-31, maybe it is worth separating them so discovering which to use doesn't require parsing the string names and looking for `-sender` and `-receiver` strings.

For example:
```javascript
{
   "receive":{
      "USD":{
        "sender_sep12_types": {
          "sep31-sender": {
            "description": "U.S. citizens limited to sending payments of less than $10,000 in value"
          },
          "sep31-large-sender": {
            "description": "U.S. citizens that do not have sending limits"
          },
          "sep31-foreign-sender": {
            "description": "non-U.S. citizens sending payments of less than $10,000 in value"
          }
        },
        "receiver_sep12_types": {
          "sep31-receiver": {
            "description": "U.S. citizens receiving USD"
          },
          "sep31-foreign-receiver": {
            "description": "non-U.S. citizens receiving USD"
          }
        },
        // ...
      }
   }
}
```

ecosystem/sep-0031.md
```diff
@@ -181,12 +198,15 @@ The JSON object contains an entry for each asset that the anchor supports for re
...
+* `sep12_types`: (optional) an object containing the accepted values for the `type` parameter in [SEP-12](sep-0012.md) requests. Each key should map to an object value with a human-readable `description`. This field can be omitted from the response if no KYC is necessary for both the sender and receiver.
+* `min_amount`: (optional) minimum amount. No limit if not specified.
+* `max_amount`: (optional) maximum amount. No limit if not specified.
+* `fee_fixed`: (optional) fixed (flat) fee for deposit. In units of the received asset. Leave blank if there is no fee or the fee schedule is complex.
+* `fee_percent`: (optional) percentage fee for deposit. In percentage points. Leave blank if there is no fee or the fee schedule is complex.
+* `sender_sep12_type`: (**deprecated**, optional) the default value of the `type` parameter used in `SEP-12 GET /customer` requests for customers sending funds. This field can be omitted from the response if no KYC is necessary.
```
leighmcculloch, line 208: ❗ To say this field is the default value is to change the semantics of it a bit, which I don't think needs to happen. The field can have the description it had previously and you can add a statement that it is deprecated in favor of reading this other field that will list all possible types that can be used rather than one.

ecosystem/sep-0031.md
```diff
@@ -197,8 +217,6 @@ Each `fields` sub-object contains a key for each field name and an object with t
 * `choices`: (optional) list of possible values for the field.
 * `optional`: (optional) false if not specified.
  
-If `sender_sep12_type` or `receiver_sep12_type` are present in the response, the client must register the sender and receiver via [SEP-12](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0012.md#customer-get) using the `type` argument provided. The receiving anchor uses this parameter to know what information needs to be collected on the customer.
```
leighmcculloch, line 200: 💡 I think you still needs this statement, but referencing the new fields? If this is sufficiently discussed elsewhere, no worries.

leighmcculloch (commented):
@JakeUrban I'd like to explore the item below tagged with 🤔 .

ecosystem/sep-0031.md
JakeUrban: > In the cases that the _required fields are false, would there still be optional types defined for them, or there would be no types?

There would be no types. I don't see a use case for optionally KYC'ing a customer.

> If there are two types of SEP-12 types for SEP-31, maybe it is worth separating them so discovering which to use doesn't require parsing the string names and looking for `-sender` and `-receiver` strings.

I agree, I'll make that change. Using `-sender` and `-receiver` suffixes is just a pattern I've developed too -- the `type` values defined here could potentially not follow that pattern. 

> How does an implementer communicate that KYC is required for amounts over a certain amount, but not required for under? Or is that not relevant for this SEP?

Ohh thats a good question, that could be a possibility. The first thing that comes to mind is to define types for all customers -- even those that don't need KYC -- and return an empty `fields` object from `GET /customer` for that no-KYC-necessary type. I can't think of a better way at the moment. The other option is to keep the `_required` field `true` and simply omit a type that would apply to the customer's situation, but that is more confusing than my first suggestion.

ecosystem/sep-0031.md
JakeUrban: I think we can omit it given that the definitions of the new `_required` fields have this language:
> If `true`, the sending anchor should use the appropriate value from `sep12_types` in a [SEP-12 GET /customer?type=](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0012.md#customer-get) request to determine which fields should be collected.

ecosystem/sep-0031.md
JakeUrban: I added a statement to convey how to handle that case.

ecosystem/sep-0031.md
```diff
@@ -146,47 +146,70 @@ The response should be a JSON object like:
...
+  "receive":{
+    "USD":{
+      "enabled":true,
+      "fee_fixed":5,
+      "fee_percent":1,
+      "min_amount":0.1,
+      "max_amount":1000,
+      "sender_sep12_required": true,
+      "receiver_sep12_required": true,
+      "sender_sep12_types": {
```
leighmcculloch, line 158: 💡 I think these fields would benefit from being grouped inside a sep12 object. It makes it really clear where the sep12 info begins and ends in this info.
```javascript
      "sep12": {
        "sender": {
          "required": true,
          "types": {
            // ..
          }
        },
        "receiver": {
          "required": true,
          "types": {
            // ..
          }
        }
```
I also think if types are only provided when required, you could leave the required field out, but I also think it is fine to include it and make that explicit.

leighmcculloch (commented):
@JakeUrban One suggestion (💡) below. Thoughts?

leighmcculloch (approved):
Looks great. Two suggestions (💡) that I think would be helpful, otherwise :shipit:.

ecosystem/sep-0031.md
```diff
@@ -146,47 +146,75 @@ The response should be a JSON object like:
...
 #### For each asset available for receiving, response contains:
 
-* `min_amount`: Optional minimum amount. No limit if not specified.
-* `max_amount`: Optional maximum amount. No limit if not specified.
-* `fee_fixed`: Optional fixed (flat) fee for deposit. In units of the received asset. Leave blank if there is no fee or the fee schedule is complex.
-* `fee_percent`: Optional percentage fee for deposit. In percentage points. Leave blank if there is no fee or the fee schedule is complex.
-* `sender_sep12_type`: Optional value of the `type` parameter the sending anchor should use for a `SEP-12 GET /customer` request. This field can be omitted if no KYC is necessary.
-* `receiver_sep12_type`: Optional value of the `type` parameter the sending anchor should use for a `SEP-12 GET /customer` request. This field can be omitted if no KYC is necessary.
+* `sep12`: an object containing `sender` and `receiver` keys.
+  * `sender`: an object containing a `types` key if KYC information is required for the sending client, empty otherwise.
```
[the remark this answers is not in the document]
JakeUrban: Providing flexibility for the server increases the complexity of the client's parsing logic. If we used the change you suggest, there would be several variations of the no-KYC-required case.

```python
info_resp = getInfo();
if "sender" in info_resp:  # must check for presense of 'sender' key before trying to access it
    sender_data = info_resp["sender"]
    if sender_data and "types" in sender_data:  # must check if sender_data is None (null) before checking for key presense
        ...
```

If you're thinking about returning JSON from Go, would it be easier to instead _always_ have the `types` key? And the `types` key would be empty if no KYC is necessary?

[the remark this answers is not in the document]
JakeUrban: Or another option is to specify that `sender` should be omitted if no KYC is necessary, but it KYC is necessary then `sender.types` should be present and non-empty?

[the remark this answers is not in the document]
JakeUrban: resolved offline -- we're sticking with the current change

PR #841, opened 2021-01-21
SEP-6: update SEP-6 for claimable balances

### LINKED ISSUE #796: SEP-006 Claimable Balance Op Support
Add feature support for new protocol's ClaimableBalance functionality

If an account doesnt have an established trustline for an asset deposits can be completed using the ClaimableBalance op

Related: https://github.com/stellar/stellar-protocol/issues/747

JakeUrban:
resolves #796 

Adds support to SEP-6 for using claimable balance operations instead of waiting for a trustline to be established. The changes mirror [those made to SEP-24](#757) as closely as possible.

### DIFF
--- a/ecosystem/sep-0006.md
+++ b/ecosystem/sep-0006.md
@@ -301,12 +301,6 @@ Predicates are one of the [claimable balance parameters](https://developers.stel
 
 This operation allows a user to redeem an asset currently on the Stellar network for the real asset (BTC, USD, stock, etc...) via the anchor of the Stellar asset.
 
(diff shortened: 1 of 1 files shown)


JakeUrban:
@morleyzhi you should see if Vibrant anchors are willing to adopt this feature, it makes the UX a lot smoother -- wallets don't have to request the user to add a trustline before receiving funds

Zagan202 (approved):
LGTM!
Great work!

morleyzhi:
Thanks @JakeUrban, this is good to know about! We probably don't need something like this for the next few months, since we create accounts with trustlines, but definitely will be useful in the future.

PR #851, opened 2021-01-28
SEP-10: Clarifying Participants

JakeUrban:
This change does not change any functionality related to SEP-10. Rather, it better defines the participants involved in the process and updating the document to always reference those participant definitions. This will help avoid confusion when referencing the participants, instead of using general terms like `client` and `server`.

If merged, I'll update #760 to use the term **Client Server**

### DIFF
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -162,7 +162,6 @@ To validate the challenge transaction the following steps are performed by the *
 * verify that transaction has time bounds set, and that current time is between the minimum and maximum bounds;
 * verify that transaction's first operation is a [Manage Data](https://www.stellar.org/developers/guides/concepts/list-of-operations.html#manage-data) operation that:
   * has a non-null source account
-  * has the **Home Domain** as the key of the operation
 * verify that transaction envelope has a correct signature by the **Server Account**
 * if the operation's source account exists:
   * verify that the remaining signature count is one or more;
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -160,7 +160,9 @@ To validate the challenge transaction the following steps are performed by the *
 * decode the received input as a base64-urlencoded XDR representation of Stellar transaction envelope;
 * verify that transaction source account is equal to the **Server Account**
 * verify that transaction has time bounds set, and that current time is between the minimum and maximum bounds;
-* verify that transaction's first operation is a [Manage Data](https://www.stellar.org/developers/guides/concepts/list-of-operations.html#manage-data) operation that:
+* verify that transaction contains at least one operation;
+* verify that transaction's first operation:
+  * is a Manage Data operation
   * has a non-null source account
 * verify that transaction envelope has a correct signature by the **Server Account**
 * if the operation's source account exists:
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -6,8 +6,8 @@ Title: Stellar Web Authentication
 Author: Sergey Nebolsin <@nebolsin>, Tom Quisel <tom.quisel@gmail.com>, Leigh McCulloch <@leighmcculloch>, Jake Urban <jake@stellar.org>
 Status: Active
 Created: 2018-07-31
-Updated: 2021-02-10
-Version 3.1.1
+Updated: 2021-03-03
+Version 3.1.2
 ```
 
 ## Simple Summary
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -18,57 +18,58 @@ This SEP defines the standard way for clients such as wallets or exchanges to cr
 
 This protocol is a variation of mutual challenge-response, which uses Stellar transactions to encode challenges and responses.
 
-It involves three components:
-- A domain hosting a [SEP-1 stellar.toml](sep-0001.md) containing a `WEB_AUTH_ENDPOINT` (URL) and `SIGNING_KEY` (`G...`), referred to as the home domain.
-- An endpoint providing the GET and POST operations discussed in this document. The endpoint may be hosted on the home domain, a sub-domain of the home domain, or any other domain. The domain where the endpoint is hosted is referred to as the web auth domain. 
-- A client who will authenticate using a Stellar account.
+It involves the following components:
+- A **Home Domain**: a domain hosting a [SEP-1 stellar.toml](sep-0001.md) containing a `WEB_AUTH_ENDPOINT` (URL) and `SIGNING_KEY` (`G...`).
+- A **Server**: a server providing the `WEB_AUTH_ENDPOINT` that implements the GET and POST operations discussed in this document. The server's domain may be the **Home Domain**, a sub-domain of the **Home Domain**, or any other domain.
+    - The `SIGNING_KEY` from the **Home Domain** will be refered to as the **Server Account**.
+- A **Client**: the software used by the holder of the account being authenticated by the **Server**.
+    - The account being authenticated will be refered to as the **Client Account**.
 
 The discovery flow is as follows:
 
-1. The client retrieves the `stellar.toml` from the home domain in accordance with [SEP-1 stellar.toml](sep-0001.md).
-1. The client looks up the `WEB_AUTH_ENDPOINT` and `SIGNING_KEY` from the `stellar.toml`.
+1. The **Client** retrieves the `stellar.toml` from the **Home Domain** in accordance with [SEP-1 stellar.toml](sep-0001.md).
+1. The **Client** looks up the `WEB_AUTH_ENDPOINT` and `SIGNING_KEY` (i.e. **Server Account**) from the `stellar.toml`.
 
 The authentication flow is as follows:
-1. The client obtains a unique [`challenge`](#challenge) from the `WEB_AUTH_ENDPOINT`, which is represented as specially formed Stellar transaction
-1. The client verifies that the transaction has an invalid sequence number 0.  This is extremely important to ensure the transaction isn't malicious.
-1. The client verifies that the transaction is signed by the `SIGNING_KEY` obtained through discovery flow.
-1. The client verifies that the transaction's first operation is a Manage Data operation that has its:
-   1. Source account set to the user's account.
-   1. Key set to `<home domain> auth` where the home domain is the home domain from the discovery flow.
+
+1. The **Client** obtains a unique [`challenge`](#challenge) from the **Server**, which is represented as specially formed Stellar transaction
+1. The **Client** verifies that the transaction has an invalid sequence number 0.  This is extremely important to ensure the transaction isn't malicious.
+1. The **Client** verifies that the transaction is signed by the **Server Account** obtained through discovery flow.
+1. The **Client** verifies that the transaction's first operation is a Manage Data operation that has its:
+   1. Source account set to the **Client Account**
+   1. Key set to `<home domain> auth` where the home domain is the **Home Domain**.
    1. Value set to a nonce value.
-1. The client verifies that if the transaction has a Manage Data operation with key `web_auth_domain` that it has:
-   1. Source account set to the server's account.
-   1. Key set to `web_auth_domain`.
-   1. Value set to the domain name of the SEP-10 server from which the client requested the challenge.
-1. The client verifies that if the transaction has other operations they are Manage Data operations that all have their source accounts set to the the server's account.
-1. The client signs the transaction using the secret key(s) of signers for the user's Stellar account
-1. The client submits the signed challenge back to the server using [`token`](#token) endpoint
-1. The server checks that the user's account exists
-1. If the user's account exists:
-  1. The server gets the signers of the user's account
-  1. The server verifies the client signatures count is one or more;
-  1. The server verifies the client signatures on the transaction are signers of the user's account;
-  1. The server verifies the weight provided by the signers meets the required threshold(s), if any
-  1. If the signatures check out, the server responds with a [JWT](https://jwt.io) that represents the user's session
-1. If the user's account does not exist (optional):
-  1. The server verifies the client signature count is one
-  1. The server verifies the client signature is correct for the master key of the account
-1. Any future calls to the server can be authenticated by including the JWT as a parameter
+1. The **Client** verifies that if the transaction has a Manage Data operation with key `web_auth_domain` that it has:
+   1. Source account set to the **Server Account**.
+   1. Value set to the **Server**'s domain.
+1. The **Client** verifies that if the transaction has other operations they are Manage Data operations that all have their source accounts set to the **Server Account**
+1. The **Client** signs the transaction using the secret key(s) of the signer(s) for the **Client Account**
+1. The **Client** submits the signed challenge back to the **Server** using [`token`](#token) endpoint
+1. The **Server** checks that the **Client Account** exists
+1. If the **Client Account** exists:
+  1. The **Server** gets the signers of the **Client Account**
+  1. The **Server** verifies that one or more signatures are from signers of the **Client Account**.
+  1. The **Server** verifies that there is only one additional signature from the **Server Account**
+  1. The **Server** verifies the weight provided by the signers of the **Client Account** meets the required threshold(s), if any
+1. If the **Client Account** does not exist (optional):
+  1. The **Server** verifies the signature count is two
+  1. The **Server** verifies that one signature is correct for the master key of the **Client Account**
+  1. The **Server** verified that the other signature is from the **Server Account**
+1. If the signatures check out, the **Server** responds with a [JWT](https://jwt.io) that represents the authenticated session
 
 The flow achieves several things:
 
-* Both client and server part can be implemented using well-established Stellar libraries
-* The client can verify that the server holds the secret key to a particular account
-* The server can verify that the client holds the secret key(s) to signers of their account
-* The client is able to prove their identity using a Ledger or other hardware wallet as well as by having direct access to the secret key(s)
-* The server can choose its own timeout for the user's session
-* The server can choose required threshold(s) that the user needs on the account, if any
-* The server can choose to include other application specific claims
-* The server can choose to authenticate accounts that do not yet exist
+* Both **Client** and **Server** can be implemented using well-established Stellar libraries
+* The **Client** can verify that the **Server** holds the secret key to the **Server Account**
+* The **Server** can verify that the **Client** holds the secret key(s) to signer(s) of the **Client Account**
+* The **Server** can choose its own timeout for the authenticated session
+* The **Server** can choose required signing threshold(s) that must be met, if any
+* The **Server** can choose to include other application-specific claims
+* The **Server** can choose to authenticate Stellar accounts that do not yet exist
 
 ## Authentication Endpoint
 
-A web service indicates that it supports user authentication via this protocol by specifying `WEB_AUTH_ENDPOINT` in their [`stellar.toml`](sep-0001.md) file. This is how a wallet knows where to find the authentication server. A web server is required to implement the following behavior for the web authentication endpoint:
+The organization with a **Home Domain** indicates that it supports authentication via this protocol by specifying `WEB_AUTH_ENDPOINT` in their [`stellar.toml`](sep-0001.md) file. This is how a wallet knows where to find the **Server**. A **Server** is required to implement the following behavior for the web authentication endpoint:
 
 * [`GET <WEB_AUTH_ENDPOINT>`](#challenge): request a challenge (step 1)
 * [`POST <WEB_AUTH_ENDPOINT>`](#token): exchange a signed challenge for session JWT (step 2)
@@ -85,8 +86,7 @@ In order for browsers-based wallets to validate the CORS headers, as [specified 
 
 ### Challenge
 
-This endpoint must respond with a Stellar transaction signed by the server that has an invalid sequence number (0) and thus cannot be executed on the Stellar network. The client can then sign the transaction using standard Stellar libraries and submit it to [`token`](#token) endpoint to prove that they control their account. This approach is compatible with hardware wallets such as Ledger. The client must also verify the server's signature to be sure the challenge is signed by the `SIGNING_KEY`, and the home domain in the challenge is the home domain from the discovery flow.
-
+This endpoint must respond with a Stellar transaction signed by the **Server Account** that has an invalid sequence number (0) and thus cannot be executed on the Stellar network. The **Client** can then sign the transaction using standard Stellar libraries and submit it to [`token`](#token) endpoint to prove that it controls the **Client Account**. This approach is compatible with hardware wallets such as Ledger. The **Client Application** must also verify the server's signature to be sure the challenge is signed by the **Server Account**, that the home domain in the first operation of the challenge is the **Home Domain**, and that the web auth domain in a subsequent operation is the **Server** domain.
 
 #### Request
 
@@ -98,8 +98,8 @@ Request Parameters:
 
 Name      | Type          | Description
 ----------|---------------|------------
-`account` | `G...` string | The stellar account that the wallet wishes to authenticate with the server
-`home_domain` | string | (optional) The home domain the client would like to authenticate with. Servers that generate tokens for multiple home domains can use this parameter to identify which home domain the client hopes to authenticate. If not provided by the client, the server should assume a default for backwards compatibility with older clients.
+`account` | `G...` string | The stellar account that the wallet wishes to authenticate with the **Server** 
+`home_domain` | string | (optional) a **Home Domain**. Servers that generate tokens for multiple **Home Domain**s can use this parameter to identify which home domain the **Client** hopes to authenticate with. If not provided by the **Client**, the **Server** should assume a default for backwards compatibility with older **Clients**.
 
 Example:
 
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0010.md
```diff
@@ -19,56 +19,56 @@ This SEP defines the standard way for clients such as wallets or exchanges to cr
...
 
 It involves three components:
-- A domain hosting a [SEP-1 stellar.toml](sep-0001.md) containing a `WEB_AUTH_ENDPOINT` (URL) and `SIGNING_KEY` (`G...`), referred to as the home domain.
-- An endpoint providing the GET and POST operations discussed in this document. The endpoint may be hosted on the home domain, a sub-domain of the home domain, or any other domain. The domain where the endpoint is hosted is referred to as the web auth domain. 
-- A client who will authenticate using a Stellar account.
+- The **TOML Server**: a server hosting a [SEP-1 stellar.toml](sep-0001.md) containing a `WEB_AUTH_ENDPOINT` (URL) and `SIGNING_KEY` (`G...`). The ``WEB_AUTH_ENDPOINT`` must be hosted by the **Web Auth Server** as described below.
+- The **Web Auth Server**: a domain hosting an endpoint providing the GET and POST operations discussed in this document. The endpoint may be hosted on the **TOML Server**'s domain, a sub-domain of the **TOML Server**'s domain, or any other domain.
+- The **Authenticating Account**: The public key of the Stellar account being authenticated.
+- The **Client Application**: the application used by the owner of the **Authenticating Account**.
```
leighmcculloch, line 25: 💡 I think `owner` is okay, but `account holder` is more specific and descriptive, and also a common phrase used in financial products which helps provide more depth here.
```suggestion
- The **Client Application**: the application used by the account holder of the **Authenticating Account**.
```

ecosystem/sep-0010.md
```diff
@@ -19,56 +19,56 @@ This SEP defines the standard way for clients such as wallets or exchanges to cr
 This protocol is a variation of mutual challenge-response, which uses Stellar transactions to encode challenges and responses.
 
 It involves three components:
-- A domain hosting a [SEP-1 stellar.toml](sep-0001.md) containing a `WEB_AUTH_ENDPOINT` (URL) and `SIGNING_KEY` (`G...`), referred to as the home domain.
-- An endpoint providing the GET and POST operations discussed in this document. The endpoint may be hosted on the home domain, a sub-domain of the home domain, or any other domain. The domain where the endpoint is hosted is referred to as the web auth domain. 
-- A client who will authenticate using a Stellar account.
+- The **TOML Server**: a server hosting a [SEP-1 stellar.toml](sep-0001.md) containing a `WEB_AUTH_ENDPOINT` (URL) and `SIGNING_KEY` (`G...`). The ``WEB_AUTH_ENDPOINT`` must be hosted by the **Web Auth Server** as described below.
```
leighmcculloch, line 22: 🔠 I think we probably want to keep `The` => `A` for all of these, because we're not referring to a definite article, but an indefinite article. We're referring to any toml server participating in SEP-10.

ecosystem/sep-0010.md
```diff
@@ -19,56 +19,56 @@ This SEP defines the standard way for clients such as wallets or exchanges to cr
...
+- The **Web Auth Server**: a domain hosting an endpoint providing the GET and POST operations discussed in this document. The endpoint may be hosted on the **TOML Server**'s domain, a sub-domain of the **TOML Server**'s domain, or any other domain.
```
leighmcculloch, line 23: 💡  I think TOML Server and Web Auth Server are being a little too prescriptive and distancing the definition of the participants from what matters for SEP-10. SEP-10 doesn't need you to have these servers in the modern sense of the word, but needs you to provides:
- A home domain that has a toml file at a URL at that domain.
- An web auth endpoint.

Whatever approach an implementer takes to providing those is out-of-scope, and making the home domain a participant is critical to the rest of the specification where the home domain is a thing.

```suggestion
- A **Home Domain**: a domain hosting a [SEP-1 stellar.toml](sep-0001.md) containing a `WEB_AUTH_ENDPOINT` (URL) and `SIGNING_KEY` (`G...`).
- A **Web Auth Endpoint**: an endpoint providing the GET and POST operations discussed in this document. The endpoint may be hosted on the **Home Domain**, a sub-domain of the **Home Domain**, or any other domain. 
- A **Web Auth Domain**: the domain hosting the **Web Auth Endpoint**.
```

Thoughts?

leighmcculloch (commented):
Some thoughts below just on the definitions, I didn't read the full PR yet, but can do so when we agree on if we prefer TOML Server / Web Auth Server, or Home Domain and Web Auth Endpoint,

ecosystem/sep-0010.md
JakeUrban: I like the change from `TOML Server` to `Home Domain` because you're right, we use the word 'home domain' throughout the doc and in request parameters. 

However I think the change from `Web Auth Server` to `Web Auth Domain` & `Web Auth Endpoint` is a bit unnecessary. Unlike the home domain hosting a static file at a path, an actual API server is required to process authentication requests. I don't see the value of trying to de-couple the endpoint and domain of the web auth server.

leighmcculloch: I see your point. How about we replace Web Auth Endpoint with Web Auth Server, but keep Web Auth Domain in the list because it is an important attribute elsewhere in the SEP?

JakeUrban: Hmm if the goal is to define a component for the `web_auth_domain` key mentioned in the SEP, then I think we should go with what you had: `Web Auth Endpoint` & `Web Auth Domain`, since the TOML uses `WEB_AUTH_ENDPOINT`.

However, I'm not sure it worth it to explicitly define a `Web Auth Domain`, since there are only two places where it would be used. Do you think the current phrasing is confusing?

(in the abstract)
> 1. The **Client Application** verifies that if the transaction has a Manage Data operation with key `web_auth_domain` that it has:
>       1. Source account set to the ``SIGNING_KEY`` from the **Home Domain**'s [SEP-1 stellar.toml](sep-0001.md).
>       1. Value set to the **Web Auth Server**'s domain.

(in the response)
> * `manage_data(source: server_account, key: 'web_auth_domain', value: web_auth_domain)`
>        * The source account is the **Home Domain**'s ``SIGNING_KEY``, the value is the domain of the **Web Auth Server**. It can be at most 64 characters.

leighmcculloch: I see your point. It's not anymore ambiguous to refer to the domain of the Web Auth Server.

I also agree for Web Auth Endpoint so as to remain consistent with the stellar.toml, so there is one less new concept and we continue to reuse the existing concept.

JakeUrban: Ok, so the take-away is to change `Web Auth Server` to `Web Auth Endpoint`, but not use the additional `Web Auth Domain` term, since we can easily say `the domain of the **Web Auth Endpoint**`?

leighmcculloch: Yup, that's how I interpreted what you were saying and I think that strikes the balance.

ecosystem/sep-0010.md
```diff
@@ -19,56 +19,56 @@ This SEP defines the standard way for clients such as wallets or exchanges to cr
 This protocol is a variation of mutual challenge-response, which uses Stellar transactions to encode challenges and responses.
 
 It involves three components:
-- A domain hosting a [SEP-1 stellar.toml](sep-0001.md) containing a `WEB_AUTH_ENDPOINT` (URL) and `SIGNING_KEY` (`G...`), referred to as the home domain.
-- An endpoint providing the GET and POST operations discussed in this document. The endpoint may be hosted on the home domain, a sub-domain of the home domain, or any other domain. The domain where the endpoint is hosted is referred to as the web auth domain. 
-- A client who will authenticate using a Stellar account.
+- A **Home Domain**: a domain hosting a [SEP-1 stellar.toml](sep-0001.md) containing a `WEB_AUTH_ENDPOINT` (URL) and `SIGNING_KEY` (`G...`). The ``WEB_AUTH_ENDPOINT`` must be hosted by a **Web Auth Endpoint** as described below.
```
leighmcculloch, line 22: 💡 I don't think you need this, it is rather self explanatory especially given the proximity to the next line.
```suggestion
- A **Home Domain**: a domain hosting a [SEP-1 stellar.toml](sep-0001.md) containing a `WEB_AUTH_ENDPOINT` (URL) and `SIGNING_KEY` (`G...`).
```

ecosystem/sep-0010.md
```diff
@@ -19,56 +19,56 @@ This SEP defines the standard way for clients such as wallets or exchanges to cr
...
+- A **Web Auth Endpoint**: a domain hosting an endpoint providing the GET and POST operations discussed in this document. The endpoint may be hosted on a **Home Domain**, a sub-domain of a **Home Domain**, or any other domain.
```
leighmcculloch, line 23: ❗ The endpoint isn't a domain, but it leads by saying that it is a domain.

```suggestion
- A **Web Auth Endpoint**: an endpoint providing the GET and POST operations discussed in this document. The endpoint may be hosted on the **Home Domain**, a sub-domain of the **Home Domain**, or any other domain.
```

❗ After re-reading this I think it is still necessary to either keep the sentence `The domain where the endpoint is hosted is referred to as the web auth domain.` which is in the original, or to add that statement to a new list item. A list item would be more explicit:

```suggestion
- A **Web Auth Endpoint**: an endpoint providing the GET and POST operations discussed in this document. The endpoint may be hosted on a **Home Domain**, a sub-domain of a **Home Domain**, or any other domain.
- A **Web Auth Domain**: the domain where the **Web Auth Endpoint** is hosted. 
```

ecosystem/sep-0010.md
```diff
@@ -19,56 +19,56 @@ This SEP defines the standard way for clients such as wallets or exchanges to cr
...
+- An **Authenticating Account**: The public key of the Stellar account being authenticated.
+- The **Client Application**: the application used by the account holder of the **Authenticating Account**.
```
leighmcculloch, line 25: ❗ We should be consistent in this list with the A vs The leads:
```suggestion
- A **Client Application**: the software used by the account holder of the **Authenticating Account**.
```

ecosystem/sep-0010.md
```diff
@@ -19,56 +19,56 @@ This SEP defines the standard way for clients such as wallets or exchanges to cr
 This protocol is a variation of mutual challenge-response, which uses Stellar transactions to encode challenges and responses.
 
 It involves three components:
-- A domain hosting a [SEP-1 stellar.toml](sep-0001.md) containing a `WEB_AUTH_ENDPOINT` (URL) and `SIGNING_KEY` (`G...`), referred to as the home domain.
-- An endpoint providing the GET and POST operations discussed in this document. The endpoint may be hosted on the home domain, a sub-domain of the home domain, or any other domain. The domain where the endpoint is hosted is referred to as the web auth domain. 
-- A client who will authenticate using a Stellar account.
+- A **Home Domain**: a domain hosting a [SEP-1 stellar.toml](sep-0001.md) containing a `WEB_AUTH_ENDPOINT` (URL) and `SIGNING_KEY` (`G...`). The ``WEB_AUTH_ENDPOINT`` must be hosted by a **Web Auth Endpoint** as described below.
+- A **Web Auth Endpoint**: a domain hosting an endpoint providing the GET and POST operations discussed in this document. The endpoint may be hosted on a **Home Domain**, a sub-domain of a **Home Domain**, or any other domain.
+- An **Authenticating Account**: The public key of the Stellar account being authenticated.
```
leighmcculloch, line 24: ❗ The component is the account not the public key. In the API we might reference the public key but the component in play is the account itself.
```suggestion
- An **Authenticating Account**: a Stellar account being authenticated.
```

ecosystem/sep-0010.md
JakeUrban: I agree, its ultimately better to be explicit even if `Web Auth Domain` is seldom used. Editing your 2nd suggestion so that it doesn't conflict with the first -- I'm committing both.

ecosystem/sep-0010.md
JakeUrban: I don't feel strongly either way, but the account may not actually exist, which is why I chose to mention the public key specifically.

ecosystem/sep-0010.md
```diff
@@ -85,8 +86,7 @@ In order for browsers-based wallets to validate the CORS headers, as [specified
...
-
+This endpoint must respond with a Stellar transaction signed by the ``SIGNING_KEY`` from the **Home Domain**'s [`stellar.toml`](sep-0001.md) file that has an invalid sequence number (0) and thus cannot be executed on the Stellar network. The **Client Application** can then sign the transaction using standard Stellar libraries and submit it to [`token`](#token) endpoint to prove that it controls the **Authenticating Account**. This approach is compatible with hardware wallets such as Ledger. The **Client Application** must also verify the home domain in the first operation of the challenge is the **Home Domain**.
```
leighmcculloch, line 89: ❗️In the last sentence of this section it called out that the client must verify the servers signature. That's been removed, but it seems unnecessary to me to remove it if this change is scoped to clarifying the participants. We kept the statement to verify the home domain, which seems like we're playing more importance on that when everything that needs verifying is important.

ecosystem/sep-0010.md
```diff
@@ -112,19 +112,19 @@ GET https://auth.example.com/?account=GCIBUCGPOHWMMMFPFTDWBSVHQRT4DIBJ7AD6BZJYDI
...
-  * source account set to server's signing account
+  * source account set to the **Home Domain**'s ``SIGNING_KEY``
   * invalid sequence number (set to 0) so the transaction cannot be run on the Stellar network
-  * time bounds: `{min: now(), max: now() + 900 }` (we recommend expiration of 15 minutes to give user time to sign transaction)
+  * time bounds: `{min: now(), max: now() + 900 }` (we recommend expiration of 15 minutes to give the **Client Application** time to sign transaction)
   * operations:
-    * `manage_data(source: client_account, key: '<home domain> auth', value: random_nonce())`
-      * The value of key is the home domain the client is authenticating with, followed by `auth`. It can be at most 64 characters.
+    * `manage_data(source: authenticating_account, key: '<home domain> auth', value: random_nonce())`
+      * The value of key is the **Home Domain**, followed by `auth`. It can be at most 64 characters.
```
leighmcculloch, line 120: 💡 I think it is worth adding a dot point here above the key dot point, that states that the source account is the `**Authenticating Account**` so you make use of the participant bold text here.

ecosystem/sep-0010.md
```diff
@@ -112,19 +112,19 @@ GET https://auth.example.com/?account=GCIBUCGPOHWMMMFPFTDWBSVHQRT4DIBJ7AD6BZJYDI
...
       * The value must be 64 bytes long. It contains a 48 byte cryptographic-quality random string encoded using base64 (for a total of 64 bytes after encoding).
     * subsequent operations, order unimportant:
       * `manage_data(source: server_account, key: 'web_auth_domain', value: web_auth_domain)`
-        * The value is the web auth domain, the domain hosting the web auth endpoint and that the client connected to when requesting the challenge transaction. It can be at most 64 characters.
+        * The source account is the **Home Domain**'s ``SIGNING_KEY``, the value is the **Web Auth Domain**. It can be at most 64 characters.
```
leighmcculloch, line 124: 💡 I think it is worth splitting the source account, and value into separate dot points.

ecosystem/sep-0010.md
```diff
@@ -146,38 +146,36 @@ Every other HTTP status code will be considered an error. For example:
...
-* if the first operation's source account exists:
-  * verify that client signatures count is one or more;
-  * verify that client signatures on the transaction are signers of the first operation's source account;
-  * verify that client signatures are correct;
-  * verify that client signatures provide weight that meets the required threshold(s), if any;
-* if the first operation's source account does not exist:
-  * verify that client signature count is one;
-  * verify that client signature is correct for the master key of the first operation's source account address;
-* verify that transaction containing additional Manage Data operations have their source account as the server signing key;
+  * has the **Home Domain** as the key of the operation
```
leighmcculloch, line 163: 💡 It's unnecessary for the server to verify this because they signed the transaction. Our SDKs might do it, but there's little need in the specification to specify it. It doesn't hurt to specify it, but this is why we don't specify verifying the web auth domain here as well.

leighmcculloch (commented):
I've got some more changes I'd like to make. Sorry to bouncing on some of these things, but there's a lot of content, subtlety and words and semantics here. I've outlined some requested suggestions below. I've also opened a PR making the suggestions: https://github.com/JakeUrban/stellar-protocol/pull/1. I'm happy to talk synchronously through these changes if that would be helpful too.

---

👏🏻 I'm not sure I said this yet, but this PR goes a long way to clearing up ambiguities and making the document more mature. Thanks for identifying that this was needed and taking charge of it.

ecosystem/sep-0010.md
```diff
@@ -18,57 +18,58 @@ This SEP defines the standard way for clients such as wallets or exchanges to cr
...
-  1. The server verifies the client signature count is one
-  1. The server verifies the client signature is correct for the master key of the account
-1. Any future calls to the server can be authenticated by including the JWT as a parameter
+1. The **Client Application** verifies that if the transaction has a Manage Data operation with key `web_auth_domain` that it has:
+   1. Source account set to the ``SIGNING_KEY`` from the **Home Domain**'s [SEP-1 stellar.toml](sep-0001.md).
+   1. Value set to the **Web Auth Domain**.
+1. The **Client Application** verifies that if the transaction has other operations they are Manage Data operations that all have their source accounts set to the the ``SIGNING_KEY`` from the **Home Domain**'s [SEP-1 stellar.toml](sep-0001.md).
+1. The **Client Application** signs the transaction using the secret key(s) of the signer(s) for the **Authenticating Account**
+1. The **Client Application** submits the signed challenge back to the **Web Auth Endpoint** using [`token`](#token) endpoint
+1. The **Web Auth Endpoint** checks that the **Authenticating Account** exists
```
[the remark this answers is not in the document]
JakeUrban: I think having a `Web Auth Endpoint`, `Web Auth Server`, and `Web Auth Domain` adds more potential for confusion than having a single `Web Auth` participant and referencing related properties of that participant. Listing each as participants makes it seem like they are independent of each other, even if the definitions at the top of the document clarify that they're all related.

My preference would be to only define `Web Auth Server` in the participants list, but as you suggest we can make sure we're referring to the `Web Auth Server`'s endpoint when appropriate, and same for it's domain.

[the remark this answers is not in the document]
leighmcculloch: I see your point. 👍

I think this also raises the question if the authenticating account is necessary, given that it is just a component of the client and it might be better to return to Client and Server, but do as you've done here and make them explicit identifiers by holding them and using them intentionally. Then attribution introduces only one new concept, the Client Server.

Thoughts?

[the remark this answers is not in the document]
JakeUrban: So you're suggesting to have the following components defined in the abstract:
- Home Domain
- Client Application
- Web Auth Server

But still use the bolded style when referring to properties related to the components defined in the abstract? I think we should reserve the bolded style to explicitly defined components, because bolding effectively implies that they are defined.

Do you think any of the following phrases are ambiguous?

> Source account set to the account the **Client Application** is authenticating

> The **Client Application** obtains a unique [`challenge`](#challenge) from the endpoint implemented by the **Web Auth Server**

> The source account is the **Home Domain**'s ``SIGNING_KEY``, the value is the domain on which the **Web Auth Server** is hosted.

The above phrases replaces usage of **Authenticating Account**, **Web Auth Endpoint**, and **Web Auth Domain** terms. It may be more verbose but I don't think there is potential for misunderstanding.

Client Attribution would then introduce the **Client Server**, along with language like the following:

> The **Client Server** should implement an endpoint that accepts a challenge transaction from the **Client Application** and signs it using the `SIGNING_KEY` listed on the [SEP-1 stellar.toml](...) file hosted on it's domain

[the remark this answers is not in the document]
leighmcculloch: I was meaning what you outlined, only three bolded components and rely on as brief identification for others, then the client server becomes a new component.

I'd like to find a more succinct form for the server account/signing key though. 

I think if we rename Web Auth Server to simply Server, then to say Client account and Server account is sufficient, then Client Server is still distinct?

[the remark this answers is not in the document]
JakeUrban: Sounds good to me!

ecosystem/sep-0010.md
JakeUrban: Aren't all verifications above this comment unnecessary though? The only thing the server needs to do is verify that it's account signed the transaction and then verify the other signatures depending on whether the client account exists. Should I remove all verification points above this comment?

leighmcculloch: You're right, but I don't think there is significant value in removing existing statements made here, I just wouldn't go out of your way to add new ones.

ecosystem/sep-0010.md
```diff
@@ -18,57 +18,58 @@ This SEP defines the standard way for clients such as wallets or exchanges to cr
...
-- A domain hosting a [SEP-1 stellar.toml](sep-0001.md) containing a `WEB_AUTH_ENDPOINT` (URL) and `SIGNING_KEY` (`G...`), referred to as the home domain.
-- An endpoint providing the GET and POST operations discussed in this document. The endpoint may be hosted on the home domain, a sub-domain of the home domain, or any other domain. The domain where the endpoint is hosted is referred to as the web auth domain. 
-- A client who will authenticate using a Stellar account.
+It involves the following components:
+- A **Home Domain**: a domain hosting a [SEP-1 stellar.toml](sep-0001.md) containing a `WEB_AUTH_ENDPOINT` (URL) and `SIGNING_KEY` (`G...`).
+- A **Server**: a server providing the `WEB_AUTH_ENDPOINT` that implements the GET and POST operations discussed in this document. The server's domain may be the **Home Domain**, a sub-domain of the **Home Domain**, or any other domain.
+    - The `SIGNING_KEY` from the **Home Domain** will be refered to as the **Server Account**.
+- A **Client**: the software used by the holder of the account being authenticated by the **Server**.
+    - The account being authenticated will be refered to as the **Client Account**.
```
leighmcculloch, line 26: 🔠 
```suggestion
    - The account being authenticated is the **Client Account**.
```

ecosystem/sep-0010.md
```diff
@@ -18,57 +18,58 @@ This SEP defines the standard way for clients such as wallets or exchanges to cr
...
 This protocol is a variation of mutual challenge-response, which uses Stellar transactions to encode challenges and responses.
 
-It involves three components:
-- A domain hosting a [SEP-1 stellar.toml](sep-0001.md) containing a `WEB_AUTH_ENDPOINT` (URL) and `SIGNING_KEY` (`G...`), referred to as the home domain.
-- An endpoint providing the GET and POST operations discussed in this document. The endpoint may be hosted on the home domain, a sub-domain of the home domain, or any other domain. The domain where the endpoint is hosted is referred to as the web auth domain. 
-- A client who will authenticate using a Stellar account.
+It involves the following components:
+- A **Home Domain**: a domain hosting a [SEP-1 stellar.toml](sep-0001.md) containing a `WEB_AUTH_ENDPOINT` (URL) and `SIGNING_KEY` (`G...`).
+- A **Server**: a server providing the `WEB_AUTH_ENDPOINT` that implements the GET and POST operations discussed in this document. The server's domain may be the **Home Domain**, a sub-domain of the **Home Domain**, or any other domain.
+    - The `SIGNING_KEY` from the **Home Domain** will be refered to as the **Server Account**.
```
leighmcculloch, line 24: 🔠 
```suggestion
    - The `SIGNING_KEY` from the **Home Domain** is the **Server Account**.
```

ecosystem/sep-0010.md
```diff
@@ -18,57 +18,58 @@ This SEP defines the standard way for clients such as wallets or exchanges to cr
 
 This protocol is a variation of mutual challenge-response, which uses Stellar transactions to encode challenges and responses.
 
-It involves three components:
-- A domain hosting a [SEP-1 stellar.toml](sep-0001.md) containing a `WEB_AUTH_ENDPOINT` (URL) and `SIGNING_KEY` (`G...`), referred to as the home domain.
-- An endpoint providing the GET and POST operations discussed in this document. The endpoint may be hosted on the home domain, a sub-domain of the home domain, or any other domain. The domain where the endpoint is hosted is referred to as the web auth domain. 
-- A client who will authenticate using a Stellar account.
+It involves the following components:
+- A **Home Domain**: a domain hosting a [SEP-1 stellar.toml](sep-0001.md) containing a `WEB_AUTH_ENDPOINT` (URL) and `SIGNING_KEY` (`G...`).
+- A **Server**: a server providing the `WEB_AUTH_ENDPOINT` that implements the GET and POST operations discussed in this document. The server's domain may be the **Home Domain**, a sub-domain of the **Home Domain**, or any other domain.
```
leighmcculloch, line 23: 💡 This is sort of out-of-scope of your PR but I think this tiny tweak is worthwhile. The 'any other domain' doesn't quite communicate the same intent as saying it can be a different domain, although they're similar.
```suggestion
- A **Server**: a server providing the `WEB_AUTH_ENDPOINT` that implements the GET and POST operations discussed in this document. The server's domain may be the **Home Domain**, a sub-domain of the **Home Domain**, or a different domain.
```

ecosystem/sep-0010.md
```diff
@@ -18,57 +18,58 @@ This SEP defines the standard way for clients such as wallets or exchanges to cr
...
+    - The `SIGNING_KEY` from the **Home Domain** will be refered to as the **Server Account**.
+- A **Client**: the software used by the holder of the account being authenticated by the **Server**.
+    - The account being authenticated will be refered to as the **Client Account**.
 
 The discovery flow is as follows:
 
-1. The client retrieves the `stellar.toml` from the home domain in accordance with [SEP-1 stellar.toml](sep-0001.md).
-1. The client looks up the `WEB_AUTH_ENDPOINT` and `SIGNING_KEY` from the `stellar.toml`.
+1. The **Client** retrieves the `stellar.toml` from the **Home Domain** in accordance with [SEP-1 stellar.toml](sep-0001.md).
+1. The **Client** looks up the `WEB_AUTH_ENDPOINT` and `SIGNING_KEY` (i.e. **Server Account**) from the `stellar.toml`.
```
leighmcculloch, line 31: 🎉 This referencing here of both the signing key and the server account at the first mention is helpful, great idea.

ecosystem/sep-0010.md
```diff
@@ -18,57 +18,58 @@ This SEP defines the standard way for clients such as wallets or exchanges to cr
...
-  1. The server verifies the client signatures on the transaction are signers of the user's account;
-  1. The server verifies the weight provided by the signers meets the required threshold(s), if any
-  1. If the signatures check out, the server responds with a [JWT](https://jwt.io) that represents the user's session
-1. If the user's account does not exist (optional):
-  1. The server verifies the client signature count is one
-  1. The server verifies the client signature is correct for the master key of the account
-1. Any future calls to the server can be authenticated by including the JWT as a parameter
+1. The **Client** verifies that if the transaction has a Manage Data operation with key `web_auth_domain` that it has:
+   1. Source account set to the **Server Account**.
+   1. Value set to the **Server**'s domain.
```
leighmcculloch, line 44: ❗ I think we need to keep the text that was removed. It is important that the client verify it matches the domain they requested from. I don't want that distinction to be lost.
```suggestion
   1. Value set to the **Server**'s domain that the client requested the challenge from.
```

ecosystem/sep-0010.md
```diff
@@ -112,19 +112,21 @@ GET https://auth.example.com/?account=GCIBUCGPOHWMMMFPFTDWBSVHQRT4DIBJ7AD6BZJYDI
...
-      * zero or more `manage_data(source: server_account, ...)` reserved for future use
-  * signature by the service's stellar.toml `SIGNING_KEY`
-* `network_passphrase`: (optional but recommended) Stellar network passphrase used by the server. This allows the client to verify that it's using the correct passphrase when signing and is useful for identifying when a client or server have been configured incorrectly.
+      * `manage_data(source: server account, key: 'web_auth_domain', value: web_auth_domain)`
+        * The source account is the **Server Account**
+        * The value is the **Server**'s domain. It can be at most 64 characters.
+      * zero or more `manage_data(source: server account, ...)` reserved for future use
+  * signature by the **Server Account**
+* `network_passphrase`: (optional but recommended) Stellar network passphrase used by the **Home Domain**. This allows a **Client** to verify that it's using the correct passphrase when signing and is useful for identifying when a **Client** or **Server** have been configured incorrectly.
```
leighmcculloch, line 129: ❗ I think this still needs to say that the intention is to check what the server is using, not the home domain. If the client wanted to check what the home domain used, it'd look at the toml.
```suggestion
* `network_passphrase`: (optional but recommended) Stellar network passphrase used by the **Server**. This allows a **Client** to verify that it's using the correct passphrase when signing and is useful for identifying when a **Client** or **Server** have been configured incorrectly.
```

leighmcculloch (commented):
This is awesome and so much easier to read with these changes. I've got a two small things that needs tweaking (❗) before merging, and three small suggestions for wording (💡🔠) that I'll defer to you but I think read clearer and more direct.

Mention me when these are addressed and I'll approve.

JakeUrban:
@leighmcculloch no issues with your suggestions, I committed them all 👍

leighmcculloch (approved):
Looks great!

Before merging, can you update the meta data? Version 3.1.2.

PR #854, opened 2021-01-29
SEP-0008: Address ambiguities in the action_required use case

leighmcculloch:
### What
Add and change statements that discuss the action_required response to state that the use of `action_fields` is optional for both the server to provide, and the client to use, and exists to allow the client to improve the UX by circumventing the need for the user to enter information.

Removed statements about the POST flow that were confusing and did not add a lot of value.

### Why
@howardtw pointed out some gaps in understanding when reviewing SEP-8 since #_ was merged.

This makes no functional changes to the specification, and only clarifies existing behavior.

### DIFF
--- a/ecosystem/sep-0008.md
+++ b/ecosystem/sep-0008.md
@@ -234,7 +234,7 @@ Name | Type | Description
 `message` | string | A human readable string containing information regarding the action required.
 `action_url` | string | A URL that allows the user to complete the actions required to have the transaction approved.
 `action_method` | string | (optional) `GET` or `POST`, indicating the type of request that should be made to the `action_url`. If not provided, `GET` is assumed.
-`action_fields` | string[] | (optional) An array of additional fields defined by [SEP-9 Standard KYC / AML fields] that the client may optionally provide to the approval service when sending the request to the `action_url`.
+`action_fields` | string[] | (optional) An array of additional fields defined by [SEP-9 Standard KYC / AML fields] that the client may optionally provide to the approval service when sending the request to the `action_url` so as to circumvent the need for the user to enter the information manually.
 
 ###### Example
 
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0008.md
```diff
@@ -248,14 +248,20 @@ Name | Type | Description
...
+- In the case that `action_method` is `GET`, or not defined:
+
+   Fields requested may be passed as query parameters in the `action_url`. The URL should be opened in a browser. Any fields passed should allow the server to proceed without collecting user information, or at the least pre-fill the provided information.
+
+- In the case that `action_method` is `POST`:
+
+   Fields requested may be passed as JSON in the request body, using content type `application/json`. The server should respond to the `POST` with a HTTP response with status code `200`, content-type `application/json`. The body should contain either an acknowledgement that the `POST` was sufficient and no further action is required, or the next URL if action is still required. If a next URL is provided, it should be loaded in a browser for the user to complete any action required.
```
JakeUrban, line 264: I don't see how a client would know how to determine if the request was sufficient or not,

- What is the attribute name of the URL provided in the response? 
- What are the attribute(s) that indicate the request was sufficient?

JakeUrban: Making changes to answer the above questions is probably best done in a separate PR since this one is just for rephrasing the existing functionality.

JakeUrban (commented):
From a phrasing perspective, LGTM. However I think we need to better define the responses for `POST <action_url>`, which should probably be done in a separate PR.

ecosystem/sep-0008.md
```diff
@@ -248,14 +248,20 @@ Name | Type | Description
...
+There are two possible values for `action_method`.
 
-   The next URL in the POST case is necessary to support products that are websites and products that are mobile applications. Products that run entirely in the browser can redirect the user to the next URL and the redirect prevents odd user experiences where a user refreshes a page and is prompted to re-post the data. Products that are mobile applications cannot open a system browser with a POST request. They can make the POST request internally, extract the redirect in the response, and open a system browser with the next URL. Separating the next URL also makes it possible for the server to indicate that the fields provided in the POST request were sufficient and no further user interaction is required.
+- In the case that `action_method` is `GET`, or not defined:
+
+   Fields requested may be passed as query parameters in the `action_url`. The URL should be opened in a browser. Any fields passed should allow the server to proceed without collecting user information, or at the least pre-fill the provided information.
```
howardtw, line 260: > The URL should be opened in a browser.

Is it in the context of web applications? In the case of mobile application, a webview might be more relevant? Perhaps we shouldn't dictate where (browser or mobile apps) and how (popups/webview or another page) the URL is opened?

howardtw (commented):
Thanks for the update 🙏 Left one comment about the mention of browser.

ecosystem/sep-0008.md
howardtw: Just using my best knowledge of [this](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0008.md#action-url) section to answer your questions:

> What is the attribute name of the URL provided in the response?

I think it's named `next_url`.

> What are the attribute(s) that indicate the request was sufficient?

If the response doesn't contain `next_url`, it means no further action is required.

JakeUrban: Ah, I missed that, never mind then!

JakeUrban (approved):

ecosystem/sep-0008.md
leighmcculloch: A webview is a type of browser, an embeddable browser. I agree SEP-8 shouldn't dictate UX, it should defer that to implementers as much as possible which is why it doesn't discuss popups, webviews, etc.

leighmcculloch: Any tips for alternative words to use than browser are welcome, let me know if you have a better term.

ecosystem/sep-0008.md
leighmcculloch: There is also the `result` response parameter that indicates what type of further action is required or not.

ecosystem/sep-0008.md
howardtw: I was thinking we can remove that particular sentence. Is the purpose of including it to differentiate the case of POST, which can happen behind the scene? For non-sensitive information or GET requests in general, can't it happen behind the scene?

howardtw (approved):

ecosystem/sep-0008.md
leighmcculloch: I added it because you expressed some confusion on our call about which request was displayed to the user and which wasn't, so I added the statements about which are opened in a browser to try and address that. I think it adds clarity.

>For non-sensitive information or GET requests in general, can't it happen behind the scene?

No, the action_required response is to indicate some user action is required. That could be to provide information, or to agree/sign an agreement. The POST case just adds the ability to potentially circumvent that early.

If I'm missing what you mean could you open a PR against this PR to demonstrate?

howardtw: Feel free to merge it.

Just a final confirmation:
Is it correct that we expect the clients to always GET in the front in a browser and POST in the background? I understand the interactive flow with GET is the original way of doing things. It's just that with the addition of POST, which is expected to happen in the background, it wasn't super clear to me whether clients can use GET in the background and use POST in the front (browser). Readers would have to understand that purpose of the addition of POST is to circumvent the interactive flow, not just to send sensitive information.

leighmcculloch: What you said is correct. There is no change to the use case of GET in the foreground. The doc mentions it exists to circumvent the interactive flow, but let me know if we can say more to make that more obvious.

PR #855, opened 2021-02-01
Update cap-0035.md

rice2000:


rice2000:
Move CAP-35 to `FCP: Acceptance`

PR #857, opened 2021-02-02
CAP-0035: Define what happens if an issuer creates a claimable balance

leighmcculloch:
### What
Define what happens if an issuer creates a claimable balance.

### Why
@jonjove pointed out in https://github.com/stellar/stellar-core/pull/2893#discussion_r567064267 that the issuer can create a claimable balance but doesn't have a trustline to itself, and we need to define what the clawback enabled state of the claimable balance is in this case.

This was discussed on the mailing list starting [here](https://groups.google.com/g/stellar-dev/c/hPhkXhrl5-Y/m/Eg6VwKiCAQAJ) and consensus appears to be to set the claimable balances clawback enabled state based on the account flags in the case of the issuer account.

### DIFF
--- a/core/cap-0035.md
+++ b/core/cap-0035.md
@@ -71,9 +71,10 @@ trustline is created to authorize a `ClawbackOp` operation submitted by the
 issuer account.
 
 A claimable balance inherits its clawback enabled status from the account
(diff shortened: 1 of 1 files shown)


tomerweller (approved):
lgtm

PR #858, opened 2021-02-02
SEP-8: add next step for action_required status

howardtw:
**What**

This PR adds the next step for the client to take after there is no more action to be done.

**Why**

It is unclear about what to do and what to expect when a user completes all the action required by the approval service.

### DIFF
--- a/ecosystem/sep-0008.md
+++ b/ecosystem/sep-0008.md
@@ -254,6 +254,8 @@ If the `action_method` field is provided, the client uses the method specified w
 
 If the `action_fields` field is provided, the client may optionally supply the fields if they already have the information so as to reduce the user's need to re-enter the information. If the requested fields contain sensitive or personal information, the server should use `action_method` `POST` so that sensitive information is not transmitted in a URL.
 
+When no further action is to be taken by the user, the client can re-submit the transaction to the approval server. However, there is no guarantee to receive a `"success"` status in the following call. Clients should expect to receive a response with any of the status.
+
 There are two possible values for `action_method`.
 
 - In the case that `action_method` is `GET`, or not defined:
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0008.md
```diff
@@ -305,6 +306,8 @@ There are two possible values for `action_method`.
    }
    ```
 
+When no further actions can be taken by the user, the client can re-submit the transaction to the approval server. However, there is no guarantee to receive a `"success"` status in the following call. Clients should expect to receive a response with any of the status.
```
leighmcculloch, line 309: This statement looks like it relates only to the POST case, and should be indented with the POST details.

I also think you should move this paragraph up to line 267 and put it before the parameter lists and examples, because that's where the discussion about how to use it is and this is part of that.

```suggestion
  When no further actions can be taken by the user, the client can re-submit the transaction to the approval server. However, there is no guarantee to receive a `"success"` status in the following call. Clients should expect to receive a response with any of the status.
```

leighmcculloch (approved):
Looks good, but needs indenting and moving upwards.

Version change looks good.

ecosystem/sep-0008.md
howardtw: It's actually meant to be used by both GET and POST. That's why I didn't have the indentation there.

I thought adding it after `Following the Action URL` makes more sense, but it's indeed not so clear which case it applies. Any suggestion?

howardtw: How about another sub-title `After Completing the Action(s) Required`?

leighmcculloch: Ah gotcha. Can you move it to line 255? I think it belongs there. 
```suggestion
When no further action is to be taken by the user, the client can re-submit the transaction to the approval server. However, there is no guarantee to receive a `"success"` status in the following call. Clients should expect to receive a response with any of the status.
```

howardtw: https://github.com/stellar/stellar-protocol/pull/858/commits/174d7fb5b7c735a807303c7b916f9fb25479d0c6

Issue #775, opened 2020-12-03
SEP-6: match SEP-24's kyc_verified functionality


JakeUrban:
Match the functionality provided in the [SEP-24 PR](https://github.com/stellar/stellar-protocol/pull/739)

PR #860, opened 2021-02-03
SEP-6: added on_change_callback parameter to /withdraw and /deposit requests

JakeUrban:
resolves stellar/stellar-protocol#775

Adds the `on_change_callback` request parameter to `/deposit` and `/withdraw` requests to match the functionality added SEP-24 ([PR](https://github.com/stellar/stellar-protocol/pull/739)).

The `kyc_verified` attribute is not relevant in SEP-6 because requests to `/deposit` and `/withdraw` return 400 until KYC has been verified. Only after KYC is verified do `/deposit` and `/withdraw` requests succeed and records are created and retrievable from `GET /transaction(s)`.

One thing that is not clear is if the anchor should return 400 if `on_change_callback` is not supported. I think it should, but its more likely that anchors would simply ignore the parameter (especially since it's a new addition and they don't know to check for it)

### DIFF
--- a/ecosystem/sep-0006.md
+++ b/ecosystem/sep-0006.md
@@ -140,6 +140,7 @@ Name | Type | Description
 `wallet_name` | string | (optional) In communications / pages about the deposit, anchor should display the wallet name to the user to explain where funds are going.
 `wallet_url` | string | (optional) Anchor should link to this when notifying the user that the transaction has completed.
 `lang` | string | (optional) Defaults to `en`. Language code specified using [ISO 639-1](https://en.wikipedia.org/wiki/ISO_639-1). `error` fields in the response should be in this language.
+`on_change_callback` | string | (optional) A URL that the anchor should `POST` a JSON message to when the `status` or properties change. The JSON message should be identical to the response format for the [/transaction](#single-historical-transaction) endpoint.
 
 Example:
 
@@ -255,6 +256,7 @@ Name | Type | Description
 `wallet_name` | string | (optional) In communications / pages about the withdrawal, anchor should display the wallet name to the user to explain where funds are coming from.
 `wallet_url` | string | (optional) Anchor can show this to the user when referencing the wallet involved in the withdrawal (ex. in the anchor's transaction history).
 `lang` | string | (optional) Defaults to `en`. Language code specified using [ISO 639-1](https://en.wikipedia.org/wiki/ISO_639-1). `error` fields in the response should be in this language.
+`on_change_callback` | string | (optional) A URL that the anchor should `POST` a JSON message to when the `status` or properties change. The JSON message should be identical to the response format for the [/transaction](#single-historical-transaction) endpoint.
 
 Example:
 
--- a/ecosystem/sep-0006.md
+++ b/ecosystem/sep-0006.md
@@ -357,7 +357,9 @@ Example:
 
 #### Transaction Polling
 
-Once the deposit or withdraw request has been made and KYC information has been passed to the anchor via [SEP-12](sep-0012.md), the wallet will need to detect that the anchor has completed the flow by sending funds to the user. The wallet needs to poll the [Transaction History](#transaction-history) endpoint response from the anchor and match the transaction objects returned using the `memo` and `memo_type` parameters used in the deposit or withdraw success response.
+Once the deposit or withdraw request has been made and KYC information has been passed to the anchor via [SEP-12](sep-0012.md), the wallet will need to detect that the anchor has completed the flow by sending funds to the user. 
+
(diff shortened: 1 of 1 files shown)


morleyzhi (approved):

ecosystem/sep-0006.md
```diff
@@ -6,8 +6,8 @@ Title: Anchor/Client interoperability
 Author: SDF
 Status: Active (Interactive components are deprecated in favor of SEP-24)
 Created: 2017-10-30
-Updated: 2021-01-19
-Version 3.1.2
+Updated: 2021-02-03
+Version 3.2.2
```
leighmcculloch, line 10: ❗ The patch version should reset when the minor version changes.
```suggestion
Version 3.2.0
```

ecosystem/sep-0006.md
```diff
@@ -355,7 +357,9 @@ Example:
 
 #### Transaction Polling
 
-Once the deposit or withdraw request has been made and KYC information has been passed to the anchor via [SEP-12](sep-0012.md), the wallet will need to detect that the anchor has completed the flow by sending funds to the user. The wallet needs to poll the [Transaction History](#transaction-history) endpoint response from the anchor and match the transaction objects returned using the `memo` and `memo_type` parameters used in the deposit or withdraw success response.
+Once the deposit or withdraw request has been made and KYC information has been passed to the anchor via [SEP-12](sep-0012.md), the wallet will need to detect that the anchor has completed the flow by sending funds to the user.
```
leighmcculloch, line 360: >the wallet will need to detect that the anchor has completed the flow by sending funds to the user. 

❓ This says that the wallet will detect by sending funds to the user. Is the intention that the anchor sends funds, and the wallet detects that has occurred? If so I think it should be reworded.
```suggestion
Once the deposit or withdraw request has been made and KYC information has been passed to the anchor via [SEP-12](sep-0012.md), the wallet will need to detect that the anchor has completed the flow. The wallet can detect the anchor has completed the flow by checking if the users account has received the funds. 
```

I'm not sure how this sentence interacts with the paragraph below too, so I might be completely misunderstanding what is happening here.

ecosystem/sep-0006.md
```diff
@@ -355,7 +357,9 @@ Example:
 
 #### Transaction Polling
 
-Once the deposit or withdraw request has been made and KYC information has been passed to the anchor via [SEP-12](sep-0012.md), the wallet will need to detect that the anchor has completed the flow by sending funds to the user. The wallet needs to poll the [Transaction History](#transaction-history) endpoint response from the anchor and match the transaction objects returned using the `memo` and `memo_type` parameters used in the deposit or withdraw success response.
+Once the deposit or withdraw request has been made and KYC information has been passed to the anchor via [SEP-12](sep-0012.md), the wallet will need to detect that the anchor has completed the flow by sending funds to the user. 
+
+The wallet either needs to provide the anchor with an `on_change_callback` or to poll the [Transaction History](#transaction-history) endpoint response from the anchor and match the transaction objects returned using the `memo` and `memo_type` parameters used in the deposit or withdraw success response.
```
leighmcculloch, line 362: ❔  Is there no way for the client to look at the specific transaction and its status?

leighmcculloch (commented):
One thing needs fixing (❗) and I have a question (❓) that might change it a bit. I also dropped a question (❔) that probably has no bearing on this change.

ecosystem/sep-0006.md
JakeUrban: There is. The client can pass the `on_change_callback` and get updates on that specific transaction, or the client can poll `GET /transaction?id=`.  I think you confusion is understandable though, because we mention the `GET /transactions` (plural) endpoint. Instead of `GET /transaction?id=`.

ecosystem/sep-0006.md
JakeUrban: I agree we could reword this, but I think both the current and suggested phrasing only outline the deposit case. I figured the language here was outside of the scope of this PR, but I can reword it here.

ecosystem/sep-0006.md
JakeUrban: Oh I see why. Its because the success response for `/deposit` and `/withdraw` do not include the transaction ID. So there's no way for the client to query the transaction using `GET /transaction?id=` until it matches the `memo`

JakeUrban: I'll make a separate PR to add the `id` attribute to the response. I don't know why it isn't there right now. It really should be required in the response, but I'll have to make it optional in order to be non-breaking.

ecosystem/sep-0006.md
JakeUrban: I've rephrased the language to be more clear. Let me know if you think what I have works

ecosystem/sep-0006.md
JakeUrban: stellar/stellar-protocol#860

leighmcculloch (approved):

marcelosalloum (approved):
LGTM!

PR #863, opened 2021-02-04
SEP-24: anchors or users should close interactive popups

### LINKED ISSUE #862: SEP-24: Users should indicate when interactive flow window should be closed
SEP-24 currently instructs wallets to close the interactive popup when the anchor makes a callback request. However, the anchor is completely capable of closing their own popup, or instructing the user to do so manually.

If anchors close their own popups when the user has indicated to do so, wallets can stop listening for callback requests, simplifying their implementation.

JakeUrban:
resolves #862

Adds a new section outlines how the interactive flow popups should be closed. Previously it was implied that the wallet would close interactive flows, but this is unnecessary and having the anchor or user close the popup reduced complexity and provides a better UX.

Once merged, we'll announce the change to the ecosystem and set a date on which we expect anchors to update by.

@pitsevich
@dmgmyza

### DIFF
--- a/ecosystem/sep-0024.md
+++ b/ecosystem/sep-0024.md
@@ -7,7 +7,7 @@ Author: SDF
 Status: Active
 Created: 2019-09-18
 Updated: 2021-01-19
-Version 1.2.2
+Version 1.3.0
 ```
 
 ## Simple Summary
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0024.md
```diff
@@ -7,7 +7,7 @@ Author: SDF
 Status: Active
 Created: 2019-09-18
 Updated: 2021-01-19
-Version 1.2.2
+Version 1.3.0
```
leighmcculloch, line 10: ❓ I'm not sure what version this should be. It's not a patch because it's not simply fixing a bug. It's not really new functionality that stands alone, it is a change to existing functionality that is sort of breaking.

What happens if an anchor does not fulfill its responsibilities to close the window or provide a way for the user to close it? Can client integrations become stuck?

leighmcculloch (commented):
This looks good, and is clear. 👏🏻 

I'm unclear if this change is non-breaking and have a question (❓) about that.

ecosystem/sep-0024.md
JakeUrban: I believe it's non-breaking because the user can always exit the interactive flow, both browser popups and mobile webviews have close buttons regardless of the content served in them.

leighmcculloch (approved):

Issue #832, opened 2021-01-16
SEP-0012: Format of PUT with SEP-9 fields is not defined


leighmcculloch:
[SEP-12] does not specify, as far as I can tell, how to format the [SEP-9] fields in the Customer PUT (`PUT [KYC_SERVER || TRANSFER_SERVER]/customer`) request.

This can probably be guessed accurately for simple SEP-9 fields such as `email_address`:

```json
{
  "email_address": "user@example.com"
}
```

But it is harder to guess accurately for complex SEP-9 fields that have sub-fields, such as `organization.name`.

For example, should the field be provided using the full field name as the key:
```json
{
  "organization.name": "SDF"
}
```

Or, should the field be provided with sub-fields as sub-keys to an object:
```json
{
  "organization": {
    "name": "SDF"
  }
}
```

I don't see this documented in SEP-12 but I assume people have implemented use of these fields and so there must be a common approach.

I'm asking this question because I'm modeling how SEP-9 fields are provided to a SEP-12 server for new functionality we are adding to SEP-8 in https://github.com/stellar/stellar-protocol/pull/758 and https://github.com/DSTOQ/stellar-protocol/pull/3/files#r558710763.

[SEP-12]: https://stellar.org/protocol/sep-12
[SEP-9]: https://stellar.org/protocol/sep-9

accordeiro:
I'm not sure what is most common implementation, but I've always interpreted this spec as using nested fields, as in:

```javascript 
payload = {
  "organization": {
    "name": "SDF"
  }
}
```

I think this is the approach that leads to less confusion to whomever is writing the parsing code (e.g. a server-side SEP-12 implementation), since some languages (like JavaScript) use `.`s to traverse the JSON structure, so you'd actually have to work around the syntax to be able to parse an attribute called `"organization.name"`. This [SO question](https://stackoverflow.com/questions/2577172/how-to-get-json-objects-value-if-its-name-contains-dots) should provide further context on the issue I'm talking about.

leighmcculloch:
@istrau2 @marcinx @JakeUrban I see you all have committed to [SEP-12](https://stellar.org/protocol/sep-12) in the past. Do you know the answer to how sub-fields should be rendered in the Customer PUT request?

JakeUrban:
[...]

PR #867, opened 2021-02-10
SEP-9: Clarify organization dot notation

JakeUrban:
resolves stellar/stellar-protocol#832

Adds a clarification to the organization fields section on how field names should be specified.

### DIFF
--- a/ecosystem/sep-0009.md
+++ b/ecosystem/sep-0009.md
@@ -28,6 +28,27 @@ date | 10 | [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) date-only format
 phone number | varies | [E.164](https://en.wikipedia.org/wiki/E.164)
 occupation | 3 | [ISCO08](https://en.wikipedia.org/wiki/International_Standard_Classification_of_Occupations)
 
+## Field Naming Conventions
+
+Where possible we use field names from [schema.org](https://schema.org/Person). Words are separated with underlines as that convention has previously been established in Stellar protocols.
+
+### Dot Notation
+
+Field names should always be used as strings. For example:
+```json
+{
+	"organization.name": "Stellar Development Foundation"
+}
+```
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0009.md
```diff
@@ -70,6 +70,14 @@ Name | Type | Description
...
 ### Organization KYC fields
 
+These field names should always be used as strings. For example:
+```json
+{
+	"organization.name": "Stellar Development Foundation"
+}
+```
+The dot notation (`organization.<field>`) is _not_ an indication that the fields described should be nested under a top-level `"organization"` object.
+
```
leighmcculloch, line 80: 💡 I think this could be a new section that isn't tied to the organization fields. e.g. `## Sub-fields and dot-notation`

leighmcculloch (approved):
I think this change is good, simple, and portable. As @JakeUrban pointed out in https://github.com/stellar/stellar-protocol/issues/832#issuecomment-763216845 this is bringing SEP-9 into consistency with the reference implementation, which is Polaris. 

I think this change is worthy of 👍🏻s from an additional person, such as @tomerweller or @accordeiro.

❗ Before merging could you update the meta information with a patch version increment, and set the updated date?

ecosystem/sep-0009.md
```diff
@@ -28,6 +28,27 @@ date | 10 | [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) date-only format
...
+}
+```
+The dot notation is _not_ an indication that the fields described should be contained in a nested object under a top-level key. For example, the following would not be compliant:
+```json
+{
+	"organization": {
+		"name": "Stellar Development Foundation"
+	}
+}
+```
```
leighmcculloch, line 50: 💡 I think stating what it shouldn't look like might backfire if someone quickly reading the spec thinks this is an example. But if you think it is necessary to prevent mistakes or assumptions, looks good.

leighmcculloch (approved):
👏🏻

leighmcculloch (approved):

PR #868, opened 2021-02-10
SEP-0010: Fix inconsistencies in flow defined by the token endpoint (v3.1.1)

leighmcculloch:
### What
Make clear in the token endpoint flow that there are multiple operations in the challenge transaction and that the verification steps are focused on the first operation.

### Why
In SEP-10 v3.1 we added a new operation that is included but the token endpoint flow was not updated to reflect that this was the case. The token endpoint flow talks about the transaction having a single operation and verifying that operation which is confusing.

This change makes no functional changes and updates the token endpoint section to align with the rest of the document.


JakeUrban (approved):
Thanks! I made these changes in the client attribution PR so now I can rebase on master once this is merged.

PR #869, opened 2021-02-10
SEP-6: Added optional 'amount' & 'country_code' parameters

JakeUrban:
Originally discussed in stellar/stellar-protocol#808 

Adds two optional parameters to both `/deposit` and `/withdraw`: `amount` and `country_code`. These fields may be necessary for the anchor to know what SEP-9 `fields` to return in the response.

One change I didn't make, is adding a `sep12_type` parameter that is associated with the `fields` list returned. We could add this attribute to enable the client to make `GET /customer?type=` requests. However, specifying both `fields` and `sep12_type` is redundant, so `fields` may need to be deprecated if we wanted to add `sep12_type`.

### DIFF
--- a/ecosystem/sep-0006.md
+++ b/ecosystem/sep-0006.md
@@ -453,7 +453,8 @@ The response should be a JSON object like:
           "description": "amount in USD that you plan to deposit"
         },
         "country_code": {
-          "description": "The ISO 3166-1 alpha-3 code of the user's current address"
+          "description": "The ISO 3166-1 alpha-3 code of the user's current address",
+          "choices": ["USA", "PRI"]
         },
         "type" : {
           "description": "type of deposit to make",
@@ -483,7 +484,10 @@ The response should be a JSON object like:
               "dest_extra": { "description": "your routing number" },
               "bank_branch": { "description": "address of your bank branch" },
               "phone_number": { "description": "your phone number in case there's an issue" },
-              "country_code": { "description": "The ISO 3166-1 alpha-3 code of the user's current address"}
+              "country_code": { 
+                "description": "The ISO 3166-1 alpha-3 code of the user's current address",
+                "choices": ["USA", "PRI"]
+              }
           }
         },
         "cash": {
--- a/ecosystem/sep-0006.md
+++ b/ecosystem/sep-0006.md
@@ -6,8 +6,8 @@ Title: Anchor/Client interoperability
 Author: SDF
 Status: Active (Interactive components are deprecated in favor of SEP-24)
 Created: 2017-10-30
-Updated: 2021-02-03
-Version 3.2.0
+Updated: 2021-02-10
+Version 3.3.0
 ```
 
 ## Simple Summary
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0006.md
```diff
@@ -475,7 +482,8 @@ The response should be a JSON object like:
               "dest": {"description": "your bank account number" },
               "dest_extra": { "description": "your routing number" },
               "bank_branch": { "description": "address of your bank branch" },
-              "phone_number": { "description": "your phone number in case there's an issue" }
+              "phone_number": { "description": "your phone number in case there's an issue" },
+              "country_code": { "description": "The ISO 3166-1 alpha-3 code of the user's current address"}
```
leighmcculloch, line 486: ❓ Is this the country code of the users address or the bank?

ecosystem/sep-0006.md
```diff
@@ -257,6 +259,8 @@ Name | Type | Description
 `wallet_url` | string | (optional) Anchor can show this to the user when referencing the wallet involved in the withdrawal (ex. in the anchor's transaction history).
 `lang` | string | (optional) Defaults to `en`. Language code specified using [ISO 639-1](https://en.wikipedia.org/wiki/ISO_639-1). `error` fields in the response should be in this language.
 `on_change_callback` | string | (optional) A URL that the anchor should `POST` a JSON message to when the `status` property of the transaction created as a result of this request changes. The JSON message should be identical to the response format for the [/transaction](#single-historical-transaction) endpoint.
+`amount` | string | (optional) The amount of the asset the user would like to deposit with the anchor. This field may be necessary for the anchor to determine what KYC information is necessary to collect.
+`country_code` | string | (optional) The [ISO 3166-1 alpha-3](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-3) code of the user's current address. This field may be necessary for the anchor to determine what KYC information is necessary to collect.
```
leighmcculloch, line 263: ❓ What happens if a country is provided that the anchor is not allowed to accept deposits from?

leighmcculloch (approved):
This looks like a safe and good change to make. KYC can be contingent on these things. Making them optional is good for backwards compatibility, and also shouldn't break anything if not provided because a server can always respond assuming a large amount. The country is interesting because it might result in an error stating that deposit is unsupported for individuals in that country.

Questions (❓) below, but otherwise LGTM.

ecosystem/sep-0006.md
JakeUrban: In that case there is no way to help the user, the service is simply unavailable for them and the response includes a `400 Bad Request` or more specifically `422 Unprocessable Entity`.

ecosystem/sep-0006.md
JakeUrban: User, as defined in the `description` and request parameter list.

ecosystem/sep-0006.md
JakeUrban: The anchor could provide a `choices` array too, which would help clarify whats supported. I'll add that to the example.

ecosystem/sep-0006.md
leighmcculloch: Got it. Why does it appear under the bank account field for withdrawal but not for deposit?

JakeUrban: Thats just the way `/info` is defined, assets under the `withdraw` section specify a `types` key, and each defined type has it's own set of `fields`. Assets under the `deposit` section just have a single `fields` object.

PR #870, opened 2021-02-17
SEP-24: deprecated postMessage callbacks

JakeUrban:
Deprecates the use of `postMessage` callbacks.

Question: Do we want to deprecate the `callback` parameter altogether? `callback` is used for when the interactive flow has completed. But now that ~it is the anchor's responsibility to close the popup~ the user must close the popup, I can't think of a reason why `callback` is valuable. `on_change_callback` can be used for all status changes, so the wallet could get an update when the transaction moves to `pending_user_transfer_start`.

### DIFF
--- a/ecosystem/sep-0024.md
+++ b/ecosystem/sep-0024.md
@@ -396,7 +396,11 @@ The URL supplied by both callback parameters should receive the full transaction
 
 If the wallet wants to be notified that the user has completed the anchor's interactive flow (either success or failure), it can add this parameter to the URL. If the user abandons the process, the anchor does not need to report anything to the wallet. If the `callback` value is a URL, the anchor must `POST` to it with a JSON message as the body. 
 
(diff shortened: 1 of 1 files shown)


tomerweller (changes requested):
I think we need to add a more explicit deprecation notice that instructs wallets and anchors what to do. For example: if I'm an anchor receiving a `postMessage` callback parameter - what's my expected behavior? If I'm a wallet - what's my alternative?

tomerweller:
I agree that the existence of both `callback` and `on_change_callback` looks pretty vestigial. @JakeUrban can you open a different issue for that? No need to overload this pr

JakeUrban:
> @JakeUrban can you open a different issue for that?

I'll can open the PR when this is merged 👍

tomerweller (commented):
lgtm

tomerweller (approved):

PR #873, opened 2021-02-23
SEP-24: users must close the popups, anchors cannot

JakeUrban:
SEP-24 was [recently updated](https://github.com/stellar/stellar-protocol/pull/863) to instruct anchors to close the popup instead of the wallet. However, modern browsers only allow the opener of the popup to close it via javascript's `window.close()`, meaning anchors cannot close their own popups. See the [window.close()](https://developer.mozilla.org/en-US/docs/Web/API/Window/close#closing_the_current_window) spec.

Therefore SEP-24 needs to be updated so that it no longer implies the anchor is able to close the popup.

### DIFF
--- a/ecosystem/sep-0024.md
+++ b/ecosystem/sep-0024.md
@@ -426,13 +426,13 @@ As an alternative to using the `callback` or `on_change_callback` parameters, th
 
 #### Guidance for anchors: closing the interactive popup
 
-After the user has provided all necessary information to the anchor through the interactive popup (`url`), the anchor should either close the popup automatically or instruct the user to close it themselves. If the anchor needs to provide instructions to the user for sending funds to it's off-chain account (in the deposit case), the anchor should not close the popup until the user has indicated to do so.
+After the user has provided all necessary information to the anchor through the interactive popup (`url`), the anchor should instruct the user to close the popup. This is necessary for deposit flows because the user must read and understand the instructions provided by the anchor for sending deposit funds. While this is not the case for withdrawal flows, anchors cannot assume the wallet will close the popup in a timely manner after updating the transaction's status.
 
-When the popup is closed, the transaction's status should be `pending_user_transfer_start`, signaling to the wallet that the anchor is waiting to receive the funds.
+When the popup is closed, the transaction's status should have transitioned from `incomplete` to `pending_user_transfer_start`, signaling to the wallet that the anchor is waiting to receive the funds.
 
 #### Guidance for wallets: completing an interactive withdrawal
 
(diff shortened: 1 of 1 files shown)


JakeUrban:
cc @leighmcculloch

ecosystem/sep-0024.md
```diff
@@ -426,13 +426,13 @@ As an alternative to using the `callback` or `on_change_callback` parameters, th
 
 #### Guidance for anchors: closing the interactive popup
 
-After the user has provided all necessary information to the anchor through the interactive popup (`url`), the anchor should either close the popup automatically or instruct the user to close it themselves. If the anchor needs to provide instructions to the user for sending funds to it's off-chain account (in the deposit case), the anchor should not close the popup until the user has indicated to do so.
+After the user has provided all necessary information to the anchor through the interactive popup (`url`), the anchor should instruct the user to close the popup. This is necessary for deposit flows because the user must read and understand the instructions provided by the anchor for sending deposit funds. While this is not the case for withdrawal flows, anchors cannot assume the wallet will close the popup in a timely manner after updating the transaction's status.
```
leighmcculloch, line 429: 💡 I think we should write this to be more specific. It occurs to me reading this that it is in a user's best interest for the wallet not to try to close the interactive flow, even if the wallet has received some event that indicates the flow is complete. As you point out the deposit flow will display meaningful information to the user. In a similar way it could be jarring for the user for the withdrawal flow to disappear on its own.

```suggestion
After the user has provided all necessary information to the anchor through the interactive popup (`url`), the anchor should instruct the user to close the popup. The anchor or wallet should not close the interactive popup if they detect the payment is complete, so that the user has ample time to read the final deposit instructions, or in the case of withdrawals, has ample time to read any information the anchor displays at the end of the withdrawal flow.
```

I know we've discussed recently how closing automatically could be a better UX, but I also think that could be a more confusing UX. Since we already need no auto-close on deposit it is simpler to keep both deposit and withdrawal the same, and I suspect it will be less jarring as well.

Thoughts?

leighmcculloch (approved):
Missing meta updates, otherwise one suggestion (💡) below. If you feel strongly against it feel free to merge without changing, but also happy to discuss more.

ecosystem/sep-0024.md
JakeUrban: For sure, auto-closing (i.e. not initiated by user action) is bad UX in any case.

However, prior to deprecating the `postMessage` callback, anchors could provide a 'Close' button that would trigger a postMessage call when clicked by the user, and the wallet would consequently close the window. This isn't jarring because the action is user-initiated, and could be used for both deposit and withdraw flows. 

This is the ideal UX, but it depends on anchors being able to make a postMessage call. Depending on the results of our internal discussions, we may end up un-deprecating postMessage to enable this UX.

If we decide to continue on this route (deprecating postMessage) then I agree with the change is phrasing here 👍

YoutacRSVARs (approved):

PR #874, opened 2021-02-23
SEP-26: reject in favor of SEP-6

### LINKED ISSUE #793: SEP-26: Reject SEP-26 because SEP-6 has become de-facto non-interactive standard
SEP-26 was originally a replacement for SEP-6 back when we had SEP-6 deprecated. Now that SEP-6 is back in active development, I don't see any reason to keep SEP-26.

JakeUrban:
resolves #793

SEP-6 is already used in the wild and it would be a burden on client applications to support SEP-26 as well. SEP-6 is still in _Active_ status and can be improved upon if desired.


leighmcculloch (approved):

leighmcculloch:
Do we know of any SEP-26 implementations that need to weigh in on this? If no, I think this is 👍🏻 .

JakeUrban:
Confirmed via @tomerweller SEP-26 never got past the exploration phrase. Merging.

PR #875, opened 2021-02-24
SEP-24: deprecate 'callback' in favor of 'on_change_callback'

JakeUrban:
Originally discussed in #870

The original `callback` parameter has become redundant with the introduction of `on_change_callback`. 

Clients who specify an `on_change_callback` can expect to receive an updated transaction body on each change to the transaction's status, including the initial transition away from `incomplete`. This means `callback`, which is used when the interactive flow completes (i.e. when the status changes from `incomplete`) is no longer necessary.

**Side note:** I'd like to confirm whether or not `incomplete` _always_ transitions to `pending_user_transfer_start` (if not `error`). I can see a case where `incomplete` transitions to `pending_anchor` if the KYC information needs to be manually verified -- in which case the SEP should be updated to clarify this. Currently, the SEP states `incomplete` _must_ transition to `pending_user_transfer_start`.

### DIFF
--- a/ecosystem/sep-0024.md
+++ b/ecosystem/sep-0024.md
@@ -387,18 +387,20 @@ The basic parameters are summarized in the table below.
 
 Name | Type | Description
 -----|------|------------
-`callback` | string | (optional) A URL that the anchor should `POST` a JSON message to when the user successfully completes the interactive flow. The use of `postMessage` is **deprecated**.
+`callback` | string | (**deprecated**, optional) A URL that the anchor should `POST` a JSON message to when the user successfully completes the interactive flow. The use of `postMessage` is **deprecated**.
 `on_change_callback` | string | (optional) A URL that the anchor should `POST` a JSON message to when the `status` or `kyc_verified` properties change. The use of `postMessage` is **deprecated**.
 
 The URL supplied by both callback parameters should receive the full transaction object.
 
 **`callback` details**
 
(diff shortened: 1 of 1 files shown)


JakeUrban:
@dmgmyza is my understanding of the following correct? Do you use `on_change_callback` today?

tomerweller:
What's the price of making this change? In other words, are there any backward compatibility issues that might arise?

Is there a functional reason to remove this callback or is it just to tidy up the API?

JakeUrban:
I'm sure most anchors do not yet support `on_change_callback`, which was added in early December, so wallets will still have to use `callback` if they have functionality that relies on it. This change then won't have much immediate effect. It serves as a strong to signal to anchors to add support for `on_change_callback` though, and lays the groundwork for `callback`'s eventual removal.

dmgmyza:
@JakeUrban 

Hi. Everything looks good to me.

Just a small remark:
We don’t see a big advantage of ```on_change_callback``` over the usual polling for now, so we’ve implemented polling system.  One of the benefits of getting transaction on client side (in case of polling) and not on the server (```on_change_callback```) - is the easy way to debug it on a client side in a quick way and performance should stay the same in both cases.
But it’s clear that if we would like to move to ```on_change_callback```, that should work well, PR is good for both cases.
Btw, do you see any advantages of ```using on_change_callback``` that we probably missing?

Regarding your side note:
As you know, it’s important for us to force transaction creation with ```incomplete``` status. And I think that would be good to highlight that transaction status might transition from incomplete to ```pending_anchor``` status. That works even for our current implementation.

JakeUrban:
@dmgmyza the one advantage is that you can get updates on transaction after the interactive flow is complete, whereas I assume you stop polling on the client side once the status transitions away from `incomplete`. With `on_change_callback`, you could send notifications to users on deposit/withdraw success/error.

PR #876, opened 2021-02-25
SEP-24: clarify 'incomplete' status transition

JakeUrban:
Originally discussed in #875 

The SEP was recently updated to re-enforce the expectation that transaction begin in the `incomplete` status. However in that change we also incorrectly made the assumption that `pending_user_transfer_start` would always be the next status.

In fact, there are cases where other status are used directly after `incomplete`, such as if the anchor needs additional time to verify KYC information provided before accepting funds.

### DIFF
--- a/ecosystem/sep-0024.md
+++ b/ecosystem/sep-0024.md
@@ -6,8 +6,8 @@ Title: Interactive Anchor/Wallet Asset Transfer Server
 Author: SDF
 Status: Active
 Created: 2019-09-18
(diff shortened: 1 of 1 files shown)


JakeUrban:
@dmgmyza thanks for clarifying, CC'ing you here

tomerweller (dismissed):
lgtm

tomerweller (approved):
lgtm

PR #879, opened 2021-03-04
SEP-0008: Fix inconsistency by updating snippet JSON field from "message" to "error"

marcelosalloum:
### What

Fix inconsistency by updating snippet JSON field from `message` to `error`.

### Why

While the table recommended using `error` the JSON snippet had the `message` field instead. 
Also, it was already called `error` in previous protocol versions.


howardtw (approved):

leighmcculloch (approved):
Thanks for identifying this mistake in the example and for fixing it. 🎉

PR #881, opened 2021-03-04
add updated field to cap21

tomerweller:


core/cap-0021.md
```diff
@@ -6,6 +6,7 @@ Title: Generalized transaction preconditions
 Author: David Mazières
 Status: Draft
 Created: 2019-05-24
+Updated: 2021-03-04
```
leighmcculloch, line 9: CAPs do not usually have Updated fields.

PR #883, opened 2021-03-05
SEP-0001: Update toml-check URL and remove deprecated AUTH_SERVER from example

sui77:
Update toml-check URL
Removed deprecated AUTH_SERVER from example


leighmcculloch:
Thanks @sui77!

PR #885, opened 2021-03-08
Change the stale issue comment message.

leighmcculloch:
### What
Change the message that is posted as a comment when issues are identified as stale to remove the statement about leaving a comment.

### Why
I've noticed a pattern that when we want to keep issues alive we post brief and mostly meaningless messages to issues like "Not stale". We should recommend folks just remove the stale label so we don't add unnecessary comments to peoples inboxes and to the issue history.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -14,8 +14,8 @@ jobs:
           debug-only: false
           days-before-stale: 30
           days-before-close: 30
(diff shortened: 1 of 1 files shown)


ire-and-curses:
I'm conflicted on this.

* The existing message correctly describes the implementation
* Some people use an email workflow so replying is the fastest way to action

On the other hand the messages are annoying clutter. I guess I'm open to giving this a try.

ire-and-curses (approved):

leighmcculloch:
I'm conflicted for the same reasons, but I think worth still trying.

Issue #861, opened 2021-02-03
SEP-6: Add transaction ID's to /deposit and /withdraw success responses


JakeUrban:
Right now, the client must figure out which object returned from `GET /transactions` represents a particular deposit or withdraw request  by matching the it's listed `memo` to the `memo` request parameter for deposit or `memo` response attribute for withdraw. This is not ideal and overly complex.

The `/deposit` and `/withdraw` success responses should include the ID of the transaction created as a result of the request.

Originally discussed in #860

PR #886, opened 2021-03-08
SEP-6: added 'id' response attributes for /deposit & /withdraw

JakeUrban:
resolves stellar/stellar-protocol#861

Adds optional `id` attributes to the `/deposit` & `/withdraw` success responses. This removes need for clients to infer which `/transactions` record is associated with a particular request.

### DIFF
--- a/ecosystem/sep-0006.md
+++ b/ecosystem/sep-0006.md
@@ -6,8 +6,8 @@ Title: Anchor/Client interoperability
 Author: SDF
 Status: Active (Interactive components are deprecated in favor of SEP-24)
 Created: 2017-10-30
-Updated: 2021-02-10
-Version 3.3.0
+Updated: 2021-03-08
+Version 3.4.0
 ```
 
 ## Simple Summary
@@ -168,6 +168,7 @@ The response body should be a JSON object with the following fields:
 Name | Type | Description
 -----|------|------------
 `how` | string | Terse but complete instructions for how to deposit the asset. In the case of most cryptocurrencies it is just an address to which the deposit should be sent.
+`id` | string | The anchor's internal ID for this deposit request. The wallet will use this ID to query the [`/transaction`](#single-historical-transaction) endpoint to check status of the request.
 `eta` | int | (optional) Estimate of how long the deposit will take to credit in seconds.
 `min_amount` | float | (optional) Minimum amount of an asset that a user can deposit.
 `max_amount` | float | (optional) Maximum amount of asset that a user can deposit.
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0006.md
```diff
@@ -168,6 +168,7 @@ The response body should be a JSON object with the following fields:
 Name | Type | Description
 -----|------|------------
 `how` | string | Terse but complete instructions for how to deposit the asset. In the case of most cryptocurrencies it is just an address to which the deposit should be sent.
+`id` | string | (optional) The anchor's internal ID for this deposit request. The wallet will use this ID to query the [`/transaction`](#single-historical-transaction) endpoint to check status of the request.
```
leighmcculloch, line 171: 💡  Internal ID implies to me it is an ID internal to the anchor that would not be exposed publicly, but in this case it is being exposed to another party. Stating that it is the anchor's ID for the deposit is sufficient.
```suggestion
`id` | string | (optional) The anchor's ID for this deposit. The wallet will use this ID to query the [`/transaction`](#single-historical-transaction) endpoint to check status of the request.
```

ecosystem/sep-0006.md
```diff
@@ -352,6 +356,7 @@ Name | Type | Description
 `account_id` | `G...` string | The account the user should send its token back to.
 `memo_type` | string | (optional) Type of memo to attach to transaction, one of `text`, `id` or `hash`.
 `memo` | string | (optional) Value of memo to attach to transaction, for `hash` this should be base64-encoded.
+`id` | string | (optional) The anchor's internal ID for this withdraw request. The wallet will use this ID to query the [`/transaction`](#single-historical-transaction) endpoint to check status of the request.
```
leighmcculloch, line 359: :bulb Same as above.
```suggestion
`id` | string | (optional) The anchor's ID for this withdrawal. The wallet will use this ID to query the [`/transaction`](#single-historical-transaction) endpoint to check status of the request.
```

ecosystem/sep-0006.md
```diff
@@ -6,8 +6,8 @@ Title: Anchor/Client interoperability
 Author: SDF
 Status: Active (Interactive components are deprecated in favor of SEP-24)
 Created: 2017-10-30
-Updated: 2021-02-10
-Version 3.3.0
+Updated: 2021-03-08
+Version 3.4.0
```
leighmcculloch, line 10: 👏🏻 Perfect.

leighmcculloch (approved):
One suggestion (💡) to do with how we refer to the anchor's ID, but regardless, looks great.

PR #890, opened 2021-03-14
SEP-0033: add project url of Go implementation.

overcat:
- add project url of [Golang implementation](https://github.com/StellarCN/stellar-identicon-go).

PR #891, opened 2021-03-15
SEP-0033: Set version and make active

leighmcculloch:
### What
Make SEP-33 active and set the version to v1.

### Why
The SEP is being used in several products in the ecosystem, namely stellar.expert, the Lobstr.co wallet, and I think some other places. There are several implementations, and the number of implementations is growing (#890). These are all indicators that this proposal/protocol is more than just a draft at this point. It's actively in use and the status should reflect that.

@pitsevich You're the author of this proposal. Do you agree?


pitsevich:
@leighmcculloch My apologies for missing the notifications.

Sure, let’s move this to active!
Happy to see Go implementation.

pitsevich (approved):

JakeUrban (approved):
I'm unfamiliar with this SEP but the change looks good 👍

PR #892, opened 2021-03-16
CAP-0021: Fix diff

leighmcculloch:
### What
Fix the XDR diff so it can be applied successfully by git.

### Why
The diff is corrupted and git errors when applying on the last line, because the last line has a space at the beginning of it.

Note that an extra line is required in the code block so that the rendered markdown on GitHub includes the necessary last empty line. This is because GitHub gobbles the last empty line in code blocks.


leighmcculloch:
This was fixed in #900.

PR #894, opened 2021-03-23
SEP-23: Improve proposed Horizon API changes

leighmcculloch:
### What
Improve the proposed Horizon API changes to rename the suffix `_id` to `_muxed_id`.

### Why
OrbitLens was talking to me about how the _id_id example highlights how using an _id suffix is not very pretty or specific to this new data we're displaying. _muxed_id would more clearly identify the field apart from others. 

It's possible there are better improvements, but this at least better than what's here.

### DIFF
--- a/ecosystem/sep-0023.md
+++ b/ecosystem/sep-0023.md
@@ -118,7 +118,7 @@ accounts in a backwards-compatible way:
       account.  (For non-multiplexed accounts, this will be the same
       as `account_id`.)
 
-    - Base field name + `_id` is the integer.
+    - Base field name + `_muxer` is the integer.
 
   For example, given the MuxedAccount
   `MAAAAAAAAAAAJURAAB2X52XFQP6FBXLGT6LWOOWMEXWHEWBDVRZ7V5WH34Y22MPFBHUHY`,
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0023.md
```diff
@@ -118,15 +118,15 @@ accounts in a backwards-compatible way:
       account.  (For non-multiplexed accounts, this will be the same
       as `account_id`.)
 
-    - Base field name + `_id` is the integer.
+    - Base field name + `_muxer` is the integer.
```
ire-and-curses, line 121: how about `mid`? I'm not really sure what a `muxer` is.

leighmcculloch: A muxer is a thing that does muxing, which highlights that we shouldn't use that term, since the identifier doesn't do the muxing. 🤔 

MID is okay, but in payments an MID is a merchant ID, and it'd be nice not to overload that.

leighmcculloch: How about `_muxed_address` and `_muxed_id`?

e.g.:
In transaction response (unrelated fields omitted):
```json
{
  "source_account": "GAAZI4TCR3TY5OJHCTJC2A4QSY6CJWJH5IAJTGKIN2ER7LBNVKOCCWN7",
  "source_account_sequence": "1",
  "source_account_muxed_address": "M...",
  "source_account_muxed_id": "1",
}
```

ire-and-curses: That looks fine to me.

ire-and-curses (approved):

leighmcculloch:
Updated to use the suggestion from https://github.com/stellar/stellar-protocol/pull/894#discussion_r599969396.

leighmcculloch:
@ire-and-curses Do you think I need anyone else's approval on this before I can merge this? I think you and your team probably have the strongest opinions about how the API would look like.

@stanford-scs Do you have an opinion on this?

ire-and-curses:
I think it's free to be merged. This is still a recommendation, it seems better, and we can revisit when we get here if needed.

PR #898, opened 2021-03-26
SEP-0008: Add discussion about when to use a revised status and when not to

leighmcculloch:
### What
Add discussion about when to use a revised status and when not to. Specifically distinguish that a revised transaction should never change the intent or semantics of the operations the user provided.

### Why
There was some conversation about this recently within SDF and we realize it is not clear in SEP-8 when to use reject or when to use revised. In general revised shouldn't alter the intent. If the user wants to perform an operation and it can succeed if the server adds operations, that's the perfect use case for revised. If the user wants to perform an operation that is not compliant no matter what operations can be added, it should just reject the transaction rather than try and modify the transaction.

### DIFF
--- a/ecosystem/sep-0008.md
+++ b/ecosystem/sep-0008.md
@@ -6,8 +6,8 @@ Title: Regulated Assets
 Author: Tomer Weller <@tomerweller>, Howard Chen <@howardtw>, Christian Peters (DSTOQ) <@shredding>, Leigh McCulloch <@leighmcculloch>
 Status: Active
 Created: 2018-08-22
-Updated: 2021-03-04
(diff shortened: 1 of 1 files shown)


JakeUrban (approved):
This definitely helps make the use of `revised` more clear. Thanks for adding

leighmcculloch:
@shredding @dmgmyza Does this align with your understanding and make sense to you?

Zagan202 (approved):
LGTM!
Much clearer on the intent.

shredding:
@leighmcculloch yes, it's exactly how we do it

tomerweller (approved):
excellent.

PR #902, opened 2021-04-08
SEP-????: Add a protocol defining data entries for NFTs

leighmcculloch:
Add a protocol defining data entries for NFTs.

See the motivation section within the SEP for why.

This protocol has an example implementation at [makenft.web.app](https://makenft.web.app). An example NFT using the protocol can be viewed [here](https://makenft.web.app/view#testnet:FALL-GA7BZCS4VE7DK2WJ3JGMJOAUQUUDATBIN2W4HLR3KWJSDMSN64SJAMYG).

This protocol is in draft state and is intended to start conversation. If there is little or no desire by the folks experimenting with NFTs on Stellar to standardize on the things that this protocol captures, it can be swiftly moved into a rejected state. Collaboration welcome, and if anyone else wanted to take this and own it I'd be more than happy to hand off.

@tyvdh I'd be interested to know if you think this would be useful or worth discussing, and the best forum for discussing it.


msfeldstein:
LGTM, i like that its staying simple.

tyvdh:
I love this save for my one major use case which is storing NFT data within the issuing account itself. This SEP requires data be stored externally, which may often be the case but certainly isn't and shouldn't be required. Particularly when you have 1,000 manage data attrs available for use on the account itself.


[GCO2YMX3FF4ITMJFSBANPESOIN676GENIWIKRSGK7JBYIZ3H7H4UW25M](https://horizon.stellar.org/accounts/GCO2YMX3FF4ITMJFSBANPESOIN676GENIWIKRSGK7JBYIZ3H7H4UW25M)


[GBVVYEFWUIG7RXX6RBYJOFMGKEEEEN3TVEZYB5J6HNIQ5UON5O4NH23F](https://horizon.stellar.org/accounts/GBVVYEFWUIG7RXX6RBYJOFMGKEEEEN3TVEZYB5J6HNIQ5UON5O4NH23F)


[GBSTM6QKNYQND77XZR3CJN6Y5JALYGNLE5AL5CUB4OYPY2X6C62VPXOI](https://horizon.stellar.org/accounts/GBSTM6QKNYQND77XZR3CJN6Y5JALYGNLE5AL5CUB4OYPY2X6C62VPXOI)

The image data for each one of those NFTs is baked right into the issuing account itself.

Examples for the code used for generating and viewing this method of storing arbitrary NFT data can be seen here:
[Issuing](https://runkit.com/tyvdh/png-to-xdr)
[Viewing](https://runkit.com/tyvdh/sq-badge)

How I'm accomplishing this is by using the key and value entries to store data vs just the value. The key of the manage data attribute is capped at 64 characters or 48 bytes. The first 2 bytes of the key are an indexing value from 0 to 999. The last 46 bytes are the first/next chunk of the data buffer. The value is just a simple slice of 64 bytes of the next chunk of the data buffer. You carry on creating chunks and adding them to the transaction(s) up to 1,000 key:value pairs.

The one item I haven't included yet is a `Content-Type` declaration for the interpreter for decoding the NFT, but that could be another simple data attribute following the syntax described here. 
`nft.asset.content-type: image/png`
We could also include the sha but not sure it's necessary as the data itself is stored on the Stellar chain. This `nft.asset.content-type` would also be the signal for wallets that it's dealing with a long format "embedded" nft and could actually decode it right here on the fly without needing to call or lookup the toml file for an external interpreter url. 

Stellar isn't a good database for arbitrary data but you can and should store some things and in cases where it makes sense this is a really elegant and preferable solution to splitting your asset across multiple chains. I don't think the implementation is overly complex and considering the nature of the SEP I think it would make sense to introduce and formalize how we store long format arbitrary data on chain.

leighmcculloch:
>This SEP requires data be stored externally

The SEP allows the storage of images directly on the issuer account by encoding the image as a base64 image url. This format includes the content type and everything a client needs to know to render it. This isn't as efficient as you the method you described, but it works.

I added an example to the SEP that shows how to do that in c0c9f826d19ca756bb6b7f9c68b2bdfad94a2a91.

As an example of a base64 URL, the base64 URL for the second image you shared (created using https://base64-image.de) is this:
```
data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAEAAAABACAYAAACqaXHeAAAFsElEQVR4nM2bv2sbSRTHtxFoVS1bxotTJUU4SC4CWyBQHHBcCMyBSSMMboK4f+AK3R8QUKE+TVqjVkGN/gG3rlKl2CLFwRXu3M8Vxxue3r43M288u4rhsfqxu57vZ77z3uzsKvvx7ZPpIvr5wEzGczMZz01RVuw+32+n5vvt1Pzz15H5/fJrsnj1/rMY2Y9vn8y/n09bj34+sBD6+WDvu9Fsa0azrcmyzGRZZt/7IgUICwAa2EUAhKKsTFFW1hn9fGDFDZe1ehsD4yAAsPjRbGvFx4h2baMB1PcPnYg/uXts9H5qCLCVgCRzwGa9C9oPGnNy92iGy9p+PlzWUaGFQGEkA1CUVdB+0PAQ4ZdffprLLz+TwqCRDEDMsPEJ0gIIhdEKAI39Q8U+BUAohM6rABVHI1asBoIKQGhy0/Z+VwB+KQdw9u4CgAuCF8BqcdtZ77cJQILgnQhhAL5M7/o+pPe7dsBwWad1ALcvHm+Hsr/aAdoA4ZAwKQip9HEBM0S6bQuCF4BmguNyQEjj8HFwfQBbTYLVQEhaBWIB4P3h8rgoK3Nzvdp7D1su+vnAridAcGDUDnhK+MRj0TRurldmMp43IIREr5fvAaEgWnNAKASX8LdvrmxgCHhLQwLU6+UWguSE1h0Q2utYOBangcCBADcc1AGSeEk0jadA4AAkcUBIdQgR7xIeC4FuOQCiA1JMfSkE/B7KmhYAiH754p0ZnX40z579tmd1V05QOSDleiD8E9r7AEASOr1YiPYenX40L1+8M1d//N0QLrlB5YDUAEA4/tzV8OnFogEAjpeGQ8wQ6KQK0F7HMCQIWDw+ltpdA0HlgNSxWe9EB7icgMX3evlexYDhEwrhoA6o7x/EFWOawKYXi4ZjQDB1Q+hwkHq/MwcUZcUuq2FBuNG9Xm5rNxaPYeLpsaskvn1z5bR/tAO0pVIaBiCWDgUK4OZ6ZUVLw4aDIAGIngmCcKgUoSC4YQDvb65X9lyuUhY6T/ABSHI1qHUAAHAlQwmANgBCiP2jc0DMbFFyAXyWCgBACLF/p8vieAxTALFC+/nA+71LfKcAIBFKw4BLYj6BuNRxCbAoK6f4zm+McNUAGs8tcbkumCjMkOTX6t3hGBdwArAzAAInHM8HfAAk8SKA1JfEWgiT8dxkWdZwABZPz8MBcAkXAdT3D2az3iW5KuRAQt33QaBTYNhns941ZoLc2A8RPxnPu3cAhoDFcxBovlgtbp3LaHjaHCK+tRxAZ4qcm/BnPgjQ21g45wCteNEBqcLnJI0TaKQQ33oVCBlK+EoxpDrA91xl0IqPdkDq5wh9TsArO9QBseLhXAd7UtQFlYNw9Py1FXv0/HWQeK7H6Sr1LwMgxAkgHMRrRNNhdDAAo9nWFGVljs/O2UbBPMQ1HEC8SyzeYsvT0nowAHgcc07YrHeNpTQ4RiOewqNAW70vwImnPdnP3Q9BxIiWzkMXSU/uHp/uAO2skesRyQkxwZ0fA8AXR0kAaOL47FxcEJGOgVtrXILUQMFJEz+QFQ0Aep5upXAJ4JITtip3jpDjIUC4MWZvjSCJA0IepcVT1eOzc3N8du61r/RITVFWJsuyLMQNw2VtjDFmNNva1xRCa0MAL3vTFRwJQIh46MmQ4z/8eWeFQ48HAUhxOQyN5VzA7UvXAkEI2BfO5zoHXgMcLms71jEACmEPAJTA+v4hqhzirE4BuI6hosAh0EgsHuzM9TrAAvEYgAQhmQNoXY/J1NznIQ6g4ikA7AIKoQFA23BqY9c+3Gqu7//SYYB7H7+GH0FRCHhIUAgsgNhwiZcWNaRJiwsC9x0nntZ7rgokmwj5EpMrfOemS9uSeAyADgeXC5IBgCs8mO2FBgdLcgBNfkX5/y12AEDvAXAAKIQkP57mBGgAaPIOnRdgAFS8BAL/Zal+Gi+Jx5e/ml73QcDnBQiccB+E/wATCT7creMvYwAAAABJRU5ErkJgggAA
```

---

>We could also include the sha but not sure it's necessary as the data itself is stored on the Stellar chain.

I think it including the sha is still important so that an application can ensure the reconstructed file was reconstructed correctly.

---

This makes me think there will probably be many ways to represent NFTs and rather attempt to group different types of NFTs together, maybe we should just try to standardize NFTs with similar implementations. Maybe the optimized version you present should be a separate SEP?

tyvdh:
My apologies on misreading the data storage bit, however due to the inefficiencies I don't think it's a good method for storing longer format arbitrary base64 data. You'll run out of space very quickly and each image will be nearly 2x as expensive to store and your storage size is also nearly halved. 

I also doubt that there will by many ways to store NFT data. It will either be an external pointer or stored as arbitrary data on-chain. If on-chain the reasonable assumption is you'll want that storage mechanic to be as compact (i.e. cheap) as possible.

My fear with splitting my recommendation into a separate SEP is that one or the other could get lost or preferred in favor of the other when both are solving identical problems. I really wish SEPs had sub or supporting SEPs or a concept of further details. Things are bound to get messy with all of this "well actually" or "alternatively" SEPs which aren't really closely linked together.

One route we might be able to take though is keeping this SEP as a draft until I have another SEP drafted for storing long format arbitrary data and then we can link to that from this for anyone not keen to use the recommended on-chain storage mechanic described here.

I do think my method is important to call out and should be the preferred method for storing long format base64 data as the only way storing on-chain makes sense is if it's as compact as possible.

leighmcculloch:
@tyvdh we could update the url field in the SEP to point to another account or data entry field that uses your format. Do the two work together.

Another option is we propose to allow the key and value be a Mac of 128 bytes combined rather than 64 each. Such a change might be relatively simple.

tyvdh:
> we could update the url field in the SEP to point to another account or data entry field that uses your format. Do the two work together.

Not quite sure I follow. Or are you simply saying like my original comment we add `nft.asset.content-type: image/png` which would queue the proposal to assume this account is dealing with my format.

> Another option is we propose to allow the key and value be a max of 128 bytes combined rather than 64 each. Such a change might be relatively simple.

Like a CAP change? Or.

leighmcculloch:
I was suggesting specifying a URI scheme that when used can reference data stored in the format you're using. The scheme would specify account(s), and maybe a key prefix. For assets stored in the same account it'd be the current address and a prefix. The `nft.asset.url` and `nft.meta.url` could then use this new scheme. An advantage of such an approach is you'd be no longer limited to the number of data entries in a single account. This also keeps the `nft.asset.url` fields common across both approaches so apps only need to inspect it to then discover the location of the data, regardless of whether it is on Stellar, IPFS, etc.

I was suggesting a CAP proposal that proposes that the size limit on data entries become based on the size of the key and value combined (128 bytes) rather than the size of the key and value independently (64+64 bytes). There would be no cost difference to the ledger.

tyvdh:
I think I'm understanding, could you detail out an example though?
---
Gotcha, yeah, if we could switch to both key and value being the same accepted type vs `string : buffer` that would be great.

leighmcculloch:
> Gotcha, yeah, if we could switch to both key and value being the same accepted type vs `string : buffer` that would be great.

They'd stay the same type, key as string and value as buffer, but if string takes up 6 bytes, value has 122 bytes available to it, so no need to mix value data into the key field.

tyvdh:
> They'd stay the same type, key as string and value as buffer, but if string takes up 6 bytes, value has 122 bytes available to it, so no need to mix value data into the key field.

Oh I see, yeah, ideal.

leighmcculloch:
> I think I'm understanding, could you detail out an example though?

Maybe something like the following would allow a client to go check the listed accounts for entries that begin with `nft.asset.data`.

```
stellardata:prefix=nft.asset.data&account=G...&account=G...
```

leighmcculloch:
@tyvdh I've updated the SEP to include the compact format you described above. Can you take another look?

leighmcculloch:
@FredericRezeau linked me to this document which is the approach Litemint uses:
https://docs.ipfs.io/how-to/best-practices-for-nft-data/#persistence-and-availability

I misunderstood how Litemint captured data on and off-network, so what I captured here is incorrect. I think Litemint's approach, which is based on IPFS's recommendations, and newer stuff on Ethereum is simpler and better. So, I think we should focus this SEP on capturing that approach, and then figure out how to add @tyvdh's compact approach back and an extendable way.

FredericRezeau:
Hi everyone!

Thanks for the opportunity to provide input.

As mentioned, Litemint follows the recommendations by Protocol Labs (IPFS) which uses the Multiformats [Content Identifier (CID) specification](https://github.com/multiformats/cid) to address NFT data. The protocol is broadly used in the industry and its adoption is fast growing with services like [web3.storage](https://web3.storage) and [nft.storage](https://nft.storage) that use IPFS hash (CID), therefore such services could be used too by NFT issuers on Stellar.

The CID stored as a single data entry against the account issuer on Stellar is a cheaper way than using url+sha256.

Regarding the JSON content of the meta data file, Litemint has mostly followed the spirit of [ERC-721 standard](https://eips.ethereum.org/EIPS/eip-721) on Ethereum, only adapting the mandatory `symbol` entry with Stellar `issuer` and `code` and keeping the `name` entry—no entries, nor the presence of such file, is mandatory on Litemint. Note that ERC-721, technically, makes these 2 entries also optional by allowing empty values.

I would further recommend for the Stellar specification to remain as free as possible provided the innovation in this space is bringing unforeseen use cases.

For reference here is the Meridian 2021 asset minted on Litemint with the data entry pointing to a CID addressing a persistent meta data file in JSON format:

Asset on Litemint: [GDOA...VBC/MERIDIAN2021](https://litemint.com/items/GDOA2FECYKBVG2RQYG5JQO6EBLCDBWZ7MFBEZY6AIBJ4NDXV3XEVCVBC/MERIDIAN2021)
Meta data on IPFS: [ipfs://QmaydGEshbL7HMQkD6FZPcMeEt6o15aC4p4P82VxQ7amjy](https://ipfs1.litemint.com/ipfs/QmaydGEshbL7HMQkD6FZPcMeEt6o15aC4p4P82VxQ7amjy)

**Comments on SEP:**

I find the Abstract and Motivation slightly ambiguous.

Information addressed by home domains and stored on stellar.toml, besides being transient, is not inherent to the NFT asset. Since this SEP standardizes the use of data entries for addressing persistent NFT data i.e., data inherent to the asset, there is no overlap of purpose possible and so the goal is not related to relying on stellar.toml for addressing meta data (we already cannot as stellar.toml data is impermanent).

Distinctively also, stellar.toml files are conveniently used to maximize interoperability in presenting assets with ecosystem services and more importantly so for NFTs, with IPFS gateways often being rate limited and slow, stellar.toml files would remain the most practical way to convey information to clients while guarantying a good user experience. By analogy, ecosystem clients do not lookup native tokens or physical goods for anchored assets and would only need to look up the intrinsic NFT data if they need a provably authentic instance of it. Mixing both is confusing, in scope and purpose.

**Notes on Backward Compatibility**

Litemint NFT marketplace is now hosting thousands of NFTs, and entire collections have been issued and traded daily since May 2021. With hundreds of committed artists issuing their work and doing business on Stellar, it would be crucial to consider maintaining backward compatibility with the current landscape which—I really want to point out—moved much beyond _experimental NFTs_, for realistic adoption of the standard.

Should we use the Multiformats Content Identifier (CID) specification with a single data entry on the Stellar issuer account to address the persistent NFT meta data:
- `ipfshash` is a common naming convention by contracts on Ethereum or APIs (e.g. Pinata) to refer to the CID and if we could use it, all NFTs previously minted on Litemint would remain be compatible with this SEP.
- If a change is needed for some reason, we should ideally provide a fallback mechanism for pointing clients to the appropriate data entry, the mechanism should be generic enough to encompass Stellar Quest NFT storage.

Thanks again, I am happy to provide any additional insights on the use cases we are currently seeing on the marketplace.

Shaptic:
First, defining some terminology to get things straight: I'll use "_NFT_" to refer to the actual Stellar asset that is issued/held by accounts on the network, and "_asset_" to refer to the actual digital trinket whose "ownership" is reflected by the NFT.

I'd personally like to see two different standards based on where the asset is stored: **on chain** and **off chain**.

--------------

#### Off-Chain Assets
The latter is the one mostly described here: the asset is stored on IPFS and the NFT is just metadata about it. This is all well and good and is a **pure way to abstract ownership into the ledger** while the asset itself can exist anywhere. We can philosophically argue about what it means if other NFTs reference to the same asset, but that's beside the point here. For this, I think we should follow existing standards, and I like what @FredericRezeau defines in terms of alignment with [EIP-721](https://eips.ethereum.org/EIPS/eip-721). No critiques there; it's not my domain and in some ways the larger crypto community has spoken.

--------------

#### On-Chain Assets
For actually storing assets on-chain, like is done for Stellar Quest (see, for example, [this account](https://horizon.stellar.org/accounts/GAOVDKTAKB4VEDN4CUY6F3WP4Q2O3ME46XRVHO67M3645JMF5A4PH276)), I think we need a different standard because 

 - (a) it's more expensive, so every byte counts, and 
 - (b) it reflects a different level of ownership: it's "all inclusive."

Take these two snippets from IPFS' [recommendations](https://docs.ipfs.io/how-to/best-practices-for-nft-data/#summary) on NFTs:

> IPFS allows NFTs to represent data of any size and format in a secure, verifiable, and distributed way that can stand the test of time.

At some point in the infinite future, either Stellar xor IPFS will "stand" longer. In the former case, NFTs that point to IPFS become completely useless; in the latter, we keep our cruft off of IPFS.

> Data persistence should be part of your platform's design. Running your own IPFS infrastructure [...]

This is often prohibitive to individuals, unlike paying $X to store it in the ledger.

**Warning, napkin math ahead**: If we take into account @leighmcculloch's earlier [comment](https://github.com/stellar/stellar-protocol/pull/902#issuecomment-843348998) about viewing a `ManageData` entry an arbitrary 128 bytes, then we come to a cost of 0.5 XLM per entry (according to the [minimum balance spec](https://developers.stellar.org/docs/glossary/minimum-balance/)). Today's XLM price of ~$0.33 means $1.32 / 1kiB or $1352 / 1 MiB ([calc 1](https://www.wolframalpha.com/input/?i=ceil%281024%2F128%29*%280.5*0.33%29), [calc 2](https://www.wolframalpha.com/input/?i=ceil%282%5E20%2F128%29*%280.5*0.33%29)) to store directly on chain.

These numbers are large but not prohibitive, and I think that means we need to define an on-chain way to represent these types of assets. For this, I think @leighmcculloch's original proposal is actually quite nice: we can consolidate the URL-based method with the "compact v1" method and use something like `mime-type|<data>`. There's no need to encode indices because order is preserved on insertion, and the mime-type a trivial prefix to what'll otherwise be a large payload but succinctly hints how readers should interpret what follows.

#### Composability
In some ways, these disjoint representation methods are actually completely compatible: store an asset on-chain, then point to it via the off-chain convention.

--------------

Fundamentally, I think we need both a philosophical and conventional distinction between on-chain and "meta" NFTs. These can be within one SEP, but need to, imo, be distinct.

leighmcculloch:
Thanks for the feedback @FredericRezeau, this is very helpful, and I agree 💯 on backwards compatibility, and that the motivation for this SEP is outdated and no longer relevant. I think your comments and @Shaptic's comments are highlighting that for off-chain, we should just capture what you all are doing, which is already based on solid reasoning and standards, with the goal of encouraging interoperability.

> I'd personally like to see two different standards based on where the asset is stored: **on chain** and **off chain**.

I think there's value in capturing these in separate SEPs that are hyper-focused on each. This will make their evolution clearer. It will also mean a wallet can clearly indicate which they support. The SEPs will be small and focused, and people could still implement both in some sort of stellar-nft SDK.

@FredericRezeau @tyvdh @Shaptic What do you think?

@FredericRezeau Would you be interested in writing a SEP that captures the pattern you're using? Or interested in collaborating on that?

@Shaptic @tyvdh In a separate proposal would you be interested in iterating on and documenting the compact format?

FredericRezeau:
@leighmcculloch Definitely, I would be very interested in writing a SEP that captures the ideas for the pattern we are using and shaping a formal proposal on the first SEP.

@Shaptic Thank you for your input. I just wanted to clarify a possible misconception about IPFS. Content addressing does not rely on network consensus and NFTs data fortunately is still verifiable, regardless i.e. you are never required to upload/pin your data and could send them over by email (although inconvenient) to the next owner so they can [compute the CID locally](https://www.npmjs.com/package/ipfs-only-hash).

Also, from a logical standpoint, since you cannot repudiate the data that the CID is referencing, there is no logical difference between storing data and storing a CID. This is still ownership, not an abstraction of it.

And from a practical standpoint, I agree, any blockchain that could accommodate arbitrary data storage, scalable in size and costs, would be very convenient. This is only a practical advantage.

Hope this clarifies.

leighmcculloch:
CIDs don't quite the same identity properties that a sha256 had does.

For example, if you send me a file I can very simply verify the sha256 and I'll always get the same value as you for the same file.

On the contrary, if I try to verify the CID, I may not arrive at the same value as you. That's because a CID is not simply a hash of the file data. Clients choose how the data is chunked, and that affects the final hash for the CID. There are also other settings clients can choose that affect the CID. And so two clients uploading the same file may generate different CIDs. There's some helpful discussion about these things here:
- https://stackoverflow.com/questions/59183147/are-cid-different-for-the-same-file-but-uploaded-by-2-users-in-ipfs
- https://github.com/alanshaw/ipfs-only-hash/issues/21

So I believe what you said is correct, the receiver of the file can verify the file matches the CID but only if both participants happen to be using the same settings, or if they communicate about what settings need to be used. All this means is interoperability is a little more challenging. This is one of the reasons that had led me to include a sha256 hash in this proposal. But @tyvdh made some good points that every data entry is increased cost and a CID does provide you with many guarantees which probably make it sufficient on its own.

FredericRezeau:
@leighmcculloch Good point. Interoperability could be improved using the meta data, worth discussing for the SEP.

tyvdh:
Splitting this into two SEPs makes sense and I'm happy to be a part of the on-Stellar's-chain one as it's the one I make the most use of and have the most experience with.

I have some questions @Shaptic about your comments here: https://github.com/stellar/stellar-protocol/pull/902#issuecomment-973298878 but we should maybe move this discussion to the new SEP

alejomendoza:
Hey there, interested in contributing to the off-chain proposal, is there a node package for https://github.com/leighmcculloch/makenft @leighmcculloch ? 

This SEP should be backwards compatible with ERC-721 as @FredericRezeau pointed out.  Also, I don't see the off-chain PR SEP proposal open, should we start one to keep the off-chain conversation there?

leighmcculloch:
> is there a node package for https://github.com/leighmcculloch/makenft @leighmcculloch ?

There isn't. And the makenft example is based on an earlier version of what was in this proposal written back in April 2021, and probably doesn't align with the latest proposal here, or what @FredericRezeau is working on. A node package that implements the new on-chain proposal #1090 or the new proposal @FredericRezeau is working on for off-chain sounds interesting!

alejomendoza:
> Hi everyone!
> 
> Thanks for the opportunity to provide input.
> 
> As mentioned, Litemint follows the recommendations by Protocol Labs (IPFS) which uses the Multiformats [Content Identifier (CID) specification](https://github.com/multiformats/cid) to address NFT data. The protocol is broadly used in the industry and its adoption is fast growing with services like [web3.storage](https://web3.storage) and [nft.storage](https://nft.storage) that use IPFS hash (CID), therefore such services could be used too by NFT issuers on Stellar.
> 
> The CID stored as a single data entry against the account issuer on Stellar is a cheaper way than using url+sha256.
> 
> Regarding the JSON content of the meta data file, Litemint has mostly followed the spirit of [ERC-721 standard](https://eips.ethereum.org/EIPS/eip-721) on Ethereum, only adapting the mandatory `symbol` entry with Stellar `issuer` and `code` and keeping the `name` entry—no entries, nor the presence of such file, is mandatory on Litemint. Note that ERC-721, technically, makes these 2 entries also optional by allowing empty values.
> 
> I would further recommend for the Stellar specification to remain as free as possible provided the innovation in this space is bringing unforeseen use cases.
> 
> For reference here is the Meridian 2021 asset minted on Litemint with the data entry pointing to a CID addressing a persistent meta data file in JSON format:
> 
> Asset on Litemint: [GDOA...VBC/MERIDIAN2021](https://litemint.com/items/GDOA2FECYKBVG2RQYG5JQO6EBLCDBWZ7MFBEZY6AIBJ4NDXV3XEVCVBC/MERIDIAN2021) Meta data on IPFS: [ipfs://QmaydGEshbL7HMQkD6FZPcMeEt6o15aC4p4P82VxQ7amjy](https://ipfs1.litemint.com/ipfs/QmaydGEshbL7HMQkD6FZPcMeEt6o15aC4p4P82VxQ7amjy)
> 
> **Comments on SEP:**
> 
> I find the Abstract and Motivation slightly ambiguous.
> 
> Information addressed by home domains and stored on stellar.toml, besides being transient, is not inherent to the NFT asset. Since this SEP standardizes the use of data entries for addressing persistent NFT data i.e., data inherent to the asset, there is no overlap of purpose possible and so the goal is not related to relying on stellar.toml for addressing meta data (we already cannot as stellar.toml data is impermanent).
> 
> Distinctively also, stellar.toml files are conveniently used to maximize interoperability in presenting assets with ecosystem services and more importantly so for NFTs, with IPFS gateways often being rate limited and slow, stellar.toml files would remain the most practical way to convey information to clients while guarantying a good user experience. By analogy, ecosystem clients do not lookup native tokens or physical goods for anchored assets and would only need to look up the intrinsic NFT data if they need a provably authentic instance of it. Mixing both is confusing, in scope and purpose.
> 
> **Notes on Backward Compatibility**
> 
> Litemint NFT marketplace is now hosting thousands of NFTs, and entire collections have been issued and traded daily since May 2021. With hundreds of committed artists issuing their work and doing business on Stellar, it would be crucial to consider maintaining backward compatibility with the current landscape which—I really want to point out—moved much beyond _experimental NFTs_, for realistic adoption of the standard.
> 
> Should we use the Multiformats Content Identifier (CID) specification with a single data entry on the Stellar issuer account to address the persistent NFT meta data:
> 
> * `ipfshash` is a common naming convention by contracts on Ethereum or APIs (e.g. Pinata) to refer to the CID and if we could use it, all NFTs previously minted on Litemint would remain be compatible with this SEP.
> * If a change is needed for some reason, we should ideally provide a fallback mechanism for pointing clients to the appropriate data entry, the mechanism should be generic enough to encompass Stellar Quest NFT storage.
> 
> Thanks again, I am happy to provide any additional insights on the use cases we are currently seeing on the marketplace.

Looking into how the metadata is stored on IPFS for litemint assets I found some inconsistencies with ERC-721, using [nft.storage](https://nft.storage/#docs) the following metadata is created when storing an NFT with video. 

[https://ipfs.io/ipfs/bafyreifbchb5iwiioqofga56k5cdgs2orzcl6vit5pev5fwefwnjdvkkua/metadata.json](https://ipfs.io/ipfs/bafyreifbchb5iwiioqofga56k5cdgs2orzcl6vit5pev5fwefwnjdvkkua/metadata.json)

1. `/metadata.json` is present at the end of the ipfs url (See ipfs [docs)](https://docs.ipfs.io/how-to/mint-nfts-with-ipfs/#minty)
2. The image asset has the file name at the end of the ipfs url `ipfs://bafybeicwxymkrmvkzciu2go6uyk757zy7ut4tyus26z3fv2jvblzbxylem/test.png`
3. The video asset goes under properties and has the name of the file at the end. [nft.storage js lib](https://github.com/nftstorage/nft.storage/blob/e8f7bed8953a6894421f1a72105d7775b4e72f2e/packages/client/src/lib.js#L595)

I like how clean it looks for the Meridian 2021 asset minted on Litemint but I'm not sure what it would take to create a standard that is based on ERC-721 and works on Stellar.  As more NFT marketplaces are built on Stellar, it is crucial to have interoperability where large offchain NFT assets can be found across different platforms.  As an example we would be thrilled to have [Skyhitz](https://communityfund.stellar.org/projects/skyhitz) music NFT assets show up on Litemint as well and index Litemint music NFT assets on our platform.  

I'm very interested in helping out with the off-chain SEP.

PR #908, opened 2021-04-13
SEP-0038: Request for Quote API

### LINKED ISSUE #901: Anchor Quote API
We've recently seen an increased demand for a standardized API specification to facilitate off-chain liquidity in Anchor flows. We'd love to hear your thoughts about this.

Currently, our deposit/withdraw (sep6/24) and cross-border payments (sep31) standards assume that off-chain assets being deposited/withdrawn are the same as those pegged on-chain. For example, when you withdraw BRLT (by nTokens) using nTokens, you get BRL in your bank account.

However, there's nothing stopping an anchor from converting currencies during deposits and withdrawals, as long as they can provide the FX. For example, nTokens could also, theoretically, offer a withdrawal from USDC (by Circle) directly to BRL in a bank account in Brazil.

The missing part of this puzzle is that nTokens doesn't have an interoperable way to provide quotes for their users and, presumably, users will not initiate a withdrawal without knowing how much they're going to get in the end.

What would this standard API be good for? Some anchors are able to provide FX and prefer to skip asset issuance due to operational and/or regulatory complexity.  

cc. @tomerweller 

This issue serves as a place for discussion and TODO

JakeUrban:
resolves #901 

This protocol enables anchors to accept off-chain assets in exchange for different on-chain assets, and vice versa. Specifically, it enables anchors to provide quotes that can referenced within the context of existing Stellar Ecosystem Proposals. How the exchange of assets is facilitated is outside the scope of this document.

Once the protocol has been merged, SEP-6 & SEP-31 will be updated to accept quote IDs provided by this API.

### DIFF
--- a/ecosystem/sep-0038.md
+++ b/ecosystem/sep-0038.md
@@ -3,9 +3,9 @@
 ```
 SEP: 0038
 Title: Anchor RFQ API
-Author: Jake Urban <@jakeurban>
+Author: Jake Urban <@jakeurban> and Leigh McCulloch <@leighmcculloch>
 Track: Standard
-Status: Final Comment Period
+Status: Draft
 Created: 2021-04-09
 Discussion: https://github.com/stellar/stellar-protocol/issues/901
 Version 1.0.0
--- a/ecosystem/sep-0038.md
+++ b/ecosystem/sep-0038.md
@@ -145,7 +145,7 @@ Name | Type | Description
 
 ### GET Price
 
-This endpoint can be used to fetch the indicative price for a given base asset and counter asset pair.
+This endpoint can be used to fetch the indicative price for a given asset pair.
 
 #### Request
 
--- a/ecosystem/sep-0038.md
+++ b/ecosystem/sep-0038.md
@@ -198,3 +198,4 @@ Name | Type | Description
 `id` | string | The unique identifier for the quote to be used in other Stellar Ecosystem Proposals (SEPs).
 `expires_at` | [UTC ISO 8601](https://www.iso.org/iso-8601-date-and-time-format.html) string | The date and time by which the anchor must receive funds from the client.
 `price` | string | The price offered by the anchor for one unit of `buy_asset` in terms of `sell_asset`.
+`buy_amount` | string | The amount of `buy_asset` to be exchanged for `sell_asset`. `price * sell_amount = buy_amount` must be true up to the number of decimals required for `buy_asset`.
--- a/ecosystem/sep-0038.md
+++ b/ecosystem/sep-0038.md
@@ -17,7 +17,7 @@ This protocol enables anchors to accept off-chain assets in exchange for differe
 
 ## Motivation
 
-Anchoring an asset and issuing an asset are distinct functions that have different buisness and technical requirements. However, issuing an asset has often been a prerequisite for anchoring an asset. This protocol enables anchors to transfer value on and off the Stellar network for their clients regardless of whether a one-for-one reserve-backed Stellar asset existed for the off-chain asset held by the anchor's clients.
+Anchoring an asset and issuing an asset are distinct functions that have different business and technical requirements. However, issuing an asset has often been a prerequisite for anchoring an asset. This protocol enables anchors to transfer value on and off the Stellar network for their clients regardless of whether a one-for-one reserve-backed Stellar asset exists for the off-chain asset held by the anchor's clients.
 
 Removing this requirement for anchors also provides downstream benefits to ecosystem participants generally. Enabling anchors to accept any Stellar asset will naturally decrease liquidity fragmentation on the decentralized exchange, leading to greater market depth and tigher spreads between trading pairs.
 
@@ -25,7 +25,11 @@ Removing this requirement for anchors also provides downstream benefits to ecosy
 
 ### Authentication
 
-All endpoints require authentication in the form of a [SEP-10](sep-0010.md) JSON Web Token (JWT) in the `Authorization` header of the request. Authenticated accounts must exist on the network in order to make requests to `POST /quote`, as described in the [Protecting Against Bad Actors](#protecting-against-bad-actors) section.
+All endpoints require authentication in the form of a [SEP-10](sep-0010.md) JSON Web Token (JWT) in the `Authorization` header of the request. 
+
+```
+Authorization: Bearer <jwt>
+```
 
 ### Content Type
 
@@ -34,6 +38,27 @@ All endpoints accept in requests the following `Content-Type`:
 
 All endpoints respond with content type:
 - `application/json`
+
+### Asset Identification Format
+
+This protocol can be used to provide quotes for any class of asset in exchange for a Stellar asset. The following format is used to identify an asset in API requests and responses.
+
+```
+namespace:symbol:issuer
+```
+
+The currently accepted `namespace` values are:
+
+Name | Description
+-----|------------
+`stellar` | Used for Stellar assets.
+[`iso4217`](https://www.iso.org/iso-4217-currency-codes.html) | Used for fiat currencies.
+
+The `symbol` value should be the Stellar `asset_code` or ISO-4217 currency code.
+
+The `issuer` value is optional because it may not be relevant for a given asset class, such as fiat currencies. For Stellar assets, `issuer` should be the public key of the asset's issuing account.
+
+When presenting assets to end users, applications should replace the Stellar public key string with the domain hosting the issuer's [SEP-1 `stellar.toml`](sep-0001.md).
 
 ### Endpoints
 
@@ -44,7 +69,7 @@ All endpoints respond with content type:
 
 ### GET Info
 
-This endpoint describes the supported Stellar assets and off-chain currencies available for trading. Note that the anchor may not support a trading pair between every Stellar asset and off-chain currency listed. Use the [`GET /prices`](#get-prices) endpoint to see which pairs are supported.
+This endpoint describes the supported Stellar assets and off-chain assets available for trading. Note that the anchor may not support a trading pair between every Stellar asset and off-chain asset listed. Use the [`GET /prices`](#get-prices) endpoint to see which pairs are supported.
 
 #### Request
 
@@ -54,21 +79,21 @@ No request arguments required.
 
 Name | Type | Description
 -----|------|------------
-`assets` | array | An array of objects describing the Stellar assets available in exchange for one or more of the `currencies` listed.
-`currencies` | array | An array of objects describing the off-chain currencies available in exchange for one or more of the `assets` listed.
+`stellar` | array | An array of objects describing the Stellar assets available in exchange for one or more of the assets listed under other namespaces.
+`iso4217` | array | An array of objects describing the fiat currencies available in exchange for one or more of the `stellar` assets listed.
 
 `assets` Object Schema
 
 Name | Type | Description
 -----|------|------------
-`asset` | string | The [SEP-32](sep-0032.md) address of the Stellar asset. Lumens should be specified as `XLM`.
+`asset` | string | The `asset-code:issuer-account` of the Stellar asset. Lumens should be specified as `native`.
 
-`currencies` Object Schema
+`iso4217` Object Schema
 
 Name | Type | Description
 -----|------|------------
-`currency` | string | The [ISO 4217](https://www.iso.org/iso-4217-currency-codes.html) code for the currency.
-`delivery_methods` | array | An array of objects describing the methods a client can use to deliver funds to the anchor for a given `currency`. The method of delivery may affect the expiration and/or price provided in a [`POST /quote`](#post-quote) response. The array can be empty if the method of delivery is irrelevant to the anchor.
+`symbol` | string | The [ISO 4217](https://www.iso.org/iso-4217-currency-codes.html) code for the currency.
+`delivery_methods` | array | An array of objects describing the methods a client can use to deliver funds to the anchor. The method of delivery may affect the expiration and/or price provided in a [`POST /quote`](#post-quote) response. The array can be empty if the method of delivery is irrelevant to the anchor.
 
 `delivery_methods` Object Schema
 
@@ -87,8 +112,8 @@ Only one of the following parameters should be included in the request.
 
 Name | Type | Description
 -----|------|------------
-`asset` | string | (optional) The [SEP-32](sep-0032.md) address of the Stellar asset you want to get the price of available off-chain currencies in. Lumens should be specified as `XLM`.
-`currency` | string | (optional) The [ISO 4217](https://www.iso.org/iso-4217-currency-codes.html) code of the currency you want to get the price of available Stellar assets in.
+`stellar` | string | (optional) The Stellar asset you want to get the price of available off-chain assets in. Lumens should be specified as `native`.
+`iso4217` | string | (optional) The ISO 4217 code of the currency you want to get the price of available Stellar assets in.
 
 #### Response
 
@@ -96,23 +121,23 @@ Only one of the following attributes should be included in the response.
 
 Name | Type | Description
 -----|------|------------
-`assets` | array | (optional) An array of objects containing the available Stellar assets and their prices in terms of the `currency` provided in the request.
-`currencies` | array | (optional) An array of objects containing the available currencies and their prices in terms of the `asset` provided in the request.
+`stellar` | array | (optional) An array of objects containing the available Stellar assets and their prices in terms of the `iso4217` currency code provided in the request.
+`iso4217` | array | (optional) An array of objects containing the available currencies and their prices in terms of the Stellar asset provided in the request.
 
-`assets` Object Schema
+`stellar` Object Schema
 
 Name | Type | Description
 -----|------|------------
-`asset` | string | Same as the definition of `asset` in the request.
-`price` | string | The price of one unit of `asset` in terms the `currency` sent in the request.
+`asset` | string | Same as the definition of `asset` in the [GET /info](#get-info) response.
+`price` | string | The price of one unit of `asset` in terms the currency sent in the request.
 `decimals` | integer | The number of decimals needed to represent asset.
 
-`currencies` Object Schema
+`iso4217` Object Schema
 
 Name | Type | Description
 -----|------|------------
-`currency` | string | Same as the definition of `currency` in the request.
-`price` | string | The price of one unit of `currency` in terms of the `asset` sent in the request.
+`symbol` | string | Same as the definition of `symbol` [GET /info](#get-info) response.
+`price` | string | The price of one unit of this currency in terms of the Stellar asset sent in the request.
 `decimals` | integer | The number of decimals needed to represent the currency.
 
 ### GET Price
@@ -123,7 +148,7 @@ This endpoint can be used to fetch the indicative price of a Stellar asset and c
 
 Name | Type | Description
 -----|------|------------
-`base` | string | The base asset or currency. Ex. `USDC:circle.io`, `ARS`
+`base` | string | The base asset or currency. Ex. `USDC:circle.io`, `iso4217:ARS`
 `counter` | string | The counter asset or currency.
 `base_amount` | string | The amount of the base asset or currency to be traded for the counter asset or currency.
 
@@ -143,8 +168,8 @@ In contrast with the `GET price(s)` endpoints, the amount requested must be held
 
 To protect against bad actor clients reserving all available captial without following through with the requested trades, servers should only accept requests for which the following is true:
 
-1. The Stellar account authenticated via SEP-10 is owned by a known and trusted entity, or
-2. The owner of the Stellar account has been properly KYC'ed, either via [SEP-12](sep-0012.md) or some other mechanism.
+1. The Stellar account authenticated via SEP-10 is owned by an entity that is known and trusted, or
+2. The Stellar account authenticated via SEP-10 is owned by an individual and has been properly KYC'ed either via [SEP-12](sep-0012.md) or some other mechanism.
 
 Servers may deny access to the API if misuse is detected.
 
@@ -155,8 +180,6 @@ It should be expected that the provided quote includes a margin for the provider
 If the client requests some amount of off-chain currency for providing some amount of a Stellar asset, the client will submit a Stellar transaction delievering funds to the anchor before the expiration included in the response, paying a fee as determined by state of the Stellar network. 
 
 In the reverse scenario, the anchor will submit a Stellar transaction to deliver funds to the client as long as the client delivered off-chain funds to the anchor before the expiration. In this case, the anchor will likely increase their margin to cover the cost of submitting the transaction.
-
-This specification could be extended in the future to include functionality where the anchor collects a transaction signature  from the client prior to submitting the transaction, enabling the anchor to have the client pay the network fees and decrease their margin. This feature would increase the technical complexity of the specification and hence will only be added if there is enough support from the ecosystem.
 
 #### Request
 
--- a/ecosystem/sep-0038.md
+++ b/ecosystem/sep-0038.md
@@ -121,7 +121,8 @@ This endpoint can be used to fetch the indicative price of a Stellar asset and c
 
 Name | Type | Description
 -----|------|------------
-`pair` | string | The base asset or currency and counter asset or currency, concatenated with a slash. Ex. `USDC:circle.io/ARS`, `USD/NGNT:cowrie.exchange`
+`base` | string | The base asset or currency. Ex. `USDC:circle.io`, `ARS`
+`counter` | string | The counter asset or currency.
 `base_amount` | string | The amount of the base asset or currency to be traded for the counter asset or currency.
 
 #### Response
@@ -160,7 +161,8 @@ This specification could be extended in the future to include functionality wher
 
 Name | Type | Description
 -----|------|------------
-`pair` | string | Same as the definition of `pair` in `GET /price`.
+`base` | string | Same as the definition of `base` in `GET /price`.
+`counter` | string | Same as the definition of `counter` in `GET /price`.
 `base_amount` | string | Same as the definition of `base_amount` in `GET /price`.
 `delivery_method` | string | (optional) One of the `name` values specified by the `delievery_methods` array for the relevant currency returned from `GET /info`. If the array is empty, this parameter is not required.
 
--- a/ecosystem/sep-0038.md
+++ b/ecosystem/sep-0038.md
@@ -82,7 +82,7 @@ The [SEP-11](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/s
 
 `iso4217` Identifier Format
 
-The [ISO 4217](https://www.iso.org/iso-4217-currency-codes.html) code for the fiat currency.
+The [ISO 4217](https://www.iso.org/iso-4217-currency-codes.html) three-character currency code for the fiat currency.
 
 ### Endpoints
 
@@ -133,7 +133,7 @@ Name | Type | Description
 
 Name | Type | Description
 -----|------|------------
-`buy_assets` | array | An array of objects containing information on the assets that can be exchanged for `sell_asset`.
+`buy_assets` | array | An array of objects containing information on the assets that the client will receive when they provide `sell_asset`.
 
 `buy_assets` Object Schema
 
--- a/ecosystem/sep-0038.md
+++ b/ecosystem/sep-0038.md
@@ -39,6 +39,28 @@ All endpoints accept in requests the following `Content-Type`:
 All endpoints respond with content type:
 - `application/json`
 
+### Errors
+
+If an error occurs when calling any endpoint, an appropriate HTTP status code
+will be returned along with an error response.
+
+#### Status Code
+
+Common HTTP status codes may be returned for a server. In particular the following are expected:
+
+Status Code | Name | Reason
+-----|------|------------
+`400` | Bad Request | The request is invalid in any way.
+`403` | Permission Denied | No `Authorization` header has been provided or the contents of the header are not accepted as valid.
+
+The response body must contain a human-readable description of the reason for error:
+
+```json
+{
+  "error": "The requested asset is not supported. See GET /prices for supported assets."
+}
+```
+
 ### Asset Identification Format
 
 This protocol can be used to provide quotes for any class of asset in exchange for a Stellar asset. The following format is used to identify an asset in API requests and responses.
@@ -188,6 +210,7 @@ Name | Type | Description
 `base` | string | Same as the definition of `base` in `GET /price`.
 `counter` | string | Same as the definition of `counter` in `GET /price`.
 `base_amount` | string | Same as the definition of `base_amount` in `GET /price`.
+`expires_after` | [UTC ISO 8601](https://www.iso.org/iso-8601-date-and-time-format.html) string | (optional) The earliest date and time the client can deliver funds to the anchor. Anchors should return `400 Bad Request` if the requested expiration cannot be provided.
 `delivery_method` | string | (optional) One of the `name` values specified by the `delivery_methods` array for the relevant currency returned from `GET /info`. If the array is empty, this parameter is not required.
 
 #### Response
--- a/ecosystem/sep-0038.md
+++ b/ecosystem/sep-0038.md
@@ -146,6 +146,8 @@ To protect against bad actor clients reserving all available captial without fol
 1. The Stellar account authenticated via SEP-10 is owned by a known and trusted entity, or
 2. The owner of the Stellar account has been properly KYC'ed, either via [SEP-12](sep-0012.md) or some other mechanism.
 
+Entities or individuals that misuse the service can be placed on a denylist that should be checked before processing requests for quotes.
+
 #### Transaction Fees
 
 It should be expected that the provided quote includes a margin for the provider as a service fee, and this margin will vary depending on the directional flow of funds.
--- a/ecosystem/sep-0038.md
+++ b/ecosystem/sep-0038.md
@@ -188,10 +188,7 @@ In contrast with the `GET price(s)` endpoints, the amount requested must be held
 
 #### Protecting Against Bad Actors
 
-To protect against bad actor clients reserving all available captial without following through with the requested trades, servers should only accept requests for which the following is true:
-
-1. The Stellar account authenticated via SEP-10 is owned by an entity that is known and trusted, or
-2. The Stellar account authenticated via SEP-10 is owned by an individual and has been properly KYC'ed either via [SEP-12](sep-0012.md) or some other mechanism.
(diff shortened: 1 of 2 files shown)


ecosystem/sep-0038.md
```diff
@@ -0,0 +1,176 @@
...
+
+Anchoring an asset and issuing an asset are distinct functions that have different buisness and technical requirements. However, issuing an asset has often been a prerequisite for anchoring an asset. This protocol enables anchors to transfer value on and off the Stellar network for their clients regardless of whether a one-for-one reserve-backed Stellar asset existed for the off-chain asset held by the anchor's clients.
+
+Removing this requirement for anchors also provides downstream benefits to ecosystem participants generally. Enabling anchors to accept any Stellar asset will naturally decrease liquidity fragmentation on the decentralized exchange, leading to greater market depth and tigher spreads between trading pairs.
+
+## Specification
+
+### Authentication
+
+All endpoints require authentication in the form of a [SEP-10](sep-0010.md) JSON Web Token (JWT) in the `Authorization` header of the request. Authenticated accounts must exist on the network in order to make requests to `POST /quote`, as described in the [Protecting Against Bad Actors](#protecting-against-bad-actors) section.
```
leighmcculloch, line 28: ❗ We should provide an example of how to specify the token in the `Authorization` header. I assume it would be as:

```
Authorization: Bearer <jwt>
```

I realize in some other SEPs we have that have a similar section to this don't provide an example, but they should, because there are multiple forms to specify an authorization header value.

ecosystem/sep-0038.md
```diff
@@ -0,0 +1,176 @@
...
+#### Request
+
+No request arguments required.
+
+#### Response
+
+Name | Type | Description
+-----|------|------------
+`assets` | array | An array of objects describing the Stellar assets available in exchange for one or more of the `currencies` listed.
+`currencies` | array | An array of objects describing the off-chain currencies available in exchange for one or more of the `assets` listed.
```
leighmcculloch, line 58: ❓ Can `currencies` be non-currencies, such as Bitcoin? The distinction seems to be on-ledger assets and off-ledger assets. Thoughts on if there's a better term than `currencies`?

`assets` => `stellar-assets`
`currencies` => `external-assets`?

ecosystem/sep-0038.md
```diff
@@ -0,0 +1,176 @@
...
+
+`assets` Object Schema
+
+Name | Type | Description
+-----|------|------------
+`asset` | string | The [SEP-32](sep-0032.md) address of the Stellar asset. Lumens should be specified as `XLM`.
```
leighmcculloch, line 64: ❓ The way we've been standardizing lumens is to refer to it as `native`. That's the SEP-11 form and I believe some of our other APIs use this form, although ticker has used native as the issuer account instead, essentially `XLM:native`. Motivation for using only `XLM` here?

SEP-32 isn't adopted yet, and I need to do some work on it. For now I'd prefer it if you could run with the SEP-11 asset form, `<code>:<issuer-account>`, and then let's modify this to state SEP-32 or SEP-11 once SEP-32 is solid.

If you would prefer SEP-32, I can put some energy into firming it up quickly so it is ready to go for this.

ecosystem/sep-0038.md
```diff
@@ -0,0 +1,176 @@
...
+
+### GET Price
+
+This endpoint can be used to fetch the indicative price of a Stellar asset and currency [pair](https://en.wikipedia.org/wiki/Currency_pair).
+
+#### Request
+
+Name | Type | Description
+-----|------|------------
+`base` | string | The base asset or currency. Ex. `USDC:circle.io`, `ARS`
```
leighmcculloch, line 126: ❓ What's the significance of the `ARS` example that has no issuer? Is issuer optional?

ecosystem/sep-0038.md
```diff
@@ -0,0 +1,176 @@
...
+`counter` | string | The counter asset or currency.
+`base_amount` | string | The amount of the base asset or currency to be traded for the counter asset or currency.
+
+#### Response
+
+Name | Type | Description
+-----|------|------------
+`price` | string | The price of one unit of the base asset or currency in terms of the counter asset or currency for the `base_amount` specified in the request.
```
leighmcculloch, line 134: ❔  I want to make sure I understand how these three fields will be used, specially the amount. If I request `base=USDC`, `counter=USDC2`, where `USDC/USDC2 = 2` (i.e. `1 USDC = 2 USDC2`), with `base_amount=50`, I should expect `price=200`?

leighmcculloch, line 134: 💡  What's the motivation for the term `price`? When I worked on Vibrant's internal API for quotes I remember seeing that exchange rate APIs often referred to the resulting value as the `rate` and we used the same term there. If `rate` is a more common term than `price` it may be easier for readers to understand.

ecosystem/sep-0038.md
```diff
@@ -0,0 +1,176 @@
...
+This endpoint can be used to request a [firm](https://www.investopedia.com/terms/f/firmquote.asp) quote for a Stellar asset and currency pair. 
+
+In contrast with the `GET price(s)` endpoints, the amount requested must be held in reserve and not used in calculations of subsequent quotes until the expiration provided in the response.
+
+#### Protecting Against Bad Actors
+
+To protect against bad actor clients reserving all available captial without following through with the requested trades, servers should only accept requests for which the following is true:
+
+1. The Stellar account authenticated via SEP-10 is owned by a known and trusted entity, or
+2. The owner of the Stellar account has been properly KYC'ed, either via [SEP-12](sep-0012.md) or some other mechanism.
```
leighmcculloch, line 147: 🤔 These two points seem like the same thing to me. If the account is owned by a known entity then KYC has been performed, because KYC is know-your-customer.

ecosystem/sep-0038.md
```diff
@@ -0,0 +1,176 @@
...
+
+Entities or individuals that misuse the service can be placed on a denylist that should be checked before processing requests for quotes.
```
leighmcculloch, line 149: ❗ It's unclear from this sentence who owns this denylist. I suggest rewording to specify what the server can do within the bounds of this SEP to deny a user.

```suggestion
Servers may deny access to the API if misuse is detected.
```

We should probably also specify an appropriate HTTP error code to use in this case.

ecosystem/sep-0038.md
```diff
@@ -0,0 +1,176 @@
...
+
+It should be expected that the provided quote includes a margin for the provider as a service fee, and this margin will vary depending on the directional flow of funds.
+
+If the client requests some amount of off-chain currency for providing some amount of a Stellar asset, the client will submit a Stellar transaction delievering funds to the anchor before the expiration included in the response, paying a fee as determined by state of the Stellar network. 
+
+In the reverse scenario, the anchor will submit a Stellar transaction to deliver funds to the client as long as the client delivered off-chain funds to the anchor before the expiration. In this case, the anchor will likely increase their margin to cover the cost of submitting the transaction.
+
+This specification could be extended in the future to include functionality where the anchor collects a transaction signature  from the client prior to submitting the transaction, enabling the anchor to have the client pay the network fees and decrease their margin. This feature would increase the technical complexity of the specification and hence will only be added if there is enough support from the ecosystem.
```
leighmcculloch, line 159: 💡 I don't think it is worth including the paragraph about some possible extension, as it distracts from the rest of the content in this section. I don't feel strongly about this, just my perspective. Businesses pay fees to move money, it's part of doing business, and the fees are likely larger for the off-chain movements of money, so I don't think we should encourage making the UX and process more complex for it.

ecosystem/sep-0038.md
```diff
@@ -0,0 +1,176 @@
...
+
+#### Request
+
+Name | Type | Description
+-----|------|------------
+`base` | string | Same as the definition of `base` in `GET /price`.
+`counter` | string | Same as the definition of `counter` in `GET /price`.
+`base_amount` | string | Same as the definition of `base_amount` in `GET /price`.
+`delivery_method` | string | (optional) One of the `name` values specified by the `delievery_methods` array for the relevant currency returned from `GET /info`. If the array is empty, this parameter is not required.
```
leighmcculloch, line 168: ❗ 
```suggestion
`delivery_method` | string | (optional) One of the `name` values specified by the `delivery_methods` array for the relevant currency returned from `GET /info`. If the array is empty, this parameter is not required.
```

ecosystem/sep-0038.md
```diff
@@ -0,0 +1,176 @@
...
+
+#### Response
+
+Name | Type | Description
+-----|------|------------
+`id` | string | The unique identifier for the quote to be used in other Stellar Ecosystem Proposals (SEPs).
+`expiration` | [UTC ISO 8601](https://www.iso.org/iso-8601-date-and-time-format.html) string | The date and time by which the anchor must receive funds from the client.
```
leighmcculloch, line 175: 💡 I'm inclined to suggest naming the `expiration` field `expires_at`. The use of `_at` is reasonably common in APIs to signal timestamps.
```suggestion
`expires_at` | [UTC ISO 8601](https://www.iso.org/iso-8601-date-and-time-format.html) string | The date and time by which the anchor must receive funds from the client.
```

leighmcculloch (commented):
This is great, very easy to read and understand the motivation, and the shape of the API and what's required.

I think it is worth addressing some things (❗). I also have some questions (❓) that might influence things. I also left some suggestions (💡) that I don't feel strongly about.

ecosystem/sep-0038.md
JakeUrban: I was also thinking this when drafting it. I didn't love the names I thought of but thats not a reason to implicitly exclude a set of assets that could be used here.

I think `stellar` and `external` are good prefixes, although I think we should stick with underscores to match our existing APIs.

ecosystem/sep-0038.md
JakeUrban: Admittedly I initially thought SEP-32 was the standardization of the `<code>:<issuer-account>` syntax and decided to stick with the protocol anyway since I assumed you were trying to move the ecosystem in that direction.

I can change both points 👍

ecosystem/sep-0038.md
JakeUrban: ARS is the "external" asset as defined by [ISO 4217](https://en.wikipedia.org/wiki/ISO_4217).

ecosystem/sep-0038.md
JakeUrban: To be clear, this API is not built for exchanging two stellar assets. There may be value in such a standard, but the request/response schemas would be fairly different. For example, we could returned signed transactions consisting of two payment operations with the quote expiration as the `maxTime`.

Maybe we should make sure we name this protocol something that doesn't conflict with such a standard in the future?

JakeUrban: But to answer your question, this `base/counter` syntax has a specific [definition](https://www.investopedia.com/terms/c/currencypair.asp):

> The quotation EUR/USD = 1.2500 means that one euro is exchanged for 1.2500 U.S. dollars. In this case, EUR is the base currency and USD is the quote currency (counter currency). This means that 1 euro can be exchanged for 1.25 U.S. dollars. Another way of looking at this is that it will cost you $125 to buy 100 euros.

JakeUrban: The inspiration  was the [0x API](https://0x.org/docs/api#request-1), but I'm not married to the term.

ecosystem/sep-0038.md
JakeUrban: I can see the possibility for confusion, I was thinking about it in the context of SEP-31 v. SEP-6.

In SEP-31, the owner of the Stellar account is a business that would have an agreement with the quote-providing anchor and would be liable for misuse of the service

In SEP-6, the owner of a Stellar account an individual (unless its a custodial service, in which case an agreement could be made with the custodian) who couldn't realistically be held liable but would have to provide personal identification info.

I tried making this business vs. individual distinction with the word `entity` but maybe thats not clear enough.

ecosystem/sep-0038.md
leighmcculloch: 💯 on underscores. Not sure what I was thinking there.

leighmcculloch: Another thought is could we distinguish stellar assets from external assets by their value/identity rather than the field. I'm thinking of a company who provides an API across any number of network's. There might be existing standards we can buy into such that we could say something like base `stellar:USDC:circle.com` counter `ethereum:...`. there was some discussion about namespacing like this on the SEP-32 discussion thread, so I think this is a thing. Maybe we could leverage it.

ecosystem/sep-0038.md
```diff
@@ -0,0 +1,176 @@
...
+Version 1.0.0
+```
+
+## Summary
+
+This protocol enables anchors to accept off-chain assets in exchange for different on-chain assets, and vice versa. Specifically, it enables anchors to provide quotes that can referenced within the context of existing Stellar Ecosystem Proposals. How the exchange of assets is facilitated is outside the scope of this document.
+
+## Motivation
+
+Anchoring an asset and issuing an asset are distinct functions that have different buisness and technical requirements. However, issuing an asset has often been a prerequisite for anchoring an asset. This protocol enables anchors to transfer value on and off the Stellar network for their clients regardless of whether a one-for-one reserve-backed Stellar asset existed for the off-chain asset held by the anchor's clients.
```
marcinx, line 20: typo "buisness"

ecosystem/sep-0038.md
leighmcculloch: This was the discussion: https://groups.google.com./g/stellar-dev/c/rWThI_BIs0Y/m/H9EBrlsHBwAJ

ecosystem/sep-0038.md
leighmcculloch: On second thought I don't think SEP-32 is appropriate here. SEP-32 is targeting Uuser facing interfaces and should almost never be used in programmatic interfaces. A client should use SEP-32 to resolve an asset address to a code and issuer then use the code and issuer with any API and services it interacts. This is because a client should rey only on itself to resolve and verify an issuer account and asset address domain match.

ecosystem/sep-0038.md
leighmcculloch: Got it. This is not clear because there could be an ARS asset on Stellar too. What do you think about `iso4217:ARS` for referencing currencies? I'm not sold on the idea myself but it would make the API explicit when referring to assets or currencies or other things. Similar to my other comment.

ecosystem/sep-0038.md
leighmcculloch: Thanks, that's what I wanted to confirm that we are using that definition. We made some mistakes in the past where we mixed up counter and base in the ticket API and just wanted to check. 

Also +1 you're right price would be 100, I messed that up. 😄.

I don't think we need to worry too much about stellar asset to stellar asset. I was just using them as and sample, a bad example given that's not what this api is for.

leighmcculloch: I don't know what is better. Stick with price. I'll resolve this one.

ecosystem/sep-0038.md
JakeUrban: I like the idea of namespacing assets by provider (or network). The existing cross-chain API I looked at isn't built for off-chain assets, which is requirement for us. I like the syntax suggested in the google group thread:

```
namespace:symbol:[domain or 'native']
```

For now we don't need to define `tranche` or `instance` but we could extend the convention in the future.

For now we can document the convention in this protocol but I could see it becoming its own SEP.

leighmcculloch: We could start with only defining `stellar:` and `iso4217:` for now, and expand as needed. I don't think we need an exhaustive list.

ecosystem/sep-0038.md
```diff
@@ -35,6 +39,49 @@ All endpoints accept in requests the following `Content-Type`:
...
+  "error": "The requested asset is not supported. See GET /prices for supported assets."
+}
+```
+
+### Asset Identification Format
+
+This protocol can be used to provide quotes for any class of asset in exchange for a Stellar asset. The following format is used to identify an asset in API requests and responses.
+
+```
+namespace:symbol:issuer
```
leighmcculloch, line 69: ❗ I think you should define the format as the scheme and identifier, then for each known scheme define the identifier. It helps make this clearer to separate the definitions and layers. Scheme is the term used by URIs for that initial namespace and it helps to use the same terms.

```suggestion
<scheme>:<identifier>
```

ecosystem/sep-0038.md
```diff
@@ -35,6 +39,49 @@ All endpoints accept in requests the following `Content-Type`:
...
+Name | Description
+-----|------------
+`stellar` | Used for Stellar assets.
+[`iso4217`](https://www.iso.org/iso-4217-currency-codes.html) | Used for fiat currencies.
+
+The `symbol` value should be the Stellar `asset_code` or ISO-4217 currency code.
+
+The `issuer` value is optional because it may not be relevant for a given asset class, such as fiat currencies. For Stellar assets, `issuer` should be the public key of the asset's issuing account.
+
+When presenting assets to end users, applications should replace the Stellar public key string with the domain hosting the issuer's [SEP-1 `stellar.toml`](sep-0001.md).
```
leighmcculloch, line 83: ❗  I think this paragraph is out-of-scope for SEP-38. This SEP is concerned with the API between products and shouldn't dictate the UI of user facing applications. It should be something we leave for SEP-32 to recommend. Also because this one sentence is an entire specification on its own that needs fleshing out, and that is essentially SEP-32. Thoughts?

ecosystem/sep-0038.md
```diff
@@ -0,0 +1,219 @@
+## Preamble
+
+```
+SEP: 0038
+Title: Request for Quote API
```
tomerweller, line 5: I think the title should qualify this as an Anchor specific API. Not to be confused with an OTC style RFQ (on-chain<>on-chain) API.
Perhaps `Anchor RFQ API`? The RFQ abbreviation is fairly known term.

ecosystem/sep-0038.md
```diff
@@ -44,7 +91,7 @@ All endpoints respond with content type:
 
 ### GET Info
 
-This endpoint describes the supported Stellar assets and off-chain currencies available for trading. Note that the anchor may not support a trading pair between every Stellar asset and off-chain currency listed. Use the [`GET /prices`](#get-prices) endpoint to see which pairs are supported.
+This endpoint describes the supported Stellar assets and off-chain assets available for trading. Note that the anchor may not support a trading pair between every Stellar asset and off-chain asset listed. Use the [`GET /prices`](#get-prices) endpoint to see which pairs are supported.
```
leighmcculloch, line 94: ❔  Is on-chain/off-chain the appropriate terms to use in the Stellar ecosystem, or on-ledger/off-ledger? I don't know the answer to this myself. @tomerweller ?
```suggestion
This endpoint describes the supported Stellar assets and off-chain assets available for trading. Note that the anchor may not support a trading pair between every Stellar asset and off-ledger asset listed. Use the [`GET /prices`](#get-prices) endpoint to see which pairs are supported.
```

ecosystem/sep-0038.md
```diff
@@ -54,21 +101,21 @@ No request arguments required.
 
 Name | Type | Description
 -----|------|------------
-`assets` | array | An array of objects describing the Stellar assets available in exchange for one or more of the `currencies` listed.
-`currencies` | array | An array of objects describing the off-chain currencies available in exchange for one or more of the `assets` listed.
+`stellar` | array | An array of objects describing the Stellar assets available in exchange for one or more of the assets listed under other namespaces.
+`iso4217` | array | An array of objects describing the fiat currencies available in exchange for one or more of the `stellar` assets listed.
```
leighmcculloch, line 105: ❓ Could this be a single array, given that the values will indicate if they are stellar, or iso4217, or some other type? If they need to be separated then I think we should rename iso4217 here to 'other' or 'external' so that it will contain whatever other scheme in its values and not only iso4217 currencies.

ecosystem/sep-0038.md
```diff
@@ -54,21 +101,21 @@ No request arguments required.
...
 
 `assets` Object Schema
 
 Name | Type | Description
 -----|------|------------
-`asset` | string | The [SEP-32](sep-0032.md) address of the Stellar asset. Lumens should be specified as `XLM`.
+`asset` | string | The `asset-code:issuer-account` of the Stellar asset. Lumens should be specified as `native`.
```
leighmcculloch, line 111: 💡 Can you reference SEP-11 here?

ecosystem/sep-0038.md
```diff
@@ -156,21 +200,20 @@ If the client requests some amount of off-chain currency for providing some amou
...
-
 #### Request
 
 Name | Type | Description
 -----|------|------------
 `base` | string | Same as the definition of `base` in `GET /price`.
 `counter` | string | Same as the definition of `counter` in `GET /price`.
 `base_amount` | string | Same as the definition of `base_amount` in `GET /price`.
-`delivery_method` | string | (optional) One of the `name` values specified by the `delievery_methods` array for the relevant currency returned from `GET /info`. If the array is empty, this parameter is not required.
+`expires_after` | [UTC ISO 8601](https://www.iso.org/iso-8601-date-and-time-format.html) string | (optional) The earliest date and time the client can deliver funds to the anchor. Anchors should return `400 Bad Request` if the requested expiration cannot be provided.
```
leighmcculloch, line 210: >The earliest date and time the client can deliver funds to the anchor.

❓ Should this description say the latest time?

leighmcculloch, line 210: ❗ Because this request is declaring or asking for the quote to expire after a particular time, I think this field should be `expire_after` rather than `expires_after`. `expires_after` sounds like the request for the quote "expire**s** after" that time, when the request is signaling that it wishes the quote that will be delivered to "expire after" the time.
```suggestion
`expire_after` | [UTC ISO 8601](https://www.iso.org/iso-8601-date-and-time-format.html) string | (optional) The earliest date and time the client can deliver funds to the anchor. Anchors should return `400 Bad Request` if the requested expiration cannot be provided.
```

leighmcculloch (commented):
Questions (❓). Changes I think are necessary (❗). 

Also suggestions I don't feel strongly about (💡).

ecosystem/sep-0038.md
JakeUrban: No, this comes from a request for the client to be able to request a longer expiration period.

https://github.com/stellar/stellar-protocol/issues/901#issuecomment-819495250

ecosystem/sep-0038.md
JakeUrban: It could be a single array, but I think its helpful to separate `stellar` from `external` assets because an asset in one group can only be exchanged for an asset in the other group, and putting all objects in a single array does not communicate this distinction.

I don't feel strongly about this though.

JakeUrban: On second thought, i think using a single list is acceptable, because we still have the `GET /prices` endpoint that provides the assets that can be exchanges for an asset provided in the request.

ecosystem/sep-0038.md
JakeUrban: I'm un-resolving this thread because I think I was misunderstanding `base` and `counter`.

For a given currency pair, say `EUR/USD`, there are two parties involved. One party is buying EUR and selling USD, and the other is selling EUR and buying USD. `EUR/USD = 1.25` simply means 1 unit of `EUR` is equal to 1.25 units of USD. 

This syntax does _not_ indicate which party is selling the base currency and buying the counter currency, but this API assumes its the client. I think this can be confusing for people who understand what `base/counter` syntax actually is.

The 0x API uses `sellToken`/`sellAmount` and `buyToken`/`buyAmount`, written in the client's perspective. I think we should use this pattern as well. (`sellAsset`/`sellAmount`) Thoughts?

ecosystem/sep-0038.md
JakeUrban: I've personally never heard the term "off-ledger" but I hear "on-chain" used by Tomer frequently. Let me know if you feel strongly about using "off-ledger".

leighmcculloch: I don't feel strongly, I was just curious as much for my own vocab as for this SEP. If @tomerweller says on-chain, 👍🏻.

ecosystem/sep-0038.md
leighmcculloch: 👍🏻 I think being explicit about the sell and buy tokens would be a great idea. In that case I also wonder if we should return not only a price, but a total amount.

ecosystem/sep-0038.md
```diff
@@ -0,0 +1,200 @@
...
+`stellar` | Used for Stellar assets.
+`iso4217` | Used for fiat currencies.
+
+`stellar` Identifier Format
+
+The [SEP-11](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0011.md#asset) asset format.
+
+`iso4217` Identifier Format
+
+The [ISO 4217](https://www.iso.org/iso-4217-currency-codes.html) code for the fiat currency.
```
leighmcculloch, line 85: ❗ I think it is worth calling out that it is three character alphabetic codes. Technically ISO4217 defines both alphabetic and numeric codes, and we should be explicit here about which we want people to use. I don't think we should support both.
```suggestion
The [ISO 4217](https://www.iso.org/iso-4217-currency-codes.html) three-character currency code for the fiat currency.
```

ecosystem/sep-0038.md
```diff
@@ -0,0 +1,200 @@
...
+
+### GET Prices
+
+This endpoint can be used to fetch the [indicative](https://www.investopedia.com/terms/i/indicativequote.asp) prices of available off-chain assets in exchange for a Stellar asset and vice versa.
+
+#### Request
+
+Name | Type | Description
+-----|------|------------
+`sell_asset` | string | The asset you want to sell, using the [Asset Identification Format](#asset-identification-format).
```
leighmcculloch, line 130: ❓ Do you think we need to support `buy_asset` too? For example, what if I want to buy off-chain asset A, and I don't care what asset I have because I could use path payments to give you whatever asset you're willing to receive.

ecosystem/sep-0038.md
```diff
@@ -0,0 +1,200 @@
...
+
+#### Response
+
+Name | Type | Description
+-----|------|------------
+`buy_assets` | array | An array of objects containing information on the assets that can be exchanged for `sell_asset`.
```
leighmcculloch, line 136: 💡 Given this field is named in the clients perspective I think the description should speak from their perspective too. It reads clearer to me.
```suggestion
`buy_assets` | array | An array of objects containing information on the assets that the client will receive when they provide `sell_asset`.
```

ecosystem/sep-0038.md
```diff
@@ -0,0 +1,200 @@
...
+
+`buy_assets` Object Schema
+
+Name | Type | Description
+-----|------|------------
+`asset` | string | The [Asset Identification Format](#asset-identification-format) value.
+`price` | string | The price offered by the anchor for one unit of `asset` in terms of `sell_asset`.
```
leighmcculloch, line 143: 💡 For those coming from a financial background that understand base and counter well, this description could be a good place to use it and to say which asset is the base and which is the counter. I'd still keep the description about which asset is one unit, etc.

ecosystem/sep-0038.md
```diff
@@ -0,0 +1,200 @@
...
+
+This endpoint can be used to fetch the indicative price for a given asset pair.
+
+#### Request
+
+Name | Type | Description
+-----|------|------------
+`sell_asset` | string | The asset the client would like to sell. Ex. `USDC:circle.io`, `iso4217:ARS`
+`buy_asset` | string | The asset the client would like to exchange `sell_asset` for.
+`sell_amount` | string | The amount of `sell_asset` the client would like to sell for `buy_asset`.
```
leighmcculloch, line 156: ❓ I think there will be cases where the client wants to buy a fixed amount, should we add buy_amount too and require one of?

ecosystem/sep-0038.md
JakeUrban: It can't hurt 👍

ecosystem/sep-0038.md
```diff
@@ -0,0 +1,200 @@
...
+In the reverse scenario, the anchor will submit a Stellar transaction to deliver funds to the client as long as the client delivered off-chain funds to the anchor before the expiration. In this case, the anchor will likely increase their margin to cover the cost of submitting the transaction.
+
+#### Request
+
+Name | Type | Description
+-----|------|------------
+`sell_asset` | string | Same as the definition of `sell_asset` in `GET /price`.
+`buy_asset` | string | Same as the definition of `buy_asset` in `GET /price`.
+`sell_amount` | string | Same as the definition of `sell_amount` in `GET /price`.
+`expire_after` | [UTC ISO 8601](https://www.iso.org/iso-8601-date-and-time-format.html) string | (optional) The earliest date and time the client can deliver funds to the anchor. Anchors should return `400 Bad Request` if the requested expiration cannot be provided.
```
leighmcculloch, line 191: 🤔 I find expire_after confusing. I mentioned this [here](https://github.com/stellar/stellar-protocol/pull/908#discussion_r613611275) but I think I did a poor job of explaining.

When I use money transfer services to transfer money between countries it is really common for that service to provide two methods:
1. Immediate transfer where the price I'm given starts now and expires after X hours.
2. A scheduled transfer where the price is unknown. An example of a product that supports this is Wise, they discuss how it works [here](https://wise.com/help/articles/2978063/what-are-scheduled-transfers). One thing they say in their FAQ is:
   >...you won’t get a guaranteed rate. This is because you’re requesting that we send the money out later, and we can’t predict the future — as much as we wish we could.

The `expires_after` field reads like a scheduled transfer. I'm not convinced a SEP-38 implementer will be able to predict the future and provide a guaranteed rate in the future. 

I think `expires_after` was added in response to @erasmus's feedback [here](https://github.com/stellar/stellar-protocol/issues/901#issuecomment-819495250) but I think we may have misunderstood their request:
> I was thinking that the quote requester should be able to specify a desired expiration and get a quote from provider based on that time horizon. As a requester I may want to pay more for a longer expiry time.

I think @erasmus is requesting that the client can specify how long till the rate expires from now as a way to signal to the server they're willing to pay a premium to get longer to pay.

@erasmus please correct me if I'm incorrect.

JakeUrban: > I'm not convinced a SEP-38 implementer will be able to predict the future and provide a guaranteed rate in the future.

I'm a bit confused by this statement, because the price an anchor provides is guaranteed until `expires_at` as the protocol is written now. It is very possible that the market price for the asset pair changes before `expires_at`, such that the market price would be better for the anchor that the price quoted in the API response, but thats the name of the game. 

Anchors will likely provide more conservative rates for more volatile assets, or in cases where the `delivery_method` specified by the client is slower, allowing for more time for the market price to change.

> I think @erasmus is requesting that the client can specify how long till the rate expires from now as a way to signal to the server they're willing to pay a premium to get a rate that will be guaranteed for longer.

That was my intention with `expire_after`; the client using this parameter would be asking the anchor to provide an `expires_at` value on or after `expires_after`. The rate is still provided immediately and only valid before `expires_at`.

We could be more explicit by renaming the attribute to something like `desired_expiration`?

leighmcculloch: The name `expire_after` is great, it sounds like it does what I was asking, and not what I was thinking it did.

I was confused by the description and your comment [here](https://github.com/stellar/stellar-protocol/pull/908#discussion_r613619017).

❗ I think the description needs changing. It is not the earliest the client can provide the funds, but the latest the clients wishes to be able to provide the funds. The clients wants up to that amount of time at least to provide the funds.

ecosystem/sep-0038.md
JakeUrban: That is a possibility, but in this case the client would be at the whim of the rates offered on the DEX between the asset held and the asset the anchor expects to receive. For context, most (if not all) SEP-31 sending anchors do not use path payments for this reason, and opt to instead hold a float of the destination asset instead.

This case is also only realistic for the on-chain-to-off-chain flow, since clients likely only have one fiat currency (or any asset) off-chain that they're interested in exchanging for `buy_asset`.

ecosystem/sep-0038.md
JakeUrban: I think you're right 👍

ecosystem/sep-0038.md
leighmcculloch: Got it. I was just curious. This is something that could be added if we needed it, but we'd be better off leaving it out if there is no convincing evidence it'd be used. Thanks.

ecosystem/sep-0038.md
JakeUrban: Ah I see, thanks for poking at that again. I think the term 'latest' is confusing because the anchor could provide an `expires_at` value that is later than the requested `expire_after` value. How about this description?

> The client's desired `expires_at` date and time for the quote. Anchors should return `400 Bad Request` if the an expiration on or after the requested value cannot be provided.

leighmcculloch: I love it that description.

ecosystem/sep-0038.md
```diff
@@ -0,0 +1,210 @@
...
+The client must provide either `sell_amount` or `buy_amount`, but not both.
+
+#### Request
+
+Name | Type | Description
+-----|------|------------
+`sell_asset` | string | The asset the client would like to sell. Ex. `USDC:G...`, `iso4217:ARS`
+`buy_asset` | string | The asset the client would like to exchange for `sell_asset`.
+`sell_amount` | string | The amount of `sell_asset` the client would exchange for `buy_asset`.
+`buy_amount` | string | The amount of `buy_asset` the client would like to purchase for `sell_asset`.
```
leighmcculloch, line 159: 🔠 
```suggestion
`buy_amount` | string | The amount of `buy_asset` the client would like to purchase with `sell_asset`.
```

ecosystem/sep-0038.md
```diff
@@ -0,0 +1,210 @@
...
+
+#### Response
+
+Name | Type | Description
+-----|------|------------
+`price` | string | The price offered by the anchor for one unit of `buy_asset` in terms of `sell_asset`. In traditional finance, `buy_asset` would be referred to as the base asset and `sell_asset` as the counter asset.
+`sell_amount` | string | The amount of `sell_asset` the anchor would exchange for `buy_asset`.
+`buy_amount` | string | The amount of `buy_asset` the anchor would provide for `sell_asset`.
```
leighmcculloch, line 167: 🔠 
```suggestion
`buy_amount` | string | The amount of `buy_asset` the anchor would provide with `sell_asset`.
```

leighmcculloch (approved):
💯 I think this looks great. I look forward to seeing the feedback we get from implementers.

JakeUrban:
@leighmcculloch Looks like committing your suggestions marked your review as 'stale', can you re-approve?

I'm also wondering if we should merge this PR without feedback, given that our README states that SEPs can be merged but in a 'Draft' and 'Final Comment Period' status. I've updated this SEPs status to 'Final Comment Period' regardless.

leighmcculloch (approved):
👏🏻

leighmcculloch:
> I'm also wondering if we should merge this PR without feedback, given that our README states that SEPs can be merged but in a 'Draft'

I agree, better to merge this as a draft now, and garner feedback on the stellar-dev mailing list.

leighmcculloch (approved):

PR #911, opened 2021-04-20
SEP-31: improve grammar, phrasing, and structure

JakeUrban:
While working on a change to SEP-31, I noticed how many grammar & punctuation mistakes there were, as well as how information could be better presented. This is a low-priority PR but would be nice to get merged.

### DIFF
--- a/ecosystem/sep-0031.md
+++ b/ecosystem/sep-0031.md
@@ -30,14 +30,17 @@ At a high level, the following steps are performed to complete a transaction:
 2. The Sending Anchor sends the associated amount of funds to the Receiving Anchor's Stellar account
 3. The Receiving Anchor sends funds to the Receiving Client's financial account
 
+Typically, the Sending and Receiving Clients reside in different regulatory jurisdictions and therefore a payment between their financial accounts must be facilitated by two business entities, the Sending and Receiving Anchors, who have the necessary licenses in their respective jurisdictions.
+
 ## Example
 
-Alice in Nigeria wants to send money to Bob in Europe. Alice signs up with NigeriaPay to make this payment to send money directly into Bob’s bank account. Bob doesn’t need to do anything, or know anything about this payment, besides letting Alice know what his bank account information is. Alice only needs to deal with her anchor (NigeriaPay). NigeriaPay will utilize its European rail, enabled with EuroPay Anchor service, to move the money to EuroPay in order to deposit it into Bob’s bank account.
+Alice in Nigeria wants to send money to Bob in Europe. Alice signs up with NigeriaPay to make this payment to send money directly into Bob’s bank account. Bob doesn’t need to do anything, or know anything about this payment, besides letting Alice know what his bank account information is. Alice only needs to deal with her anchor (NigeriaPay). Alice passes this information and her money to NigeriaPay, and NigeriaPay sends those funds to the EuroPay Anchor service, and Europay deposits those funds into Bob’s bank account.
 
 ## Prerequisites
 
 * The Receiving Anchor must define `DIRECT_PAYMENT_SERVER` in their [`stellar.toml`](sep-0001.md).
 * The Sending and Reciving Anchors must create bi-lateral agreements to interoperate with each other. This differs from other protocols in that there is no concept of 'discoverability'.
+* If the Receiving Anchor requires KYC information for the Sending or Receiving Clients, the Receiving Anchor must implement [SEP-12](sep-0012.md) and define a `KYC_SERVER` in their [stellar.toml](sep-0001.md).
 
 ## Authentication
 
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0031.md
```diff
@@ -7,36 +7,47 @@ Title: Direct Payments
...
+- A **Receiving Anchor**: The business offering inbound payment processing.
+- A **Receiving Client**: The owner of the destination financial account.
 
-Alice in Nigeria wants to send money to Bob in Europe. Alice signs up with NigeriaPay to make this payment to send money directly into Bob’s bank account. Bob doesn’t need to do anything, or know anything about this payment, besides letting Alice know what his bank account information is. Alice only needs to deal with her anchor (NigeriaPay).
+At a high level, the following steps are performed to complete a transaction:
 
-NigeriaPay will utilize its European rail, enabled with EuroPay Anchor service, to move the money to EuroPay in order to deposit it into Bob’s bank account.
+1. The Sending Client sends funds from a financial account to the Sending Anchor
+2. The Sending Anchor sends the associated amount of funds to the Receiving Anchor's Stellar account
+3. The Receiving Anchor sends funds to the Receiving Client's financial account
```
leighmcculloch, line 31: 💡 Is it true that steps 1 and 3 are always off-chain, and step 2 is always on-chain? It might be worth calling out where the Stellar is and where it isn't in these steps. 

I also think it is worth being specific that the sending client will send funds from their financial account rather than a financial account. Money transfer businesses, in my experience, discourage and avoid customers sending funds from other peoples accounts.

```suggestion
1. The Sending Client sends funds from their financial account to the Sending Anchor's financial account (off-Stellar)
2. The Sending Anchor sends the funds to the Receiving Anchor's Stellar account (on-Stellar)
3. The Receiving Anchor sends funds to the Receiving Client's financial account (off-Stellar)
```

ecosystem/sep-0031.md
```diff
@@ -7,36 +7,47 @@ Title: Direct Payments
...
-* Each anchor registers a Stellar key with each counterparty anchor they interact with in order to identify themselves via [SEP-10 Web Authentication](sep-0010.md).
+## Example
 
+Alice in Nigeria wants to send money to Bob in Europe. Alice signs up with NigeriaPay to make this payment to send money directly into Bob’s bank account. Bob doesn’t need to do anything, or know anything about this payment, besides letting Alice know what his bank account information is. Alice only needs to deal with her anchor (NigeriaPay). Alice passes this information and her money to NigeriaPay, and NigeriaPay sends those funds to the EuroPay Anchor service, and Europay deposits those funds into Bob’s bank account.
+
+## Prerequisites
+
+* The Receiving Anchor must define `DIRECT_PAYMENT_SERVER` in their [`stellar.toml`](sep-0001.md).
+* The Sending and Reciving Anchors must create bi-lateral agreements to interoperate with each other. This differs from other protocols in that there is no concept of 'discoverability'.
```
leighmcculloch, line 42:
```suggestion
* The Sending and Receiving Anchors must create bi-lateral agreements to interoperate with each other. This differs from other protocols in that there is no concept of 'discoverability'.
```

leighmcculloch, line 42: 💭 If there is no concept of discoverability then the use of the stellar.toml for the API endpoint seems irrelevant. The two parties could establish their endpoints with each other without needing to lookup URLs in this file. The two parties may choose to operate over a VPN or some other mechanism. Or is the sending client an independent piece of software that knows how to talk to the SEP-31 parties and needs it?

ecosystem/sep-0031.md
```diff
@@ -385,7 +390,7 @@ Example response:
 {
   "transaction": {
       "id": "82fhs729f63dh0v4",
-      "status": "pending_info_update",
+      "status": "pending_transaction_info_update",
```
leighmcculloch, line 393: Was this a bug in the spec?

leighmcculloch (dismissed):
I think this update looks good. I included some thoughts about the leading sections below. All my comments are suggestions and I defer to you on whether they're the best thing to do or not.

I skimmed over the API changes because it looks largely like a reorganization of existing copy.

ecosystem/sep-0031.md
JakeUrban: `2` is always on Stellar and `3` is always off Stellar, but `1` could theoretically be either.

- If the Sending Client wanted to send funds from their Stellar account to a Receiving Client's Stellar account, any wallet can do that.
- If the Sending Client wanted to send funds from their off-chain account to a Receiving Client's Stellar account, only one anchor is needed and therefore this case isn't relevant to this protocol.
- If the Sending Client wanted to send funds from their on-chain account to a Receiving Client's off-chain account, its not any different from an off-chain to off-chain account. The Sending Anchor provides an interface that allows the Sending Client to provide funds.

leighmcculloch: Got it, makes sense, so like:

```suggestion
1. The Sending Client sends funds from their financial account to the Sending Anchor's financial account (off/on-Stellar)
2. The Sending Anchor sends the funds to the Receiving Anchor's Stellar account (on-Stellar)
3. The Receiving Anchor sends funds to the Receiving Client's financial account (off-Stellar)
```

ecosystem/sep-0031.md
JakeUrban: Yes.

ecosystem/sep-0031.md
JakeUrban: Maybe saying "there is no concept of 'discoverability'" is too strong. 

Signaling to the ecosystem that you support SEP-31 (i.e. making your service discoverable) is still valuable even if a bi-lateral agreement is necessary before the anchors can transact. So I think it makes sense to remove this sentence.

leighmcculloch (approved):

PR #912, opened 2021-04-20
SEP-1: Add ORG_SUPPORT_EMAIL, updated definition of ORG_OFFICIAL_EMAIL

### LINKED ISSUE #903: SEP-1: Add ORG_SUPPORT_EMAIL to TOML file
`ORG_OFFICIAL_EMAIL` is usually not a support email, so wallets usually asks the anchor directly for a support email and stores it on their service's backend. This is not idea because the anchor has to reach out to the wallet to change it, and some wallet services are frontend-only.

This is why an additional email field, `ORG_SUPPORT_EMAIL` would be helpful for anchors and clients.

JakeUrban:
resolves #903 

Adds an explicit support email attribute to the organization fields for SEP-1, and redefines the `ORG_OFFICIAL_EMAIL` to be for business inquiries.

### DIFF
--- a/ecosystem/sep-0001.md
+++ b/ecosystem/sep-0001.md
@@ -6,8 +6,8 @@ Title: stellar.toml
 Author: stellar.org
 Status: Active
 Created: 2017-10-30
-Updated: 2021-03-05
-Version: 2.2.1
+Updated: 2021-04-20
+Version: 2.3.0
 ```
 
 ## Simple Summary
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0001.md
```diff
@@ -80,7 +80,8 @@ ORG_PHONE_NUMBER_ATTESTATION | `https://` url | URL on the same domain as your `
 ORG_KEYBASE | string | A [Keybase](https://keybase.io/) account name for your organization. Should contain proof of ownership of any public online accounts you list here, including your organization's domain.
 ORG_TWITTER | string | Your organization's Twitter account
 ORG_GITHUB | string | Your organization's Github account
-ORG_OFFICIAL_EMAIL | string | An email where clients can contact your organization. Must be hosted at your `ORG_URL` domain.
+ORG_OFFICIAL_EMAIL | string | An email that business partners such as wallets, exchanges, or anchors can use to contact your organization. Must be hosted at your `ORG_URL` domain.
+ORG_SUPPORT_EMAIL | string | An email that users can use to request support regarding your Stellar assets or applications.
```
leighmcculloch, line 84: ❓ Should this also be hosted at the `ORG_URL`? I'm not advocating for it, more wondering why we require it on one email and not the other.

leighmcculloch (approved):
Looks great. One question, but regardless I think this is 👍🏻.

ecosystem/sep-0001.md
JakeUrban: I left that statement because I also wasn't sure why its there in the first place. It seems to be overly opinionated.

JakeUrban: It was added by Tom Q in https://github.com/stellar/stellar-protocol/commit/f7f69bb8f02142b38a0abb7bfc9483d36822feb0, which has no commit body or PR explaining. I think it was likely to encourage organizations to have official emails.

leighmcculloch: Agreed, and not particularly useful.

JakeUrban: I could easily see a support email going to a ticketing system that doesn't use the same domain as `ORG_URL`.

leighmcculloch: The goal might be authenticity and encouraging good best practices for building trust.

leighmcculloch: Support email seems less critical to have this. I think it is fine to leave it off for it.

dmgmyza:
Perfect! Thank you, @JakeUrban @leighmcculloch

PR #913, opened 2021-04-21
SEP: Record dependencies in template

leighmcculloch:
### What
Record dependencies in the template of SEPs. Add a `Dependencies` section to the template that should be completed for new SEPs and may be added to existing SEPs if the authors of those SEPs choose.

### Why
To make it clear from first-read when there are other SEPs or CAPs that an implementer needs to be aware of and prepared to implement. In some SEPs these dependencies are buried in the text and you have to build a mental model of the dependencies you need to read an understand as you go. I think it would be convenient if these were called out up-front.

### Notes
I've updated one SEP in this PR as an example.

### DIFF
--- a/ecosystem/sep-0030.md
+++ b/ecosystem/sep-0030.md
@@ -10,6 +10,7 @@ Discussion: https://groups.google.com/forum/#!topic/stellar-dev/SFr2dHBZlsY
 Created: 2019-12-20
 Updated: 2020-11-17
 Version: 0.8.0
+Dependencies: SEP-10
 ```
 
 ## Summary
--- a/ecosystem/sep-0030.md
+++ b/ecosystem/sep-0030.md
@@ -10,12 +10,15 @@ Discussion: https://groups.google.com/forum/#!topic/stellar-dev/SFr2dHBZlsY
 Created: 2019-12-20
 Updated: 2020-11-17
 Version: 0.8.0
-Depends-On: SEP-10
 ```
 
 ## Summary
 
 This protocol defines an API that enables an individual (e.g., a user or wallet) to regain access to a Stellar account that it owns after the individual has lost its private key without providing any third party control of the account. Using this protocol, the user or wallet will preregister the account and a phone number, email, or other form of authentication with one or more servers implementing the protocol and add those servers as signers of the account. If two or more servers are used with appropriate signer configuration no individual server will have control of the account, but collectively, they may help the individual recover access to the account. The protocol also enables individuals to pass control of a Stellar account to another individual.
+
+## Dependencies
+
+Implementers and clients of this protocol must also implement [SEP-10].
 
 ## Motivation
 
--- a/ecosystem/sep-0030.md
+++ b/ecosystem/sep-0030.md
@@ -10,7 +10,7 @@ Discussion: https://groups.google.com/forum/#!topic/stellar-dev/SFr2dHBZlsY
 Created: 2019-12-20
 Updated: 2020-11-17
 Version: 0.8.0
-Dependencies: SEP-10
+Depends-On: SEP-10
 ```
 
 ## Summary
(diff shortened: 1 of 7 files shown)


leighmcculloch:
This PR is intended to launch discussion on if this idea is worthwhile or not.

JakeUrban:
Great idea, I agree dependencies are most often mentioned in-passing, and it would be better to be explicit.

I think the `Depends-On` field could work, but I think a `Dependencies` section in each SEP is better because the relationships between SEPs are more complex than what can be modeled using lists.

For example, SEP-31 does not directly depend on SEP-9. It depends on SEP-12, which depends on SEP-9. Should SEP-9 still be listed in SEP-31 `Depends-On` list? 

SEPs can also be optional dependencies. SEP-10 and SEP-12 aren't required in SEP-6, but they could be used. 
These nuances can be explained in their own section.

ire-and-curses:
Definitely in favour of identifying dependent CAPs and SEPs. I don't have a strong opinion on the implementation of that.

leighmcculloch:
I've updated this change to move the data to a section rather than meta.

sep-template.md
```diff
@@ -14,6 +14,11 @@ Discussion: <link to where discussion for this SEP is taking place, typically th
 "If you can't explain it simply, you don't understand it well enough." Provide a simplified and
 layman-accessible explanation of the SEP.
 
+## Dependencies
+Should clearly list the dependencies this SEP has on other SEPs or CAPs. May
+go into further discussion about the circumstances of the dependency,
+although a list sufficient.
+
```
JakeUrban, line 21: nit: formatting

```suggestion
## Dependencies

Should clearly list the dependencies this SEP has on other SEPs or CAPs. May go into further discussion about the circumstances of the dependency, although a list sufficient.
```

ecosystem/sep-0030.md
```diff
@@ -16,6 +16,10 @@ Version: 0.8.0
 
 This protocol defines an API that enables an individual (e.g., a user or wallet) to regain access to a Stellar account that it owns after the individual has lost its private key without providing any third party control of the account. Using this protocol, the user or wallet will preregister the account and a phone number, email, or other form of authentication with one or more servers implementing the protocol and add those servers as signers of the account. If two or more servers are used with appropriate signer configuration no individual server will have control of the account, but collectively, they may help the individual recover access to the account. The protocol also enables individuals to pass control of a Stellar account to another individual.
 
+## Dependencies
+
+Implementers and clients of this protocol must also implement [SEP-10].
```
JakeUrban, line 21:
```suggestion
## Dependencies

Implementers and clients of this protocol must also implement [SEP-10](sep-0010.md).
```

JakeUrban (approved):
Approved with some formatting suggestions. Ping me when you merge them for a re-approve.

ecosystem/sep-0030.md
leighmcculloch: The `(sep-0010.md)` is unnecessary. The `[SEP-10]` link is defined in a footnote.

JakeUrban (dismissed):

sep-template.md
leighmcculloch: I committed this, then removed it again, because the new line at the end is needed, and in the template there are no new lines under headings so this would be inconsistent.

sep-template.md
```diff
@@ -14,6 +14,9 @@ Discussion: <link to where discussion for this SEP is taking place, typically th
 "If you can't explain it simply, you don't understand it well enough." Provide a simplified and
 layman-accessible explanation of the SEP.
 
+## Dependencies
+Should clearly list the dependencies this SEP has on other SEPs or CAPs. May go into further discussion about the circumstances of the dependency, although a list sufficient.
+
```
lydiat, line 19: a list "is" sufficient

leighmcculloch: Thanks, fixed in 4f7b7f9f9877d37af7f8e481c903e3fe4d350b7f.

rice2000 (dismissed):
I think this will prove to be really helpful.  And I like the `Dependencies` section that allows for elaboration if and when necessary.  Smart!

JakeUrban (dismissed):

lydiat (approved):

PR #919, opened 2021-04-28
Updating SEP-1 title for clarity and usability

lydiat:
**Context**

**Before: stellar.toml
After: Stellar Info File**
"Stellar Info File" clearly indicates the purpose of this protocol, which was originally named simply after the file name itself.

Other names considered were Stellar Discovery File and Stellar Lookup; however, “discovery” and “lookup” don’t cover all of the information presented. 

While there was some concern over the similarity of “Info” to the `/info` endpoint, in the end we went with what felt most natural to say and was most descriptive.


JakeUrban (approved):
This PR is about updating SEP-1, but other unrelated changes are made here too. It would've been ideal if the `README.md` file was updated in each PR instead of all at once here. Not a blocker but FYI.

PR #921, opened 2021-04-28
Updating SEP-9 title for clarity and usability

lydiat:
**Context**

**Before: Standard KYC / AML fields
After: Standard KYC Fields**
This title changes only slightly in that “AML” will be removed, as Anti Money Laundering (AML) is only one of many reasons KYC is collected. 

Other names considered were KYC Data and Know Your Customer (KYC) Data. Including KYB was also considered, but KYC was identified as the most broadly used terminology. It was noted that in the future there may be a broader user for this info, in which case referring to the user as a customer would no longer be correct, but it was agreed the title could easily be revisited at that time. Additionally, the use of “Standards” was questioned, but as these are in fact SDF’s requirements and not determined by the companies using them it was deemed valid to refer to them as standards.

### DIFF
--- a/ecosystem/sep-0009.md
+++ b/ecosystem/sep-0009.md
@@ -2,7 +2,7 @@
 
 ```
 SEP: 0009
-Title: Standard KYC / AML fields
+Title: Standard KYC Fields
 Author: stellar.org
 Status: Active
 Created: 2018-07-27


leighmcculloch (approved):

JakeUrban (dismissed):

lydiat (commented):
Not sure why I'm re-approving my own changes

JakeUrban (approved):

PR #922, opened 2021-04-28
Updating SEP-10 title for clarity and usability

lydiat:
**Context**

**Before: Stellar Web Authentication
After: Stellar Authentication**
 This title would change only slightly, removing the “Web” as it’s superfluous since all authorization of this type occurs on the web. SEP-10 wasn’t in the original list of SEPs being considered until discussion moved towards supporting shorter titles that increased usability. 

Another name considered was StellarAuth. However, that’s also the name of a Stellar 2FA system, and the full “Authentication” was chosen in order to differentiate. However, it’s likely it will still be shortened colloquially in the same vein as OAuth.

### DIFF
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -2,7 +2,7 @@
 
 ```
 SEP: 0010
-Title: Stellar Web Authentication
+Title: Stellar Authentication
 Author: Sergey Nebolsin <@nebolsin>, Tom Quisel <tom.quisel@gmail.com>, Leigh McCulloch <@leighmcculloch>, Jake Urban <jake@stellar.org>
 Status: Active
 Created: 2018-07-31


leighmcculloch (approved):

leighmcculloch:
Reposting here what I posted about the name on the mailing list:
>I wonder if we should consider another name.
>
>Stellar Authentication sounds accurate to me, but since the term authentication could be used to describe other aspects of the Stellar protocol, such as signing transactions, I wonder if it is not specific enough.
>
>Having 'Web' as part of the name helped to set it apart. If we had settled on Stellar Auth that probably would have also helped set it apart, as it would have sounded similar to OAuth.
>
>Another name that might focus more on the use case and less on the technology is "Login with Stellar" which describes in reality what SEP-10 is used for. Thoughts?

lydiat:
This is a good point - `Stellar Authentication` could cover an array of operations. My concern with `Login with Stellar` is first that it's a phrase/imperative (a request or demand) instead of a descriptive noun and second that I don't find it obvious what's being logged INTO with Stellar (which would also negate the non-imperative `Stellar Login`).

In a Slack conversation @leighmcculloch  noted that other entities that offer authentication have similar names e.g. `Facebook Login` and `Google Sign-in.` I believe these names fit as they allow users of these applications to reuse their accounts with other applications that require login. While SEP-10 has the same mechanism of exchanging challenges, in my mind an anchor isn't an application to be logged into as the token received is used for API calls to its endpoints. 

During the naming discovery process @tyvdh offered a couple alternatives: `Stellar OAuth` to recognize the open aspect and `Stellar Session` or `Stellar Auth Token`, which identifies the end result. I think any of these could work, but I'm open to being persuaded by other options!

leighmcculloch:
Stellar Session is a little too generic like Stellar Authentication, and  OAuth is a [specific thing](https://oauth.net) that SEP-10 is not, so I don't think Steller OAuth is suitable.

Stellar Auth Token is okay, but SEP-10 is more than a specification about the format of a token. For example, RFC7519 is a specification about the JSON Web Token, and it is entirely about the format and contents of the token. There are other specifications that define what to do with the token.

In regards to `Login with Stellar` being a phrase. I agree, that's a downside of it. I think we should consider that a tradeoff though, and therefore consider does it deliver more value than it trades off. Speaking subjectively I think the name is clear about what the SEP is about, the use case, and the scope.

leighmcculloch:
@lydiat Was the plan to still rename SEP-10?

PR #923, opened 2021-04-28
Updating SEP-12 title for clarity and usability

lydiat:
**Context**

**Before: Anchor/Client customer info transfer
After: KYC API**
KYC API is much easier to say than “Anchor/Client customer info transfer” and while it’s comprised solely of acronyms, they’re both widely understood in the financial development industry. 

Spelling out Know Your Customer (KYC) in the title was considered as an educational opportunity, but it was decided that this work could be done in the body of the protocol.


leighmcculloch (approved):

JakeUrban (changes requested):
Title looks good, but can you update the meta info as well?

lydiat (commented):
👍

JakeUrban (approved):

PR #924, opened 2021-04-28
Updating SEP-24 title for clarity and usability

lydiat:
**Context**

**Before: Interactive Anchor/Wallet Asset Transfer Server
After: Hosted Deposit and Withdrawal**
The naming considerations of SEP-24 were strongly tied to that of SEP-6, as they’re similar protocols with one major difference, that of which party hosts the collection of KYC. 

Other names considered were Interactive Asset Transfer, Webview Deposit and Withdrawal API, Anchor/Wallet Deposit API for Web, and Deposit and Withdrawal with Web UI. The usage of “on/off ramp” was noted as easier to say but less descriptive. Both “interactive” and “web ui” were considered as ways to indicate that the user would leave the client (wallet) in order to interact with the server (anchor) while submitting KYC info; “hosted” was chosen as it’s commonly used in the financial industry as “hosted checkout.” Additionally, using the words “simple” and “complex” to distinguish between the protocols was explored but found to be inaccurate as they could be seen to indicate that one protocol is more difficult to build. Finally, it should be noted that while “deposit and withdrawal” is unwieldy, there’s no perfect English word that covers those transactions in a way that meaningfully distinguishes this protocol from payments.

A common explanation made during discussions was the comparison of SEP-6 to Stripe (a payments API) and SEP-24 to PayPal (a web UI exposed to users). More succinctly, Stripe is a Payments API and PayPal is a hosted payments page. (Thanks to @JakeUrban and @kylemccollom for introducing these useful metaphors.)

### DIFF
--- a/ecosystem/sep-0024.md
+++ b/ecosystem/sep-0024.md
@@ -6,8 +6,8 @@ Title: Hosted Deposit and Withdrawal
 Author: SDF
 Status: Active
 Created: 2019-09-18
-Updated: 2021-03-29
-Version 1.3.1
+Updated: 2021-05-17
+Version 1.3.2
 ```
 
 ## Simple Summary
--- a/ecosystem/sep-0024.md
+++ b/ecosystem/sep-0024.md
@@ -2,7 +2,7 @@
 
 ```
 SEP: 0024
-Title: Interactive Anchor/Wallet Asset Transfer Server
+Title: Hosted Deposit and Withdrawal
 Author: SDF
 Status: Active
 Created: 2019-09-18


JakeUrban (approved):
I think this is the right approach.

Participants already active in the Stellar ecosystem will likely continue to use the 'SEP-24' name, but the 'Hosted' keyword helps organizations outside the ecosystem relate this protocol to other payment solutions that use the same word.

The gap is organizations who are unfamiliar with the 'Hosted' term and are new to the ecosystem. I think this is acceptable because these types of organizations likely need to come up to speed on a range of topics, both related to Stellar and the payments space generally.

JakeUrban (dismissed):
Actually, can you please update the `Updated` and `Version` to reflect the change? The patch version should be incremented.

lydiat:
> Actually, can you please update the `Updated` and `Version` to reflect the change? The patch version should be incremented.

Absolutely - should I do this across all the change PRs?

leighmcculloch (dismissed):

leighmcculloch:
I don't think we need to change updated/version for meta changes, but If others prefer them to be bumped, I think that is fine too.

@lydiat One thing that is missing from these PRs is updates to the `ecosystem/README.md` file which has a table that lists all of the SEPs and their titles.

Linopena318 (dismissed):

JakeUrban:
@lydiat 

> Absolutely - should I do this across all the change PRs?

Yes please

lydiat:
I'll wait to change the update timestamp until we're ready to merge all of these. If I recall correctly patch versioning is usually for bugs, so if I edit anything wouldn't it be the minor version?

leighmcculloch:
We usually change the patch version when something changes that doesn't functionally change the specification, because the change is much like an internal change in software. A title change would fit under that I think because there is nothing functionally changing. I also think it is fine to think of this as the title is changing because it had a problem.

leighmcculloch (approved):

PR #925, opened 2021-04-28
Updating SEP-31 title for clarity and usability

lydiat:
**Context**

**Before: Direct Payments
After: Cross-border Payments API**
The proposed new title of SEP-31 is “Cross-border Payments API.” The usage of "direct" didn't provide any additional clarity and could be confused with the type of direct payment performed by consumers with traditional banks.

Other names considered were Anchor-Facilitated Cross Border Payments, Anchor-Facilitated Cross-border Transfer, Mediated Payments, and Bilateral Payments. The emphasis on anchor, while clearly indicating that a payment is facilitated by anchors, is unwieldy and can be inaccurate as a party doesn’t have to be an anchor (e.g. they can be a custodial wallet with a bilateral relationship with the receiver). While “cross-border” technically at some point in the future could be a misnomer, right now it’s an important indication that the payments must be sent between two independent parties as neither has jurisdiction in the other’s country. The addition of “API” is meant to make clear the scope of the protocol.

### DIFF
--- a/ecosystem/sep-0031.md
+++ b/ecosystem/sep-0031.md
@@ -3,7 +3,7 @@
 
 ```
 SEP: 0031
-Title: Direct Payments
+Title: Cross-border Payments API
 Author: SDF
 Status: Active
 Created: 2020-04-07
--- a/ecosystem/sep-0031.md
+++ b/ecosystem/sep-0031.md
@@ -7,8 +7,8 @@ Title: Cross-Border Payments API
 Author: SDF
 Status: Active
 Created: 2020-04-07
-Updated: 2021-04-20
-Version 1.2.1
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0031.md
```diff
@@ -3,7 +3,7 @@
 
 ```
 SEP: 0031
-Title: Direct Payments
+Title: Cross-border Payments API
```
leighmcculloch, line 6: 🔠  I believe border should be capitalized. Ref: https://capitalizemytitle.com
```suggestion
Title: Cross-Border Payments API
```

leighmcculloch (approved):
One typo 🔠  noted below, otherwise looks great 🎉 .

PR #927, opened 2021-04-28
SEP-38: add a GET /quote endpoint

JakeUrban:
In the original PR for creating this protocol, we opted to not add a `GET /quote/:id` endpoint with the rationale that the client could store previously provided quotes in their own persistent storage.

However, this assumption places an unnecessary requirement on client-side implementations. It also means that a user would only be able to fetch quote details from the application used to originally request the quote.

### DIFF
--- a/ecosystem/sep-0038.md
+++ b/ecosystem/sep-0038.md
@@ -7,8 +7,9 @@ Author: Jake Urban <@jakeurban> and Leigh McCulloch <@leighmcculloch>
 Track: Standard
 Status: Draft
 Created: 2021-04-09
+Updated: 2021-04-28
 Discussion: https://github.com/stellar/stellar-protocol/issues/901
-Version 1.0.0
+Version 1.1.0
 ```
 
 ## Summary
--- a/ecosystem/sep-0038.md
+++ b/ecosystem/sep-0038.md
@@ -213,7 +213,7 @@ Name | Type | Description
 
 ### GET Quote
 
-This endpoint can be used to fetch a previously-provided firm quote. Expired quotes should still be retrievable using this endpoint.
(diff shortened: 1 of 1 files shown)


Linopena318 (approved):

ecosystem/sep-0038.md
```diff
@@ -208,3 +210,26 @@ Name | Type | Description
 `price` | string | The price offered by the anchor for one unit of `buy_asset` in terms of `sell_asset`. In traditional finance, `buy_asset` would be referred to as the base asset and `sell_asset` as the counter asset.
 `buy_amount` | string | The amount of `buy_asset` to be exchanged for `sell_asset`. `price * sell_amount = buy_amount` must be true up to the number of decimals required for `buy_asset`.
 `sell_amount` | string | The amount of `sell_asset` to be exchanged for `buy_asset`.
+
+### GET Quote
+
+This endpoint can be used to fetch a previously-provided firm quote. Expired quotes should still be retrievable using this endpoint.
```
leighmcculloch, line 216: >Expired quotes should still be retrievable using this endpoint.

❓ What's the motivation for this? It is convenient for clients but at a cost for servers. I think quotes should be transient so that there is not undue stress on servers to maintain large amounts of historical data.

leighmcculloch (approved):
One question ❓ that might result in a small change, but regardless 👍🏻 looks great.

Linopena318 (dismissed):

ecosystem/sep-0038.md
JakeUrban: The primary motivation is to keep record of quotes used by other protocols, but we could allow unused quotes to be removed at expiration.

leighmcculloch (approved):

Linopena318 (commented):
It seems all

PR #931, opened 2021-04-30
SEP-38: Add request examples

JakeUrban:
I thought it would be helpful to include examples of the requests and responses for each endpoint, similar to what we have in SEP-31. 

In the process of writing these, I realized that the formula described in the `buy_amount` attribute description was incorrect based on the definition of `price` we've defined. I updated the patch version, but I could see an argument for a minor version increment.

### DIFF
--- a/ecosystem/sep-0038.md
+++ b/ecosystem/sep-0038.md
@@ -7,9 +7,9 @@ Author: Jake Urban <@jakeurban> and Leigh McCulloch <@leighmcculloch>
 Track: Standard
 Status: Draft
 Created: 2021-04-09
-Updated: 2021-04-28
+Updated: 2021-04-30
 Discussion: https://github.com/stellar/stellar-protocol/issues/901
-Version 1.1.0
+Version 1.1.1
 ```
 
 ## Summary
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0038.md
```diff
@@ -121,6 +133,40 @@ Name | Type | Description
 `name` | string | The value to use when making `POST /quote` requests.
 `description` | string | A human readable description of the method identified by `name`.
 
+##### Examples
+
+```
+GET /info
+
```
leighmcculloch, line 140: 💡 I think it is easier to read these if you make them valid json.
```suggestion
`GET /info`:

```json
```

ecosystem/sep-0038.md
```diff
@@ -121,6 +133,40 @@ Name | Type | Description
 `name` | string | The value to use when making `POST /quote` requests.
 `description` | string | A human readable description of the method identified by `name`.
 
+##### Examples
```
leighmcculloch, line 136: 💡 Looks like there is only one.
```suggestion
##### Example
```

leighmcculloch (approved):

leighmcculloch (approved):

Issue #917, opened 2021-04-26
SEP-6: Integrate SEP-38


JakeUrban:
Now that we have a standard for providing clients quotes between on and off chain assets, we should integrate this functionality into SEP-6.

For withdrawals, the user will provide the anchor with an on-chain asset A and receive a different off-chain asset B. For deposits, the reverse will occur. This will be made possible by the client fetching a quote and referencing the quote ID in the SEP-6 deposit or withdraw request.

shanzzam:
Removing the stale label and putting this in Product Eng team backlog for planning

PR #932, opened 2021-04-30
SEP-31: integrate SEP-38

### LINKED ISSUE #916: SEP-31: Integrate SEP-38 RFQ Quotes
Now that we have a standard for providing clients quotes between on and off chain assets, we should integrate this functionality into SEP-31.

The basic idea is that the sending anchor would send an on-chain asset A and the receiving anchor would deliver an off-chain asset B at a rate the receiving anchor provided the sending anchor in a previous request.

JakeUrban:
Resolves #916

Integrates SEP-38 into SEP-31.

### Changes

- Clarifies the assets listed in a `GET /info` response to be the Stellar assets the Receiving Anchor accepts from a Sending Anchor. This makes the API consistent with SEP-24 and SEP-6
- Adds a "Asset Conversions and Liquidity Strategies" section outlining the three methods possible for sending and converting assets
- Adds the following parameters to `POST /transactions`:
    - `destination_asset`: used when the Receiving Anchor does not support providing firm quotes
    - `quote_id`: used when the Receiving Anchor supports firm quotes
- Makes the following `POST /transactions` parameters optional:
    - `amount_in`
    - `asset_code`
        - neither of these parameters are needed when `quote_id` is provided
- Adds the following attributes to the `GET /transactions/:id` response:
    - `amount_in_asset`
    - `amount_out_asset`
    - `amount_fee_asset`


ecosystem/sep-0031.md
```diff
@@ -68,6 +68,30 @@ All endpoints respond with content type:
...
+
+This approach requires a sufficiently liquid market to exist between the source and destination assets on the Stellar Decentralized Exchange (SDEX). Note that the exchange rate between the two assets is dynamic due to market fluctuations on the SDEX.
+
+### Receiving Anchor Asset Conversion
+
+Finally, the Receiving Anchor can receive payments of the source asset on Stellar and deliver payments of the destination asset to the Receiving Client. 
+
+This requires the Receiving Anchor to implement the [SEP-38 Anchor RFQ API](sep-0038.md), which enables the Receiving Anchor to provide quotes to the Sending Anchor. These quotes can be [indicative](https://www.investopedia.com/terms/i/indicativequote.asp) or [firm](https://www.investopedia.com/terms/f/firmquote.asp). If a firm quote is used, the Stellar transaction _must_ be submitted to the Stellar network prior to the quote's expiration. More specifically, the `created_at` timestamp returned in the [Horizon GET /transactions/:id](https://developers.stellar.org/api/resources/transactions/) request must be earlier than the quote's expiration.
+
```
tomerweller, line 94: This is really great. We should have @phil-stellar and @lisaclaire828 take a look as well. I would consider the following renaming of strategies to reflect *when* the asset conversion happens: 

`Sending Anchor Asset Conversion` -> `Pre-payment asset conversion`

`Stellar Network Asset Conversion`-> `Real time asset conversion` 

`Receiving Anchor Asset Conversion` -> `Post-payment asset conversion`

ecosystem/sep-0031.md
```diff
@@ -68,6 +68,30 @@ All endpoints respond with content type:
 
 - `application/json`
 
+## Asset Conversions and Liquidity Strategies
+
+This protocol can be used if the Sending and Receiving Client wish to send and receive the same asset, but this is typically not the case. It is more common for the Sending Client to provide the Sending Anchor with a particular asset (the source asset) and have the Receiving Client receive a different asset (the destination asset).
```
leighmcculloch, line 73: 💡 This paragraph is a little difficult to digest because it leads with the least important information and finishes with the most important information. The most important information is what the protocol is intended to be used for. Suggest reordering:
```suggestion
This protocol is designed for the common case where the Sending Client provides the Sending Anchor with a particular asset (the source asset) and the Receiving Client receives a different asset (the destination asset). It is possible to use this protocol when the Sending and Receiving clients wish to send and receive the same asset, but that is not the common use case it has been designed.
```

ecosystem/sep-0031.md
```diff
@@ -68,6 +68,30 @@ All endpoints respond with content type:
...
+
+In this case, the source asset must convert into the receiving asset at some point prior to delivering funds to the Receiving Client. This conversion must be done by either the Sending Anchor, the Stellar Network, or the Receiving Anchor. The subsections below outline the stratgies used for each method.
```
leighmcculloch, line 75: >In this case,

💡  It's ambiguous at the start of the sentence what "this case" is because the prior paragraph discusses two cases. The reader has to look ahead towards to rest of the paragraph to figure it out. I think this would make the paragraph clearer:

```suggestion
When the source asset and destination assets are different, the source asset must convert into the receiving asset at some point prior to delivering funds to the Receiving Client. This conversion must be done by either the Sending Anchor, the Stellar Network, or the Receiving Anchor. The subsections below outline the strategies used for each method.
```

🔠 I also fixed a typo in this paragraph.

ecosystem/sep-0031.md
```diff
@@ -68,6 +68,30 @@ All endpoints respond with content type:
...
+
+### Sending Anchor Asset Conversion
+
+The Sending Anchor can collect the source asset from the Sending Client and send the destination asset to the Receiving Anchor. 
+
+This approach requires the Sending Anchor to provide an exchange rate, or quote, to the Sending Client. It also requies the Sending Anchor to hold a balance of the destination asset on Stellar. The Sending Anchor would use Payment operations to send the quoted amount of the destination asset to the Receiving Anchor.
+
+### Stellar Network Asset Conversion
+
+Instead of holding a balance of the destination asset on Stellar, the Sending Anchor can instead us Path Payments to convert the source asset to the destination asset on the Stellar Network.
```
leighmcculloch, line 85: 🔠  There's a double instead in this sentence, and a typo.
```suggestion
Instead of holding a balance of the destination asset on Stellar, the Sending Anchor can use Path Payments to convert the source asset to the destination asset on the Stellar Network. 
```

ecosystem/sep-0031.md
```diff
@@ -68,6 +68,30 @@ All endpoints respond with content type:
...
+
+The Sending Anchor can collect the source asset from the Sending Client and send the destination asset to the Receiving Anchor. 
+
+This approach requires the Sending Anchor to provide an exchange rate, or quote, to the Sending Client. It also requies the Sending Anchor to hold a balance of the destination asset on Stellar. The Sending Anchor would use Payment operations to send the quoted amount of the destination asset to the Receiving Anchor.
+
+### Stellar Network Asset Conversion
+
+Instead of holding a balance of the destination asset on Stellar, the Sending Anchor can instead us Path Payments to convert the source asset to the destination asset on the Stellar Network. 
+
+This approach requires a sufficiently liquid market to exist between the source and destination assets on the Stellar Decentralized Exchange (SDEX). Note that the exchange rate between the two assets is dynamic due to market fluctuations on the SDEX.
```
leighmcculloch, line 87: >Note that the exchange rate between the two assets is dynamic due to market fluctuations on the SDEX.

❓ This is mentioned here as if it is unique to the SDEX, but in all cases the rate between two assets will be fluid. What's the intention with calling it out for only the SDEX?

ecosystem/sep-0031.md
```diff
@@ -68,6 +68,30 @@ All endpoints respond with content type:
...
+
+### Receiving Anchor Asset Conversion
+
+Finally, the Receiving Anchor can receive payments of the source asset on Stellar and deliver payments of the destination asset to the Receiving Client. 
+
+This requires the Receiving Anchor to implement the [SEP-38 Anchor RFQ API](sep-0038.md), which enables the Receiving Anchor to provide quotes to the Sending Anchor. These quotes can be [indicative](https://www.investopedia.com/terms/i/indicativequote.asp) or [firm](https://www.investopedia.com/terms/f/firmquote.asp). If a firm quote is used, the Stellar transaction _must_ be submitted to the Stellar network prior to the quote's expiration. More specifically, the `created_at` timestamp returned in the [Horizon GET /transactions/:id](https://developers.stellar.org/api/resources/transactions/) request must be earlier than the quote's expiration.
```
leighmcculloch, line 93: 💭 There are some subtleties regarding submitting transactions, the created at timestamp, and how long they are valid and can be resubmitted that we might need to consider in this scenario.

The transaction not only needs to be submitted before the quote's expiration, but it needs to be included in a closed ledger. The created at timestamp on transactions is the nomination time of the ledger (the time the ledger closed) that the transaction was included in. It will be possible for a transaction to be queued, and so it could be submitted prior to the expiry time but not included in the next ledger but rather a future ledger until after the close time. This may be confusing to implementers. This is an edge case, but we should probably still document what the two parties should do if this happens. Implementers could also reduce the chance of this happening by setting the time bounds on the transaction to be the expiry time or earlier.

ecosystem/sep-0031.md
leighmcculloch: I think those names somewhat obfuscate who is responsible for asset conversion, and if I read them without context of what's in each section I don't really know what they mean.

leighmcculloch (commented):
I've left some comments inline.

 - Typos 🔠 
 - Suggestions 💡 I defer to you on, they help me with reading it, but may be subjective.
 - Questions ❓ I defer to you on.
 - Thoughts 💭 about an area of transaction submission that I don't think this addresses. I think a change is needed, but I defer to you if this is needed right now.

ecosystem/sep-0031.md
JakeUrban: > Implementers could also reduce the chance of this happening by setting the time bounds on the transaction to be the expiry time or earlier.

If they do this, there is actually _no_ chance a transaction would be submitted earlier and included later than the expiration right? That detail may be important enough to explicitly mention here.

leighmcculloch: I think it's a little fuzzy, I'd verify it with the core team, but essentially yes.

ecosystem/sep-0031.md
JakeUrban: Fair point. I can remove it 👍

ecosystem/sep-0031.md
JakeUrban: I agree with Leigh, I think the _who_ is more important than the _when_

ecosystem/sep-0031.md
```diff
@@ -189,7 +215,9 @@ The response should be a JSON object like:
 }
 ```
 
-The JSON object contains an entry for each currency that the anchor supports for receiving and completing a direct payment.
+The JSON object contains an entry for each Stellar asset that the Receiving Anchor supports receiving from the Sending Anchor. 
+
+If the Receiving Anchor supports [SEP-38](sep-0038.md), the Sending Client can check the [`SEP-38 GET /prices`](sep-0038.md#get-prices) endpoint to determine which off-chain assets the Receiving Anchor can deliver to recipients in exchange for an asset listed in the [`SEP-31 GET /info` response](#get-info). If the Receiving Anchor does not provide exchange rates for an asset, it is assumed that the Stellar asset can be exchanged 1-for-1 with the corresponding off-chain asset.
```
marcelosalloum, line 220: > If the Receiving Anchor does not provide exchange rates for an asset, it is assumed that the Stellar asset can be exchanged 1-for-1 with the corresponding off-chain asset.

Is that safe to assume? Are there anchors in the ecosystem operating that way?

JakeUrban: All existing anchors operate this way, since the Stellar asset they use is pegged to the fiat currency their users have.

marcelosalloum (commented):
This plan sounds good but there are a few details missing IMO:

In SEP-38 how will we indicate the anchor supports the SEP? I don't see any proposed toml parameter [there](https://github.com/marcelosalloum/stellar-protocol/blob/97dd8972672d0c44eea5f6ee3a1c3d3663b3b25f/ecosystem/sep-0038.md).

If a Receiving Anchor's toml file indicates both SEP-38 and SEP-31 are implemented separately, how do we guarantee SEP-31 is supporting SEP-38? According to this proposal the Sending Anchor would need to submit a `POST /transactions` request with `quote_id` to check if it succeeds.

JakeUrban:
@marcelosalloum responding to your comments:

> In SEP-38 how will we indicate the anchor supports the SEP? I don't see any proposed toml parameter there.

Good point. We can add a prerequisites section like we do in SEP-31 mentioning the need to have the `ANCHOR_QUOTE_SERVER` attribute in the stellar.toml, which is already defined in SEP-1.

> how do we guarantee SEP-31 is supporting SEP-38?

Thats a good question, I think there are actually two questions here:
- Does SEP-31 support SEP-38 quotes at all?
- For what assets does SEP-31 support using quotes for?

Personally, I don't see why an anchor would support SEP-38 but not support quotes to be used in the context of SEP-31 transactions. This was an assumption I made when writing this PR: if the anchor has an `ANCHOR_QUOTE_SERVER` attribute in the toml (indicating SEP-38 support), then the quotes returned from that protocol could be used in any other protocol also defined (i.e. SEP-31 and SEP-6). Maybe this is not a safe assumption to make, in which case we could add some kind of `quotes_supported: true` attribute to the SEP-31 and SEP-6 `/info` endpoints.

But then there is also the question of whether a particular asset pair is supported for exchange. For example, if I list USDC<-->BRL in my SEP-38 `GET /prices` endpoint, and I list USDC in my SEP-31 `/info` endpoint, is it possible that the anchor doesn't accept quote `id`'s for `USDC-->BRL` in a SEP-31 transaction? I don't see why that would be the case. If we wanted to be careful, we would either need to list the assets that could be exchanged for USDC in the `/info` endpoint, effectively duplicating the info in SEP-38. Or, we could adjust SEP-38 to list the protocols that can use quotes between each asset pair.

marcelosalloum:
> Personally, I don't see why an anchor would support SEP-38 but not support quotes to be used in the context of SEP-31 transactions.

My concern is if SEP-38 is available through `ANCHOR_QUOTE_SERVER`, should we assume all other protocols that can implement SEP-38 are already implementing it? I believe an anchor should be able to support SEP-31 (with quotes) and  SEP-6 (without quotes), for instance. A boolean like `quotes_supported: true` would do the trick.

JakeUrban:
@marcelosalloum I added `quotes_supported` as an attribute under each asset listed in `/info`. I also clarified the section header and `fee` attribute as related to the stellar asset, not the fiat currency delivered to the receiving client.

marcelosalloum (approved):
LGTM!

marcelosalloum:
I'm having second thoughts on the approaches based on quotes... when posting to `/transactions` should we suggest that the Receiving Anchor could optionally implement the `quotes_supported` approach but mandatorily implement the older approach? 

Or should we make it clear that the Sending anchor should follow the `quotes_supported` approach if `quotes_supported:true`, otherwise fall back to the older SEP-31 approach?

My concern here is we could have Sending Anchors trying one approach only while Receiving Anchors are implementing the other approach, thus making their communication inconsistent.

marcelosalloum:
Cool, using `quote_required` makes more sense to me too @JakeUrban. This way it's clearer which kind of implementation should be used for each case

ecosystem/sep-0031.md
```diff
@@ -256,16 +292,63 @@ Content-Type: application/json
...
+`amount` | number | (optional) Amount of the Stellar asset sent to the Receiving Anchor.
+`asset_code` | string | (optional) Code of the asset the Sending Anchor intends to send. This must match one of the entries listed in the receiving anchor's `GET /info` endpoint.
 `asset_issuer` | string | (optional) The issuer of the Stellar asset the Sending Anchor intends to send. If not specified, the asset sent must be issued by the Receiving Anchor.
+`destination_asset` | string | (optional) The off-chain asset the Receiving Anchor will deliver to the Receiving Client. The value must match one of the `asset` values included in a [`SEP-38 GET /prices?sell_asset=stellar:<asset_code>:<asset_issuer>`](sep-0038.md#get-prices) response using [SEP-38 Asset Identification Format](sep-0038.md#asset-identification-format).
+`quote_id` | string | The `id` returned from a `SEP-38 POST /quote` response. If this attribute is specified, none of the attributes listed above should also be included in the request. The Stellar transaction used to send the asset _must_ have a `created_at` timestamp that is earlier than the quote's `expires_at` attribute.
```
marcelosalloum, line 351: 🤔  Should we be including this `created_at` attribute you're referring to? Or should it actually be `started_at`?

```suggestion
`quote_id` | string | The `id` returned from a `SEP-38 POST /quote` response. If this attribute is specified, none of the attributes listed above should also be included in the request. The Stellar transaction used to send the asset _must_ have a `started_at` timestamp that is earlier than the quote's `expires_at` attribute.
```

marcelosalloum (commented):
Added a couple more suggestions below.

ecosystem/sep-0031.md
JakeUrban: I'm referring to the `created_at` attribute returned from Horizon's `GET /transactions/:id` endpoint. This value is the close time of the ledger the transaction was included in. This gives us a source of trust to use so that there is never a dispute about whether the payment was made prior to the quote's expiration.

JakeUrban:
Hello @leighmcculloch & @marcelosalloum, now that we've settled on the technical design decisions for #917, I'd like to take what we've discussed and adjusted there and apply it to this PR. It seems like the primary decision to be made relates to whether or not we should create a new endpoint for the quote-style transactions as we did for SEP-6.

This SEP also doesn't have a `/fee` endpoint so we don't need to worry about this.

ecosystem/sep-0031.md
```diff
@@ -343,9 +426,12 @@ Name | Type | Description
...
+`amount_in_asset` | string | (optional) The asset received or to be received by the Receiving Anchor. Must be present if `quote_id` or `destination_asset` was included in the `POST /transactions` request. The value must be in [SEP-38 Asset Identification Format](sep-0038.md#asset-identification-format).
+`amount_out` | string | (optional) The amount sent by the Receiving Anchor to Receiving Client.
+`amount_out_asset` | string | (optional) The asset delivered to the Receiving Client. Must be present if `quote_id` or `destination_asset` was included in the `POST /transactions` request. The value must be in [SEP-38 Asset Identification Format](sep-0038.md#asset-identification-format).
 `amount_fee` | string | (optional) The amount of fee charged by the Receiving Anchor.
+`amount_fee_asset` | string | (optional) The asset in which fees are calculated in. Must be present if `quote_id` or `destination_asset` was included in the `POST /transactions` request. The value must be in [SEP-38 Asset Identification Format](sep-0038.md#asset-identification-format).
```
marcelosalloum, line 434: I believe the idea now is all conversion fees are included in the quote returned from SEP-38 /price and /quote endpoints, right? In that case, we don't need the attribute `amount_fee_asset` here.

ecosystem/sep-0031.md
```diff
@@ -68,6 +69,30 @@ All endpoints respond with content type:
...
+
+This approach requires the Sending Anchor to provide an exchange rate, or quote, to the Sending Client. It also requires the Sending Anchor to hold a balance of the destination asset on Stellar. The Sending Anchor would use Payment operations to send the quoted amount of the destination asset to the Receiving Anchor.
+
+### Stellar Network Asset Conversion
+
+Instead of holding a balance of the destination asset on Stellar, the Sending Anchor can use Path Payments to convert the source asset to the destination asset on the Stellar Network. 
+
+This approach requires a sufficiently liquid market to exist between the source and destination assets on the Stellar Decentralized Exchange (SDEX).
+
+### Receiving Anchor Asset Conversion
```
marcelosalloum, line 90: Q: is this type of conversion a new addition in this version of the protocol? How many Anchors (ballpark estimation) do you think are using SEP-31 already?

marcelosalloum (commented):
TBH I don't feel very strongly towards creating a new /transactions-exchange endpoint for SEP-31. Looking back at SEP-6, the meaning of `asset_code` was pretty confusing for deposits and withdrawals with non-equivalent assets but that's not the case here.

Both `asset_code` and `destination_asset` are clear concepts in SEP-31 so I think it's fine to handle all three use cases with the same existing endpoint.

---

Also, I've left a few suggestions below.

ecosystem/sep-0031.md
JakeUrban: I think it's worth keeping as an optional attribute in cash a fixed fee is also charged on top of the margin included in the exchange rate. If only the margin on the exchange rate is used, then `amount_fee` will be 0 anyway.

ecosystem/sep-0031.md
JakeUrban: Yes, prior to the development of SEP-38 and these changes to SEP-31 it wasn't possible for the receiving anchor to convert assets in a standardized way. Currently, every anchor that implements SEP-31 is interested in this capability, with many more in the pipe.

ecosystem/sep-0031.md
```diff
@@ -100,11 +126,12 @@ All endpoints respond with content type:
 1. The Sending Anchor makes a `SEP-12 GET /customer` request for the Sending and Receiving Clients if required.
 1. The Sending Anchor makes a `SEP-12 PUT /customer` request for the Sending and Receiving Clients if required.
   1. The Receiving Anchor must validate the KYC data provided and reject the request with useful error messages if invalid.
+1. The Sending Anchor may request a quote using the `SEP-38 GET /price` or `SEP-38 POST /quote` endpoints.
 1. The Sending Anchor makes a `POST /transactions` request.
-  1. The Receiving Anchor must ensure the asset, amount, transaction fields, and customers IDs are valid.
+  1. The Receiving Anchor must ensure the asset, amount, transaction fields, and customers IDs are valid. The Sending Anchor may provide a `quote_id` instead of asset and amount information if [SEP-38](sep-0038.md) is supported.
```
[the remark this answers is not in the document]
JakeUrban: I don't feel strongly either way. If we made asset & amount mandatory in all cases, the other optional parameters would be `quote_id` and `destination_asset`, which would be mutually exclusive. I don't like the idea of also requiring `destination_asset` in all requests since requiring a new field is a breaking change for existing implementations.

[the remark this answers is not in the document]
marcelosalloum: On second thought, requiring the user to repeat the fields that are already part of the quote goes against the [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) principle. 

@leighmcculloch, @jacekn: if you both agree with that I'll approve this PR and update #973 to use the same strategy [here](https://github.com/stellar/stellar-protocol/pull/973/#discussion_r674033709).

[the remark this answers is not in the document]
leighmcculloch: It may be simpler, I think, to keep the API stable regardless if a quote is used, and to have the quote ID be an optional parameter. DRY isn't a truth, just a tool, and doesn't necessarily result in the simplest to understand APIs. In this specific case I do not feel strongly though and defer to you both.

Something that some APIs do to separate these changing input parameters from other parameters is to wrap them in a subfield and introduce a concept to wrap up those changing inputs. A good example of this is Stripe's source/payment intent interface. We could do something similar by saying the changing fields are the ExchangeParams, and for taht field you can either provide RawExchangeParams or QuoteExchangeParams.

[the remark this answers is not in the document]
JakeUrban: I really like this idea of a subfield, especially because the SEP-31 endpoint accepts `application/json` in a POST request. This pattern doesn't work well with SEP-6's `GET /*-exchange` endpoints though.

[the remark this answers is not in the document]
leighmcculloch: Even if we don't use a subfield, thinking about it more the subfield identifies a conceptual oddity with using mutual exclusion on destination_asset. For example:

<table>
<tr><th>non-quote</th><th>quote</th></tr>
<tr>
<td>

```json
{
  "asset": "...",
  "amount": "...",
  "subfield": {
    "destination_asset": "...",
  },
}
```

</td>
<td>

```json
{
  "asset": "...",
  "amount": "...",
  "subfield": {
    "quote_id": "...",
  },
}
```

</td>
</tr>
</table>

It's not intuitive to me that destination asset is inside the subfield when asset is outside. In the same way it's not intuitive that destination_asset is optional when asset isn't.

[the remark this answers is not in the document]
JakeUrban: Since we're not creating a `POST /transactions-exchange` endpoint and instead using the original `POST /transactions` endpoint we cannot require `destination_asset` on every request, otherwise that would be a breaking change for existing applications.

Personally I don't think having two mutually exclusive parameters is too confusing, especially if we have a subsection below the request parameters list explaining when to use each one.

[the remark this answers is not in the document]
JakeUrban: Updated language:

> The Receiving Anchor must ensure the asset, amount, transaction fields, and customers IDs are valid. The Sending Anchor may provide either, but not both, a `quote_id` or `destination_asset` attribute if [SEP-38](sep-0038.md) is supported.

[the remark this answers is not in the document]
JakeUrban: Actually, I just realized that `destination_asset` can be specified if `quote_id` is also specified. There is no need for mutual exclusion.

So the three combinations are:
- `asset_code`, `amount`
- `asset_code`, `amount`, `destination_asset`
- `asset_code`, `amount`, `destination_asset`, `quote_id`

And if `quote_id` is specified, the information must match that of the quote.

Thoughts?

marcelosalloum (approved):
LGTM!

PR #936, opened 2021-05-04
CAP-21: Add seqSuccess

leighmcculloch:
### What
Add a seqSuccess to the account entry extension v3, and corresponding precondition to transaction general preconditions.

### Why
The top-up flow described for two-way payment channels does not handle the case where a top-up fails. In that case the channel will have progressed past the sequence number of the last agreed C/D, but won't progress forward. There needs to be a second set of C/D that preserves the existing state but that would be bumped past if the top-up is successful. Even with this the channel still needs repair at some point if it is to stay open, with a third set of C/D. We can avoid the need to rely on the bump sequences and a third set if we introduce `seqSuccess` since the fork is completely finalized in the top-up's success or failure without need for tidy up on-chain.

### DIFF
--- a/core/cap-0021.md
+++ b/core/cap-0021.md
@@ -489,7 +489,7 @@ to correct an abnormal situation (e.g., one end of a payment channel
 failing, or an account owner losing the key) and so don't actually get
 submitted in the common case.
 
-An earlier version of the proposal did not contain the `seqSuccess
+An earlier version of the proposal did not contain the `seqSuccess`
 field either.  However, payment channels and contracts inadvertently
 need to handle transaction sequences where when transactions fail,
 only one path is possible ideally without executing another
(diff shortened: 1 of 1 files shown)


leighmcculloch:
Closing because we can address this by specifying the source account of the deposit transaction is not the escrow account: https://github.com/stellar/stellar-protocol/pull/937.

leighmcculloch:
Reopening because even though the safety case is addressed by #937, it isn't optimal because it will probably require participants of the channel to use additional sequence number accounts for any coordinated change of the channel. The two-way payment channel is intended to work well with hubs where one party may be a participant in many channels and such a party would probably need sequence number pooling. This change allows the escrow account to be the sequence of all transactions even those that may fail.

So this PR is now an optimization, rather than fixing a safety issue.

Thoughts @stanford-scs ?

PR #938, opened 2021-05-06
CAP-21: Add another type of two-way payment channel that uses two escrow accounts

leighmcculloch:
### What
Add another type of two-way payment channel design to the rationale that uses two escrow accounts and does not use claimable balances.

### Why
I've been expanding on the definition of the existing two-way payment channel in CAP-21, writing out all of the different steps, how fees would work, etc. The full write up I've done is here:
https://github.com/stellar/experimental-payment-channels/pull/3

After doing the write up I'm reasonably certain that even though on the surface it appears that a single escrow account is required, two 2-of-2 multisig accounts would likely be necessary in a real product. I think it will be required because claimable balances require a sponsor, and the sponsor is bound to them and can't be merged until they are claimed. To keep claimable balance sponsorship from binding the initiator and responder together after the channel closes a second reserve account would be required to sponsor the claimable balances safely. See the full write up in the PR above for how it would work and why.

If a two-way payment channel is likely to have two 2-of-2 multisig accounts, then we may as well construct the channel to make the most of those two accounts. If we use two accounts to hold I and R's contributions both parties can deposit without coordination. The channel is still not symmetric because I's escrow account EI is the account providing sequence numbers, but I don't think this is a big deal.

As an extra bonus, channel close does not create any ledger entries, which means the channel never has to deal with the oddities of claimable balances and the oddities of ledger entry creation which can fail because of sponsorship low reserve errors. These things are not insurmountable in the existing two-way payment channel design, but the design is simpler without them.

### DIFF
--- a/core/cap-0021.md
+++ b/core/cap-0021.md
@@ -567,7 +567,7 @@ constructed as follows:
 * D_i, the _declaration transaction_, declares an intent to execute
   the corresponding closing transaction C_i.  D_i has source account
   EI, sequence number 2i, and `minSeqNum` set to s.  Hence, D_i can
-  execute at any time, so long as E's sequence number n satisfies s <=
+  execute at any time, so long as EI's sequence number n satisfies s <=
   n < 2i.  D_i always leaves EI's sequence number at 2i after
   executing.  Because C_i has source account E and sequence number
   2i+1, D_i leaves EI in a state where C_i can execute.  Note that D_i
@@ -594,7 +594,7 @@ payment into them directly.
 I and R may withdraw excess funds from the escrow accounts by skipping
 a generation. They set s = 2(i+1), and i = i+2. They then exchange C_i
 and D_i (which unlike the update case, can be exchanged in a single
-phase of communication because D_i is not yet executable while E's
+phase of communication because D_i is not yet executable while EI's
 sequence number is below the new s).  Finally, they create a withdraw
 transaction that atomically adjusts EI's and/or ER's balance, as
 required, and uses `BUMP_SEQUENCE` to increase EI's sequence number to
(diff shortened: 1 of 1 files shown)


stanford-scs (dismissed):
It's kind of too bad that claimable balances don't solve this problem.  If the main problem is that R can sit on the reserve entry, how bad is this?  I mean I can still recycle the escrow account for another payment channel, just might need to keep some extra minimum balance.

Is the problem that it's hard to figure out how many base reserves an account is consuming?  If so, that might be a problem for all payment channels, because R could be tricked into approving something on the assumption that a create claimable balance won't fail.

leighmcculloch:
> If the main problem is that R can sit on the reserve entry, how bad is this?

It's not that bad in the scale of things that could be bad, but it's inconvenient that at the end of a channel the two parties are intertwined with one another.

The main issue imo for usability is that accounts are intertwined. It increased the error conditions implementations need to handle.


> Is the problem that it's hard to figure out how many base reserves an account is consuming?

This isn't that hard, it's pretty easy to calculate, so I don't think that's the issue. Even though it isn't hard, it could be a variable that could make an implementation exploitable if the implementation doesn't think about it right, but this is highly theoretical.

We could say that calculating the reserve amount is problematic if the reserve changes. Participants must monitor base reserve cost and not accidentally close the channel without topping up first in that case.

>R could be tricked into approving something on the assumption that a create claimable balance won't fail

I hadn't considered this, and I think this is a problem, but not with all payment channels. Only with payment channels that _create ledger entries like claimable balances_. The design being added in this PR doesn't suffer from this.

stanford-scs (approved):

nikhilsaraf (commented):
LGTM!

PR #940, opened 2021-05-12
SEP-0012: make payload of POST callback_url clear

howardtw:
When I read the description of the payload POST callback_url, it wasn't clear to me whether it should be the same as the GET /customer request or response although the GET request body should not have any meaning.


leighmcculloch (commented):
This looks right to me. Could you update the version and updated meta fields, bumping the patch version?

leighmcculloch (approved):

PR #942, opened 2021-05-14
CAP-21: Add comment about unit of duration

leighmcculloch:
### What
Add a comment stating that the unit of time the `Duration` type is expressed in is in seconds.

### Why
It is unclear looking at the XDR what unit of time the new `Duration` type is expressed in. Seconds is mentioned at one location it will be used, in claimable balances, and it would be helpful if this is defined on the type too.


acharb (approved):
lgtm 👍

PR #947, opened 2021-05-19
SEP-0009: update "birth_place" field type from date to string

marcelosalloum:
### What

Update SEP-9 `birth_place` field type from "date" to "string".

### Why

The field description says _"Place of birth (city, state, country; as on passport)"_ so I'm assuming its type should actually be "string".


marcelosalloum:
Please let me know if this fix should trigger a version bump from `1.3.1` to `1.3.2`

JakeUrban:
Thanks for catching this @marcelosalloum , lets bump the patch version and I'll approve 👍

marcelosalloum:
> lets bump the patch version and I'll approve 👍

Done 😄

JakeUrban (approved):

leighmcculloch (approved):
Nice catch. :shipit:

PR #948, opened 2021-05-21
ecosystem/sep-0024.md: Claimable balance flow where stellar account doesn't exist

Zagan202:
### What 
Adds clarity to the `Anchor Claimable Balance Flow` section to describe what to do when a user makes a deposit to an anchor and doesn't have a created Stellar account.

### Why
To decrease the _"trust line must be established first"_ blockers that exist in the traditional SEP24 deposit flow we leverage the claimable balance operation and detail the deposit flow that anchors and wallets should follow.
However the scenario where a wallet user attempts a deposit without having a stellar account created needs to be detailed to remove ambiguity of this edge case.

### DIFF
--- a/ecosystem/sep-0024.md
+++ b/ecosystem/sep-0024.md
@@ -263,9 +263,10 @@ const result = await this.server.submitTransaction(transaction)
 1. Wallets make a request to `/transactions/deposit/interactive` providing the `claimable_balance_supported` request parameter.
 2. Anchors update their internal database record of the transaction to indicate the wallet supports receiving a claimable balance.
 3. Users send the external asset to the anchor's off-chain account.
-4. Anchors detect that the user's Stellar account doesn't have a trustline.
-5. Anchors submits a Stellar transaction containing a claimable balance operation.
-6. Anchors update the `/transaction(s)` attriutes `status` to `completed` and `claimable_balance_id` to the ID returned in the Horizon response.
+4. If anchors detect that the user's Stellar account doesn't exist then the anchor should [create the Stellar account](#stellar-account-does-not-exist) but skip the step where the anchor should listen for the user to establish their trust line(skip to step 6).
+5. If anchors detect that the user's Stellar account exists then anchors must check if the user's Stellar account doesn't have a trust line(if trust line exists then proceed with the traditional [deposit flow](#deposit) where anchors then proceed to send payment to the user's Stellar account; skips "Anchor Claimable Balance Flow" steps 6 and 7).
+6. Anchors submits a Stellar transaction containing a claimable balance operation.
+7. Anchors update the `/transaction(s)` attributes `status` to `completed` and `claimable_balance_id` to the ID returned in the Horizon response.
 
 ##### Claimable Balance Claimants and Predicates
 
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0024.md
```diff
@@ -263,9 +263,10 @@ const result = await this.server.submitTransaction(transaction)
...
 2. Anchors update their internal database record of the transaction to indicate the wallet supports receiving a claimable balance.
 3. Users send the external asset to the anchor's off-chain account.
-4. Anchors detect that the user's Stellar account doesn't have a trustline.
-5. Anchors submits a Stellar transaction containing a claimable balance operation.
-6. Anchors update the `/transaction(s)` attriutes `status` to `completed` and `claimable_balance_id` to the ID returned in the Horizon response.
+4. If anchors detect that the user's Stellar account doesn't exist then the anchor should [create the Stellar account](#stellar-account-does-not-exist) but skip the step where the anchor should listen for the user to establish their trust line(skip to step 6).
+5. If anchors detect that the user's Stellar account exists then anchors must check if the user's Stellar account doesn't have a trust line(if trust line exists then proceed with the traditional [deposit flow](#deposit) where anchors then proceed to send payment to the user's Stellar account; skips "Anchor Claimable Balance Flow" steps 6 and 7).
```
JakeUrban, line 267: I think we can condense this into one list item: 

```
4. Anchors check if the Stellar account exists, and if it doesn't, creates the account using a [CreateAccount operation](link)
```

Then the 5th step can be adjusted to say "if the Anchor created the account immediately prior, the account will not have a trustline."

JakeUrban: The issue I have with this change is that it assumes anchors create the account, which the rest of this document doesn't assume. It is an option, but it is up to the anchor to decide if it should support it. What should the behavior be if the anchor doesn't create accounts?

JakeUrban (changes requested):

ecosystem/sep-0024.md
Zagan202: > The issue I have with this change is that it assumes anchors create the account, which the rest of this document doesn't assume. It is an option, but it is up to the anchor to decide if it should support it. What should the behavior be if the anchor doesn't create accounts?

Thats a really good point.
I believe if the anchor doesn't support creating accounts it should skip straight step 6 `6. Anchors submits a Stellar transaction containing a claimable balance operation.`

CB operation is not blocked by having the recipient be a stellar address to a not created account.

Zagan202: If the deposit and CB was a success for a user without a created account the wallet needs to do one of two things.

- Create the account with at least enough XLM for the minimum reserve and a trust line and claim balance to the requested asset
- Or, Communicate to the user that then need to create their account on their own. (which I do not know how they do so on main net).

JakeUrban: Oh interesting, I didn't know that. The issue here is that the wallet doesn't know if the anchor will create the account or not. If both the anchor and wallet don't support creating accounts, the user is left hanging and it would seem like they got scammed.

Maybe the anchor should signal whether or not they create accounts via the `/info` endpoint?

PR #950, opened 2021-05-25
meta: add cap diff check

### LINKED ISSUE #893: meta: diff checker for CAPs
I think we could benefit from a small scoped investment in a diff checker on CAPs.

On two or three occasions diffs have been committed that are corrupt. This is usually discovered by someone consuming the diff, and not by the person publishing the diff. Diffs are fiddly and it's easy for copy paste errors to occur, or stray characters, or missing new lines at the end to result in a corrupt diff.

We could build a small utility that runs in a GitHub Action that:
1. Grabs all CAPs in a PR that contain a diff block that begins with ` ```diff` and ends with ` ``` `
2. Extract the diff within the block, minus its last new line (see https://github.com/stellar/stellar-protocol/pull/892 for why the last line should be ignored).
3. Checks out the stellar-core at the version of stellar-core the CAP provides (requires a new meta entry in CAPs to define what protocol they are proposed against).
4. Attempts to run git apply, and posts any errors to the PR, notifying the diff author it is corrupt.

cc @MonsieurNicolas @stanford-scs @jonjove @sisuresh

leighmcculloch:
### What
Add a GitHub Action that check the diffs in CAPs are valid.

### Why
On a few occasions diffs have been committed that are corrupt. This is usually discovered by someone consuming the diff, and not by the person publishing the diff. This consumes that other persons time, as they need to debug the diff. Diffs are fiddly and it's easy for copy paste errors to occur, or stray characters, or missing new lines at the end to result in a corrupt diff. Instructions introduced in 9b0d752c7ac4d42c9735789cb190479ee8744a33 should help us generate good diffs, but this check will catch the corrupt diffs when the author is submitting them, before someone else is consuming them.

Some of the existing CAPs were easy to update to pass the check. Others not, some just don't seem to apply cleanly to any version as far as I could see, which is the type of situation this check will prevent. For those cases I added a parameter to ignore them for the moment until their authors can fix them.

This GitHub Action uses a tool that I wrote, `mddiffcheck`, which lives at https://github.com/stellar/mddiffcheck. It took more effort than I expected or planned to give to it to write the tool, but it is done now.

Close #893

### DIFF
--- a/.github/workflows/mddiffcheck.yml
+++ b/.github/workflows/mddiffcheck.yml
@@ -11,5 +11,5 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
-    - run: go get github.com/stellar/mddiffcheck@latest
+    - run: go get github.com/stellar/mddiffcheck
     - run: mddiffcheck -repo https://github.com/stellar/stellar-core core/*.md
--- a/.github/workflows/mddiffcheck.yml
+++ b/.github/workflows/mddiffcheck.yml
@@ -15,5 +15,5 @@ jobs:
       uses: actions/setup-go@v2
       with:
         go-version: 1
-    - run: go get github.com/stellar/mddiffcheck@latest
+    - run: GOPROXY=direct go get github.com/stellar/mddiffcheck@latest
     - run: mddiffcheck -repo https://github.com/stellar/stellar-core core/*.md
--- a/.github/workflows/mddiffcheck.yml
+++ b/.github/workflows/mddiffcheck.yml
@@ -11,5 +11,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
+    - name: Set up Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: ^1.16
     - run: go get github.com/stellar/mddiffcheck
     - run: mddiffcheck -repo https://github.com/stellar/stellar-core core/*.md
--- a/.github/workflows/mddiffcheck.yml
+++ b/.github/workflows/mddiffcheck.yml
@@ -11,9 +11,5 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
-    - name: Set up Go
-      uses: actions/setup-go@v2
-      with:
-        go-version: 1
     - run: go get github.com/stellar/mddiffcheck@latest
     - run: mddiffcheck -repo https://github.com/stellar/stellar-core core/*.md
--- a/.github/workflows/mddiffcheck.yml
+++ b/.github/workflows/mddiffcheck.yml
@@ -1,0 +1,19 @@
+name: 'Check CAP Diffs'
+
+on:
+  push:
+    branches:
+    - master
+  pull_request:
+
+jobs:
+  mddiffcheck:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: 1
+    - run: go get github.com/stellar/mddiffcheck@latest
+    - run: mddiffcheck -repo https://github.com/stellar/stellar-core core/*.md
--- a/.github/workflows/mddiffcheck.yml
+++ b/.github/workflows/mddiffcheck.yml
@@ -15,5 +15,5 @@ jobs:
       uses: actions/setup-go@v2
       with:
         go-version: 1
-    - run: GOPROXY=direct go get github.com/stellar/mddiffcheck@latest
+    - run: go get github.com/stellar/mddiffcheck@latest
     - run: mddiffcheck -repo https://github.com/stellar/stellar-core core/*.md
--- a/core/cap-0035.md
+++ b/core/cap-0035.md
@@ -97,7 +97,7 @@ flags. This operation can be used to unset `TRUSTLINE_CLAWBACK_ENABLED_FLAG`.
 
 This patch of XDR changes is based on the XDR files in tag/commit `v15.1.0` (`90b2780584c6390207bf09291212d606896ce9f8`) of [stellar-core].
 
-```diff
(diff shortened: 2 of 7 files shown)


.github/workflows/mddiffcheck.yml
```diff
@@ -0,0 +1,19 @@
...
+jobs:
+  mddiffcheck:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: 1
+    - run: GOPROXY=direct go get github.com/stellar/mddiffcheck@latest
```
leighmcculloch, line 18: Using `GOPROXY=direct` due to https://github.com/golang/go/issues/46369.

core/cap-0035.md
```diff
@@ -97,7 +97,7 @@ flags. This operation can be used to unset `TRUSTLINE_CLAWBACK_ENABLED_FLAG`.
 
 This patch of XDR changes is based on the XDR files in tag/commit `v15.1.0` (`90b2780584c6390207bf09291212d606896ce9f8`) of [stellar-core].
 
-```diff
+```diff mddiffcheck.ignore=true
```
leighmcculloch, line 100: I would have liked to fix this CAP-35 diff, but I can't figure out what is corrupted. It's not just whitespace, and even using the reject mode in git apply wasn't particularly useful.

marcelosalloum (approved):
LGTM!

.github/workflows/mddiffcheck.yml
```diff
@@ -0,0 +1,19 @@
...
+    branches:
+    - master
+  pull_request:
+
+jobs:
+  mddiffcheck:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Go
```
MonsieurNicolas, line 14: do you need to setup go? I see from https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md that go should already be installed?

.github/workflows/mddiffcheck.yml
```diff
@@ -0,0 +1,19 @@
...
+      uses: actions/setup-go@v2
+      with:
+        go-version: 1
```
MonsieurNicolas, line 17: If you need it, I _think_ that you may want `go-version: '^1.'` ? I looked at https://github.com/actions/setup-go and I am not sure what kind of matching is happening here

core/cap-0035.md
MonsieurNicolas: Let's try to figure out that problem before merging this otherwise things are going to get frustrating very fast and people are going to leave this disabled.

I agree there is probably a bug in mddiffcheck - maybe using `patch` directly would help (`git apply` may do a bunch of things we don't care about)

MonsieurNicolas (changes requested):
Nice! A few questions and suggestions

core/cap-0035.md
leighmcculloch:
```sh
$ git patch
git: 'patch' is not a git command. See 'git --help'.
```
What do you mean use patch?

I don't think we should use patches, that's a different thing to diffs, and the CAP process calls for diffs, not patches. Patches include all the commit data, not just the content change, and would bloat a CAP.

---

>I agree there is probably a bug in mddiffcheck

Okay, so it looks like this is a line endings problem. The patch contains windows line ending characters CRLF, and mddiffcheck is preserving them because the markdown parser's AST preserves them, but when I copy the patch via vim on my machine it removes them and applies fine.

A quick check shows that CAP-35 is the only cap that has CRLF line endings. Maybe we should ensure we don't commit CRLFs so as to keep all the CAPs consistent.

Either way, I can probably improve mddiffcheck so it deals with line endings better, but we should also keep CAPs consistent.

.github/workflows/mddiffcheck.yml
leighmcculloch: The tools are cached, but not setup. The setup-go action will setup the tool for use, it won't download it unless the cache hasn't provided it. The setup-go action uses the tool-cache library to do that.

.github/workflows/mddiffcheck.yml
leighmcculloch: This uses the https://www.npmjs.com/package/semver package and will resolve 1 to the latest v1 that is installed, which is 1.16. But you're right, we should be using the caret to specify 1.16 or newer, because the `mddiffcheck` tool uses functions that were added to the stdlib of 1.16. Updated in e46bce9.

core/cap-0035.md
leighmcculloch: After a little more digging I'm not sure the line endings issues can be really fixed. Git apply isn't designed to handle line endings discrepancies, and we should just make sure our CAPs are all committed as LF not CRLF.

MonsieurNicolas: `patch` as in [the unix command](https://linux.die.net/man/1/patch) (not a git subcommand). It can process uni-diffs as a parameter.
Line ending: can't you just normalize line ending in the tool so that `patch` will work? You're already extracting the diff from the md file?

leighmcculloch: I could normalize the line endings to LF, but if we make sure files in the repo are all LF to begin with, this isn't an issue and has other benefits. We can do that with #952 and #956.

Also, mddiffcheck doesn't know if a file should be CRLF or LF, and it'd be a bit weird for it to have that responsibility.|

leighmcculloch: > `patch` as in [the unix command](https://linux.die.net/man/1/patch) (not a git subcommand)

`patch` will work instead of `git apply` because by default it replaces CRLFs with LFs, but the tool is harder to use. We can fix this issue so that both `patch` and `git apply` will work, and this test should make sure that the basic `git apply` should work.

MonsieurNicolas: agree let's try to enforce encoding and line ending first

MonsieurNicolas (approved):

PR #954, opened 2021-05-26
SEP-8: Be clear about the action_required step being interactive

leighmcculloch:
### What
Make the action_required text clearer upfront that this state is intended to require user interaction with a browser, however that there are capabilities to allow a client to skip it if it has the necessary information.

### Why
It has been surprising to a few folks that user interaction with a browser is required because one of the two options for replying is a POST request.

### DIFF
--- a/ecosystem/sep-0008.md
+++ b/ecosystem/sep-0008.md
@@ -226,7 +226,7 @@ Name | Type | Description
 
 This response means that the user must complete an action before this transaction can be approved. The approval service will provide a URL that facilitates the action. Upon completion, the user will resubmit the transaction.
 
-The action the user must complete is intended to be displayed in a browser. The client may be able to skip the need to display the URL in the browser if it can supply the requested details.
+The action the user must complete is intended to be displayed in a browser. The client may be able to skip displaying the URL in the browser if the server supports the `POST` `action_method`, the client can supply the requested details, and the server's final response to the `POST` indicates no further action is required.
 
 An `action_required` response will have a `200` HTTP status code and `action_required` as the `status` value.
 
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0008.md
```diff
@@ -226,6 +226,8 @@ Name | Type | Description
 
 This response means that the user must complete an action before this transaction can be approved. The approval service will provide a URL that facilitates the action. Upon completion, the user will resubmit the transaction.
 
+The action the user must complete is interactive, and is intended to be displayed in a browser. The client may be able to skip the need to display the URL in the browser if it can supply the requested details.
```
howardtw, line 229:
```suggestion
The default action the user must complete is interactive, and is intended to be displayed in a browser. The client may be able to skip the need to display the URL in the browser if it can supply the requested details.
```
Iirc it's only interactive if the `action_method` is `GET`, which is the default method if undefined.

howardtw (approved):
LGTM

ecosystem/sep-0008.md
leighmcculloch: It's also interactive in the POST case if the data included is not sufficient.

howardtw: That's right, but it won't be if the data included is sufficient. Just wanted to point this out.

Or we can do this or something similar
```suggestion
The action the user must complete is interactive unless the client is able to provide sufficient data with a POST request, and is intended to be displayed in a browser. The client may be able to skip the need to display the URL in the browser if it can supply the requested details.
```

marcelosalloum, line 229: How well-known is the concept of an "interactive" flow? I had never heard of this concept before Stellar so maybe we could include a description about what it means.

leighmcculloch: Great point. The term `interactive` is unnecessary and raises more questions than the text has without it. Rather than define the word I removed its use in 1421ab9.

leighmcculloch: I took your edit and extended it in ec706df. Thanks!

howardtw (approved):
LGTM

PR #955, opened 2021-05-27
SEP-24: improved memo description, removed successful response

yuriescl:
- Improved `memo` and `memo_type` descriptions about their use cases
- Removed successful response since SEP-6 must be used for non-interactive transactions
- Interactive response documentation now doesn't account for other types of responses. It focuses on explaning the interactive URL response.
- Updated link to Demo Wallet

### DIFF
--- a/ecosystem/sep-0024.md
+++ b/ecosystem/sep-0024.md
@@ -6,8 +6,8 @@ Title: Hosted Deposit and Withdrawal
 Author: SDF
 Status: Active
 Created: 2019-09-18
-Updated: 2021-05-17
-Version 1.3.2
+Updated: 2021-05-26
+Version 1.3.3
 ```
 
 ## Simple Summary
@@ -159,8 +159,8 @@ Name | Type | Description
 `asset_issuer` | string | (optional) The issuer of the stellar asset the user wants to receive for their deposit with the anchor.  If asset_issuer is not provided, the anchor should use the asset issued by themselves as described in their TOML file.
 `amount` | number | (optional) Amount of asset requested to deposit.  If this is not provided it will be collected in the interactive flow.
 `account` | `G...` string | The stellar account ID of the user that wants to deposit. This is where the asset token will be sent.
-`memo_type` | string | (optional) Type of memo that anchor should attach to the Stellar payment transaction, one of `text`, `id` or `hash`.
-`memo` | string | (optional) Value of memo to attach to transaction, for `hash` this should be base64-encoded.
+`memo` | string | (optional) Value of memo to attach to transaction, if `memo_type` is `hash`, this must be the base64-encoded hash. A wallet will send this to uniquely identify a user if the wallet has multiple users sharing one Stellar account. The anchor can use this along with `account` to look up the user's KYC info.
+`memo_type` | string | (optional) Type of `memo`. One of `text`, `id` or `hash`.
 `wallet_name` | string | (optional) In communications / pages about the deposit, anchor should display the wallet name to the user to explain where funds are going.
 `wallet_url` | string | (optional) Anchor should link to this when notifying the user that the transaction has completed.
 `lang` | string | (optional) Defaults to `en`. Language code specified using [ISO 639-1](https://en.wikipedia.org/wiki/ISO_639-1). `error` fields in the response, as well as the interactive flow UI and any other user-facing strings returned for this transaction should be in this language.
@@ -180,8 +180,6 @@ asset_code=USD&email_address=myaccount@gmail.com&account=GACW7NONV43MZIFHCOKCQJA
 ```
 
 ### Response
-
-There are several possible kinds of response, depending on whether the anchor needs more information about the user, how it should be sent to the anchor, and if there are any errors.
 
 Responses are detailed in the [Deposit and Withdraw shared responses](#deposit-and-withdraw-shared-responses) section below.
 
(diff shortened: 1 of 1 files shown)


yuriescl:
@JakeUrban These changes are inspired by Polaris behavior, and are related to https://github.com/stellar/django-polaris/issues/428

ecosystem/sep-0024.md
```diff
@@ -159,8 +159,8 @@ Name | Type | Description
 `asset_issuer` | string | (optional) The issuer of the stellar asset the user wants to receive for their deposit with the anchor.  If asset_issuer is not provided, the anchor should use the asset issued by themselves as described in their TOML file.
 `amount` | number | (optional) Amount of asset requested to deposit.  If this is not provided it will be collected in the interactive flow.
 `account` | `G...` string | The stellar account ID of the user that wants to deposit. This is where the asset token will be sent.
-`memo_type` | string | (optional) Type of memo that anchor should attach to the Stellar payment transaction, one of `text`, `id` or `hash`.
-`memo` | string | (optional) Value of memo to attach to transaction, for `hash` this should be base64-encoded.
+`memo` | string | (optional) Value of memo to attach to transaction, if `memo_type` is `hash`, this must be the base64-encoded hash. A wallet will send this to uniquely identify a user if the wallet has multiple users sharing one Stellar account. The anchor can use this along with `account` to look up the user's KYC info.
```
JakeUrban, line 162: If we're going to be explicit about this use case, we should also mention that the `memo` field could be used for cases not defined here.

JakeUrban: Lets make sure we add this clause to both `memo` definitions (deposit & withdraw)

ecosystem/sep-0024.md
```diff
@@ -316,48 +314,15 @@ When uploading data for fields specificed in [SEP-9](sep-0009.md), `binary` type
 
 ### Response
 
-There are several possible kinds of response, depending on whether the anchor needs more information about the user, how it should be sent to the anchor, and if there are any errors.
-
-The first response, the success response, is explained below. The other possible responses are shared with the deposit endpoint, and are explained in the [Deposit and Withdraw shared responses](#deposit-and-withdraw-shared-responses) section directly below.
-
-#### 1. Success: no additional information needed
```
JakeUrban, line 323: I agree we should remove this. Its not used by any known anchor and SEP-6 provides a non-interactive alternative. There was a PR that turned stale to do the same thing: https://github.com/stellar/stellar-protocol/issues/539

JakeUrban (approved):
Thanks for doing this. Left a couple comments regarding the `memo` field and I'd like @leighmcculloch to take a look too.

ecosystem/sep-0024.md
leighmcculloch: This info looks useful, but I agree, if we're going to discuss one use we should discuss all uses otherwise it appears that this field has a specific use when it currently doesn't?

I think it'd be helpful to be more general:

```suggestion
`memo` | string | (optional) Value of memo the user would like the anchor to attach to the transaction. If `memo_type` is `hash`, this must be the base64-encoded hash. This value may be used by the user or wallet to identify the transaction, or in the case this wallet is a pooled account, to uniquely identify a user if the wallet has multiple users sharing one Stellar account. The anchor can use this along with `account` to look up the user's KYC info.
```

This does raise the question, how should the anchor know to lookup the user's KYC info or not based on the memo?

leighmcculloch (commented):
@JakeUrban I left some thoughts and a question, I defer to you though.

ecosystem/sep-0024.md
JakeUrban: > This does raise the question, how should the anchor know to lookup the user's KYC info or not based on the memo?

That is a good question. I think that in order for this to work properly we would need #726 or muxed accounts. The anchor could guess that a memo is required based on whether or not a memo was used in the request, but if the memo isn't used per-user and instead per-transaction, the anchor would require the user to provide KYC per-transaction.

JakeUrban: @yuriescl How were you planning on handling this ambiguity?

yuriescl: @JakeUrban I've given this some more thought and here's a suggestion:
The `memo` and `memo_type`, in any operation related to transactions (SEP-6 `/deposit`, SEP-24 `/transactions/deposit/interactive`, should represent the transaction memo, so I think they should remain there, but with only one purpose: contain the memo for the transaction, not for KYC/subaccounts.
SEP-10 at the moment is just used to authenticate a pubkey, but it could maybe be used to store information required to uniquely identify the account. For example, SEP-10 could have two more claims in the JWT: `memo` and `memo_type`. These would be considered private claims, which are supported by the JWT protocol. They'd store the memo that represents the subaccount and if they are present, the Anchor will use them to uniquely identify the user.
SEP-10 is required by most transaction endpoints, and it seems cleaner to just store the memo there since it will be required for transaction calls anyway.
The `memo` and `memo_type` could be passed to SEP-10 `GET <WEB_AUTH_ENDPOINT>`, and in the future, more complex authentication parameters could be passed as well to validate subaccounts.
One downside of this idea is that it can cause issues with Wallets and Anchors that are currency relying on the `memo` for subaccounts/KYC. They'd have to know beforehand about the change, and make sure to transition their systems to use SEP-10 instead.

Another idea is to add two more parameters to the transaction calls: `transaction_memo` and `transaction_memo_type`. They would be used for the transaction memo, and `memo` and `memo_type` would be used as account identifier.

I've thought about adding a parameter to inform the Anchor if the memo is for the transaction or the account, but this would keep the limitation of having only one memo purpose. There could be a memo for the account and also a memo for the transaction, and they should not collide.

Either change can cause issues with systems already in production. But at the moment, the memo fields are already ambiguous, we'd be just removing that ambiguity.

JakeUrban: Hi @yuriescl, we're planning to move forward with `memo` claims in the SEP-10 JWT, but in the meantime lets postpone clarifying the `memo` definition for `/deposit` requests. Can you open a new PR that only removes the success response? Thats a change we don't need to wait on anything for.

JakeUrban:
When the memo-based pooled account support PR is merged, we'll come back to this PR.

PR #956, opened 2021-05-27
meta: add action that requires LF line endings

### LINKED ISSUE #953: meta: add check that requires documents are eol=lf when committed, and UTF-8
### What
Add a GitHub action check that requires documents committed are eol=lf, and not crlf, and that they are UTF-8.

### Why
It is inconvenient and not obvious when someone commits a CAP with CRLF line endings. It's inconvenient when multiple people are working on a CAP. An example of this is CAP-35. I noticed it's line endings when working on it a couple months ago, and it made diffs confusing and difficult to work with, but I didn't do anything about it then. It later caused issues with #950. We should make #950 more resilient to this specific issue, but I think it is also worthwhile to require some consistency in the CAPs that get committed.

I've also noticed we're not consistent with the encoding of our documents either. Some are UTF-8, others are ASCII. This probably matters much less and hasn't actually caused a problem because I think any of the ASCII file are ASCII-7BIT and therefore a subset of UTF-8, and not ASCII-8BIT.

leighmcculloch:
### What
Add action that requires LF line endings, and errors if CRLF line endings are present in CAPs.

### Why
Different line endings are inconvenient when folks on different operating systems are working together on documents. I found this when working on CAP-35 but ignore it at the time. Discovered in #950 that the CRLF line endings can affect diffs.

Close #953

### Known Limitations
This action is limited to checking CAPs as a starting point. We can extend it to SEPs too, but let's start here.


leighmcculloch:
This test will pass once #952 is merged.

.github/workflows/lineendings.yml
```diff
@@ -0,0 +1,15 @@
...
+    - master
+  pull_request:
+
+jobs:
+  lineendings:
+    runs-on: ubuntu-latest
+    steps:
+    - run: sudo apt-get update && sudo apt-get install dos2unix
+    - uses: actions/checkout@v2
+    - run: (ls core/cap-*.md | xargs -I '{}' sh -c 'cat {} | dos2unix | cmp - {}') || (echo 'Documents listed above contain CRLF line endings, and should be LF.'; exit 1)
```
MonsieurNicolas, line 15: Do we also need to enforce encoding or crlf is enough?

I see that dos2unix allows to enforce utf8bom for example.

leighmcculloch: I'd like to enforce both, but I'd like to start with only line endings in this PR and follow up with utf8 in a separate pr once I understand the state of that better I'm the repo.

MonsieurNicolas (approved):

.github/workflows/lineendings.yml
leighmcculloch: I looked more at this, I don't think we need to enforce utf8bom, just utf8 is sufficient whether the bom is present or not. Many of our CAPs only used ASCII characters and so the distinction is not particularly useful. If any file strays outside utf8 the dos2unix tool should error on that I think.

Unless you think we should require utf8bom, or need to do more testing here, I think we're okay.

PR #957, opened 2021-05-27
SEP-8: remove requirement where incoming transactions needed to be signed by the wallet

marcelosalloum:
### What

Remove SEP-8 requirement where incoming transactions needed to be signed by the wallet.

### Why

I don't really see any benefit of enforcing that rule.

### DIFF
--- a/ecosystem/sep-0008.md
+++ b/ecosystem/sep-0008.md
@@ -31,7 +31,7 @@ Implementing a regulated asset requires these parts:
 
 ## Regulated Assets Approval Flow
 
-1. Wallet creates and signs a transaction.
+1. Wallet creates a transaction.
 2. Wallet resolves asset information and detects that it's a regulated asset.
 	1. Wallet can detect whether an asset is a regulated asset by checking the [authorization flags] of the asset issuer's account.
 3. Wallet sends the transaction to the approval server.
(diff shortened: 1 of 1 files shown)


leighmcculloch (commented):
Could you update the meta? Bump the minor version and update the updated at timestamp.

I think we need to discuss this with @tomerweller. I have requested his review.

howardtw:
Indeed, it's up to the approval server what kind of rules it wants to enforce.

When I worked on SEP-8, I often assumed that an approval server might be more willing to sign a transaction that has been signed by the user, but it doesn't have to be a requirement.

leighmcculloch (commented):
I think this change is good to make, since it reflects the implementations we're aware of today. I think it would be okay to merge this on that basis, but I'd like @tomerweller to review it before you do, for this reason I'm just leaving a comment and defer to him to approve.

As Nick pointed out, since the idea is we would recommend using SEP-10 in place of this, there might be some value in thinking about that use case more before merging this. But on the same note, right now SEP-8 promotes an API that people aren't implementing, and it'd be better for new implementers to be compatible with the ecosystem existing implementations. So I think I'd rather see this merged sooner than later.

leighmcculloch (approved):
cc @tomerweller

tomerweller:
> When I worked on SEP-8, I often assumed that an approval server might be more willing to sign a transaction that has been signed by the user

Yes, that's the reason why the wallet signature is there. SEP-8 predates SEP-10. If we want to remove this requirement we can mandate or recommend SEP-10.

leighmcculloch:
@tomerweller All known implementers don't require the transaction to be signed first, and don't use SEP10. I think we should move ahead with this change and introduce SEP10 when the need arises.

PR #960, opened 2021-05-28
SEP-12: clarify the need for 'type' parameters in GET and PUT requests

JakeUrban:
Recommends clients always pass the customer `type` in `GET` & `PUT` requests.

### Context

I discovered that our `testanchor.stellar.org` SEP-12 implementation was returning `NEEDS_INFO` status values for customers who had passed all required fields for a SEP-6 deposit and received a customer `id`.

Upon further inspection I realized that this was because our `sep6-withdraw` customer `type` requires additional fields that the `sep6-deposit` `type` value does not, and the server was being conservative by assuming the request was for the `type` that required the most information.

Therefore, in order to ensure the correct `status` is returned for a customer in `GET /customer`, `type` should always be passed unless no types are specified in the SEP-31 or SEP-6 implementation.

### DIFF
--- a/ecosystem/sep-0012.md
+++ b/ecosystem/sep-0012.md
@@ -6,8 +6,8 @@ Title: KYC API
 Author: Interstellar
 Status: Active
 Created: 2018-09-11
-Updated: 2021-05-19
-Version 1.6.2
+Updated: 2021-05-28
+Version 1.6.3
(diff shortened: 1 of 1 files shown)


leighmcculloch (approved):
:nods: Good catch. :shipit:

PR #962, opened 2021-06-02
SEP-31: include some details about the Sending and Receiving Anchors to make the SEP clearer

marcelosalloum:
### What

Disambiguate parts of the SEP and add further recommendation about refunded transactions.

### Why

When reading the SEP, some details and interactions between Sending and Receiving Anchors were not very clear.

### DIFF
--- a/ecosystem/sep-0031.md
+++ b/ecosystem/sep-0031.md
@@ -351,7 +351,7 @@ Name | Type | Description
 `stellar_memo` | string | The memo to attach to the Stellar payment.
 `started_at` | UTC ISO 8601 string | (optional) Start date and time of transaction.
 `completed_at` | UTC ISO 8601 string | (optional) Completion date and time of transaction.
-`stellar_transaction_id` | string | (optional) The transaction_id on Stellar network of the transfer that initiated the payment from the Sending Anchor to the Receiving Anchor.
+`stellar_transaction_id` | string | (optional) The transaction_id on Stellar network of the transfer that initiated the payment.
 `external_transaction_id` | string | (optional) The ID of transaction on external network that completes the payment into the receivers account.
 `refunded` | boolean | (optional) Should be true if the transaction was refunded. Refunded transactions should also have their `status` set to `error`. Not including this field means the transaction was not refunded.
 `required_info_message` | string | (optional) A human-readable message indicating any errors that require updated information from the sender.
--- a/ecosystem/sep-0031.md
+++ b/ecosystem/sep-0031.md
@@ -346,12 +346,12 @@ Name | Type | Description
 `amount_in` | string | (optional) The amount received or to be received by the Receiving Anchor. Up to 7 decimals are permitted. Excludes any fees charged after Receiving Anchor receives the funds.
 `amount_out` | string | (optional) The amount sent by the Rending Anchor to Receiving Client. Up to 7 decimals are permitted.
 `amount_fee` | string | (optional) The amount of fee charged by the Receiving Anchor.
-`stellar_account_id` | string | The Stellar account to send the payment to.
+`stellar_account_id` | string | The Receiving Anchor's Stellar account the Sending Anchor will be making the payment to.
 `stellar_memo_type` | string | The type of memo to attach to the Stellar payment: `text`, `hash`, or `id`.
 `stellar_memo` | string | The memo to attach to the Stellar payment.
 `started_at` | UTC ISO 8601 string | (optional) Start date and time of transaction.
 `completed_at` | UTC ISO 8601 string | (optional) Completion date and time of transaction.
-`stellar_transaction_id` | string | (optional) The transaction_id on Stellar network of the transfer that initiated the payment.
+`stellar_transaction_id` | string | (optional) The transaction_id on Stellar network of the transfer that initiated the payment from the Sending Anchor to the Receiving Anchor.
 `external_transaction_id` | string | (optional) The ID of transaction on external network that completes the payment into the receivers account.
 `refunded` | boolean | (optional) Should be true if the transaction was refunded. Not including this field means the transaction was not refunded.
 `required_info_message` | string | (optional) A human-readable message indicating any errors that require updated information from the sender.
--- a/ecosystem/sep-0031.md
+++ b/ecosystem/sep-0031.md
@@ -7,8 +7,8 @@ Title: Cross-Border Payments API
 Author: SDF
 Status: Active
 Created: 2020-04-07
-Updated: 2021-05-17
-Version 1.2.2
+Updated: 2021-06-02
+Version 1.2.3
 ```
 
 ## Simple Summary
--- a/ecosystem/sep-0031.md
+++ b/ecosystem/sep-0031.md
@@ -86,7 +86,7 @@ All endpoints respond with content type:
 1. The Sending Anchor makes [SEP-12 PUT /customer](sep-0012.md#customer-put) requests containing the collected information for each client.
 1. On successful registration (`202` HTTP status), the Sending Anchor makes a `POST /transactions` request to create a transaction record with the Receiving Anchor. This request contains the `id`s returned by the `PUT /customer` requests, the `transaction.fields` values collected from the Sending Client, as well as the transaction amount and asset information. The response will include a `id` that will be used to check the status of the transaction.
 1. The Sending Anchor makes `GET /transactions/:id` requests until the transaction's `status` is `pending_sender`. 
-1. The Sending Anchor submits the payment transaction to Stellar using the memo and memo_type specified by the Receiving Anchor `GET /transactions/:id` response fields `stellar_memo` and `stellar_memo_type`.
+1. The Sending Anchor submits the payment transaction to Stellar using the information provided in the `POST /transactions` response.
 1. The Sending Anchor makes `GET /transactions/:id` requests until the transaction's `status` is `completed`, `error`, `pending_customer_info_update`, or `pending_transaction_info_update`.
 1. If `completed`, the Sending Anchor should notify the Sending Client that funds have been delivered to the Receiving Client.
 1. If `error`, the Receiving Anchor should be contacted to resolve the situation.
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0031.md
```diff
@@ -86,7 +86,7 @@ All endpoints respond with content type:
 1. The Sending Anchor makes [SEP-12 PUT /customer](sep-0012.md#customer-put) requests containing the collected information for each client.
 1. On successful registration (`202` HTTP status), the Sending Anchor makes a `POST /transactions` request to create a transaction record with the Receiving Anchor. This request contains the `id`s returned by the `PUT /customer` requests, the `transaction.fields` values collected from the Sending Client, as well as the transaction amount and asset information. The response will include a `id` that will be used to check the status of the transaction.
 1. The Sending Anchor makes `GET /transactions/:id` requests until the transaction's `status` is `pending_sender`. 
-1. The Sending Anchor submits the payment transaction to Stellar.
+1. The Sending Anchor submits the payment transaction to Stellar using the memo and memo_type specified by the Receiving Anchor `GET /transactions/:id` response fields `stellar_memo` and `stellar_memo_type`.
```
JakeUrban, line 89: Since this clarification leaves out the `stellar_account_id` response parameter, I think it's better to say the following:

```suggestion
1. The Sending Anchor submits the payment transaction to Stellar using the information provided in the `POST /transactions` response.
```

ecosystem/sep-0031.md
```diff
@@ -346,14 +346,14 @@ Name | Type | Description
...
 `amount_out` | string | (optional) The amount sent by the Rending Anchor to Receiving Client. Up to 7 decimals are permitted.
 `amount_fee` | string | (optional) The amount of fee charged by the Receiving Anchor.
-`stellar_account_id` | string | The Stellar account to send the payment to.
+`stellar_account_id` | string | The Receiving Anchor's Stellar account the Sending Anchor will be making the payment to.
 `stellar_memo_type` | string | The type of memo to attach to the Stellar payment: `text`, `hash`, or `id`.
 `stellar_memo` | string | The memo to attach to the Stellar payment.
 `started_at` | UTC ISO 8601 string | (optional) Start date and time of transaction.
 `completed_at` | UTC ISO 8601 string | (optional) Completion date and time of transaction.
-`stellar_transaction_id` | string | (optional) The transaction_id on Stellar network of the transfer that initiated the payment.
+`stellar_transaction_id` | string | (optional) The transaction_id on Stellar network of the transfer that initiated the payment from the Sending Anchor to the Receiving Anchor.
```
JakeUrban, line 354: I think this change is a bit redundant. The only Stellar transaction involved in a SEP-31 flow is the payment between these two parties, I don't see a reason to mention them here in addition to the `stellar_account_id` definition.

ecosystem/sep-0031.md
```diff
@@ -346,14 +346,14 @@ Name | Type | Description
...
 `external_transaction_id` | string | (optional) The ID of transaction on external network that completes the payment into the receivers account.
-`refunded` | boolean | (optional) Should be true if the transaction was refunded. Not including this field means the transaction was not refunded.
+`refunded` | boolean | (optional) Should be true if the transaction was refunded, and accompanied by the status `"error"` to make sure the Sending Anchor won't try to handle this transaction in any way. Not including this field means the transaction was not refunded.
```
JakeUrban, line 356:
```suggestion
`refunded` | boolean | (optional) Should be true if the transaction was refunded. Refunded transactions should also have their `status` set to `error`. Not including this field means the transaction was not refunded.
```

JakeUrban (changes requested):
Thanks @marcelosalloum, I added a few suggestions to phrasing. Once updated I think we can merge.

ecosystem/sep-0031.md
```diff
@@ -86,7 +86,7 @@ All endpoints respond with content type:
 1. The Sending Anchor makes [SEP-12 PUT /customer](sep-0012.md#customer-put) requests containing the collected information for each client.
 1. On successful registration (`202` HTTP status), the Sending Anchor makes a `POST /transactions` request to create a transaction record with the Receiving Anchor. This request contains the `id`s returned by the `PUT /customer` requests, the `transaction.fields` values collected from the Sending Client, as well as the transaction amount and asset information. The response will include a `id` that will be used to check the status of the transaction.
 1. The Sending Anchor makes `GET /transactions/:id` requests until the transaction's `status` is `pending_sender`. 
-1. The Sending Anchor submits the payment transaction to Stellar.
+1. The Sending Anchor submits the payment transaction to Stellar using the information provided in the `POST /transactions` response.
```
leighmcculloch, line 89: ❗ What information will it use? This is somewhat vague. If the intent is to say to send the an address that was specified in the post, it might be worth being more specific about that.

ecosystem/sep-0031.md
marcelosalloum: I agree that's somewhat implicit but it wasn't very clear to me in my first read, that's why I added it. I think it's ok to leave it out of this update though.

ecosystem/sep-0031.md
marcelosalloum: https://github.com/stellar/stellar-protocol/pull/962/commits/993c85e7f02c1bf8ddc336b142df5104690f2747

ecosystem/sep-0031.md
marcelosalloum: https://github.com/stellar/stellar-protocol/pull/962/commits/37e9d9030c066f8aa6826509a17a68f0eda64fd6

ecosystem/sep-0031.md
```diff
@@ -346,14 +346,14 @@ Name | Type | Description
 `amount_in` | string | (optional) The amount received or to be received by the Receiving Anchor. Up to 7 decimals are permitted. Excludes any fees charged after Receiving Anchor receives the funds.
 `amount_out` | string | (optional) The amount sent by the Rending Anchor to Receiving Client. Up to 7 decimals are permitted.
 `amount_fee` | string | (optional) The amount of fee charged by the Receiving Anchor.
-`stellar_account_id` | string | The Stellar account to send the payment to.
+`stellar_account_id` | string | The Receiving Anchor's Stellar account the Sending Anchor will be making the payment to.
```
leighmcculloch, line 349: 💡 I could be wrong, but when I read this my head automatically inserted this word between these two, then I realized it wasn't there. 
```suggestion
`stellar_account_id` | string | The Receiving Anchor's Stellar account that the Sending Anchor will be making the payment to.
```

ecosystem/sep-0031.md
```diff
@@ -279,7 +279,7 @@ This is the successful case where a Receiving Anchor confirms that they can fulf
 
 Name | Type | Description
 -----|------|------------
-`id` | string | The persistent identifier to check the status of this payment.
+`id` | string | The persistent identifier to check the status of this payment transaction.
```
leighmcculloch, line 282: ❔ What distinction are we making by specifying this as the `Stellar payment transaction` vs the `Stellar payment`? We refer to the payment as simply a Stellar payment elsewhere in the doc, like two lines down.

ecosystem/sep-0031.md
marcelosalloum: https://github.com/stellar/stellar-protocol/pull/962/commits/c7d4c260aa3c31b095a203a1fdc0138a5c271c53

ecosystem/sep-0031.md
```diff
@@ -346,14 +346,14 @@ Name | Type | Description
...
 `stellar_memo_type` | string | The type of memo to attach to the Stellar payment: `text`, `hash`, or `id`.
 `stellar_memo` | string | The memo to attach to the Stellar payment.
 `started_at` | UTC ISO 8601 string | (optional) Start date and time of transaction.
 `completed_at` | UTC ISO 8601 string | (optional) Completion date and time of transaction.
-`stellar_transaction_id` | string | (optional) The transaction_id on Stellar network of the transfer that initiated the payment.
+`stellar_transaction_id` | string | (optional) The transaction_id on Stellar network of the transfer that initiated the payment from the Sending Anchor to the Receiving Anchor.
 `external_transaction_id` | string | (optional) The ID of transaction on external network that completes the payment into the receivers account.
-`refunded` | boolean | (optional) Should be true if the transaction was refunded. Not including this field means the transaction was not refunded.
+`refunded` | boolean | (optional) Should be true if the transaction was refunded. Refunded transactions should also have their `status` set to `error`. Not including this field means the transaction was not refunded.
```
leighmcculloch, line 356: ❔ Why do refund transactions have a status of error?

JakeUrban (dismissed):

leighmcculloch (commented):
One thing I think should be addressed (❗), and some not-critical suggestions (💡). I also asked some questions (❔).

ecosystem/sep-0031.md
JakeUrban: There isn't a better status for it. In retrospect we should have used a `refunded` status instead of a boolean here.

ecosystem/sep-0031.md
JakeUrban: The response includes the destination account and the memo to use in the transaction. I felt we should refer to the endpoint that describes those attributes instead of explicitly mentioning them here, which we don't do for the other calls described in this section.

marcelosalloum: IMO if we describe all fields here the text tends to become outdated as new fields are included so it's easier to maintain this spec if we direct the user to the endpoint containing the desired information.

leighmcculloch: Makes sense, thanks for explaining.

leighmcculloch (approved):
I think this is good. Some questions and suggestions in my prior review, but none critical.

ecosystem/sep-0031.md
marcelosalloum: For this SEP there's kind of an intersection between payment and transaction definitions even though they are distinct Stellar concepts.

The motivation for that change is the text said this is a "payment id" but it's returned from the `/transactions` endpoint, so it made sense from a syntax perspective to clarify that. That change may be unnecessary though.

PR #963, opened 2021-06-03
SEP-23: Update status to Active

leighmcculloch:
### What
Update the status of SEP-23 to Active, set version to v1.0.0.

### Why
The SEP is implemented in Horizon, the Go SDK, the JS SDK, stc, and more. At this point the inertia that would be required to change SEP-23 is significantly high because changing it would require breaking changes to all these things. Even though in some of these things the functionality SEP-23 defines is specified as opt-in the implementation carries weight that'll be difficult to coordinate in changing. Therefore I think this SEP should be considered in active status, and not draft. It has moved past being a draft because we would not make arbitrary changes to it at this point.

cc @stanford-scs @tomerweller @ire-and-curses @shanzzam @2opremio


tomerweller (approved):
lgtm

tomerweller (approved):
ok

PR #972, opened 2021-06-11
SEP-6: fix typos and update wording

marcelosalloum:
### What

Fix typos and update wording.

### DIFF
--- a/ecosystem/sep-0006.md
+++ b/ecosystem/sep-0006.md
@@ -131,7 +131,7 @@ Request Parameters:
 
 Name | Type | Description
 -----|------|------------
-`asset_code` | string | The code of the off-chain asset the user wants to deposit with the anchor. The value passed must match one of the codes listed in the [/info](#info) response's deposit object.
+`asset_code` | string | The code of the asset the user wants to deposit with the anchor. The value passed must match one of the codes listed in the [/info](#info) response's deposit object.
 `account` | `G...` string | The stellar account ID of the user that wants to deposit. This is where the asset token will be sent.
 `memo_type` | string | (optional) Type of memo that anchor should attach to the Stellar payment transaction, one of `text`, `id` or `hash`.
 `memo` | string | (optional) Value of memo to attach to transaction, for `hash` this should be base64-encoded.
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0006.md
```diff
@@ -131,7 +131,7 @@ Request Parameters:
 
 Name | Type | Description
 -----|------|------------
-`asset_code` | string | The code of the asset the user wants to deposit with the anchor. The value passed must match one of the codes listed in the [/info](#info) response's deposit object.
+`asset_code` | string | The code of the off-chain asset the user wants to deposit with the anchor. The value passed must match one of the codes listed in the [/info](#info) response's deposit object.
```
JakeUrban, line 134: This is actually the on-chain asset, as specified in the `/info` description:

> The JSON object contains an entry for each Stellar asset that the anchor supports for SEP6 deposit and/or withdrawal.

ecosystem/sep-0006.md
```diff
@@ -814,7 +814,7 @@ Request parameters:
 
 Name | Type | Description
 -----|------|------------
-`id` | string | (optional) The id of the transaction.
+`id` | (optional) string | The id of the transaction.
```
JakeUrban, line 817: I think we should keep this as-is, since it matches the syntax used in our other endpoints and SEPs

JakeUrban (changes requested):
Most of these changes are great, I made 2 comments for minor adjustments and then it should be good to merge 👍

ecosystem/sep-0006.md
marcelosalloum: Reverted the change in https://github.com/stellar/stellar-protocol/pull/972/commits/2672aba7e7d62e3cc830f7b5cc02b6c6326915bf.

ecosystem/sep-0006.md
marcelosalloum: Reverted this change in https://github.com/stellar/stellar-protocol/pull/972/commits/2672aba7e7d62e3cc830f7b5cc02b6c6326915bf.

JakeUrban (approved):

PR #973, opened 2021-06-11
SEP-6: integrate SEP-38

### LINKED ISSUE #917: SEP-6: Integrate SEP-38

marcelosalloum:
### What

Integrate SEP-38 into SEP-6.

### Changes

* Add reference to SEP-38 `ANCHOR_QUOTE_SERVER` variable that should be included in the toml file if quotes are supported
* Add the "Asset Conversions" section explaining SEP-6 now supports conversions between non-equivalent assets, like `BRL <> USDC` for instance.
* Add endpoints `GET /deposit-exchange` and `GET /withdraw-withdraw`.
* Update the "transaction" object by adding the following fields:
  * `amount_in_asset`
  * `amount_out_asset`
  * `amount_fee_asset`

Closes #917


ecosystem/sep-0006.md
```diff
@@ -69,6 +71,12 @@ Access-Control-Allow-Origin: *
...
 
+## Asset Conversions
+
+This protocol was originally designed to provide conversions between off-chain and on-chain equivalent assets – for instance `BRL <> BRLT`, `USD <> USDC`, `NGN <> NGNT`, etc. – but it also describes how to make conversions between off-chain and on-chain non-equivalent assets like `BRL <> USDC`.
+
+To enable the conversion of non-equivalent assets the Anchor needs to: (i) mark the asset with `"quotes_supported": true` in its [`/info`](#info) endpoint; and (ii) implement the [SEP-38 Anchor RFQ API](sep-0038.md), which enables the Anchor to provide quotes to the Wallet. These quotes can be [indicative](https://www.investopedia.com/terms/i/indicativequote.asp) or [firm](https://www.investopedia.com/terms/f/firmquote.asp). If a firm quote is used, the Stellar transaction _must_ be submitted to the Stellar network prior to the quote's expiration. More specifically, the `created_at` timestamp returned in the [Horizon GET /transactions/:id](https://developers.stellar.org/api/resources/transactions/) request must be earlier than the quote's expiration.
```
JakeUrban, line 78: 💡  I think it would be more helpful to outline the flow of funds (user --> anchor --> user) for both the deposit and withdrawal case where non-equivalent assets are being exchanged, instead of listing the technical requirements in the same section that introduces the concept.

The reason differentiating between deposit and withdrawal is necessary is because in the deposit case the user has to take steps to deliver funds to the anchor's off-chain account before the quotes expiration. In the withdrawal case the wallet just needs to send funds from the user's Stellar account before the quotes expiration.

I also think we can leave the details of listing `quotes_supported` outside this section.

ecosystem/sep-0006.md
```diff
@@ -144,6 +154,8 @@ Name | Type | Description
...
 `claimable_balance_supported` | string | (optional) `true` if the client supports receiving deposit transactions as a claimable balance, `false` otherwise.
+`destination_asset` | string | (optional) The on-chain asset the Anchor will deliver to the user's Stellar account. The value must match one of the `asset` values included in a [`SEP-38 GET /prices?sell_asset=iso4217:<asset_code>`](sep-0038.md#get-prices) response using [SEP-38 Asset Identification Format](sep-0038.md#asset-identification-format). If neither this field nor `quote_id` are set, it's assumed the Anchor will be providing a bridge between equivalent assets, for instance receiving fiat USD and depositing USDC to the user's Stellar account.
+`quote_id` | string | (optional) The `id` returned from a `SEP-38 POST /quote` response. If this attribute is specified, none of the attributes `asset_code`, `amount` or `destination_asset` should be included in the request. The Stellar transaction used to send the asset _must_ have a `started_at` timestamp that is earlier than the quote's `expires_at` attribute.
```
JakeUrban, line 158: > The Stellar transaction used to send the asset _must_ have a `started_at` timestamp that is earlier than the quote's `expires_at` attribute.

❗  This is not true, since the `expires_at` attribute from the quote describes the time by which the user must deliver funds to the anchor, not the opposite. If the user has delivered funds to the anchor before the quote's expiration is should be acceptable for the anchor to deposit funds to the user's stellar account after the expiration.

ecosystem/sep-0006.md
```diff
@@ -330,6 +344,8 @@ Name | Type | Description
...
+`destination_asset` | string | (optional) The off-chain asset the Anchor will deliver to the user's account. The value must match one of the `asset` values included in a [`SEP-38 GET /prices?sell_asset=stellar:<asset_code>:<asset_issuer>`](sep-0038.md#get-prices) response using [SEP-38 Asset Identification Format](sep-0038.md#asset-identification-format). If neither this field nor `quote_id` are set, it's assumed the anchor will be providing a bridge between equivalent assets, for instance receiving USDC in the Stellar network and depositing fiat USD to the user's bank account.
+`quote_id` | string | (optional) The `id` returned from a `SEP-38 POST /quote` response. If this attribute is specified, none of the attributes `asset_code`, `amount` or `destination_asset` should be included in the request. The Stellar transaction used to send the asset _must_ have a `started_at` timestamp that is earlier than the quote's `expires_at` attribute.
```
JakeUrban, line 348: > The Stellar transaction used to send the asset _must_ have a `started_at` timestamp that is earlier than the quote's `expires_at` attribute.

❗  There is no `started_at` attribute on the a Stellar transaction's record in Horizon, see the [docs](https://developers.stellar.org/api/resources/transactions/object/). I think you mean `created_at`. I saw you had a question about this on the SEP-31 PR, happy do discuss further.

ecosystem/sep-0006.md
```diff
@@ -445,6 +461,7 @@ The wallet must use the response fields in the following way to complete the wit
 - `withdraw_memo`: (if specified) use this memo in the payment transaction to the anchor.
 - `withdraw_memo_type`: use this as the memo type.
 - `amount_in`: the amount expected in the Stellar payment.
+- `expires_at`: (if specified) if already expired, the wallet should not allow the user to proceed with the deposit/withraw action.
```
JakeUrban, line 464: 💡  I think we can omit this from the change since this field is retrievable from the [`GET /quote`](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0038.md#get-quote) endpoint and we don't need to duplicate information. The expiration also only refers to the delivery of funds to the anchor, vs. the entire transaction completing, so I think its confusing to list it on the transaction object in this SEP.

ecosystem/sep-0006.md
```diff
@@ -605,6 +625,7 @@ The JSON object contains an entry for each Stellar asset that the anchor support
 
 * `enabled`: `true` if SEP-6 withdrawal for this asset is supported
 * `authentication_required`: Optional. `true` if client must be [authenticated](#authentication) before accessing the deposit endpoint for this asset. `false` if not specified.
+* `quotes_supported`: Optional. `true` if client can use [SEP-38] quotes to perform a bridge between non-equivalent assets, for instance by withdrawing fiat BRL from Stellar-USDC. `false` if not specified.
```
JakeUrban, line 628: ❓  Do we want to use `quote_required` vs `quotes_supported`? Or should we have both attributes?

Imagine an anchor operates in a country that only has one fiat currency (ex. the US for USD). In this case anchoring another Stellar asset would require `quote_required` since thats the _only_ method of receiving that asset. 

But imagine an anchor operates in a country that has multiple fiat currencies (or multiple state-recognized / bank-supported currencies). In this case `quotes_supported` should be used because the user _could_ want to exchange non-equivalent assets OR the user could just want to exchange equivalent assets. Thoughts?

ecosystem/sep-0006.md
```diff
@@ -720,6 +744,7 @@ Name | Type | Description
 `withdraw_memo_type` | string | (optional) Memo type for `withdraw_memo`.
 `started_at` | UTC ISO 8601 string | (optional) Start date and time of transaction.
 `completed_at` | UTC ISO 8601 string | (optional) Completion date and time of transaction.
+`expires_at` | UTC ISO 8601 string | (optional) Expiration date for the anchor to receive that payment and proceed with the deposit/withdraw at the agreed quote. This value should be present if and only if the deposit/withdraw was submitted using the `quote_id` parameter. If already expired, the wallet should not allow the user to proceed with the deposit/withraw action.
```
JakeUrban, line 747: Same point about omitting this attribute.

JakeUrban (changes requested):
Some ideas (💡 ) , questions (❓ ), and requests (❗ ), but a great start. Also remember to update the meta info.

ecosystem/sep-0006.md
marcelosalloum: Good catch, thanks for spotting that! Here's the fix: https://github.com/stellar/stellar-protocol/pull/973/commits/c1a8a926f80e7c3a22efa7cc6c35ad8d350fcb23

ecosystem/sep-0006.md
marcelosalloum: You're right, that part of the description didn't make much sense for the deposit. Here's the fix: https://github.com/stellar/stellar-protocol/pull/973/commits/3c832f6ce361e397d3632949f2a1bfc756915649

ecosystem/sep-0006.md
marcelosalloum: Agreed! Removed it in https://github.com/stellar/stellar-protocol/pull/973/commits/169e4a1408ddafe6cd87a2008f4384e3a363403f.

ecosystem/sep-0006.md
marcelosalloum: Agreed! Removed it in https://github.com/stellar/stellar-protocol/pull/973/commits/169e4a1408ddafe6cd87a2008f4384e3a363403f.

ecosystem/sep-0006.md
marcelosalloum: Cool, I like those suggestions a lot! Take a look if this updated description would make things clearer: https://github.com/stellar/stellar-protocol/pull/973/commits/66d8220a12e8325fbe523f0f632820a79ef278b5

ecosystem/sep-0006.md
marcelosalloum: 👏  I really like the idea of having both `quote_required` and `quotes_supported` parameters since it would make the implementation clearer as developers won't need to make any assumptions nor read between the lines.

---

> Imagine an anchor operates in a country that only has one fiat currency (ex. the US for USD). In this case anchoring another Stellar asset would require quote_required since thats the only method of receiving that asset.

Even in the US we could have an Anchor serving both `USD <> USDC` and `ETH <> Stellar-ETH` for instance, and that same anchor could decide to make `USD <> Stellar-ETH`.

In addition to that, with `quote_required` we can have anchors focused only on making `USD <> Stellar-ETH`, which sounds like a valuable use case.

marcelosalloum: Added the `quotes_required` parameter in https://github.com/stellar/stellar-protocol/pull/973/commits/ef5c902b7f4a1690d95322df0c5c7bd579a78099.

ecosystem/sep-0006.md
```diff
@@ -75,7 +75,51 @@ This protocol involves the transfer of value, and so HTTPS is required for all e
...
+
+### Asset Conversions With Indicative Quotes
+
+For indicative quotes, users won't have a firm guarantee of the asset price they are going to pay. Here is how the conversion should happen in deposits and withdrawals for an asset pair like `BRL <> USDC`:
+
+**Deposit:**
+
+Considering the user will be making a bank deposit of fiat `BRL` in exchange for `USDC` in the Stellar network:
+
+1. The Wallet checks the assets' conversion rate in [`SEP-38 GET /price`](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0038.md#get-price) and shows the expected price to the user.
```
JakeUrban, line 88: I would change the word "expected" to "estimated", and maybe add a note about how the rate returned in the price at that moment and can change.

ecosystem/sep-0006.md
```diff
@@ -614,6 +660,7 @@ The JSON object contains an entry for each Stellar asset that the anchor support
 * `enabled`: `true` if SEP-6 deposit for this asset is supported
 * `authentication_required`: Optional. `true` if client must be [authenticated](#authentication) before accessing the deposit endpoint for this asset. `false` if not specified.
 * `quotes_supported`: Optional. `true` if client can use [SEP-38] quotes to perform a bridge between non-equivalent assets, for instance by depositing fiat BRL in exchange for USDC on the Stellar network. `false` if not specified.
+* `quotes_required`: Optional. `true` if client _must_ use [SEP-38] quotes to perform a bridge between non-equivalent assets, for instance by depositing fiat BRL in exchange for USDC on the Stellar network. `false` if not specified.
```
JakeUrban, line 663: In SEP-31 I was planning on using `quote_required` vs. `quotes_provided` and I'd like to be consistent in SEP-6. If you like `quotes` vs. `quote` I can make sure its the same in SEP-31 too.

JakeUrban (commented):
One requested change but otherwise looks good. Lets get the `Updated At` and `Version` updated and then I'll approve 👍

ecosystem/sep-0006.md
```diff
@@ -131,7 +185,7 @@ Request Parameters:
 
 Name | Type | Description
 -----|------|------------
-`asset_code` | string | The code of the asset the user wants to deposit with the anchor. The value passed must match one of the codes listed in the [/info](#info) response's deposit object.
+`asset_code` | string | (optional) The code of the asset the user wants to deposit with the anchor. The value passed must match one of the codes listed in the [/info](#info) response's deposit object. This field is mandatory if `quote_id` is not set.
```
accordeiro, line 188: I think the `asset_code` and `destination_asset` definitions here are a bit confusing. If you're doing a **deposit**, whatever you (as a client/user) are depositing is off-chain, and what you're getting in return from the anchor is on-chain.

There are now two options here:
a) you deposit X and get TokenX (the "traditional" Stellar deposit)
a) you deposit Y and get TokenX (the novel concept introduced by this PR)

The commonality here is that both will lead to you receiving TokenX, which should be one of the tokens specified on the anchor's `/info` endpoint. So it seems to me that what we always meant by `asset_code` is in fact the **destination** asset (TokenX), and what we're introducing now is the concept of a **source asset** when it's not a 1:1 conversion.

So my take is that:
1. `asset_code` should be the destination asset, and it should be one of the assets provided on the `/info` endpoint. This field should **remain mandatory**. This also allows for better backwards API compatibility.
2. we should create a new, optional param called `source_asset` that should have similar characteristics to the ones described for `destination_asset` below and should be accompanied by a `quote_id`.

accordeiro: You could also alternatively just make `destination_asset` mandatory, but this would be a breaking API change.

ecosystem/sep-0006.md
```diff
@@ -317,7 +375,7 @@ Request parameters:
 
 Name | Type | Description
 -----|------|------------
-`asset_code` | string | Code of the asset the user wants to withdraw. The value passed must match one of the codes listed in the [/info](#info) response's withdraw object.
+`asset_code` | string | (optional) Code of the asset the user wants to withdraw. The value passed must match one of the codes listed in the [/info](#info) response's withdraw object. This field is mandatory if `quote_id` is not set.
```
accordeiro, line 378: I think here the `asset_code` and `destination_asset` definitions make sense. I think `asset_code` should be mandatory though – why change that?

Also, I'm a bit confused about the overall developer / user experience. What should the anchor provide on `/info`? Is it:

a) the list of on-chain assets that can be withdrawn, or 
b) the list of off-chain assets they convert to?

Intuitively I'd go with (a) because of the current behavior. Is that what you're thinking?

accordeiro (commented):
This is a solid start – I think there are a few details to iron out, I've left some inline comments.

ecosystem/sep-0006.md
```diff
@@ -625,6 +672,7 @@ The JSON object contains an entry for each Stellar asset that the anchor support
 * `enabled`: `true` if SEP-6 withdrawal for this asset is supported
 * `authentication_required`: Optional. `true` if client must be [authenticated](#authentication) before accessing the deposit endpoint for this asset. `false` if not specified.
 * `quotes_supported`: Optional. `true` if client can use [SEP-38] quotes to perform a bridge between non-equivalent assets, for instance by withdrawing fiat BRL from Stellar-USDC. `false` if not specified.
+* `quotes_required`: Optional. `true` if client _must_ use [SEP-38] quotes to perform a bridge between non-equivalent assets, for instance by withdrawing fiat BRL from Stellar-USDC. `false` if not specified.
```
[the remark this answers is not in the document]
marcelosalloum: Why a duplicate? There's one `quotes_required` in the [deposit response section](https://github.com/stellar/stellar-protocol/blob/ef5c902b7f4a1690d95322df0c5c7bd579a78099/ecosystem/sep-0006.md#for-each-deposit-asset-response-contains) and another for the [withdrawal response section](https://github.com/stellar/stellar-protocol/blob/ef5c902b7f4a1690d95322df0c5c7bd579a78099/ecosystem/sep-0006.md#for-each-withdrawal-asset-response-contains). All other fields also show up in both sections, like `enabled` and `authentication_required` for instance.

ecosystem/sep-0006.md
marcelosalloum: Since there are two types of quotes, firm and indicative, maybe `quotes_*` would be more accurate? I don't feel strongly towards it though, so we could go with the singular form as well.

ecosystem/sep-0006.md
marcelosalloum: Seems like my understanding of the SEP was wrong then. Judging by the existing `asset_code` description:

> `asset_code` | string | The code of the asset the user wants to deposit with the anchor. (...)

I had the understanding that the asset code would be `X` if the user is depositing X to get TokenX. So what you're saying is the asset_code for a SEP-6 deposit would actually be `TokenX`?

ecosystem/sep-0006.md
marcelosalloum: > I think here the `asset_code` and `destination_asset` definitions make sense.

Is SEP-6 `asset_code` always the code of the asset in the Stellar network?

marcelosalloum: > Also, I'm a bit confused about the overall developer / user experience. What should the anchor provide on `/info`? Is it:
> 
> a) the list of on-chain assets that can be withdrawn, or
> b) the list of off-chain assets they convert to?
> 
> Intuitively I'd go with (a) because of the current behavior. Is that what you're thinking?

We should stick to the current behavior and avoid any breaking changes.

ecosystem/sep-0006.md
accordeiro: I don't think your understanding was wrong – my point is that this definition was intentionally ambiguous because effectively `asset_code` meant **both** `X` and `TokenX` as this SEP only described 1:1 asset conversions.

What I'm proposing is that, from now on, we standardize the concept of `asset_code` on both `/withdraw` and `/deposit` flows to always be the **on-chain** asset, and continue being a mandatory parameter on both. In summary:

1. Clients should find all available `asset_code`s on an anchor's `/info` endpoint. If not listed there, the anchor does not support that on-chain asset.
1. For all "traditional" 1:1 deposits or withdrawals, the client only needs to provide `asset_code`.
1. For all SEP38-powered deposits or withdrawals, the client also needs to provide: a `source_asset` (in the case of a deposit); or a `destination_asset` (in the case of a withdrawal).

This should keep the API behavior backwards-compatible for both clients and servers, but also enable them to augment their functionality with SEP-38 FX if they'd like.

ecosystem/sep-0006.md
accordeiro: > Is SEP-6 asset_code always the code of the asset in the Stellar network?

See https://github.com/stellar/stellar-protocol/pull/973#discussion_r651919114 – I don't think that's true as of yet (it's ambiguous), but that's what I'm proposing we standardize for.

ecosystem/sep-0006.md
marcelosalloum: Good suggestion! Updated in https://github.com/stellar/stellar-protocol/pull/973/commits/e25f8b306d471b13862b7d62f94a82773899dd3e.

ecosystem/sep-0006.md
marcelosalloum: > 3\. For all SEP38-powered deposits or withdrawals, the client also needs to provide: a `source_asset` (in the case of a deposit); or a `destination_asset` (in the case of a withdrawal).

Do you think `source_asset` and `destination_asset` should be mandatory when using a firm quote with SEP-38 `quote_id`?

JakeUrban: I think both parameters should be omitted if a firm quote is used, since those assets are already determined when the quote is given via the SEP-38 endpoint.

marcelosalloum: Here are the changes I made based on Alex's suggestions in this thread:
* Make it clear the `asset_code` is the on-chain asset and use `source_asset` for de deposit: https://github.com/stellar/stellar-protocol/pull/973/commits/1844408afb37146f9271c97e19320f3ed9349859
* Make `asset_code` mandatory again: https://github.com/stellar/stellar-protocol/pull/973/commits/f87f49e9f0bdfdbb601e6fec7001042d62b416d6
* Currently, `source_asset` and `destination_asset` are optional and only used for indicative quotes. Firm quotes will use `quote_id` instead.

marcelosalloum: > I think both parameters should be omitted if a firm quote is used, since those assets are already determined when the quote is given via the SEP-38 endpoint.

From an implementations standpoint, I agree with Jake that `asset_code` doesn't need to be mandatory since the anchor should have access to the desired asset through the asset_id.

JakeUrban (dismissed):
Pending the change for `asset_code` being optional when `quote_id` is provided, this LGTM.

ecosystem/sep-0006.md
accordeiro: I think conditional optionality of parameters makes APIs confusing, because it introduces many edge cases for the server-side implementer. What if the API caller provides mutually exclusive fields? What if the information in each of these fields is contradictory? If the user sends both a `quote_id` and the other fields (e.g. `source_asset` and `asset_code`), what should take precedence?

It also becomes harder to reason about the implementation from the client perspective: as we build more functionality into the same endpoint, we create more and more conditional rules, to the point that most of our params become optional and we need a whole section in the SEP to explain the possible field combinations. Think of this as the same principle you'd use for writing functions in your code: if you're overloading functions with optional parameters and adding a bunch of `if`s, maybe it's time to break down into several well-scoped functions. APIs are basically remote function calls.

My suggestion is that we take a step back and analyze if what we really should do is to create additional endpoints in this spec, so that each endpoint has a well-defined function and set of mandatory fields, and what we leave as optional is indeed optional information.

Here's what I'm thinking:

**OPTION A (more ergonomic, but also with more changes)**

1. Leave `/deposit` and `/withdraw` as currently defined in SEP-6.
1. Create specialized `/deposit/fx/indicative` and `/withdraw/fx/indicative` endpoints, which take additional `source_asset` and `dest_asset` fields respectively.
1. Create specialized `/deposit/fx/firm` and `/withdraw/fx/firm` endpoints that replace all asset info params with a single, mandatory, `quote_id`.

**OPTION B (less ergonomic, less changes)**

1. Modify the current `/deposit` and `/withdraw` flows to take an additional optional param (`source_asset` or `dest_asset` respectively). `asset_code` is still a mandatory field. The "traditional" and the indicative quote flows will rely on the same endpoint.
1. Create specialized `/deposit/firm-quote` and `/withdraw/firm-quote` endpoints that replace all asset info params with a single, mandatory, `quote_id`.

marcelosalloum: I really like the idea of having dedicated endpoints for each case, so the API gets easier to use and it reduces the chances of mistakes from both the Anchor and Wallet sides. Also, we guarantee that the current SEP-6 behavior will remain unchanged.

> **OPTION A (more ergonomic, but also with more changes)**
> 
> 1. Leave `/deposit` and `/withdraw` as currently defined in SEP-6.
> 2. Create specialized `/deposit/fx/indicative` and `/withdraw/fx/indicative` endpoints, which take additional `source_asset` and `dest_asset` fields respectively.
> 3. Create specialized `/deposit/fx/firm` and `/withdraw/fx/firm` endpoints that replace all asset info params with a single, mandatory, `quote_id`.

About this ☝️ we should additionally update the `/info` data with the new endpoints.

Also, what does `fx` stands for in `/deposit/fx/firm`? Maybe we could use `/deposit/quote/firm` instead?

accordeiro: Awesome! FX = foreign exchange. I'm not attached to the endpoint path names though, it was just a way to illustrate my reasoning. I like using `../quote/..` as well.

marcelosalloum: Updated the approach to use dedicated endpoints in https://github.com/stellar/stellar-protocol/pull/973/commits/d1fec88ff9a18166a53b781e231d943db01c14bf. It's best to re-review the whole PR though, rather than spotting the changes in that one commit.

JakeUrban: I'm going to push back on the idea of new endpoints. I think it introduces other potential issues and increases the development/maintenance burden of the implementer.

What happens if we have another use case that requires the same debate (new endpoint vs. added & omitted params)? We may be able to create new endpoints without issue, but its more likely that this new use case is not mutually exclusive with the use of firm & indicative quotes. We would need to have the following endpoints if we used the same strategy:

- /deposit
- /deposit/quotes/indicative
- /deposit/quotes/firm
- /deposit/new-use-case/A
- /deposit/new-use-case/B
- /deposit/quotes/indicative/new-use-case/A
- /deposit/quotes/indicative/new-use-case/B
- /deposit/quotes/firm/new-use-case/A
- /deposit/quotes/firm/new-use-case/B

I also think the developers of both client and server applications would write functions shared by all endpoints to avoid duplicate code, but it's not unlikely that these shared functions would require conditional logic to handle the variance of each use case. The end result would be code that would look a lot like the function written if there was just one endpoint.

If the implementor duplicates code to avoid this conditional logic, then they increase the surface area for bugs and increase the amount of work needed to make changes that affect all endpoints.

accordeiro: > If the implementor duplicates code to avoid this conditional logic, then they increase the surface area for bugs and increase the amount of work needed to make changes that affect all endpoints.

I don't think implementing multiple endpoints means you can't reuse functions in your backend implementation. In fact I think it enforces engineers to think about what is specific to a given endpoint, and what is common logic that can be shared.

> What happens if we have another use case that requires the same debate (new endpoint vs. added & omitted params)? We may be able to create new endpoints without issue, but its more likely that this new use case is not mutually exclusive with the use of firm & indicative quotes. We would need to have the following endpoints if we used the same strategy:
> - /deposit
> - /deposit/quotes/indicative
> - /deposit/quotes/firm
> - /deposit/new-use-case/A
> - /deposit/new-use-case/B
> - /deposit/quotes/indicative/new-use-case/A
> - /deposit/quotes/indicative/new-use-case/B
> - /deposit/quotes/firm/new-use-case/A
> - /deposit/quotes/firm/new-use-case/B

Can you give a concrete example? This is an on/off ramp API, so the basic primitive actions will continue to be **deposit** and **withdraw**. We can have variants for how to accomplish those (like a new alternative method for quotes), but I don't see them layering up.

Also, the alternative to that list of new URLs would be a table that explains all possible combinations of parameters for each desired task. Something like:

* For deposit with new-use-case A, use param X and Y, and do not provide param Z.
* For deposit with new-use-case B, use param X, Z and XX, and do not provide Y
* For deposit with new-use-case-A and indicative quotes, use param X and Y, similarly to the first bullet point, but in the  X use ISO format instead of urlencoded
* For deposit with new-use-case-B using indicative quotes, use param ZZ and nevermind all other params
* ...
* And BTW: all parameters are defined as optional, but they actually aren't.


What is more, what happens if an anchor only wants to implement a subset of these features? If there are multiple endpoints, they can simply not implement the ones they don't support. When we're overloading all possibilities in a single endpoint, it's pretty hard for clients to differentiate whether there's an implementation bug, or if the feature is simply not supported.

It seems pretty clear to me that having multiple endpoints is the way to scale this, while also keeping the SEP understandable. Even if it comes with the (in my opinion, very small) burden of maintaining and integrating with multiple endpoints.

If you're still not convinced, my proposal is that we meet halfway and go with **OPTION B** that I described above. Let me know your thoughts.

marcelosalloum: > Also, the alternative to that list of new URLs would be a table that explains all possible combinations of parameters for each desired task. Something like:
> 
> * For deposit with new-use-case A, use param X and Y, and do not provide param Z.
> * For deposit with new-use-case B, use param X, Z and XX, and do not provide Y
> * For deposit with new-use-case-A and indicative quotes, use param X and Y, similarly to the first bullet point, but in the  X use ISO format instead of urlencoded
> * For deposit with new-use-case-B using indicative quotes, use param ZZ and nevermind all other params
> * ...
> * And BTW: all parameters are defined as optional, but they actually aren't.

This ☝️  sounds very likely to increase complexity in both Wallet and Anchor implementations. If we have `n` mutually exclusive not-really-optional parameters there will be `n!` combinations of parameters that the Wallet and the Anchor will need to figure out. Even though most combinations can be invalid, it'll get pretty hard to figure out which combinations should be supported.

On the other hand, by splitting it into `n` different endpoints and making those parameters mandatory, it'll get much clearer which parameters are needed. The complexity will be `~O(n)`, which should be much preferred than anything exponential like `O(n!)`.

_PS: I'm not so sure if it's `O(n!)` or `O(n^2)`, but my point should remain the same regardless._

ecosystem/sep-0006.md
```diff
@@ -301,9 +359,91 @@ const result = await this.server.submitTransaction(transaction)
...
+`wallet_name` | string | (optional) In communications / pages about the deposit, anchor should display the wallet name to the user to explain where funds are going.
+`wallet_url` | string | (optional) Anchor should link to this when notifying the user that the transaction has completed.
+`lang` | string | (optional) Defaults to `en`. Language code specified using [ISO 639-1](https://en.wikipedia.org/wiki/ISO_639-1). `error` fields in the response should be in this language.
+`on_change_callback` | string | (optional) A URL that the anchor should `POST` a JSON message to when the `status` property of the transaction created as a result of this request changes. The JSON message should be identical to the response format for the [/transaction](#single-historical-transaction) endpoint.
+`amount` | string | (optional) The amount of the asset the user would like to deposit with the anchor. This field may be necessary for the anchor to determine what KYC information is necessary to collect.
```
JakeUrban, line 389: What units is `amount` in within the context of an indicative deposit? I think it needs to be in units of the `source_asset` since thats the asset held by the user, and also gives the anchor an opportunity to return 400 if the amount is too large or small.

This also has implications for the `max_amount` and `min_amount` attributes in the `/info` endpoint for the associated `deposit_indicative_quote` object.

ecosystem/sep-0006.md
```diff
@@ -375,6 +515,87 @@ Example:
...
+`memo_type` | string | (optional) Type of `memo`. One of `text`, `id` or `hash`.
+`wallet_name` | string | (optional) In communications / pages about the withdrawal, anchor should display the wallet name to the user to explain where funds are coming from.
+`wallet_url` | string | (optional) Anchor can show this to the user when referencing the wallet involved in the withdrawal (ex. in the anchor's transaction history).
+`lang` | string | (optional) Defaults to `en`. Language code specified using [ISO 639-1](https://en.wikipedia.org/wiki/ISO_639-1). `error` fields in the response should be in this language.
+`on_change_callback` | string | (optional) A URL that the anchor should `POST` a JSON message to when the `status` property of the transaction created as a result of this request changes. The JSON message should be identical to the response format for the [/transaction](#single-historical-transaction) endpoint.
+`amount` | string | (optional) The amount of the asset the user would like to deposit with the anchor. This field may be necessary for the anchor to determine what KYC information is necessary to collect.
```
JakeUrban, line 546:
```suggestion
`amount` | string | (optional) The amount of the on-chain asset the user would like to send to the anchor's Stellar account. This field may be necessary for the anchor to determine what KYC information is necessary to collect.
```
Similar to the notes about `/deposit_indicative_quote`

ecosystem/sep-0006.md
```diff
@@ -375,6 +515,87 @@ Example:
...
```
JakeUrban, line 586: `amount` should be removed here

ecosystem/sep-0006.md
JakeUrban: Actually, I think we should make `amount` required for the indicative endpoints. I think this is especially important for the deposit case because we don't have a great way of communicating source asset deposit limits, as mentioned in my comments in `/info`.

ecosystem/sep-0006.md
```diff
@@ -540,6 +761,60 @@ The response should be a JSON object like:
...
     }
   },
+  "deposit_indicative_quote": {
+    "USD": {
+      "enabled": true,
+      "authentication_required": true,
+      "fee_fixed": 5,
+      "fee_percent": 1,
+      "min_amount": 0.1,
+      "max_amount": 1000,
```
JakeUrban, line 771: I think we should remove these attributes for `deposit_indicative_quote` objects since the limits should be imposed on the asset sent by the user, which is the off-chain asset. Instead we can make `amount` required on the `deposit_indicative_quote` endpoint and enforce limits there and communicate those limits in 400 error messages.

Its not a great design, but `min_amount` and `max_amount` in terms of the on-chain asset can't be used until the anchor determines the exchange rate, which may not happen until after the user has sent the off-chain funds.

ecosystem/sep-0006.md
```diff
@@ -540,6 +761,60 @@ The response should be a JSON object like:
...
+    }
+  },
+  "deposit_firm_quote": {
+    "USD": {
+      "enabled": true,
+      "authentication_required": true,
+      "fee_fixed": 5,
+      "fee_percent": 1,
+      "min_amount": 0.1,
+      "max_amount": 1000,
```
JakeUrban, line 798: Since the amount of the Stellar asset delivered to the user is determined when the request for quote is made, I don't think it makes sense to have the `min_amount` or `max_amount` properties here either.

JakeUrban (changes requested):

ecosystem/sep-0006.md
JakeUrban: I'm not able to come up with a concrete use case, and using separate endpoints does make the API contract cleaner.

ecosystem/sep-0006.md
```diff
@@ -575,6 +850,70 @@ The response should be a JSON object like:
       "enabled": false
     }
   },
+  "withdraw_indicative_quote": {
+    "USD": {
+      "enabled": true,
```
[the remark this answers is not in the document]
marcelosalloum: https://github.com/stellar/stellar-protocol/pull/973/commits/b3b88bad67dd5c92cc08be582c1d7c995902fff8

ecosystem/sep-0006.md
marcelosalloum: Removed in https://github.com/stellar/stellar-protocol/pull/973/commits/46ff1babb42954d81b5e6ec2682aadc423a1fa82.

ecosystem/sep-0006.md
marcelosalloum: Updated in https://github.com/stellar/stellar-protocol/pull/973/commits/46ff1babb42954d81b5e6ec2682aadc423a1fa82.

ecosystem/sep-0006.md
marcelosalloum: That makes sense! Unless the value is dynamically defined from the Anchor side, but we don't want to impose that kind of a burden, right?

Removed those fields in https://github.com/stellar/stellar-protocol/pull/973/commits/ea0a84cd45149ffc6a757c6f96d25354e987802a.

ecosystem/sep-0006.md
marcelosalloum: Removed those fields in https://github.com/stellar/stellar-protocol/pull/973/commits/ea0a84cd45149ffc6a757c6f96d25354e987802a.

ecosystem/sep-0006.md
```diff
@@ -575,6 +850,70 @@ The response should be a JSON object like:
...
+      "authentication_required": true,
+      "fee_fixed": 5,
+      "fee_percent": 0,
```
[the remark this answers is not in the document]
marcelosalloum: > I think we need to mention that these attributes should only be present if fees are charged in units of the asset on Stellar

Only for `withdraw*_quotes` or also for `deposit*_quotes`? I did it for withdraws in https://github.com/stellar/stellar-protocol/pull/973/commits/0ae235ddd62fb7e5f4bb30f26bc4b24d13fbb41a.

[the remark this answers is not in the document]
marcelosalloum: > And, we should add `source_asset` & `destination_asset` as parameters to `GET /fee` as well.
> 
> Instead of adding these parameters to `/fee` we could create new endpoints `/fee/indicative_quote` and `fee/firm_quote` too.

If we make it so fees are always charged in Stellar assets we shouldn't need these changes. But maybe that is not a real-world assumption 🤔 . WDYT?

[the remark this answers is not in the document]
JakeUrban: > Only for withdraw*_quotes or also for deposit*_quotes? I did it for withdraws in 0ae235d.

For both, since fees could be charged in either currency for deposits as well.

> If we make it so fees are always charged in Stellar assets we shouldn't need these changes

Thats true, but I think its likely that businesses will want to charge fees in terms of the off-chain asset. We could make the argument that since the entity charging deposit/withdrawal fees is the same entity doing the FX, the entity is capable of charging fees in terms of the on-chain asset while designing the UX to communicate fees in terms of the off-chain asset. However I don't think we should introduce this complication if our protocol can support it with changes.

ecosystem/sep-0006.md
```diff
@@ -375,6 +515,86 @@ Example:
...
+
+```
+GET TRANSFER_SERVER/withdraw_firm_quote
+```
+
+Request parameters:
+
+Name | Type | Description
+-----|------|------------
+`quote_id` | string | The `id` returned from a `SEP-38 POST /quote` response. The Stellar transaction used to send the asset to the Anchor _must_ have a `created_at` timestamp that is earlier than the quote's `expires_at` attribute.
```
JakeUrban, line 575:
```suggestion
`quote_id` | string | The `id` returned from a `SEP-38 POST /quote` response. The Stellar transaction used to send the asset to the Anchor _must_ have a [`created_at`](https://developers.stellar.org/api/resources/transactions/object/) timestamp that is earlier than the quote's `expires_at` attribute.
```

marcelosalloum: https://github.com/stellar/stellar-protocol/pull/973/commits/e09f89c25f48e31c0aebeadd47473287f0371ed7

ecosystem/sep-0006.md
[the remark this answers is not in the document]
marcelosalloum: > Since the fees charged can be in units of either asset, I think we need to mention that these attributes should only be present if fees are charged in units of the asset on Stellar, since it would be misleading if fees were charged in terms of the off-chain asset.

Ok, now I'm making it clear that fees should be charged in terms of the incoming assets:
* withdrawal fees are charged in units of the **on-chain** assets: https://github.com/stellar/stellar-protocol/commit/0ae235ddd62fb7e5f4bb30f26bc4b24d13fbb41a
* deposits fees are charged in units of **off-chain** assets: https://github.com/stellar/stellar-protocol/pull/973/commits/12a2e4d27a9847e3dd57a7ee68a274fe2f8372c8

Also, added the new fee endpoints in https://github.com/stellar/stellar-protocol/pull/973/commits/373a083ab58b5044da385c66b230ac83119dafc2.

JakeUrban:
Adding to my team board for tracking

ecosystem/sep-0006.md
```diff
@@ -301,9 +361,91 @@ const result = await this.server.submitTransaction(transaction)
...
 
+## Deposit With Indicative Quote
+
+If the anchor supports [SEP-38] quotes, it can provide a deposit that makes a bridge between non-equivalent tokens by receiving, for instance BRL via bank transfer and in return sending the equivalent value (minus fees) as USDC to the user's Stellar account.
+
+The `/deposit_indicative_quote` endpoint allows a wallet to get deposit information from an anchor when the user intends to make a conversion between non-equivalent tokens using an [indicative quote](https://www.investopedia.com/terms/i/indicativequote.asp). With this endpoint, a user has all the information needed to initiate a deposit and it also lets the anchor specify additional information (if desired) that the user must submit via [SEP-12](sep-0012.md).
```
leighmcculloch, line 368: Could we use hyphens instead of underscores? There's no hard technical reason for this, but over time most APIs style guides including those at larger tech companies have required hyphens as separators and they're mostly the norm. [Ref](https://www.google.com/search?&q=api+hyphen+or+underscore+in+api)

ecosystem/sep-0006.md
leighmcculloch: @accordeiro asked me to read through this conversation.

> I think conditional optionality of parameters makes APIs confusing

> If the user sends both a `quote_id` and the other fields (e.g. `source_asset` and `asset_code`), what should take precedence?

I agree with the issue of optionality, but I think the issue of optionality is mostly avoidable without adding additional deposit/withdraw endpoints if we require all fields all the time and validate the inputs against the quote if it is present.

We could also deprecate asset_code and make source_asset and dest_asset required. When asset_code is provided we can make assumptions about source and deposit to satisfy existing use cases. For new implementations this would leave only the quote_id which would be optional, and I don't think that is so bad. In terms of what to do if the assets don't match the quote, we should error, in the same way if the deposit amount doesn't match the quote amount, so I don't think we need to consider precedence.

I also think the path we've chosen to use separate endpoints is fine too, although I have some concerns discussed below.

---

The deposit indicative quote endpoint, does it provide or work off of an indicative quote? I don't see that anywhere. It looks like it will just use whatever the market rate is. Will anyone use it? Consumer products I've used usually provide a firm quote.

If it is using a market rate, I think we should refer to that instead of referring to it as deposit with an indicative quote, because we should point to what is being used for the final conversion.

---

For a firm quote, what happens if you deposit after the expiry? Are you legally required to provide the amount of the firm quote before expired, or does it cancel silently? Does it fallback to market rates? Fwiw I've seen consumer products do the first (legal requirement) and last (market rates) of those things.

---

The use of the terms indicative and firm in the SEP-6 API is a bit confusing. I like that in SEP-38 we refer to these as the price API and the quote API.

Assuming we're still going with separate endpoints, I think they would be easier to understand if they were:

- /deposit – Deposit an asset without conversion.
- /deposit-exchange-with-market-rate – Deposit an asset involving exchange using the market rate.
- /deposit-exchange-with-quote – Deposit an asset involving exchange using a quote.

ecosystem/sep-0006.md
marcelosalloum: Fixed in https://github.com/stellar/stellar-protocol/pull/973/commits/c40b687c40c81dc2e96b15c95675c1dc7bf1fdd9

ecosystem/sep-0006.md
marcelosalloum: > I agree with the issue of optionality, but I think the issue of optionality is mostly avoidable without adding additional deposit/withdraw endpoints

Let me give it some thought 🤔

> If it is using a market rate, I think we should refer to that instead of referring to it as deposit with an indicative quote, because we should point to what is being used for the final conversion.

I agree this wording is clearer. I'm updating the description in https://github.com/stellar/stellar-protocol/pull/973/commits/4c2769685f1ee625e74a97cf7b0ef18d4816ac8e

> For a firm quote, what happens if you deposit after the expiry? Are you legally required to provide the amount of the firm quote before expired, or does it cancel silently? Does it fallback to market rates? Fwiw I've seen consumer products do the first (legal requirement) and last (market rates) of those things.

I think an anchor should figure out the next steps their own way since they have the user contact. IMO they could either send another quote for approval or return the payment to the user account. I don't think it should fallback to the market price because it can be very different from the user expectations.

leighmcculloch: > > I agree with the issue of optionality, but I think the issue of optionality is mostly avoidable without adding additional deposit/withdraw endpoints
> 
> Let me give it some thought 🤔

One more thing I want to highlight, most of the optionality we're trying to work around exists due to optimizations, not requirements. We're optimizing the request by removing fields we don't really need because the quote has the same data. We don't have to optimize the request, and the overall API complexity may be lower if we don't.

> I don't think it should fallback to the market price because it can be very different from the user expectations.

I agree that the expectations aren't great, but in reality this is what some money transfer services do. For example Wise does this. I think this problem can get complicated because the user has delivered the funds so if they aren't going to go ahead with the exchange how do they get them back to the user?

marcelosalloum: > I agree with the issue of optionality, but I think the issue of optionality is mostly avoidable without adding additional deposit/withdraw endpoints

TBH I'm not a big fan of changing an existent endpoint's interface... my rationale is the endpoints `/deposit` and `/withdraw` where created with the clear purpose of creating a common interface for exchanging equivalent assets that are 1:1, like `USD<>USDC`. Deprecating an API parameter could be very troublesome and bring instability to the network, if an anchor is already implementing it I don't think they should need to make a change just to keep it consistent with the documentation.

On the other hand, I agree that adding 4 new endpoints might be too verbose. Another option that takes into account your suggestions would be to create a single new endpoint called `/exchange` that takes both a `buy_asset` and a `sell_asset` in the forms of `iso4217:USD` and/or `stellar:<asset_code>:<asset_issuer>`. It could also take a `quote_id` for a firm quote. The deposit and withdrawal concepts would kind of lose their purpose in this design because ultimately the user will be just making an exchange where s/he always sends an asset to receive another asset in return.

WDYT @leighmcculloch, @JakeUrban?

marcelosalloum: > I agree that the expectations aren't great, but in reality this is what some money transfer services do. For example Wise does this. I think this problem can get complicated because the user has delivered the funds so if they aren't going to go ahead with the exchange how do they get them back to the user?

Now I get your point. It can get very complicated for the anchor to deal with this case, returning the payment to the user would definitely bring some operational costs. Maybe we could add a statement like this for quoted exchanges: _"If the user deposits the amount to the anchor after the quote expiration, the Anchor has the autonomy to make the asset conversion at market price. The anchor should make this information clear in the `how` field so the Wallet can display it to the user."_

leighmcculloch: > Deprecating an API parameter could be very troublesome and bring instability to the network

That's a good point, this would make a lot of noise for little real advantage to implementers. If we left out the API field change and reused `/deposit` and `/withdraw` for all three use cases we'd end up with two optional parameters:
 - source_asset (for deposit) / destination_asset (for withdrawal)
 - quote_id
Is that right?

I'm not wanting to push that idea. I'm wanting to understand the options.

marcelosalloum: Yes, that's right.

Although keeping only `/deposit` and `/withdraw` and make them use `source_asset` and/or `destination_asset` and/or `quote_id` takes us back to the argument @accordeiro made [previously in this thread](https://github.com/stellar/stellar-protocol/pull/973#discussion_r652073327) when he raised that "the conditional optionality of parameters makes APIs confusing".

Personally, I'd prefer to use a new endpoint for this new feature of converting non-equivalent assets, but remain open to discussions.

leighmcculloch: True, what do you think about combining the two sets of new exchange endpoints into one, so SEP-6 has two sets of APIs: one for simple deposits/withdrawal, and one for exchange deposit/withdrawal?

We can make all the fields required except quote id. Instead of making the RFQAPI responsible for providing the input data for this API for firm quotes, we provide it only as a side-car for linking to a previously locked in rate.

marcelosalloum: > We can make all the fields required except quote id. Instead of making the RFQAPI responsible for providing the input data for this API for firm quotes, we provide it only as a side-car for linking to a previously locked in rate.

Let me know if I understand your suggestions correctly:
* Leave the existing `GET /withdraw` and `GET /deposit` API untouched.
* Create new endpoints for exchanges, say `GET /withdraw-exchange` and `GET /deposit-exchange`
* For the new endpoints, `source_asset` and `destination_asset` are always mandatory and `quote_id` is optional.

leighmcculloch: Yup, that's what I meant. It keeps exchange APIs entirely separate, doesn't affect existing APIs, but reduces the surface area of the total API by 1/3. I also think it is easier to explain: instead of needing to explain 'indicative' and 'firm' names in endpoints the different types of conversion are simply explained as you're converting with or without a quote ahead of time.

accordeiro: @leighmcculloch I like this suggestion. It's similar to my `Option B` (2 different sets of endpoints vs. 3), but with better semantics. +1 :)

marcelosalloum: Updated the RFQ strategy in https://github.com/stellar/stellar-protocol/pull/973/commits/ccff6f496e84d8d8f6ab4c340121fdd35764b004.

ecosystem/sep-0006.md
```diff
@@ -33,11 +34,14 @@ To support this protocol an anchor acts as a server and implements the specified
 
 * An anchor must define the location of their `TRANSFER_SERVER` in their [`stellar.toml`](sep-0001.md). This is how a wallet knows where to find the anchor's server.
 * Anchors and clients may support [SEP-10](sep-0010.md) web authentication to enable authenticated deposits, withdrawals, or transaction history lookups.
+* If the Anchor supports [SEP-38], `ANCHOR_QUOTE_SERVER` must be defined in their [`stellar.toml`](sep-0001.md).
```
leighmcculloch, line 37: ❔ SEP-38 doesn't mention this stellar.toml field. Should we mention it there instead or as well?

ecosystem/sep-0006.md
```diff
@@ -33,11 +34,14 @@ To support this protocol an anchor acts as a server and implements the specified
...
 
 ## API Endpoints
 
 * [`GET /deposit`](#deposit): required
+* [`GET /deposit-exchange`](#deposit-exchange): optional
 * [`GET /withdraw`](#withdraw): required
+* [`GET /withdraw-exchange`](#withdraw-exchange): optional
```
leighmcculloch, line 44: 💡 This is non-critical, and is a suggestion that could easily be subjective. I think it is useful to group related items together. The deposit endpoints are related, but the fact the exchange endpoints are implemented together and optional and serve the same category of transaction type, I think their affiliation is stronger.
```suggestion
* [`GET /deposit`](#deposit): required
* [`GET /withdraw`](#withdraw): required
* [`GET /deposit-exchange`](#deposit-exchange): optional
* [`GET /withdraw-exchange`](#withdraw-exchange): optional
```

ecosystem/sep-0006.md
```diff
@@ -69,6 +73,58 @@ Access-Control-Allow-Origin: *
 
 This protocol involves the transfer of value, and so HTTPS is required for all endpoints for security.  Wallets and anchors should refuse to interact with any insecure HTTP endpoints.
 
+## Asset Exchanges
+
+This protocol was originally designed to provide conversions between off-chain and on-chain equivalent assets – for instance `BRL <> BRLT`, `USD <> USDC`, `NGN <> NGNT`, etc. – but it also describes how to make exchanges between off-chain and on-chain non-equivalent assets like `BRL <> USDC`.
+
+To enable the conversion of non-equivalent assets, the Anchor needs to implement the [SEP-38 Anchor RFQ API](sep-0038.md), which enables the Anchor to provide quotes to the Wallet, as well as the endpoints [`GET /deposit-exchange`](#deposit-exchange) and [`GET /withdraw-exchange`](#withdraw-exchange).
```
leighmcculloch, line 80: 🔡 
```suggestion
To enable the conversion of non-equivalent assets, the Anchor must implement endpoints [`GET /deposit-exchange`](#deposit-exchange) and [`GET /withdraw-exchange`](#withdraw-exchange), and the [SEP-38 Anchor RFQ API](sep-0038.md) that enables the Anchor to provide quotes to the Wallet, as well as .
```

ecosystem/sep-0006.md
```diff
@@ -69,6 +73,58 @@ Access-Control-Allow-Origin: *
...
+
+### Asset Conversions Using the Market Rate
+
+To convert at the market rate, clients first need to get an [indicative quote](https://www.investopedia.com/terms/i/indicativequote.asp) with [SEP-38] and then proceed with the conversion flow. With market rate conversions there won't be a firm guarantee of the final price to be paid for the asset. Here is how this kind of conversion should happen in deposits and withdrawals for an asset pair like `BRL <> USDC`:
```
leighmcculloch, line 84: 🔡 
```suggestion
To convert at the market rate, clients get an [indicative quote](https://www.investopedia.com/terms/i/indicativequote.asp) with [SEP-38] and then proceed to calling the [`GET /deposit-exchange`](#deposit-exchange) and [`GET /withdraw-exchange`](#withdraw-exchange) endpoints. Market rate conversions executed in this way do not provide a guarantee of the final price to be paid for the asset. Here is how this kind of conversion should happen in deposits and withdrawals for an asset pair like `BRL <> USDC`:
```

ecosystem/sep-0006.md
```diff
@@ -69,6 +73,58 @@ Access-Control-Allow-Origin: *
...
+1. The Wallet checks the assets' market conversion rate in [`SEP-38 GET /price`](sep-0038.md#get-price) and shows the estimated price to the user.
+1. The Wallet initiates a [`GET /deposit-exchange`](#deposit-exchange) using the parameters `source_asset` and `destination_asset`.
+1. The user proceeds with the off-chain deposit, sending `BRL` to the Anchor's bank account.
+1. Upon receiving the off-chain deposit funds, the Anchor converts it to `USDC` using the market rate and sends the resulting amount to the user's Stellar account.
+
+**Withdraw:**
+
+Considering the user will be making a Stellar payment of `USDC` in exchange for withdrawing `BRL` in the bank account:
+
+1. The Wallet checks the assets' conversion rate in [`SEP-38 GET /price`](sep-0038.md#get-price) and shows the estimated price to the user.
```
leighmcculloch, line 99: 🔡 
```suggestion
1. The Wallet checks the assets' conversion rate with [`SEP-38 GET /price`](sep-0038.md#get-price) and shows the estimated price to the user.
```

ecosystem/sep-0006.md
```diff
@@ -69,6 +73,58 @@ Access-Control-Allow-Origin: *
...
+1. The Wallet initiates a [`GET /withdraw-exchange`](#withdraw-exchange) using the parameters `source_asset` and `destination_asset`.
+1. The user proceeds with the on-chain payment, sending `USDC` to the Anchor's Stellar account.
+1. Upon receiving the on-chain withdraw funds, the Anchor converts it to `BRL` using the market rate and sends the resulting amount to the user's bank account.
+
+### Asset Conversions Using a Firm Quote
+
+With firm quotes, users can have a guarantee of the asset price they are going to pay **if they submit the payment before the quote expiration**. Here is how the conversion should happen in deposits and withdrawals for an asset pair like `USDC <> BRL`:
```
leighmcculloch, line 106: 🔡 Also, avoid emphasising text in SEPs that aren't headings. All the text in SEPs is important.
```suggestion
With firm quotes, users are guaranteed an asset price they will pay if they submit the payment before the quote expiration. Here is how the conversion should happen in deposits and withdrawals for an asset pair like `USDC <> BRL`:
```

ecosystem/sep-0006.md
```diff
@@ -589,29 +793,33 @@ The response should be a JSON object like:
 }
 ```
 
-The JSON object contains an entry for each Stellar asset that the anchor supports for SEP6 deposit and/or withdrawal.
+The JSON object contains an entry for each **Stellar asset** that the anchor supports for SEP-6 deposit and/or withdrawal.
```
leighmcculloch, line 796: ❔  What's the motivation for bolding some terms throughout the doc? I think where we have seen emphasis work really well is to emphasis entities only, for example the Server and Client. I think when we start emphasizing other things it becomes difficult to read the doc. 

Thoughts @marcelosalloum @JakeUrban ?

ecosystem/sep-0006.md
```diff
@@ -589,29 +793,33 @@ The response should be a JSON object like:
...
+_Note: the attributes above mentioned are valid for both kinds of withdrawals `withdraw` and `withdraw-exchange`, exept for the `fee_*` attributes that are only valid for `withdraw`._
+
+The `fields` object allows an anchor to describe fields that are passed into `/deposit*` and `/withdraw*`. It can explain standard fields like `dest` and `dest_extra` for withdrawal, and it can also specify extra fields that should be passed into `/deposit*` or `/withdraw*` such as an email address or bank name. If a field is part of the KYC/AML flow handled by SEP-12 or the field is handled by your interactive deposit/withdrawal flow, there's no need to list it in `/info`. Only fields that are passed to `/deposit*` or `/withdraw*` need appear here.
```
leighmcculloch, line 822: ❓ What if we need to add new fields to exchange and not the other, will they be included in the shared list? Could this get confusing?

ecosystem/sep-0006.md
```diff
@@ -628,7 +836,9 @@ An anchor should also indicate in the `/info` response if they support the `fee`
 
 ## Fee
 
-The fee endpoint allows an anchor to report the fee that would be charged for a given deposit or withdraw operation. This is important to allow an anchor to accurately report fees to a user even when the fee schedule is complex. If a fee can be fully expressed with the `fee_fixed` and `fee_percent` fields in the `/info` response, then an anchor should not implement this endpoint.
+The fee endpoint allows an anchor to report the fee that would be charged for a given `/deposit` or `/withdraw` operation between equivalent assets. For non-equivalent assets the fees are included in the conversion price provided with [`SEP-38 GET /price`](sep-0038.md#get-price) or [`SEP-38 POST /quote`](sep-0038.md#post-quote).
```
leighmcculloch, line 839: 💡 I think this paragraph needs breaking up. Also, the paragraph talks about equivalent and non-equivalent assets and how fees apply to them, but I think we can make this easier to understand by saying which endpoints fees apply.

```suggestion
The fee endpoint allows an anchor to report the fee that would be charged when using the `/deposit` or `/withdraw` endpoints.

When using the `/deposit-exchange` or `/withdraw-exchange` endpoints fees are included in the conversion price provided with [`SEP-38 GET /price`](sep-0038.md#get-price) or [`SEP-38 POST /quote`](sep-0038.md#post-quote).
```

leighmcculloch (approved):
Overall this looks great. I left some suggestions (💡🔡) and questions (❓❔💭) throughout. I defer to @JakeUrban.

ecosystem/sep-0006.md
marcelosalloum: Yep, that makes sense. I emphasized some parts of the text that I considered key concepts, but this is probably just a bias I have when taking notes to myself. I'll go easier on the habit of emphasizing things.

ecosystem/sep-0006.md
marcelosalloum: Sorry, I didn't understand your point here.

ecosystem/sep-0006.md
marcelosalloum: That's a good point! It's currently being covered only in [SEP-1](https://github.com/stellar/stellar-protocol/blame/926cf0dfb9d4b8470efc7b354e2a5865892210c4/ecosystem/sep-0001.md#L64). I will add that to SEP-38 in another PR.

ecosystem/sep-0006.md
marcelosalloum: https://github.com/stellar/stellar-protocol/pull/973/commits/7283bfcc3c38b3d24185becc32543a4dc83ca4de

ecosystem/sep-0006.md
marcelosalloum: https://github.com/stellar/stellar-protocol/pull/973/commits/7283bfcc3c38b3d24185becc32543a4dc83ca4de

ecosystem/sep-0006.md
marcelosalloum: https://github.com/stellar/stellar-protocol/pull/973/commits/7283bfcc3c38b3d24185becc32543a4dc83ca4de

ecosystem/sep-0006.md
marcelosalloum: https://github.com/stellar/stellar-protocol/pull/973/commits/7283bfcc3c38b3d24185becc32543a4dc83ca4de

ecosystem/sep-0006.md
marcelosalloum: https://github.com/stellar/stellar-protocol/pull/973/commits/7283bfcc3c38b3d24185becc32543a4dc83ca4de

ecosystem/sep-0006.md
marcelosalloum: https://github.com/stellar/stellar-protocol/pull/973/commits/b6b2a59a097c557bdf52a68c0cde5711be5da54c

ecosystem/sep-0006.md
marcelosalloum: I added it to SEP-38 in https://github.com/stellar/stellar-protocol/pull/989/commits/bad6e317a04d2f51092a87213508814bf96f9a85.

I think it's ok to keep it in SEP-6 as well.

ecosystem/sep-0006.md
marcelosalloum: Removed a few uses of `**` in https://github.com/stellar/stellar-protocol/pull/973/commits/70f1a688bf23566df14f418e4c621c0fa2fa95f9.

ecosystem/sep-0006.md
leighmcculloch: The section notes that the `fields` object describes fields that are passed into all four endpoints (/deposit, /deposit-exchange, etc). What happens if we want to add a field only to the exchange version?

marcelosalloum: Good point! How does this look? https://github.com/stellar/stellar-protocol/pull/973/commits/04234d98cff2d4a13615ea10ddb49aee45dbb18e

leighmcculloch: 04234d9 looks like it changes fields that are not part of the `fields` object.  I think my comment was misunderstood. I was referring to this section about the `fields` object:
>The `fields` object allows an anchor to describe fields that are passed into `/deposit*` and `/withdraw*`. It can explain standard fields like `dest` and `dest_extra` for withdrawal, and it can also specify extra fields that should be passed into `/deposit*` or `/withdraw*` such as an email address or bank name.

marcelosalloum: Ok got it, Thanks for explaining again.

Yes, the idea here is `fields` can contain different values for any of the following: `/deposit`, `/withdraw`, `/deposit-exchange` and `/withdraw-exchange`.

I thought using the `*` in `/deposit*` to represent both `/deposit` and `/deposit-exchange` would be enough to make it clear... what do you think of this copy https://github.com/stellar/stellar-protocol/pull/973/commits/d5d591481e4802587229bd6749b5c40202ceef06?

leighmcculloch: That does read clearer by being verbose, but that wasn't actually what I had an issue with. My question was, "What happens if we want to add a field only to the exchange version?"

If we want to add a field X that is only to be used with "/deposit-exchange" and "/withdraw-exchange", and if I add field X to `fields`, will the field also automatically apply to "/deposit" and "/withdraw"?

I ask this because we went with separate endpoints to remove optionality and overlap, to separate them, but in this `fields` definition we are joining them back together.

marcelosalloum: If you look at the `GET \info` [response](https://github.com/stellar/stellar-protocol/blob/04234d98cff2d4a13615ea10ddb49aee45dbb18e/ecosystem/sep-0006.md#response-4) you'll notice there is an occurrence of `fields` for every one of those endpoints:

```js
{
  "deposit": {
    "USD": {
      "fields": {...}
    },
  },
  "deposit-exchange": {
    "USD": {
      "authentication_required": true,
      "fields": {...}
    }
  },
  "withdraw": {
    "USD": {
      "types": {
        "bank_account": {
          "fields": {...}
        }
      }
    },
  },
  "withdraw-exchange": {
    "USD": {
      "types": {
        "bank_account": {
          "fields": {...}
        }
      }
    }
  },
}
```

That should be enough to allow every one of those endpoints to have different `fields`.

leighmcculloch: Oh right, I misunderstood how `fields` works. 🎉 Thanks for explaining.

leighmcculloch (dismissed):
👏🏻 This looks great. Thanks for answering my questions. I defer to @JakeUrban for final approval.

ecosystem/sep-0006.md
```diff
@@ -589,29 +793,39 @@ The response should be a JSON object like:
...
+
+#### Response attributes for deposit assets
 
-#### For each deposit asset, response contains:
+All assets listed in a `deposit` and `deposit-exchange` can contain these attributes:
 
 * `enabled`: `true` if SEP-6 deposit for this asset is supported
 * `authentication_required`: Optional. `true` if client must be [authenticated](#authentication) before accessing the deposit endpoint for this asset. `false` if not specified.
 * `min_amount`: Optional minimum amount. No limit if not specified.
 * `max_amount`: Optional maximum amount. No limit if not specified.
```
JakeUrban, line 805: These attributes shouldn't be valid in `deposit-exchange` objects because the asset under which they're provided is the on-chain asset, and the user provides the off-chain asset.

ecosystem/sep-0006.md
```diff
@@ -688,7 +906,7 @@ Name | Type | Description
 `account` | string | The stellar account ID involved in the transactions.
 `no_older_than` | UTC ISO 8601 string | (optional) The response should contain transactions starting on or after this date & time.
 `limit` | int | (optional) The response should contain at most `limit` transactions.
-`kind` | string | (optional) The kind of transaction that is desired. Should be either `deposit` or `withdrawal`.
+`kind` | string | (optional) The kind of transaction that is desired. Should be either `deposit`, `deposit-exchange`, `withdrawal` or `withdrawal-exchange`.
```
JakeUrban, line 909: This has me wondering -- what if a client wants both types of deposits? Are we just going to require they make 2 API calls?

JakeUrban (changes requested):
Left some comments, but we're almost there. I'm pretty happy with where we've ended up, thanks for taking this to the finish line while I was gone.

ecosystem/sep-0006.md
marcelosalloum: Good point, that wasn't a problem when there were only two kinds...

If we want to allow subsets we could either add a new field called `kinds` or update the `kind` description to accept a comma-separated list, like "deposit,deposit-exchange" for instance. I'm leaning towards the comma-separated option, WDYT?

ecosystem/sep-0006.md
```diff
@@ -375,6 +435,99 @@ Example:
...
+`source_asset` | string | The off-chain asset the Anchor will receive from the user. The value must match one of the `asset` values included in a [`SEP-38 GET /prices?buy_asset=stellar:<destination_asset>:<asset_issuer>`](sep-0038.md#get-prices) response using [SEP-38 Asset Identification Format](sep-0038.md#asset-identification-format).
+`quote_id` | string | (optional) The `id` returned from a `SEP-38 POST /quote` response. If this parameter is provided and the user delivers the deposit funds to the Anchor before the quote expiration, the Anchor should respect the conversion rate agreed in that quote. Otherwise, the conversion will be performed at the market rate.
+`amount` | string | The amount of the source_asset the user would like to deposit to the anchor's off-chain account. This field may be necessary for the anchor to determine what KYC information is necessary to collect.
+`account` | `G...` string | The stellar account ID of the user that wants to deposit. This is where the asset token will be sent.
```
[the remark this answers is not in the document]
marcelosalloum: This field doesn't show only in /deposit-exchange but in /deposit as well. Should I remove both?

ecosystem/sep-0006.md
```diff
@@ -115,7 +171,9 @@ SEP-6 lays out many options for how deposit and withdrawal can work. These are r
 
 ## Deposit
 
-A deposit is when a user sends an external token (BTC via Bitcoin, USD via bank transfer, etc...) to an address held by an anchor. In turn, the anchor sends an equal amount of an equivalent token on the Stellar network (minus fees) to the user's Stellar account.
+A deposit is when a user sends an external token (BTC via Bitcoin, USD via bank transfer, etc...) to an address held by an anchor. In return, the anchor sends an equal amount of an equivalent token on the Stellar network (minus fees) to the user's Stellar account.
+
+If the anchor supports [SEP-38] quotes, it can also provide a bridge between non-equivalent tokens by receiving, for instance ARS via bank transfer and in return sending the equivalent value (minus fees) as USDC on the Stellar network to the user's Stellar account. That kind of deposit is covered in [`GET /deposit-exchange`](#deposit-exchange).
```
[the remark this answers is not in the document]
marcelosalloum: https://github.com/stellar/stellar-protocol/pull/973/commits/9366513e322865c093b5b012820fad495116d31d

ecosystem/sep-0006.md
```diff
@@ -303,7 +361,9 @@ Predicates are one of the [claimable balance parameters](https://developers.stel
 
 ## Withdraw
 
-A withdraw is when a user redeems an asset currently on the Stellar network for it's equivalent off-chain asset via the Anchor. For instance, a user redeeming their NGNT in exchange for fiat NGN.
+A withdraw is when a user redeems an asset currently on the Stellar network for its equivalent off-chain asset via the Anchor. For instance, a user redeeming their NGNT in exchange for fiat NGN.
+
+If the anchor supports [SEP-38] quotes, it can also provide a bridge between non-equivalent tokens by receiving, for instance USDC from the Stellar network and in return sending the equivalent value (minus fees) as NGN to the user's bank account. That kind of withdrawal is covered in [`GET /withdraw-exchange`](#withdraw-exchange).
```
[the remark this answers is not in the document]
marcelosalloum: https://github.com/stellar/stellar-protocol/pull/973/commits/9366513e322865c093b5b012820fad495116d31d

ecosystem/sep-0006.md
marcelosalloum: Should we update the SEP-38 description to enforce min_amount and max_amount when providing quotes?

ecosystem/sep-0006.md
```diff
@@ -375,6 +435,99 @@ Example:
...
+
+Name | Type | Description
+-----|------|------------
+`destination_asset` | string | The code of the on-chain asset the user wants to get from the Anchor after doing an off-chain deposit. The value passed must match one of the codes listed in the [/info](#info) response's `deposit-exchange` object.
+`source_asset` | string | The off-chain asset the Anchor will receive from the user. The value must match one of the `asset` values included in a [`SEP-38 GET /prices?buy_asset=stellar:<destination_asset>:<asset_issuer>`](sep-0038.md#get-prices) response using [SEP-38 Asset Identification Format](sep-0038.md#asset-identification-format).
+`quote_id` | string | (optional) The `id` returned from a `SEP-38 POST /quote` response. If this parameter is provided and the user delivers the deposit funds to the Anchor before the quote expiration, the Anchor should respect the conversion rate agreed in that quote. Otherwise, the conversion will be performed at the market rate.
+`amount` | string | The amount of the source_asset the user would like to deposit to the anchor's off-chain account. This field may be necessary for the anchor to determine what KYC information is necessary to collect.
```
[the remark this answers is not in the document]
marcelosalloum: > We should add a sentence to the quote_id definition stating that requests should be rejected if these parameters don't match the values used when creating the quote.

Cool, that makes sense to me.

> My preference would be to make (`destination_asset`, `source_asset`, `amount` ) and ( `quote_id` ) mutually exclusive in requests, where the anchor returns 400 if thats not the case.

We took that design choice based on @leighmcculloch suggestion [here](https://github.com/stellar/stellar-protocol/pull/973#discussion_r659977358). Mutually exclusive fields don't sound too bad in this context but I still think the API usage is simpler without it.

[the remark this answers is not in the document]
marcelosalloum: https://github.com/stellar/stellar-protocol/pull/973/commits/6aa404fd6e4cdaab33d95bfd76557ceaa408ede8

ecosystem/sep-0006.md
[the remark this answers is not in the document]
JakeUrban: Lets keep it in `/deposit` since it's a breaking change. In another PR is may make sense to deprecate it.

ecosystem/sep-0006.md
JakeUrban: The anchor is free to return 400 Bad Request for `POST /quote` or `GET /price` requests with too small of an amount as the SEP is now, but it may be helpful to add `min_amount`/`max_amount` to either `GET /info` or `GET /prices` for the client's convenience. I don't feel strongly either way.

ecosystem/sep-0006.md
leighmcculloch: You could allow multiple `kind` params, i.e.g `/transactions?kind=deposit&kind=deposit-exchange`, which is more likely to be backwards compatible with existing services. This would be my preference for this reason, and because you're leveraging URL queries natural expression of arrays without introducing comma-separated lists or new params.

ecosystem/sep-0006.md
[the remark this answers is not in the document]
marcelosalloum: https://github.com/stellar/stellar-protocol/pull/973/commits/4688fefc40e6de6309197755b007e24bf5e295a3

ecosystem/sep-0006.md
marcelosalloum: Applied Leigh's suggestion in https://github.com/stellar/stellar-protocol/pull/973/commits/31e833b0bd2a2726086c1a03ef2ac14ee9b7dd9f.

ecosystem/sep-0006.md
```diff
@@ -69,6 +73,58 @@ Access-Control-Allow-Origin: *
...
+
+#### Deposit:
+
+Considering the user will be making a bank deposit of fiat `BRL` in exchange for `USDC` in the Stellar network:
+
+1. The Wallet gets a firm conversion rate in [`SEP-38 POST /quote`](sep-0038.md#post-quote) and shows the price and expiration date to the user.
+1. The Wallet initiates a [`GET /deposit-exchange`](#deposit-exchange) using the parameter `quote_id` returned from the previous step, as well as `asset_code` and `destination_asset`.
+1. The user proceeds with the off-chain deposit, sending `BRL` to the Anchor's bank account before the quote expires.
+1. Upon receiving the off-chain deposit funds, the Anchor converts it to `USDC` according with the agreed quote and sends the resulting amount to the user's Stellar account.
+1. If the user sends the funds after the quote expiration, the Anchor will perform the exchange using the market rate.
```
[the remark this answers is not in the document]
JakeUrban: For now, lets state that the anchor can define their own policy on this topic. Once we get more feedback from anchors and if a consensus emerges, we can standardize it.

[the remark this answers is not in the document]
marcelosalloum: Done it in https://github.com/stellar/stellar-protocol/pull/973/commits/4817965fc7ebf82f6d5538aa0da22129d373e3d3.

JakeUrban (approved):

JakeUrban (approved):

Issue #915, opened 2021-04-25
SEP-0009: Add sex field


ymatienzo:
### What problem does your feature solve?

Sometimes anchors need to collect user gender information as part of their KYC requirements. 

### What would you like to see?

1. Add gender as an additional field for the individual
2. Gender could have the following values [male, female, undefined]

leighmcculloch:
Thanks for opening this issue @ymatienzo. Could you provide some examples of the jurisdiction (e.g. country) where gender is required for KYC and how it influences KYC?

A quick internet search turned up this [this document](https://www.pwc.com/gx/en/financial-services/publications/assets/pwc-anti-money-laundering-2016.pdf) dated 2016 that indicates it is a KYC requirement in these countries:
- South Korea
- Mexico
- Isle of Man
- Russia
- Israel

I'm wondering whether for KYC the importance in these countries is biological or behavioral/cultural, and if the former, if it would be more appropriate to use the term `sex`. A reference for more details on `sex` vs `gender` is [here](https://www.merriam-webster.com/dictionary/gender#usage-1). Passports and other identity documentation use the term `sex`. Given the data collected in KYC is verified through identity documents such as passports it probably makes most sense to mirror those and stick to the biological focus. i.e. If a field is added, have the field be `sex`.

cc @lydiat @JakeUrban Thoughts?

ymatienzo:
> this document

JakeUrban:
Sounds like a reasonable addition to the standard, thanks for filing @ymatienzo.

PR #977, opened 2021-06-17
SEP-9: add 'sex' field

JakeUrban:
resolves stellar/stellar-protocol#915 

Uses `other` instead of the suggested `undefined` since `undefined` could be interpreted to mean `null` or simply omitted.

@ymatienzo

### DIFF
--- a/ecosystem/sep-0009.md
+++ b/ecosystem/sep-0009.md
@@ -81,6 +81,9 @@ Name | Type | Description
 `notary_approval_of_photo_id` | binary | Image of notary's approval of photo ID or passport
 `ip_address` | string | IP address of customer's computer
 `photo_proof_residence` | binary | Image of a utility bill, bank statement or similar with the user's name and address
+`sex` | string | `male`, `female`, or `other`
+
+```
 
 ### Organization KYC fields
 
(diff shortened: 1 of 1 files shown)


rice2000:
I'd suggest using `gender` instead of `sex`.  What do you think?

JakeUrban:
Leigh [pointed out](https://github.com/stellar/stellar-protocol/issues/915#issuecomment-827055684) that `sex` is the term used by official documents. `gender` is also used more often in a social / psychological context whereas `sex` is generally understood to refer to someone's biology.

I don't feel strongly about this though.

ecosystem/sep-0009.md
```diff
@@ -81,6 +81,7 @@ Name | Type | Description
 `notary_approval_of_photo_id` | binary | Image of notary's approval of photo ID or passport
 `ip_address` | string | IP address of customer's computer
 `photo_proof_residence` | binary | Image of a utility bill, bank statement or similar with the user's name and address
+`sex` | string | `male`, `female`, or `other`
```
leighmcculloch, line 84: I think the values included here make sense for now, given the field is `sex` and not `gender`. If the field was `gender` as @rice2000 suggested we would probably include some other values. There's some previous discussion on this at https://github.com/stellar/stellar-protocol/issues/915#issuecomment-827055684.

The primary use case of this field is to capture data that is verifiable on official documentation such as passports or other forms of photo identification. According to wikipedia there are countries that support three markers on passports: other (X), female (F), male (M), so the values in this PR appear to align well with what we can expect on passports.

We should add other values if we discover there are forms of identification that support additional values to keep this inclusive of all countries and their KYC requirements.

leighmcculloch (commented):
I think this looks good. I left a comment below about sex vs gender and the values being included.

Can you update the meta data?

leighmcculloch (approved):

PR #982, opened 2021-06-22
SEP-10: Add requirement that server should return only one JWT token for a challenge

leighmcculloch:
### What
Add a requirement to SEP-10 that the server should return only one JWT token for a given challenge, without prescribing exactly how that is achieved.

### Why
It is a best practice to not allow multiple authentication tokens be collected through a handshake process. While this SEP does not attempt to mention all best practices relating to JWTs and other things, I think it could be worth mentioning this one.

Realistically there is probably little someone could do with an average SEP-10 server, but it may be confusing if there are multiple JWTs that were all related to one challenge.

The change doesn't prescribe how to implement this because I don't think it is particularly important, at least not yet to force a particular implementation on an implementation. A server may implement this by making the JWT deterministic, such that subsequent requests for a token with a valid not-expired challenge return the exact same JWT. Or, a server may implement this by caching or storing challenges for which tokens have been issued and refusing to supply a token more than once. There are trade-offs with different approaches. The former doesn't prevent recollection but scales well, the latter requires storage and is harder to scale and operate.

The change is backwards compatible for any sensible client.

### DIFF
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -7,7 +7,7 @@ Author: Sergey Nebolsin <@nebolsin>, Tom Quisel <tom.quisel@gmail.com>, Leigh Mc
 Status: Active
 Created: 2018-07-31
 Updated: 2021-03-04
-Version 3.2.0
+Version 3.3.0
 ```
 
 ## Simple Summary
@@ -199,6 +199,8 @@ Upon successful verification, **Server** responds with a session JWT, containing
 * `client_domain` - (optional) a nonstandard JWT claim containing the client home domain, included if the challenge transaction contained a `client_domain` (see [Verifying Client Application Identity](#verifying-client-application-identity))
 
 The JWT may contain other claims specific to your application, see [RFC7519].
+
+The **Server** should provide only a single session JWT for a specific challenge transaction.
 
 [Uniform Resource Identifier (URI)]: https://en.wikipedia.org/wiki/Uniform_Resource_Identifier
 [RFC7519]: https://tools.ietf.org/html/rfc7519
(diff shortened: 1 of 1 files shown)


JakeUrban (approved):
The reference server implements this by using the challenge envelope hex representation as the the [jti](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.7) claim. It may be helpful to point out that putting any data that is unique per JWT will ensure the JWT is unique.

leighmcculloch:
> It may be helpful to point out that putting any data that is unique per JWT will ensure the JWT is unique.

@JakeUrban Per the RFC you linked to, it is also important that the entire data object included in the JWT is consistent, not just that there is one item to make it unique.

>The identifier value MUST be assigned in a manner that ensures that there is a negligible probability that the same value will be accidentally assigned to a different data object

Ref: https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.7

accordeiro (dismissed):
Looks good to me, thanks for addressing my (offline) feedback :)

accordeiro (approved):

JakeUrban (commented):
This seems to imply that the server should return 400 bad request when the client requests a token challenge that was used in a prior post request. Don’t we want to just have the same token returned for the same challenge?

leighmcculloch:
> This seems to imply that the server should return 400 bad request when the client requests a token challenge that was used in a prior post request. Don’t we want to just have the same token returned for the same challenge?

I think that is one interpretation, and either is sufficient. Clients should only ever need to query once. If they query twice we just want to signal in the protocol that a server shouldn't generate them more than one JWT, so if they're able to serve up the same one that is probably fine (although that's up to the implementer and their risk tolerance), or if they're able to decline subsequence requests that is probably fine too.

JakeUrban (approved):

Issue #981, opened 2021-06-22
SEP-12: Clarify `fields` and `field.status` ambiguity 


JakeUrban:
related to stellar/stellar-demo-wallet#189

This issue is for discussion what steps to take in order to clarify the issues mentioned below.

The changes made for the [property verification](https://github.com/stellar/stellar-protocol/commit/2d46fe87fa9aa512610078eda51dc52573ed754b) functionality introduces some variability in the `GET /customer` response, specifically related to the presence of the `fields` and field-level `status` properties.

The commit clarified that `fields` may be present for all customer statuses, not just `NEEDS_INFO`. However, there is no reason for `fields` to be present for other customer statuses unless the field-level `status` properties are present.

Field-level `status` properties are optional, but its not clear what the client should assume if the property is omitted. Most importantly, **clients used to be able to assume all fields present needed to be collected, but that is no longer the case if field-level status values are not `NOT_PROVIDED`.**

That last point is a breaking change that we missed when making the change.

cc @leighmcculloch

leighmcculloch:
Interesting. We made a breaking change, is that a bug? If so we can probably change fix that bug with a patch change. Do we know if anyone is dependent on that buggy logic?

I don't feel strongly for how we resolve the fields presence though. I can see arguments for both approaches. It could be confusing if someone can't see the final status for a field when the status changes to needs info to something else. Especially if the user is rejected and the server wishes to signal which field was rejected and related to the overall rejection.

JakeUrban:
The change was made 2 months ago, and I'm not aware of any service who has implemented SEP-12 in that timeframe but its possible someone has.

The only way I can think of that ensures existing SEP-12 clients consume the response properly is by maintaining the rule that `fields` in a `NEEDS_INFO` response can only be for fields the anchor has not yet received. But this is directly at odds with what we're trying to achieve: letting the anchor signal that verification is required for a field that has already been provided. I also think its confusing to limit the fields listed in a `NEEDS_INFO` response but allow all fields to be listed for customers in other statuses.

leighmcculloch:
> this is directly at odds with what we're trying to achieve: letting the anchor signal that verification is required for a field that has already been provided

🤔 We could create a new `fields` that contains the new behavior and keep the old `fields` as is. Wdyt?

JakeUrban:
[...]

PR #985, opened 2021-06-24
SEP-12: Clarify existing vs. provided fields

JakeUrban:
resolves stellar/stellar-protocol#981

Changes:
- Clarifies `fields` to be an object containing SEP-9 fields that have not yet been collected by the anchor
- Adds `provided_fields` for SEP-9 fields that have been collected
- Moves field `status` and `error` properties from `fields` to `provided_fields`
- Removes `NOT_PROVIDED` as a value for field `status`

Open questions:

- Do we want to add a new `VERIFICATION_REQUIRED` customer-level status to better indicate what the anchor is waiting on?

`NEEDS_INFO` is primarily for indicating that the customer must provide more information detailed in `fields`. However, anchors that require verification of provided fields are instructed to use `NEEDS_INFO` as well. If all fields have been provided it may be confusing for the client to still see `NEEDS_INFO`. 

- What should the customer `status` be when multiple statuses apply to the customer's state? 

For example, an anchor could be `PROCESSING` a provided field but still need others (`NEEDS_INFO`) or needs a verification of another provided field `VERIFICATION_REQUIRED`.

### DIFF
--- a/ecosystem/sep-0012.md
+++ b/ecosystem/sep-0012.md
@@ -117,7 +117,8 @@ Name | Type | Description
 -----|------|------------
 `id` | string | (optional) ID of the customer, if the customer has already been created via a `PUT /customer` request.
 `status` | string | Status of the customers KYC process.
-`fields` | object | (optional) An object containing the fields for the given KYC type. See [fields](#fields) for more detailed information.
+`fields` | object | (optional) An object containing the fields the anchor has not yet received for the given customer of the `type` provided in the request. Required for customers in the `NEEDS_INFO` status. See [Fields](#fields) for more detailed information.
+`provided_fields` | object | (optional) An object containing the fields the anchor has received for the given customer. See [Provided Fields](#provided-fields) for more detailed information. Required for customers whose information needs verification via [`PUT /customer/verification`](#customer-put-verification).
 `message` | string | (optional) Human readable message describing the current state of customer's KYC process.
 
 ```js
@@ -125,7 +126,7 @@ Name | Type | Description
 {
    "id": "d1ce2f48-3ff1-495d-9240-7a50d806cfed",
    "status": "ACCEPTED",
-   "fields": {
+   "provided_fields": {
       "first_name": {
          "description": "The customer's first name",
          "type": "string",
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0012.md
```diff
@@ -6,8 +6,8 @@ Title: KYC API
 Author: Interstellar
 Status: Active
 Created: 2018-09-11
-Updated: 2021-05-28
-Version 1.6.3
+Updated: 2021-06-24
+Version 1.7.0
```
leighmcculloch, line 10: 💭 I'm not convinced this should be a minor version, but instead a patch version. This is fixing a bug we introduced, albeit a big bug. It's a bug that doesn't introduce new functionality. I'm fine with minor if you prefer since this is definitely a gray area.

ecosystem/sep-0012.md
```diff
@@ -228,11 +285,9 @@ Status | Description
 -------|------------
 ACCEPTED | The field has been validated. When all required fields are accepted, the [Customer Status](#customer-statuses) should also be accepted.
 PROCESSING | The field is being validated. The client can make [GET /customer](#customer-get) requests to check on the result of this validation in the future.
-NOT_PROVIDED | The field has not been sent by the client. The field is required unless the `optional` property is `true`. The [Customer Status](#customer-statuses) should be `NEEDS_INFO` if a field is required but has not been provided.
```
leighmcculloch, line 231: 💭 I wonder if it is worth trying to keep `NOT_PROVIDED`, and returning `provided_fields` even when they haven't been provided yet. That would make this change less breaking, although would probably require renaming `provided_fields` to `all_fields` or something more generic.

leighmcculloch (approved):
I agree that this is the best path forward to address this problem. This change is probably worthy of eyes from a second person, such as @tomerweller or @accordeiro.

ecosystem/sep-0012.md
JakeUrban: Since we're moving the field-level `status` field to a new object I don't think keeping `NOT_PROVIDED` makes this changes less breaking personally.

JakeUrban:
@leighmcculloch thanks for the quick review. I'm curious on your thoughts on the first question posed in the PR description though.

ecosystem/sep-0012.md
leighmcculloch: That's a good point. 👍🏻

tomerweller (approved):
I think this will work.

PR #989, opened 2021-06-29
SEP-38: distinguish sell/buy delivery methods for fee calculation

marcelosalloum:
### What

1. Replace `delivery_methods` with `buy_delivery_methods` and `sell_delivery_methods`.
2. Update request parameters of `GET /price`, `GET /prices` and `POST /quote` endpoints so the fee can be accurately calculated with the new parameters provided:
    * `sell_delivery_method`
    * `buy_delivery_method`
    * `country_code`
2. Add `country_codes` parameter to the `GET /info` endpoint
4. Fix request example for `GET /quote`.

### DIFF
--- a/ecosystem/sep-0038.md
+++ b/ecosystem/sep-0038.md
@@ -176,13 +176,18 @@ This endpoint can be used to fetch the [indicative](https://www.investopedia.com
 
 These prices are indicative. The actual price will be calculated at conversion time once the Anchor receives the funds from a User.
 
-The prices returned include any margin that the provider may keep as a service fee, and this margin may vary depending on the directional flow of funds, or other factors.
+The prices returned include any margin that the provider may keep as a service fee, and this margin may vary depending on the directional flow of funds, amount, delivery method, country code, or other factors.
 
 #### Request
 
 Name | Type | Description
 -----|------|------------
 `sell_asset` | string | The asset you want to sell, using the [Asset Identification Format](#asset-identification-format).
+`sell_amount` | string | The amount of `sell_asset` the client would exchange for each of the `buy_assets`.
+`sep` | string | The name of the SEP that will be executing the quoted exchange. As of now, the ones that can make use of RFQ are `SEP-6` or `SEP-31`.
+`delivery_method` | string | (optional) One of the `name` values specified by the `delivery_methods` array for the associated asset returned from [`GET /info`](#info). If the array is not provided, this parameter is not required.
+`country_code` | string | (optional) The [ISO 3166-1 alpha-3](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-3) code of the user's current address. Should be provided if there are two or more country codes available for the desired asset in [`GET /info`](#get-info).
+
 
 #### Response
 
@@ -202,7 +207,7 @@ Name | Type | Description
 
 
 ```
-GET /prices?sell_asset=stellar:USDC:GA5ZSEJYB37JRC5AVCIA5MOP4RHTM335X2KGX3IHOJAPP5RE34K4KZVN
+GET /prices?sell_asset=stellar:USDC:GA5ZSEJYB37JRC5AVCIA5MOP4RHTM335X2KGX3IHOJAPP5RE34K4KZVN&sep=SEP-6&sell_amount=100
 ```
 
 ```json
@@ -218,7 +223,7 @@ GET /prices?sell_asset=stellar:USDC:GA5ZSEJYB37JRC5AVCIA5MOP4RHTM335X2KGX3IHOJAP
 ```
 ---
 ```
-GET /prices?sell_asset=iso4217:BRL
+GET /prices?sell_asset=iso4217:BRL&sep=SEP-6&sell_amount=500&country_code=BRA&delivery_method=PIX
 ```
 
 ```json
--- a/ecosystem/sep-0038.md
+++ b/ecosystem/sep-0038.md
@@ -355,7 +355,7 @@ Name | Type | Description
 `expire_after` | [UTC ISO 8601](https://www.iso.org/iso-8601-date-and-time-format.html) string | (optional) The client's desired `expires_at` date and time for the quote. Anchors should return `400 Bad Request` if the an expiration on or after the requested value cannot be provided.
 `sep` | string | The name of the SEP that will be executing the quoted exchange. As of now, the ones that can make use of RFQ are `SEP-6` or `SEP-31`.
 `delivery_method` | string | (optional) One of the `name` values specified by the `delivery_methods` array for the associated asset returned from [`GET /info`](#info). If the array is not provided, this parameter is not required.
-`country_code` | string | (optional) The [ISO 3166-1 alpha-3](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-3) code of the user's current address.
+`country_code` | string | (optional) The [ISO 3166-1 alpha-3](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-3) code of the user's current address. Should be provided if there are two or more country codes available for the desired asset in [`GET /info`](#get-info).
 
 #### Response
 
--- a/ecosystem/sep-0038.md
+++ b/ecosystem/sep-0038.md
@@ -24,6 +24,10 @@ Removing this requirement for anchors also provides downstream benefits to ecosy
 
 ## Specification
 
+## stellar.toml
+
+An anchor must define the location of their `ANCHOR_QUOTE_SERVER` in their [`stellar.toml`](sep-0001.md). This is how a wallet knows where to find the anchor's server.
+
 ### Authentication
 
 All endpoints require authentication in the form of a [SEP-10](sep-0010.md) JSON Web Token (JWT) in the `Authorization` header of the request. 
--- a/ecosystem/sep-0038.md
+++ b/ecosystem/sep-0038.md
@@ -7,9 +7,9 @@ Author: Jake Urban <@jakeurban> and Leigh McCulloch <@leighmcculloch>
 Track: Standard
 Status: Draft
 Created: 2021-04-09
-Updated: 2021-06-28
+Updated: 2021-06-30
 Discussion: https://github.com/stellar/stellar-protocol/issues/901
-Version 1.2.0
+Version 1.3.0
 ```
 
 ## Summary
--- a/ecosystem/sep-0038.md
+++ b/ecosystem/sep-0038.md
@@ -124,10 +124,11 @@ Name | Type | Description
 Name | Type | Description
 -----|------|------------
 `asset` | string | The [Asset Identification Format](#asset-identification-format) value.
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0038.md
```diff
@@ -174,13 +176,18 @@ This endpoint can be used to fetch the [indicative](https://www.investopedia.com
...
-The prices returned include any margin that the provider may keep as a service fee, and this margin may vary depending on the directional flow of funds, or other factors.
+The prices returned include any margin that the provider may keep as a service fee, and this margin may vary depending on the directional flow of funds, amount, delivery method, country code, or other factors.
 
 #### Request
 
 Name | Type | Description
 -----|------|------------
 `sell_asset` | string | The asset you want to sell, using the [Asset Identification Format](#asset-identification-format).
+`sell_amount` | string | The amount of `sell_asset` the client would exchange for each of the `buy_assets`.
+`sep` | string | The name of the SEP that will be executing the quoted exchange. As of now, the ones that can make use of RFQ are `SEP-6` or `SEP-31`.
```
leighmcculloch, line 187: I'm hesitant for us to include a SEP name in here. Can we include an `intent` or similar instead? For example, `"intent": "transfer"` for the transfer API SEPs SEP-6/24, `"intent": "cross_border_payment"` for SEP-31. There are probably better names.

leighmcculloch (commented):
A question inline.

ecosystem/sep-0038.md
leighmcculloch: @JakeUrban wondering what your thoughts are on this.

JakeUrban (commented):
I think we may need to add an additional field used to communicate how the client would like funds to be delivered to the user. We can't redefine `delivery_method` to be used for both because there are two deliveries of funds in a SEP-31 transaction, sender --> anchor and anchor --> receiver.

In general, we need to ensure the anchor has all the information needed to calculate their necessary margin. I think having the service provided (`sep`), rails used for both flows (`delivery_method`), and country of operation (`country_code`) is a good start. If an anchor requires other information, we can try and integrate that information in the future.

ecosystem/sep-0038.md
marcelosalloum: Regarding `"intent"` vs `"sep"`, I think both have downsides: in the future, we could end up having different protocols for "cross_border_payment" or different kinds of quote-related actions inside the same SEP, so both `&sep=6` or `&intent=cross_border_payment` could become insufficient to describe the intended action. TBH I like the "intent" key for it being more generic and easier to expand, but I also like the values "sep6" and "sep31" as they give a precise definition of what it'll be used for and they're also easier to expand with edge cases.

---

> `delivery_method` is only used for off-chain --> on-chain (i.e. deposit) flows, so we may need another attribute for the opposite flow, describing how funds are delivered to the user.

@JakeUrban: maybe we could tweak the description of delivery_methods a bit so it can be used for both buy_asset and sell_asset:
* Update the description of delivery_methods to say `An array of objects describing the methods a client can use to deliver funds to or withdraw funds from the anchor.`
* Update the `/quote`, `/price` and `/prices` request parameters to have `buy_delivery_method` and `sell_delivery_method` as parameters that should be used when buy_asset or sell_asset are non-Stellar assets.

JakeUrban: > Update the description of delivery_methods to say An array of objects describing the methods a client can use to deliver funds to or withdraw funds from the anchor.

I think this approach would make the spec more confusing and make it easier to pass semantically invalid `delivery_method` values. For example, a client could potentially pass `cash_pickup` for an off-chain --> on-chain transaction, when the appropriate value would've been `cash_dropoff`.

From a design perspective, the methods a client uses to deliver funds to the anchor are disjoint from the methods an anchor supports for delivering funds to the client/user. Why describe both sets of methods under the same object?

What if we changed the existing `delivery_methods` object & parameter to be `client_delivery_method` and added a `anchor_delivery_method` object & parameter? So `/quote`, `/price`, and `/prices` requests would include a `client_delivery_method` if it was an off-chain --> on-chain exchange and an `anchor_delivery_method` if it was an on-chain --> off-chain exchange.

marcelosalloum: @JakeUrban: I applied your suggestion in https://github.com/stellar/stellar-protocol/pull/989/commits/533d52fa85965b55d87ec71b3388c0e0292278aa. 

Q: should this trigger a major version increment from `1.2.0` to `2.0.0` or are we ok going to 1.3.0?

JakeUrban: > Q: should this trigger a major version increment from 1.2.0 to 2.0.0 or are we ok going to 1.3.0?

Thats a good question. Technically this is a breaking change (renaming an existing field) but since the SEP is in a draft state I don't think we need to strictly adhere to semantic versioning. @leighmcculloch curious on your thoughts here.

leighmcculloch: I'd try and keep the SEP version under 1 until it is in an Active state. We have no rules or process for this, but I've found that useful since v0 is a break anything you want version in regards to semantic versioning.

JakeUrban (approved):
Pending the version change, this LGTM, thanks @marcelosalloum

ecosystem/sep-0038.md
```diff
@@ -149,7 +151,8 @@ Name | Type | Description
     },
     {
       "asset": "iso4217:BRL",
-      "delivery_methods": [
+      "country_codes": ["BRA"],
+      "client_delivery_methods": [
```
leighmcculloch, line 155: If these are always for deposit, could we use the term `deposit_delivery_methods`, and similar for anchors, `withdrawal_delivery_methods`? That builds on language we're already using for these two operations, where as `client` and `anchor` requires the reader to infer that during deposit they are the deliverer and therefore need to use this field.

cc @JakeUrban As I know you originally suggested client/anchor here.

JakeUrban: `deposit` and `withdraw` work too, but SEP-31 doesn't use that language. Maybe its still better than `client/anchor` since in SEP-31 the client is referred to as an anchor. I'm good with either 👍

leighmcculloch: `send` / `receive`?

marcelosalloum: What about `buy`/`sell`? They will be used along with parameters like buy_asset, buy_amount, sell_asset and sell_amount

ecosystem/sep-0038.md
```diff
@@ -124,9 +124,11 @@ Name | Type | Description
...
+`sell_delivery_methods` | array | (optional) Only for non-Stellar assets. An array of objects describing the methods a client can use to sell/deliver funds to the anchor. The method of delivery may affect the expiration and/or price provided in a [`POST /quote`](#post-quote) response.
+`buy_delivery_methods` | array | (optional) Only for non-Stellar assets. An array of objects describing the methods a client can use to buy/retrieve funds from the anchor. The method of delivery may affect the expiration and/or price provided in a [`POST /quote`](#post-quote) response.
+`country_codes` | array | (optional) Only for fiat assets. A list of [ISO 3166-1 alpha-3](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-3) codes of the countries where the Anchor operates for fiat transactions.
 
-`delivery_methods` Object Schema
+`delivery_methods` Object Schema, used for both parameters `sell_delivery_methods` and `buy_delivery_methods`.
```
leighmcculloch, line 131: It appears there is no `delivery_methods` anymore.
```suggestion
Object Schema for `sell_delivery_methods` and `buy_delivery_methods`.
```

leighmcculloch (approved):
I think this looks good. The buy/sell language works.

The PR title is not particularly specific. I suggest changing the title of the PR, and therefore the title of the final squashed commit, to:
> SEP-38: distinguish sell/buy delivery methods for fee calculation

ecosystem/sep-0038.md
marcelosalloum: Good suggestion!

ecosystem/sep-0038.md
```diff
@@ -174,13 +195,18 @@ This endpoint can be used to fetch the [indicative](https://www.investopedia.com
...
-The prices returned include any margin that the provider may keep as a service fee, and this margin may vary depending on the directional flow of funds, or other factors.
+The prices returned include any margin that the provider may keep as a service fee, and this margin may vary depending on the directional flow of funds, amount, delivery method, country code, or other factors.
 
 #### Request
 
 Name | Type | Description
 -----|------|------------
 `sell_asset` | string | The asset you want to sell, using the [Asset Identification Format](#asset-identification-format).
+`sell_amount` | string | The amount of `sell_asset` the client would exchange for each of the `buy_assets`.
+`sell_delivery_method` | string | (optional) One of the `name` values specified by the `sell_delivery_methods` array for the associated asset returned from [`GET /info`](#info). Must be provided if the user is making an off-chain deposit to the anchor.
```
JakeUrban, line 206: Can we avoid the terms `deposit`? That way we limit the number of terms we use when describing the movement of funds.

```suggestion
`sell_delivery_method` | string | (optional) One of the `name` values specified by the `sell_delivery_methods` array for the associated asset returned from [`GET /info`](#info). Must be provided if the user is delivering an off-chain asset to the anchor.
```

ecosystem/sep-0038.md
```diff
@@ -174,13 +195,18 @@ This endpoint can be used to fetch the [indicative](https://www.investopedia.com
...
+`buy_delivery_method` | string | (optional) One of the `name` values specified by the `buy_delivery_methods` array for the associated asset returned from [`GET /info`](#info). Must be provided if the user intends to receive an off-chain deposit from the anchor.
```
JakeUrban, line 207: Same thing here, and for the attributes on the `/price` endpoint.

```suggestion
`buy_delivery_method` | string | (optional) One of the `name` values specified by the `buy_delivery_methods` array for the associated asset returned from [`GET /info`](#info). Must be provided if the user intends to receive an off-chain asset from the anchor.
```

JakeUrban (changes requested):
Very minor change, but otherwise LGTM! Once the syntax is changed I'll approve

ecosystem/sep-0038.md
marcelosalloum: Fixed in https://github.com/stellar/stellar-protocol/pull/989/commits/4256e69a463f879755c70b35e3032e28a3643391

ecosystem/sep-0038.md
marcelosalloum: Fixed in https://github.com/stellar/stellar-protocol/pull/989/commits/4256e69a463f879755c70b35e3032e28a3643391

JakeUrban (approved):

marcelosalloum:
Hey @JakeUrban, the review was dismissed when I bumped the SEP update date in https://github.com/stellar/stellar-protocol/pull/989/commits/c304ac799ba2b15ac1c0ae109a6326ba0303b88e. Could you approve again?

JakeUrban (approved):

PR #990, opened 2021-06-30
SEP: Update Team

leighmcculloch:
### What
Remove Tom Q from the SEP team, and add Leigh and Jake.

### Why
I'm not really sure the SEP team definition is as distinct or important as the CAP team is, but it looks out of date, and myself and Jake make decisions about state transitions of SEPs. It's possible we should remove this list of people all together, but that seems like a larger conversation to have and once updated the list does seem somewhat accurate indication of the people who will typically approve SEPs.

### DIFF
--- a/ecosystem/README.md
+++ b/ecosystem/README.md
@@ -175,7 +175,7 @@ From there, the following process will happen:
 
 ## SEP Team Members
 
-**SEP Team**: Tomer (SDF), Nikhil (SDF) Tom Q. (SDF), Michael (SDF), Alex (SDF), orbitlens, David (SDF), Jed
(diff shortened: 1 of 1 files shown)


accordeiro:
Since we're updating this, should we either use GH handles or full names (or both)? Having a list of first names is probably not that helpful.

leighmcculloch:
> should we either use GH handles or full names (or both)?

I went with both. Both is consistent with how we typically specify authors on SEPs and CAPs.

accordeiro (approved):
Looks great! Thanks for updating this :)

JakeUrban (commented):
Lets add @marcelosalloum as well

accordeiro:
Good catch @JakeUrban – looking at this again, should we also add @rice2000?

PR #993, opened 2021-07-01
SEP: Add details about SEP versioning

leighmcculloch:
### What
Add instructions about how to set the version number in SEPs.

### Why
We have no instructions for how to use versions, and that leads to some unfortunate situations during the draft state of a SEP. It's happened where we start a draft of a new SEP, start with v1, and then a little while later while it is still in draft we want to make a breaking change and need to change to v2. Drafts are intended to be a work-in-progress where unrestrained iteration is occurring, anything can change, and in semantic versioning that's what the v0 major version is intended for.

### DIFF
--- a/ecosystem/README.md
+++ b/ecosystem/README.md
@@ -139,6 +139,7 @@ From there, the following process will happen:
 * You should continue the discussion of the draft SEP on the mailing list to gather additional
   feedback. We welcome any additional PRs that iterate on the draft.
 * Keep the version of the SEP as a v0 version while in draft.
+* Increment the minor or patch versions on each change while in draft. See [SEP Versioning].
 
 ### Draft -> Awaiting Decision -> Final Comment Period (FCP)
 * When you're ready, you should submit a PR changing the status in the draft to `Awaiting Decision`.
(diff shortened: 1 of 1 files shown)


JakeUrban (approved):
LGTM, thanks for adding this clarification.

rice2000 (dismissed):
This is great.  Thanks @leighmcculloch!

JakeUrban (approved):

ecosystem/README.md
```diff
@@ -158,24 +161,36 @@ From there, the following process will happen:
...
+* Patch changes may be made to address bugs, errors, clarifications, or to fix errata.
 * Minor changes may be made as more implementations are brought online highlighting any edge cases.
+* Major changes, and breaking changes, should be considered with care as they may reduce interoperability.
 
 ### Active -> Final
 * Once the SEP team determines that an active SEP is complete, proven, and won't be extended, the SEP can move to `Final` status.
 * This promotion can only occur once there are multiple live implementations being used in production to ensure any edge cases or incompatibilities are found.
-* No changes will be made to a finalized SEP aside from fixing minor errata.
+* No changes will be made to a finalized SEP aside from fixing errata.
+* Changes should increment the the patch version number only.
```
accordeiro, line 176: There's a duplicate "the" here.

```suggestion
* Changes should increment the patch version number only.
```

accordeiro (approved):
This is a great addition, thanks @leighmcculloch!

accordeiro (approved):

PR #996, opened 2021-07-12
CAP-21: Update discussion URL

leighmcculloch:
### What
Update the discussion URL to the new mailing list thread.

### Why
The discussion URL is pointing to the old thread. The new mailing list thread is where all discussion is happening now.


rice2000 (approved):
Good catch.  Thanks, @leighmcculloch!

PR #998, opened 2021-07-14
CAP-40: Add proposal to add a ed25519 signed payload signer

leighmcculloch:
### What
Add a proposal to add an ed25519 signed hash signer for use alongside the precondition primitives, specifically the `extraSigners` precondition, that are proposed in CAP-21.

### Why
See within the document.

Close stellar/experimental-payment-channels#165

### DIFF
--- a/core/cap-0040.md
+++ b/core/cap-0040.md
@@ -138,9 +138,9 @@ signatures.
 Signature checking is changed to include verifying that any ed25519 signed hash
 signer's have matching signatures.
 
-Ed25519 signatures are verified by performing ed25519 signature verification
-using the signature, the hash from the signer, and the ed25519 public key from
-the signer.
+Ed25519 signed hash signer signatures are verified by performing ed25519
+signature verification using the signature, the hash from the signer, and the
+ed25519 public key from the signer.
 
 ## Design Rationale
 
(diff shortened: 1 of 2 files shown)


core/cap-0040.md
```diff
@@ -0,0 +1,232 @@
...
+```diff mddiffcheck.base=v17.2.0
+diff --git a/src/xdr/Stellar-types.x b/src/xdr/Stellar-types.x
+index 8f7d5c20..820d67ff 100644
+--- a/src/xdr/Stellar-types.x
++++ b/src/xdr/Stellar-types.x
+@@ -19,6 +19,7 @@ enum CryptoKeyType
+     KEY_TYPE_ED25519 = 0,
+     KEY_TYPE_PRE_AUTH_TX = 1,
+     KEY_TYPE_HASH_X = 2,
++    KEY_TYPE_ED25519_SIGNED_HASH = 4,
```
leighmcculloch, line 81: It would probably be more accurate if this was named:
```suggestion
+    KEY_TYPE_HASH_TO_SIGN_ED25519 = 4,
```

leighmcculloch: Or:
```suggestion
+    KEY_TYPE_HASH_FOR_ED25519 = 4,
```

stanford-scs:
Hint is too expensive.  Do something without SHA.

What happened to crypto key type 3?

Be explicit about what is being signed

Instead of hash, just make it a variable-length opaque data.

Need new strkey type.  (SEP-0023)

leighmcculloch:
Thanks @stanford-scs. I addressed your comments. For the strkey, I'll address that after the CAP team has done an initial review of the CAP and it shows to have promise for being accepted.

stanford-scs (approved):
Very nice!

PR #1006, opened 2021-07-21
SEP-39: Add strkey proposal for CAP-40

leighmcculloch:
### What
Add a strkey change proposal for the new signer added in [CAP-40].

### Why
[CAP-40] introduced a new signer for disclosing transaction signatures in transactions. Some ecosystem services, such as Horizon, only refer to signers in string form. Stellar SDKs also use strings to pass around signers. A string form for the signer introduced by CAP-40 is required for these services.

Only @leighmcculloch authored this SEP, but it is a copy of [SEP-23] with modifications, so if folks don't mind I've listed all of the authors of SEP-23 too.

[CAP-40]: https://stellar.org/protocol/cap-40
[SEP-23]: https://stellar.org/protocol/sep-23

Close stellar/experimental-payment-channels#191

cc @acharb @nikhilsaraf


ecosystem/sep-0039.md
```diff
@@ -0,0 +1,274 @@
...
+length payload that is prefixed with a 4 byte length.
+
+| Key type                | Base value   | First char | Muxed | Alg  |
+| ----------------------- | ------------ | ---------- | ----- | ---- |
+| STRKEY\_PUBKEY          | 6 << 3       | G          | no    | PK   |
+| STRKEY\_MUXED           | 12 << 3      | M          | yes   | PK   |
+| STRKEY\_PRIVKEY         | 18 << 3      | S          | no    | PK   |
+| STRKEY\_PRE\_AUTH\_TX   | 19 << 3      | T          | no    | Hash |
+| STRKEY\_HASH\_X         | 23 << 3      | X          | no    | Hash |
+| STRKEY\_SIGNED_PAYLOAD  | 23 << 3      | P          | no    | PK   |
```
stanford-scs, line 54: I think you mean 15 << 3 (see https://datatracker.ietf.org/doc/html/rfc4648#section-6)

stanford-scs (commented):
The document seems okay with the fix to the base32 decoded value of 'P'.  However, I think it might be confusing to call this document SEP-39.  I think we should consider either sep-23 or sep-23bis or something, or we should deprecate sep-23.  It's going to be confusing if there isn't an obvious place to get the latest version of the strkey specification.

ecosystem/sep-0039.md
leighmcculloch: Yes I do. Thanks.

leighmcculloch:
For other SEPs we update them in place and update their version number. We could do same here except that SEP23 is not a specification, it's more of a proposal. 

For other SEPs making small incremental changes has worked okay, but it's difficult to debate or reject changes on the mailing list, and this SEP is more related to the CAP process and tied to CAP-40 so I wouldn't want to alter SEP23 until CAP-40 is implemented. 

It'd be a bit odd to deprecate SEP-23 because people are implementing it right now and it just moved to Active status. Nothing about it is invalid either with this change.

If I update SEP23 in place and change it from a proposal to a specification, we'll need to revert the change if necessary, or keep the change in a PR for a couple months.

Or we could add a new meta header, `Updated by: SEP-39` to SEP-23 to make it clear that SEP-23 is still valid, but it's updated by another SEP. I think this is similar to the RFC process?

leighmcculloch:
I'm going to close this and make the change afresh as a modification of SEP-23.

leighmcculloch:
Superseded by #1013 and #1014.

PR #1007, opened 2021-07-21
SEP-10: Advise servers to set minTime in the past

morleyzhi:
Currently, we ask SEP-10 implementors to set the minTime of their challenge transaction to now(). But if the server's clock is ahead of a client's clock, the client could try to validate the server's transaction, and validation would fail because their clock is faster than the transaction's minTime.
To fix this, we suggest setting the minTime to 5 minutes before now().

### DIFF
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -121,7 +121,9 @@ On success the endpoint must return `200 OK` HTTP status code and a JSON object 
 * `transaction`: an XDR-encoded Stellar transaction with the following:
   * source account set to the **Server Account**
   * invalid sequence number (set to 0) so the transaction cannot be run on the Stellar network
-  * time bounds: `{min: now(), max: now() + 900 }` (we recommend expiration of 15 minutes to give the **Client** time to sign transaction)
+  * time bounds: `{min: now() - 300, max: now() + 900 }`
+    * We recommend expiration of 15 minutes to give the **Client** time to sign transaction
+    * We recommend setting the min time to 5 minutes in the past, in case the **Client**'s clock is slightly behind the **Server**'s
   * operations:
     * `manage_data(source: client account, key: '<home domain> auth', value: random_nonce())`
       * The source account is the **Client Account**


ecosystem/sep-0010.md
```diff
@@ -121,7 +121,9 @@ On success the endpoint must return `200 OK` HTTP status code and a JSON object
 * `transaction`: an XDR-encoded Stellar transaction with the following:
   * source account set to the **Server Account**
   * invalid sequence number (set to 0) so the transaction cannot be run on the Stellar network
-  * time bounds: `{min: now(), max: now() + 900 }` (we recommend expiration of 15 minutes to give the **Client** time to sign transaction)
+  * time bounds: `{min: now() - 300, max: now() + 900 }`
+    * We recommend expiration of 15 minutes to give the **Client** time to sign transaction
+    * We recommend setting the min time to 5 minutes in the past, in case the **Client**'s clock is slightly behind the **Server**'s
```
leighmcculloch, line 126: I think we should consider dropping validation of the `min` timebound in our SDKs instead of making this change.

The SDKs validate the min timebound, but SEP-10 doesn't require that it be validated for clients. It requires it to be validated by servers, see the "Token" section for that, but for any distributed system that is likely going to cause similar issues.

After reading around a little, I found [a few projects](https://github.com/dgrijalva/jwt-go/issues/246) that have chosen to stop validating the issued-at field of JWTs, including auth0, which is comparable to the min timebound we use in SEP-10.

This raises the thought, should we set the min time to zero instead? No I don't think so, because we still want a stable issued-at value in our JWTs so we know when they were issued, and in some implementations of SEP-10 the JWT's issued-at value is set based on the challenge's min timebound.

@morleyzhi @JakeUrban @accordeiro Thoughts?

leighmcculloch (commented):
Thanks for highlighting this issue. I think we should take a different approach to solving the problem that is aligned with how this problem is being solved in similar auth methods. See below.

ecosystem/sep-0010.md
JakeUrban: I think we should remove the `iat` JWT claim in addition to dropping min time validation, if we chose to go that route. Otherwise it will be difficult to make the JWT returned purely derivative of the challenge used to request it without persisting JWT's until its expiration.

That being said, I'm happy with simply setting the mintime a fixed length back from "now()" as well.

leighmcculloch: > I think we should remove the `iat` JWT claim in addition to dropping min time validation

Removing the `iat` claim would be fine according to [RFC7519 Section 4.1.6](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.6) because the claim is optional.

However, it isn't clear to me how we'd implement code like the below that ensures the same JWT is issued for the same challenge by setting the JWT's lifespan based on the challenges min time. I think we still need min time.
```go
issuedAt := time.Unix(tx.Timebounds().MinTime, 0)
claims := jwt.Claims{
	Issuer:   h.JWTIssuer,
	Subject:  clientAccountID,
	IssuedAt: jwt.NewNumericDate(issuedAt),
	Expiry:   jwt.NewNumericDate(issuedAt.Add(h.JWTExpiresIn)),
```
Ref: https://github.com/stellar/go/blob/e358a1b/exp/services/webauth/internal/serve/token.go#L143-L148

JakeUrban: Ah you're right, I forgot that the expiration is dependent on the time it was issued.

leighmcculloch: I think the grace period change proposed by @morleyzhi on the JS SDK https://github.com/stellar/js-stellar-sdk/pull/667 would also be perfect without needing to change the values added to the challenge or JWT, which makes the change more backwards compatible.

JakeUrban: To confirm, this will require us reaching out to SDK maintainers right? Given that this request isn't motivated by a security concern I don't believe this change is time-sensitive. We can simply file issues on the repositories as we when we need SDKs to update for an update to the core protocol.

leighmcculloch: I think so yeah. I think it would be with embodying the best practice in this SEP first so that we can reference that change. @morleyzhi are you up for repurposing this PR for that?

JakeUrban:
I'm closing this issue but will open a new one for tracking SDK adoption of the grace period pattern we've discussed.

PR #1013, opened 2021-07-24
SEP-23: Rework to be a standalone specification for strkeys

leighmcculloch:
### What
Rework SEP-23 so that it is a standalone specification for strkeys rather than a proposal to add a muxed account strkey.

### Why
SEP-23 is currently a definition of strkeys but written from the perspective of being a proposal for adding a muxed account strkey. We could really do with a single place to define all strkeys and iterate on that SEP as new strkeys are added. That is more like how other SEPs work.

We will soon propose a new strkey that will be used for CAP-40, and so I am doing this work to SEP-23 now so that it is already in the position of being the place where strkeys are specified.

This change attempts to make as least changes possible to change the intended use of the document. There are probably ways we could improve this document by adding invalid and valid test cases. We can do that, but let's do that in a different PR to this.

This change is a patch version bump because it does not introduce any breaking changes, nor introduce, change, or remove any functionality. The changes to the document are superficial and changing the intended future use of the document.

Related to https://github.com/stellar/experimental-payment-channels/issues/191

### DIFF
--- a/ecosystem/sep-0023.md
+++ b/ecosystem/sep-0023.md
@@ -2,52 +2,48 @@
 
 ```
 SEP: 0023
-Title: Muxed Account Strkeys
+Title: Strkeys
 Author: David Mazières, Tomer Weller <@tomerweller>, Leigh McCulloch <@leighmcculloch>, Alfonso Acosta <@2opremio>
 Track: Standard
 Status: Active
 Created: 2019-09-16
-Updated: 2021-06-02
-Version: 1.0.0
+Updated: 2021-07-23
+Version: 1.0.1
 Discussion: https://groups.google.com/forum/#!forum/stellar-dev
 ```
 
 ## Simple Summary
 
 Strkey is an ASCII format for representing Stellar account IDs and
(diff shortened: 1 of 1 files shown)


tomerweller (dismissed):
I love this change. Much more coherent.

ecosystem/sep-0023.md
```diff
@@ -259,14 +217,21 @@ using the following array:
...
+https://github.com/xdrpp/stc
 
 ## Security Concerns
 
-Augmented strkey could potentially make it harder to see where
-payments are going, as different multiplexed accounts will reference
-the same underlying account.
+Implementations should not rely on the CRC to determine if a strkey is
+incomplete or has been truncated. Existing strkeys are designed to be
+fixed length, with their length determinible from their strkey type and
```
stanford-scs, line 232: determin**a**ble

stanford-scs (changes requested):
I'm not crazy about the Security Concerns section trying to prescribe things about future revisions on the specification.  It seems more like a design rationale, that the current format is prefix-free, so I'd mention it in the rationale section.

PR #1014, opened 2021-07-24
SEP-23: Add strkey type for CAP-40 signed payload signers

leighmcculloch:
### What
Introduces a new strkey type to represent ed25519 signed payload signers introduced by [CAP-40].

### Why
[CAP-40] introduces a new signer for disclosing transaction signatures in transactions. Some ecosystem services, such as Horizon, only refer to signers in string form. Stellar SDKs also use strings to pass around signers. A string form for the signer introduced by CAP-40 is required for these services.

The leading `P` prefix for the new strkey was chosen to indicate that the signer is a signed payload because the word payload begins with p.

The 4-byte length prefix is to ensure that a truncated signed payload that has a valid CRC is not considered valid.

The 4-byte length prefix and the payload zero padding to a multiple of 4 bytes makes the raw bytes encoded in the strkey identical to the signer key when XDR encoded, a property of existing strkeys, excluding muxed accounts that match their XDR encoding with fields reversed.

Close stellar/experimental-payment-channels#191

### Merging
This is intended for merging if [CAP-40] is accepted.

[CAP-40]: https://stellar.org/protocol/cap-40


ecosystem/sep-0023.md
```diff
@@ -77,14 +79,20 @@ The following steps transform a binary key into a strkey:
...
 
-4.  Compute a 16-bit CRC16 checksum of the combined version byte,
+4.  If we are encoding a signed payload, append a 4-byte length in
+    network byte order (most significant byte first) that holds the
+    length of the payload, then append the payload, and finally zero
+    padding of 0 to 3 zero bytes such that the total number of bytes of
+    the payload plus the zero padding is a multiple of four.
+
+5.  Compute a 16-bit CRC16 checksum of the combined version byte,
     optional memo ID, and binary key (using polynomial
```
stanford-scs, line 89: optional memo ID, optional payload, and binary key

stanford-scs (changes requested):
This looks good except for the one nit I put in an annotation of the diff.

ecosystem/sep-0023.md
leighmcculloch: 817c10f6cd06bc48daa9e59f6f284207d74b7804

stanford-scs (approved):

Shaptic:
@leighmcculloch Will this get merged soon? I'd like to start the relevant SDK work but don't want something to change in the middle of it.

leighmcculloch:
@Shaptic This is unlikely to be merged until the core implementation is complete, otherwise we'll have no room to make breaking changes if necessary. However, I don't think this will change substantially as strkeys are opaque so any change is unlikely to require only tweaks in the SDK and not changes downstream to Horizon.

Also, at the moment the PR has only one approval. It would be great if @stellar/horizon-committers, and anyone else, could review also.

ecosystem/sep-0023.md
```diff
@@ -77,14 +79,20 @@ The following steps transform a binary key into a strkey:
...
+    padding of 0 to 3 zero bytes such that the total number of bytes of
+    the payload plus the zero padding is a multiple of four.
 
-3.  Encode the result of the previous step using [RFC4648 base-32
+5.  Compute a 16-bit CRC16 checksum of the combined version byte,
+    optional memo ID, optional payload, and binary key (using
+    polynomial x<sup>16</sup> + x<sup>12</sup> + x<sup>5</sup> + 1).
+    Append the two-byte checksum to the result of the previous step
+    (e.g., producing a 35-byte quantity for a non-multiplexed ED25519
+    public key, or 43 byte quantity for a multiplexed one).
```
Shaptic, line 93: Obviously the `e.g.` list here is not meant to be comprehensive, but it might be worth adding the payload case anyway.

ecosystem/sep-0023.md
```diff
@@ -77,14 +79,20 @@ The following steps transform a binary key into a strkey:
 3.  If we are encoding a multiplexed address, append an 8-byte memo ID
     in network byte order (most significant byte first).
 
-4.  Compute a 16-bit CRC16 checksum of the combined version byte,
-    optional memo ID, and binary key (using polynomial
-    x<sup>16</sup> + x<sup>12</sup> + x<sup>5</sup> + 1). Append the
-    two-byte checksum to the result of the previous step (e.g.,
-    producing a 35-byte quantity for a non-multiplexed ED25519 public
-    key, or 43 byte quantity for a multiplexed one).
+4.  If we are encoding a signed payload, append a 4-byte length in
```
Shaptic, line 82: Payloads are 64 bytes, so can we just use a single byte for length here?

ecosystem/sep-0023.md
```diff
@@ -52,6 +53,7 @@ base-32 encoding of the key, are listed here:
 | STRKEY\_PRIVKEY         | 18 << 3      | S          | no    | PK   |
 | STRKEY\_PRE\_AUTH\_TX   | 19 << 3      | T          | no    | Hash |
 | STRKEY\_HASH\_X         | 23 << 3      | X          | no    | Hash |
+| STRKEY\_SIGNED\_PAYLOAD | 15 << 3      | P          | no    | PK   |
```
Shaptic, line 56: For my own edification: How are these values (`15 << 3` et. al) decided?

ecosystem/sep-0023.md
```diff
@@ -77,14 +79,20 @@ The following steps transform a binary key into a strkey:
...
-    key, or 43 byte quantity for a multiplexed one).
+4.  If we are encoding a signed payload, append a 4-byte length in
+    network byte order (most significant byte first) that holds the
+    length of the payload, then append the payload, and finally zero
+    padding of 0 to 3 zero bytes such that the total number of bytes of
+    the payload plus the zero padding is a multiple of four.
 
-3.  Encode the result of the previous step using [RFC4648 base-32
+5.  Compute a 16-bit CRC16 checksum of the combined version byte,
+    optional memo ID, optional payload, and binary key (using
```
Shaptic, line 89: The unfortunate amendment to this existing sentence makes it sound like memo IDs and payloads can both be part of an address. The earlier steps already establish this, so I don't think there's a big risk of ambiguity here, but maybe we can say `optional memo ID xor payload` as a minimal workaround?

Shaptic (approved):
Some minor optional comments :arrow_down:, but LGTM as-is :+1:

ecosystem/sep-0023.md
leighmcculloch: Yeah true. I've simplified this section in bf6f538.

ecosystem/sep-0023.md
leighmcculloch: See https://go.dev/play/p/GhbGoyh-_1O.

These values define the first character in the base 32 strkey. Base 32 encodes 5 bits into every byte/character. So the first 5 bits define the first character. The next 3 bits are used to define the alg. And that makes up the header that is common to all strkeys.

ecosystem/sep-0023.md
leighmcculloch: Today payloads are 64 bytes, but in the future they might be increased. We could reduce the number of bytes here as an optimization I guess to at least down to 2 bytes. I think originally @stanford-scs was trying to keep strkeys as generally the encoding / decoding to valid XDR, typically the full XDR for that type, but we haven't followed that strictly in the past. XDR requires a length to be 4 bytes which is where the 4 bytes comes from.

leighmcculloch: I guess if we really wanted to optimize while preserving future possible used we could encode the length into 7-bit chunks before encoding into base32. It would make the minimum length of the length 1 byte. If we ever supported 32-bit lengths for the payload it the length of the length would be 5 bytes. This might be a pain to code in most SDKs, but most SDKs could probably punt on actually implementing it on the assumption payloads are only length 64 max today.

leighmcculloch: I'm inclined to say nothing is worth optimizing here and leaving it as it is. Wdyt @stanford-scs @ire-and-curses?

ecosystem/sep-0023.md
Shaptic: Ah, of course. I expected `15 << 3` to be `P` in ascii, but forgot about the base32 aspect of it all.

ecosystem/sep-0023.md
Shaptic: > I think originally @stanford-scs was trying to keep strkeys as generally the encoding / decoding to valid XDR, typically the full XDR for that type,

If this was an initial design goal then I'm not against keeping it that way (though muxed addresses already deviate from this, with the ID coming second in the StrKey but first in the XDR). I just wanted to bring up that we might be able to shave 2-3 bytes off per strkey here. 

Over-optimizing for base32 seems excessive, and dropping to 1 byte does limit future extensions (though having signed >256 byte payloads seems fictitious: you should hash your payload and sign _that_ instead), but 2 bytes would be a reasonable middle ground. However, as you noted, it would deviate from the underlying XDR, so if that's an important element to preserve here, this micro-optimization discussion probably isn't worth it.

leighmcculloch: This is what the saving looks like:
```
4-byte len: PA7QYNF7SOWQ3GLR2BGMZEHXAVIRZA4KVWLTJJFC7MGXUA74P7UJUAAAAAQACAQDAQCQMBYIBEFAWDANBYHRAEISCMKBKFQXDAMRUGY4DUPB6IBZGM
3-byte len: PA7QYNF7SOWQ3GLR2BGMZEHXAVIRZA4KVWLTJJFC7MGXUA74P7UJUAAAEAAQEAYEAUDAOCAJBIFQYDIOB4IBCEQTCQKRMFYYDENBWHA5DYPSAOJT
2-byte len: PA7QYNF7SOWQ3GLR2BGMZEHXAVIRZA4KVWLTJJFC7MGXUA74P7UJUABAAEBAGBAFAYDQQCIKBMGA2DQPCAIREEYUCULBOGAZDINRYHI6D4QDSMY
1-byte len: PA7QYNF7SOWQ3GLR2BGMZEHXAVIRZA4KVWLTJJFC7MGXUA74P7UJUIABAIBQIBIGA4EASCQLBQGQ4DYQCEJBGFAVCYLRQGI2DMOB2HQ7EA4TG
```
Ref: https://go.dev/play/p/sLUfHJImdR4

A 2-byte length would reduce the total strkey by 3 characters.

Shaptic: I'll be honest, the only reason why I even noticed this is because I saw the `AAAA` characters in the middle of the strkey and that seemed odd to me. So reducing the length has an aesthetic benefit as well :laughing: but again I don't have a strong opinion here, especially if we want adherence to XDR, as it'd make SDK code simpler.

leighmcculloch: Unless there's a scaling implication, such as a significant cost to store the strkey in databases, I don't think we should over optimize it. The XDR type could allow a larger value in the future.

Shaptic (dismissed):

ecosystem/sep-0023.md
Lippertt: Eiiiii

ire-and-curses: Good discussion, but yeah I also don't think this is worth optimising for.

Lippertt:
Não tô te entendendo fala tu mano

Em qui, 10 de mar de 2022 2:29 PM, George ***@***.***>
escreveu:

stanford-scs (approved):
Looks good to me.

PR #1018, opened 2021-07-27
SEP-10: improve discovery flow

christian-rogobete:
for the client verification of the challenge: adds that the source account of the manage data operation must be the client domain account in case that the manage data operation key is "client_domain" and not the server account

### DIFF
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -6,8 +6,8 @@ Title: Stellar Web Authentication
 Author: Sergey Nebolsin <@nebolsin>, Tom Quisel <tom.quisel@gmail.com>, Leigh McCulloch <@leighmcculloch>, Jake Urban <jake@stellar.org>
 Status: Active
 Created: 2018-07-31
-Updated: 2021-06-24
-Version 3.2.1
+Updated: 2021-07-27
+Version 3.2.2
 ```
 
 ## Simple Summary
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0010.md
```diff
@@ -46,7 +46,9 @@ The authentication flow is as follows:
 1. The **Client** verifies that if the transaction has a Manage Data operation with key `web_auth_domain` that it has:
    1. Source account set to the **Server Account**.
    1. Value set to the **Server**'s domain that the client requested the challenge from.
-1. The **Client** verifies that if the transaction has other operations they are Manage Data operations that all have their source accounts set to the **Server Account**
+1. The **Client** verifies that if the transaction has other operations they are Manage Data operations and that their source account is set to:
+   1. The **Server Account** if the Manage Data operation key is NOT set to `client_domain`
+   1. The **Client Domain Account** if the Manage Data operation key is set to `client_domain`
```
leighmcculloch, line 51: I think this list can be simplified a little such that if we add other manage data operations with other rules they can be inserted here.
```suggestion
   1. The **Client Domain Account** if the Manage Data operation key is set to `client_domain`
   1. Otherwise, the **Server Account**.
```

leighmcculloch (commented):
This looks good to me. One request below, and can you update the version to v3.2.2 and the updated date which are both at the top of the document in the preamble?

I suggest a patch version bump because this behavior is already intended but isn't captured clearly here.

christian-rogobete:
Hi @leighmcculloch, thanks for your comment. I added your suggestions.

leighmcculloch (approved):
Looks good to me. @JakeUrban would you also 👀 this change before we merge it?

JakeUrban (approved):
Looks great, thanks for adding.

PR #1024, opened 2021-07-30
CAP-21: Add some notes about resource utilization

leighmcculloch:
### What
Add some notes about resource utilization.

### Why
I noticed that one side-effect benefit of CAP-21's `extraSigners` is we make the use cases of the existing HASH_X signer more efficient since there is now no, or little, reason to store them on ledger as an account signer. I feel like it is worth calling that out, that this proposal may in some ways improve utilization, albeit small.

While adding a brief blurb about that, it felt natural to add some other comments about utilization that were missing.

### DIFF
--- a/core/cap-0021.md
+++ b/core/cap-0021.md
@@ -731,7 +731,9 @@ the transaction at s with `minSeqNum` s-99 ensures that if any of the
 servers do not submit transactions, the gap will not prevent other
 transactions from executing.
 
-## Backwards Incompatibilities
+## Protocol Upgrade Transition
+
+### Backwards Incompatibilities
 
 Previously signed transactions containing time points greater than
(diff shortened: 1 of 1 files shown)


stanford-scs:
> The maximum number of signatures that must be verified for each
> transaction will not change.

The changes look good, except the above actually needs to be specified in the document (where you've place it does not make it a normative requirement).  I plan to do that, so I guess we could just merge in anticipation of my next changes.

stanford-scs (approved):
see my comment in the conversation.

leighmcculloch:
I didn't add a requirement because I thought it was an assumption since nothing in the CAP changed the maximum number of signers, but it would be better to make it clear as a requirement. I'll merge this, and the requirement can be added by your update. Thanks.

PR #1025, opened 2021-07-30
CAP-40: Add example of other use case

leighmcculloch:
### What
Add example of another use case to CAP-40.

### Why
At the request of @jonjove.

cc @stanford-scs @MonsieurNicolas

### DIFF
--- a/core/cap-0040.md
+++ b/core/cap-0040.md
@@ -255,6 +255,45 @@ signature for C_i, the sender who is watching the network may see D_i be
 submitted, extract the signature for C_i from the signatures on D_i, attach the
 signature to C_i, and submit C_i.
 
+### Cross-Chain Atomic Operations
+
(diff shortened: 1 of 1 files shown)


leighmcculloch:
I am not pursuing adding this use case that there are no systems that this example is meaningfully compatible with.

If we want the concepts in CAP-40 to be generally applicable outside the main use case of signing Stellar transaction hashes then we'd need to modify CAP-40 to:
1. Support ecdsa secp256k1 too, so add SIGNER_KEY_TYPE_ECDSASECP256K1_SIGNED_PAYLOAD alongside the new signer already defined. This would support signing for a wide variety of chains Bitcoin, Ethereum, etc.
2. Expand the payload max length to something like 256/512 bytes. This would support signing with a variety of products that use ed25519, such as Polkadot, Paseto, etc.

PR #1036, opened 2021-08-09
Memo-based Pooled Account Support

### LINKED ISSUE #727: SEP-10: Support pooled accounts

JakeUrban:
resolves #727 

### SEP-10

- Adds the `memo` request parameters to the `GET WEB_AUTH_ENDPOINT` request
- Updates the `account` parameter to also accept muxed accounts (`M...`)
- Adds a memo to the challenge transaction returned from `GET WEB_AUTH_ENDPOINT`
- Updates the decoded JWT's `sub` value format to include the memo or be the muxed account used as the `account` parameter

### SEP-6

- Clarifies `GET /deposit`'s `memo` and `memo_type` request parameters strictly as the memo on the transaction submitted by the anchor to deliver on-chain funds
    - Previously, there was confusion around whether the memo parameters could also be used to identify users of a shared Stellar account, see [#955](https://github.com/stellar/stellar-protocol/pull/955#discussion_r640955335)
- Clarifies that the `memo` and `memo_type` request parameters of the `GET /withdraw` endpoint should only be used when the anchor does not support SEP-10 authentication
- Mandates anchors restrict the transactions returned from `GET /transaction(s)` endpoints to only be those created by requests containing the same `sub` value as the request for transactions

### SEP-24

The changes described here are identical to those made for SEP-6, with one exception:

- Deprecates the `memo` and `memo_type` request parameters of the `GET /withdraw` endpoint, since SEP-10 is required

### SEP-12

The changes made for this protocol are more nuanced because it is used by both SEP-6 and SEP-31. In SEP-6, authenticated sessions are scoped to a single user's (or customer's) transactions. In SEP-31, authenticated sessions are scoped to a Sending Anchor's transactions and customers.

This means that a client interacting with a SEP-12 implementation may be a Wallet acting on behalf of a single user, in which case the memo may be present in the SEP-10 JWT's `sub` value and therefore the `memo` and `memo_type` request parameters are no longer necessary, but for the sake of avoiding more changes to existing implementations should still be specified.

It also means that a client interactive with a SEP-12 implementation may be a SEP-31 Sending Anchor initiating the transfer of funds between two users, in which case the SEP-10 JWT shouldn't have memo or muxed account in the `sub` value. Instead, the SEP-31 anchor should use the `memo` request parameter to identify users.

SEP-12 implementors therefore must support both types of identifying users unless the protocol is only used for SEP-31 transactions.

- Removes text suggesting SEP-10 JWT's can be specified in URL's
    - There is no use case for this
- Deprecates the use of the `account` request parameter in all applicable endpoints
- Deprecates the use of the `memo_type` request parameter -- memos should always be of type `id`
- For all endpoints, the `memo` request parameter must match the memo portion of the JWT's `sub` value if present.
    - This ensure the anchor can use a single memo to identify the user in association with the Stellar account
 - For all endpoints, the `memo` request parameter should not be accepted if the JWT's `sub` value is a muxed account.
    - This ensure the anchor does not use both a memo and muxed account to identify the same user.


ecosystem/sep-0010.md
```diff
@@ -106,7 +108,9 @@ Request Parameters:
 
 Name      | Type          | Description
 ----------|---------------|------------
-`account` | `G...` string | The stellar account that the wallet wishes to authenticate with the **Server** 
+`account` | `G...` string | The stellar account that the **Client** wishes to authenticate with the **Server** 
+`memo` | string | (optional) The memo to attach to the challenge transaction, for `hash` this should be base64-encoded. See the [Memo](#memos) section for details.
```
leighmcculloch, line 112: ❗ We should use hex encoding for the hash. That's what we do in the lab, and it is more common in general outside of Stellar to hex encode hashes when a string representation is needed.

ecosystem/sep-0010.md
```diff
@@ -106,7 +108,9 @@ Request Parameters:
...
+`memo_type` | string | (optional) The type of memo to attach to transaction, one of `text`, `id` or `hash`.
```
leighmcculloch, line 113: ❓ Could we start with supporting only a single memo type, ideally ID, since then it matches muxed accounts. See my other comment about this further up.

ecosystem/sep-0010.md
```diff
@@ -138,6 +142,7 @@ On success the endpoint must return `200 OK` HTTP status code and a JSON object
         * Add this operation if the server supports [Verifying the Client Domain](#verifying-the-client-domain) and the client provided a `client_domain` parameter in the request.
       * zero or more `manage_data(source: server account, ...)` reserved for future use
   * signature by the **Server Account**
+  * memo: derived from the `memo_type` and `memo` value passed by the **Client** if specified, omitted otherwise
```
leighmcculloch, line 145: ❗ Can you move this up to after line 129 so that it's inclusion in this list is similar to how we group that data fields in transactions in the XDR and in SDKs?

ecosystem/sep-0010.md
```diff
@@ -199,6 +204,8 @@ Upon successful verification, **Server** responds with a session JWT, containing
 * `iat` (the time at which the JWT was issued [RFC7519, Section 4.1.6](https://tools.ietf.org/html/rfc7519#section-4.1.6)) — current timestamp (`1530644093`)
 * `exp` (the expiration time on or after which the JWT must not be accepted for processing, [RFC7519, Section 4.1.4](https://tools.ietf.org/html/rfc7519#section-4.1.4)) — a server can pick its own expiration period for the token (`1530730493`)
 * `client_domain` - (optional) a nonstandard JWT claim containing the client home domain, included if the challenge transaction contained a `client_domain` (see [Verifying Client Application Identity](#verifying-client-application-identity))
+* `memo_type` — (optional) the memo type provided in the challenge, one of `text`, `id` or `hash`. See the [Memos](#memos) section for more information.
```
leighmcculloch, line 207: ❓ Are there use cases where we need to separate authenticating for an account for all memos, vs an account with memo type none? This change doesn't currently support that. How should the memo_type being unset be interpreted?

ecosystem/sep-0010.md
```diff
@@ -199,6 +204,8 @@ Upon successful verification, **Server** responds with a session JWT, containing
...
+* `memo` — (optional) the memo provided in the challenge, always of type `string` (even when `memo_type` is equal to `id`) to support parsing the value in languages that don't support big numbers.
```
leighmcculloch, line 208: 💡 I would make the why a little clearer here, the issue is actually that the format of JWTs, JSON, doesn't support 64-bit integers, not just languages. But what you have is accurate, so I don't feel strongly about this.
```suggestion
* `memo` — (optional) the memo provided in the challenge, always of type `string` even when `memo_type` is equal to `id` (unsigned 64-bit integer) because JSON and some runtimes do not support 64-bit integers.
```

ecosystem/sep-0010.md
```diff
@@ -300,6 +307,12 @@ This setup allows the **Server** to verify that the challenge returned by the **
 
 **Servers** may chose which **Client Domains** to verify. If the **Client** requests verification of its domain but the **Server** has no use for verifying that domain, the **Server** should proceed as if the **Client** did not provide the domain in the request for the challenge transaction. If the **Server** attempts but is unable to fetch the `SIGNING_KEY` from the provided **Client Domain**, the **Server** should return a `400 Bad Request` HTTP status code.
 
+### Memos
+
+Stellar transaction memos are used for a variety of purposes in the ecosystem, but in the context of this standard memos attached to challenge transactions are typically used to distinguish users who share a single Stellar account. This is especially relevant for applications who custody user funds in shared Stellar accounts.
```
leighmcculloch, line 312: ❗ A couple suggestions here, that I feel a bit stronger about, but let me know your thoughts.

- We should be as clear as possible, even if verbose, about the separation between sessions that use memos.
- Mentioning custody seems unnecessary here.

```suggestion
Stellar transaction memos are used for a variety of purposes in the ecosystem, but in the context of this standard memos attached to challenge transactions distinguish sessions that should be entirely separate, scoped, and detached from each other. Two users who are authenticated with the same Stellar account but different memos should have the same level of separation as two users who are authenticated with different Stellar accounts. Typically the sessions are unique users who share a single Stellar account.
```

ecosystem/sep-0010.md
leighmcculloch, line 208: ❗ One issue we have with making the JWT's identity a sum of `sub`, `memo_type`, and `memo`, is that it is not consistent with [RFC7519 Section 4.1.2] which states that that the `sub` **MUST** uniquely identify the principal being identified.

>The "sub" (subject) claim identifies the principal that is the
>subject of the JWT.  The claims in a JWT are normally statements
>about the subject.  The subject value MUST either be scoped to be
>locally unique in the context of the issuer or be globally unique.
>The processing of this claim is generally application specific.

-- [RFC7519 Section 4.1.2]

[RFC7519 Section 4.1.2]: https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.2

In the past I think we discussed other designs with we try to encode the memo and memo type into the `sub`, but I don't remember where we landed on that. The `sub` claim must be a string or a URI so unfortunately we can't use a JSON object. We could reconsider a `sub` value in the form `<account>,<memotype>,<memo>`, or similar.

ecosystem/sep-0012.md
```diff
@@ -49,10 +49,15 @@ Clients should submit the JWT previously obtained from the anchor via the [SEP-1
 Authorization: Bearer <JWT>
 ```
 
-Alternatively, if the client cannot add the authorization header. The JWT should be passed as a jwt query parameter:
-```
-?jwt=<token>
-```
+### Shared Accounts
+
+Client applications can use a single Stellar account to hold multiple customers' funds. To differentiate customers using the same Stellar account, the decoded SEP-10 JWT may contain `memo` and `memo_type` attributes. The anchor should use these attributes in addition to the Stellar account in their data model to represent unique customers. See the [SEP-10 Memos](sep-0010.md#memos) section for more information.
```
leighmcculloch, line 54: 💡 
```suggestion
Client applications can use a single Stellar account to hold multiple customers' funds. To differentiate customers using the same Stellar account, the decoded SEP-10 JWT may contain `memo` and `memo_type` attributes. The anchor should use these attributes in addition to the Stellar account in their data model to identify customers. See the [SEP-10 Memos](sep-0010.md#memos) section for more information.
```

ecosystem/sep-0012.md
```diff
@@ -49,10 +49,15 @@ Clients should submit the JWT previously obtained from the anchor via the [SEP-1
...
+
+Note that this scenario is not applicable to [SEP-31](sep-0031.md)'s use of this protocol. This is because the Stellar account used to authenticate in this case identifies the [Sending Anchor](sep-0031.md#abstract) itself, not a particular customer, and therefore the scope of the authenticated session should include all customers registered and transactions initiated by the Sending Anchor.
+
+To facilitate this, SEP-31 Sending Anchors should use the `memo` and `memo_type` request parameters described for the endpoints below to represent unique customers instead of using SEP-10 memos. 
+
+SEP-12 implementations should expect SEP-10 memos, if present in the decoded JWT, to match the memo request parameters and should return an error response if they do not.
```
leighmcculloch, line 60: ❗ I don't think we should say that rules apply to some SEPs and not others. SEP-12 needs to stand on its own with a single set of rules that can be applied uniformly without conditions on the SEP it is being used with, otherwise I think it is confusing to implement. I think we can tweak the language below to get the same result without presenting it as SEP-31 gets an exception to a rule.

```suggestion
SEP-12 implementations should expect memo claims in the SEP-10 JWT to match the memo request parameters and should return an error response if they do not. If the JWT does not contain memo claims then the memo request parameters may contain any value. This behavior allows shared account owners, such as [SEP-31 sending anchors](sep-0031.md#abstract), to submit user information about their users using the memos assigned to their users.
```

Thoughts?

---

Note that this section appears to answer my earlier question of what it means if no memo is provided: authentication with a parent account should provide access to all data of sub-accounts identified by memos. It may be worth calling this out explicitly in SEP-10, or at least specifying in SEP-10 that it is undefined behavior and clients cannot rely on the outcome in SEPs that use SEP-10.

leighmcculloch (commented):
👏🏻 I really like that this change was across multiple SEPs in a single PR. The broad nature of this change is easier realized applying it holistically. I reviewed SEP-6, 10, and 12. I did not review SEP-24 thoroughly given its similarity but I'll review it on my final pass. I left suggestions (💡) that I don't feel strongly about, questions (❓) that may result in changes but also may not, and concerns (❗) that I'd like to discuss further.

ecosystem/sep-0010.md
JakeUrban: Changing the format of the `sub` string means that users of the `sub` field need to update their parsing logic. In most cases, the only service using the decoded contents of the JWT is the same service that issued the JWT, but I can imagine cases where a service or client has not updated to parse the new format.

Personally I would rather deviate from the RFC spec and the `sub` format unchanged but I don't feel strongly about it.

One question I have about this change is, how would non-shared accounts be represented? Would there be separators in the `sub` string still? I would vote that if we make this change, the separators should always be present:
```
# with "sub" --> "G...,," or "G...,123,id"
>>> account, memo, memo_type = jwt.get("sub").split(",")
```

Otherwise, we'll have to implement logic like so:
```
# with "sub" --> "G..." or "G...,123,id"
>>> account, memo, memo_type = None, None, None
>>> if "," in jwt.get("sub"):
>>>     account, memo, memo_type = jwt.get("sub")
>>> else:
>>>     account = jwt.get("sub")
```

leighmcculloch: > I can imagine cases where a service or client has not updated to parse the new format

Even with the existing proposal services that aren't updated are likely to contain bugs, since the only way to accurately identify the user is to consume the `sub`, `memo_type`, and `memo` claims. In some ways changing the subject to ensure that the full identity of the user is captured within it may be more resilient to a more catastrophic bug.

---

To answer the question about how this would be implemented, I would take the approach of assuming they could always be present, but ignoring them if they are not:

```go
var account, memo_type, memo string
subParts := strings.SplitN(sub, ",", 3)
account = subParts[0]
if len(subParts) > 1 {
    if len(subParts) < 3 {
        return fmt.Errorf("invalid number of components to jwt subject: %q", sub)
    }
    memo_type = subParts[1]
    memo = subParts[2]
}
```

ecosystem/sep-0012.md
JakeUrban: > authentication with a parent account should provide access to all data of sub-accounts identified by memos.

I have concerns about this approach, both from a security and application-logic perspective.

If a client application authenticates with an account that was previously used with memos, the session should be treated as disjoint from other sessions using memos with the same account.

This ensures the service never exposes transaction or KYC information that was not explicitly requested by the client using the composite primary key of (account, memo, memo_type).

This simplified application logic as well. Services can always use the three claims to retrieve user and transaction records instead of sometimes using memos and sometimes not.
```
# .get() either returns the value or None

def create_user(jwt):
    User.objects.create(account=jwt.get("account"), memo=jwt.get("memo"), memo_type=jwt.get("memo_type"))

def get_user(jwt):
    User.objects.get(account=jwt.get("account"), memo=jwt.get("memo"), memo_type=jwt.get("memo_type"))
```

JakeUrban: I agree with the text in the suggested change though, so I'll commit it but leave this thread unresolved.

ecosystem/sep-0010.md
JakeUrban: Thats a good point about un-updated services, and the code above is certainly more resilient. 👍 

I'll update the PR accordingly. Just to make sure: this format should only be used in the `sub` field of the JWT right? Request parameters remain unchanged?

ecosystem/sep-0012.md
leighmcculloch: > I have concerns about this approach, both from a security and application-logic perspective.
> 
> If a client application authenticates with an account that was previously used with memos, the session should be treated as disjoint from other sessions using memos with the same account.

This is my preference and I share the same concern, but it sounds like the way it works with SEP-31 is that it isn't disjoint. I think we should endeavor to have our SEPs use SEP-10's scoping in the same way. How the scoping occurs is really out of scope of SEP-10 itself, but some consistency would have value.

Maybe we say in SEP-10 that it is not recommended, but then SEP-31/12 does it anyway?

ecosystem/sep-0010.md
leighmcculloch: Another way to interpret the RFC is to say that the principal is the shared account holder, who is requesting authentication for one of their users. It's odd because from an applications point-of-view the subject is usually the user, although from a key management perspective it is somewhat accurate because the key holder is the account holder.

leighmcculloch: > I'll update the PR accordingly. Just to make sure: this format should only be used in the `sub` field of the JWT right? Request parameters remain unchanged?

Yes request parameters can remain unchanged.

I wasn't really meaning to propose this format, but rather highlight maybe we could do something like this. We should maybe consider if there are better formats, although comma-separated is simple to implement.

Another option is to encode the memo into the muxed account strkey format, but that will make it not possible to distinguish muxed accounts from memo accounts in these setups which might not be great.

ecosystem/sep-0012.md
JakeUrban: I'm not sure I understand how SEP-31's use of SEP-10 does not make sessions using the same account disjoint. In SEP-31's use of SEP-10, Sending Anchors will never request memos to be in their challenge transactions and downstream JWTs, so the Sending Anchor's session will always be disjoint from transactions and KYC data created in sessions authenticated by JWTs including memos. Let me know if you think I'm missing something.

ecosystem/sep-0010.md
JakeUrban: Thats also true, although I think `sub` needs fully represent the scope of the session based on the specification's definition, and the scope would not be the same for two JWT's that use the same `sub` value but have different `memo` and `memo_type` parameters. 

> The claims in a JWT are normally statements about the subject.

You also have to squint hard to think that the `memo` and `memo_type` claims are _about_ the shared account holder, since they're really changing the identity of the subject, since (account, null, null) and (account, 123, text) describe two distinct entities from the anchor's perspective.

ecosystem/sep-0006.md
```diff
@@ -59,7 +59,11 @@ The JWT should be included in all requests as request header:
 Authorization: Bearer <JWT>
 ```
 
-It is highly recommended to require authentication for all endpoints other than `/info` (and optionally `/fee`).
+It is highly recommended to require authentication for all endpoints other than `/info` (and optionally `/fee`). Some functionality, including the [Shared Accounts](#shared-accounts) functionality described in the section below, may only be possible through the use of SEP-10.
+
+### Shared Accounts
+
+Client applications can use a single Stellar account to hold multiple users' funds. To differentiate users using the same Stellar account, the decoded SEP-10 JWT may contain `memo` and `memo_type` attributes. The anchor should use these attributes in addition to the Stellar account in their data model to represent unique users. See the [SEP-10 Memos](sep-0010.md#memos) section for more information.
```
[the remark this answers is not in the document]
JakeUrban: Ah.. I hadn't thought about the transition from this system to muxed accounts.

The problem I see with taking this approach is the transition for users of shared accounts that have already been registered with anchor services using other memo types via request parameters. 

In SEP-31, memos will continue to be specified as request parameters, so we don't have the issue there, assuming sending anchors will never use muxed accounts to identify their customers, and I don't see why they would.

But for SEP-6 and SEP-24 clients and anchors, they're using text and hash memo types if the account is shared, and would be expected to use those memos in the JWTs with this change. I think the question is, how many users are using shared accounts with anchor services today? And even if the number is small, are we willing to cause a compatibility issue for them in order to more easily migrate to a new system in the future?

Curious on your thoughts here.

ecosystem/sep-0006.md
```diff
@@ -59,7 +59,11 @@ The JWT should be included in all requests as request header:
 Authorization: Bearer <JWT>
 ```
 
-It is highly recommended to require authentication for all endpoints other than `/info` (and optionally `/fee`).
+It is highly recommended to require authentication for all endpoints other than `/info` (and optionally `/fee`). Some functionality, including the [Shared Accounts](#shared-accounts) functionality described in the section below, may only be possible through the use of SEP-10.
+
+### Shared Accounts
```
tomerweller, line 64: We've been referring to these type of accounts as "Pooled accounts". Some partners prefer the term "Ombinbus Accounts". Is there any specific reason you went with "Shared"?

ecosystem/sep-0006.md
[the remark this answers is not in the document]
tomerweller: Beyond muxed accounts, I think that removing `memo_type` will simplify these changes quite a bit

> how many users are using shared accounts with anchor services today?

I might not understand this question but how can there be users using pooled accounts today without support for that in the seps?

ecosystem/sep-0006.md
JakeUrban: "Shared" is a non-technical and accurate description of what the accounts are. "Pooled" is common in the engineering world (thread pools, worker pool, etc.) but the term isn't commonly used outside it. "Omnibus" is common in the financial world but I never heard that term until I got on calls with payment services.

I'm open to discuss the naming, but I think "shared" is going to communicate what it means to the most about of people.

ecosystem/sep-0006.md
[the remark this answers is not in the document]
JakeUrban: The changes in this PR specify how memos can be used to narrow the scope of authenticated sessions for users of a shared account, but using shared accounts have been possible with our SEPs for some time, although there are some semantic issues that this PR resolves.

For example, a service using shared accounts for their users could integrate with an anchor for withdrawals of their stellar funds simply by specifying the `memo` and `memo_type` request parameters in the `/withdraw` request. These parameters were [defined](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0024.md#request-1) for this exact purpose:

The definition for memo:

> (optional) A wallet will send this to uniquely identify a user if the wallet has multiple users sharing one Stellar account. The anchor can use this along with account to look up the user's KYC info.

So the capability existed. But it may make sense to make the breaking change now when the user base is small compared to allowing different memo types now and then requiring anchor services and client applications to migrate to id-based memos when muxed accounts come around.

yuriescl:
If I understood correctly, these are the issues being discussed regarding the addition of `memo` and `memo_type` to SEP-10:
1. JWT `sub` claim requires uniqueness, and the account pubkey alone would not provide that
2. SEP-12 `memo` and `memo_type` parameters already refer to shared accounts, not to transaction memos

Sometimes I feel that SEPs need some kind of versioning and dependency setup. The SEPs have been through changes and they depend on each other, it can be difficult to always be backward compatible and want changes at the same time.
The memo changes would affect SEP-6, SEP-10, SEP-12, SEP-24 and (maybe?) SEP-31. There has to be some kind of dependency and versioning to allow Anchors and libraries to upgrade smoothly.
Even a basic system like this would be very useful:
- Let's say the changes are applied and we version the SEPs like this: SEP-10 v3, SEP-24 v2 and SEP-12 v1
- SEP-24 v2 would require SEP-10 v3 and be incompatible with SEP-10 v2
- SEP-12 v1 would require SEP-10 v2 and be incompatible with SEP-10 v3
- Etc.

Polaris and other SEP-dependant software could then be compatible to a set of versions, instead of trying to have all versions compatible with all SEP versions, which restricts space for non-breaking real-time improvement.

About the numbered points above, this is what I suggest:
1. Add a `version` claim to the JWT indicating the SEP-10 version. Any version prior to the memo addition will not contain the `version` claim and this means the `sub` claim will contain only the pubkey. Any version starting from the memo addition can have the memo value added to the `sub` claim (formats are to be discussed - maybe `sub: {"account": "G...","memo": "SOME_MEMO","memo_type":"SOME_MEMO_TYPE"}`? - I don't see a problem with that since everything could be in JSON format).
2. With a version system it's simple to avoid conflicts.

SEPs have API-like behavior. I've yet to see a complex API system without versioning work well.

ecosystem/sep-0006.md
leighmcculloch: Omnibus accounts is the term closest to the term used in the real world both in securities for trading accounts that hold others funds, and in payments by payment facilitiators who are holding funds for merchants who don't have their own MIDs and accounts, so that term is pretty spot on really.

If I had to choose between shared or pooled, shared would be preferred. Pooled gives me _mixed_ feelings.

I think I like Omnibus more though. It aligns with the real world?

ecosystem/sep-0010.md
leighmcculloch: > > The claims in a JWT are normally statements about the subject.
> 
> You also have to squint hard to think that the `memo` and `memo_type` claims are _about_ the shared account holder, since they're really changing the identity of the subject

Claims should be attributes of the subject that have been proven. In the context of SEP-10 the memo isn't an attribute about the account, it's a modifier that changes the scope of authorization. This compels me to think embedding within the sub is more important than I had first thought.

JakeUrban: I agree. We'll update the format 👍

ecosystem/sep-0006.md
JakeUrban: I'm not convinced using "omnibus" as the primary term is the right approach. If you look up definitions, the term implies a set of use cases for an account, whereas the term "shared" describes the account's nature, not necessarily what it's used for.

For example, the [investopedia](https://www.investopedia.com/terms/o/omnibusaccount.asp) definition focuses very much on the privacy of entities whose funds are within the account. This is not a goal of changes we're making here.

I would rather avoid conflating the two subject areas, but we could mention the similarities in the `Shared Accounts` section?

yuriescl: Shared is simple and easy to understand, I upvote for Shared.

leighmcculloch: The description on investopedia is very use case specific, but in my experience the term is used broadly to refer to an account that is holding funds for multiple people where the funds are segregated.

Interestingly, in MasterCard documentation they refer to the account holding shared funds an "omnibus, segregated trust, or **pooled account**." It would seem the term pooled account is used in literature and not uncommon as I thought.

One thought about shared is that the term is so generic it could be confused with use cases that are not compatible with it. For example, with a joint or multi-person account where funds are not segregated within the account, we would say that is not a "shared account" which doesn't make sense, because it is. The term is so generic it also may not be easy to search for. For example, if an implementer is trying to find best practices about accounts used to pool multiple people's accounts, searching "stellar shared accounts" may not be fruitful.

I'd lean towards omnibus because it is a well understood term in financial products, it is unambiguous as to the meaning, it is a unique term easy to search for. I think shared and pooled are also okay. Pooled is maybe a little more specific than shared is.

leighmcculloch:
@yuriescl Thanks for the commenting on versioning. There is definitely room for improvement and it's a long time coming I think. Let's continue discussion about that in #1046.

ecosystem/sep-0010.md
JakeUrban: Now that we're using only memo IDs, there is no need for `memo_type`.

ecosystem/sep-0006.md
```diff
@@ -898,6 +902,8 @@ For example:
 
 The transaction history endpoint helps anchors enable a better experience for users using an external wallet. With it, wallets can display the status of deposits and withdrawals while they process and a history of past transactions with the anchor. It's only for transactions that are deposits to or withdrawals from the anchor.
 
+If the decoded JWT also contains the `memo` and `memo_type` parameters, the anchor must only return transactions for the user identified by the `memo`, _not all transactions for the Stellar account_.
```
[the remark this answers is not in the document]
JakeUrban: This is a good question, I had thought about the same. I agree they should absolutely be considered separate, although I don't think this case should even be allowed.

If an anchor has a user registered to a Stellar account without a memo, it shouldn't allow authenticated sessions for that same account if a memo is included. Same for the opposite scenario. If a user is registered for an account and has a memo, authenticated requests with just the account should also be rejected. This policy matches the reality of the account -- its either shared or its not.

And regarding your thoughts on "none" memos, I agree but I think its an implementation detail. When Polaris adds support for these changes, we'll always query for transactions/users/accounts using a memo condition. If a memo isn't present in the JWT then the condition will be `memo=None`.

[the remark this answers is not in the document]
leighmcculloch: Makes sense. Maybe what we need here is not new requirements for these situations, but a brief discussion about this in the security section of the SEP. What do you think?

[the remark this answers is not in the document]
JakeUrban: Good idea, I'll add it 👍

ecosystem/sep-0010.md
JakeUrban: I've updated the `sub` format to be `G...:1234567` in the case of a shared account.

ecosystem/sep-0006.md
JakeUrban: What if we tool a similar approach to MasterCard and listed applicable terms? I think only using the term omnibus is going to cause some folks to not realize that section is relevant to them.

```
## Shared, Omnibus, or Pooled Accounts

...
```

leighmcculloch: Not a bad idea. I like. This could be a good way to ensure we zero folks in who are coming from different backgrounds. We can then still pick one term to refer to them in that section.

ecosystem/sep-0006.md
```diff
@@ -59,11 +59,19 @@ The JWT should be included in all requests as request header:
 Authorization: Bearer <JWT>
 ```
 
-It is highly recommended to require authentication for all endpoints other than `/info` (and optionally `/fee`).
+It is highly recommended to require authentication for all endpoints other than `/info` (and optionally `/fee`). Some functionality, including the [Shared Accounts](#shared-omnibus-or-pooled-accounts) functionality described in the section below, may only be possible through the use of SEP-10.
+
+### Shared, Omnibus, or Pooled Accounts
+
+Client applications can use a single Stellar account to hold multiple users' funds. To distinguish users that use the same Stellar account, the decoded SEP-10 JWT's `sub` value may contain a memo value after the Stellar account (`G...:2810101841641761712`) OR the `sub` value will be a [Muxed Account (`M...`)](../core/cap-0027.md). The anchor should use this `sub` attribute in their data model to identify unique users.
```
leighmcculloch, line 66: 💡 
```suggestion
Client applications can use a single Stellar account to hold multiple users' funds. To distinguish users that use the same Stellar account, the decoded SEP-10 JWT's `sub` value may contain a memo value after the Stellar account (`G...:2810101841641761712`) OR the `sub` value will be a [Muxed Account (`M...`)](../core/cap-0027.md). The anchor should use the `sub` attribute in their data model to identify unique users.
```

ecosystem/sep-0010.md
```diff
@@ -6,14 +6,16 @@ Title: Stellar Web Authentication
...
-Version 3.2.2
+Updated: 2021-08-10
+Version 3.3.0
 ```
 
 ## Simple Summary
 
 This SEP defines the standard way for clients such as wallets or exchanges to create authenticated web sessions on behalf of a user who holds a Stellar account. A wallet may want to authenticate with any web service which requires a Stellar account ownership verification, for example, to upload KYC information to an anchor in an authenticated way as described in [SEP-12](sep-0012.md).
 
+This SEP also supports authenticating users of shared, omnibus, or pooled Stellar accounts. Clients can use [memos](#memos) or [muxed accounts](../core/cap-0027.md) to distinguish users of these accounts.
```
leighmcculloch, line 17: 💡 
```suggestion
This SEP also supports authenticating users of shared, omnibus, or pooled Stellar accounts. Clients can use [memos](#memos) or [muxed accounts](../core/cap-0027.md) to distinguish users or sub-accounts of shared accounts.
```

ecosystem/sep-0010.md
```diff
@@ -23,7 +25,7 @@ It involves the following components:
 - A **Server**: a server providing the `WEB_AUTH_ENDPOINT` that implements the GET and POST operations discussed in this document. The server's domain may be the **Home Domain**, a sub-domain of the **Home Domain**, or a different domain.
     - The `SIGNING_KEY` from the **Home Domain** is the **Server Account**.
 - A **Client**: the software used by the holder of the account being authenticated by the **Server**.
-    - The account being authenticated is the **Client Account**.
+    - The account being authenticated is the **Client Account**. This can be a Stellar account (`G...`) or a muxed account (`M...`).
```
leighmcculloch, line 28: 💡 
```suggestion
    - The account being authenticated is the **Client Account**.
    - A Stellar account (`G...`) or a muxed account (`M...`).
```

ecosystem/sep-0010.md
```diff
@@ -106,7 +108,8 @@ Request Parameters:
 
 Name      | Type          | Description
 ----------|---------------|------------
-`account` | `G...` string | The stellar account that the wallet wishes to authenticate with the **Server** 
+`account` | `G...` string | The **Client Account**, which can be a stellar account (`G...`) or muxed account (`M...`) that the **Client** wishes to authenticate with the **Server**.
+`memo` | string | (optional) The memo of type `id` to attach to the challenge transaction. Only permitted if a muxed account (`M...`) is not used. `hash` and `text` memos are not supported. See the [Memo](#memos) section for details.
```
leighmcculloch, line 112: 💡 
```suggestion
`memo` | string | (optional) The memo to attach to the challenge transaction. Only permitted if a stellar account (`G...`) is used. The memo must be of type `id`. Other memo types are not supported. See the [Memo](#memos) section for details.
```

ecosystem/sep-0010.md
```diff
@@ -124,6 +127,7 @@ On success the endpoint must return `200 OK` HTTP status code and a JSON object
   * source account set to the **Server Account**
   * invalid sequence number (set to 0) so the transaction cannot be run on the Stellar network
   * time bounds: `{min: now(), max: now() + 900 }` (we recommend expiration of 15 minutes to give the **Client** time to sign transaction)
+  * memo: derived from the `memo` value passed by the **Client** if specified, omitted otherwise
```
leighmcculloch, line 130: 💡 Derived sounds like it is transformed or modified, however it is only copied.
```suggestion
  * memo: the `memo` value passed by the **Client** if specified, omitted otherwise
```

ecosystem/sep-0010.md
```diff
@@ -195,7 +199,10 @@ The verification process confirms that the **Client** controls the **Client Acco
 Upon successful verification, **Server** responds with a session JWT, containing the following claims:
 
 * `iss` (the principal that issued a token, [RFC7519, Section 4.1.1](https://tools.ietf.org/html/rfc7519#section-4.1.1)) — a [Uniform Resource Identifier (URI)] for the issuer (`https://example.com` or `https://example.com/G...`)
-* `sub` (the principal that is the subject of the JWT, [RFC7519, Section 4.1.2](https://tools.ietf.org/html/rfc7519#section-4.1.2)) — the public key of the authenticating Stellar account (`G...`)
+* `sub` (the principal that is the subject of the JWT, [RFC7519, Section 4.1.2](https://tools.ietf.org/html/rfc7519#section-4.1.2)) — there are three possible formats:
+  * If the **Client Account** is a muxed account, the `sub` value should also be the muxed account (`M...`).
+  * If a memo was attached to the challenge transaction, the `sub` should be the Stellar account appended with the memo, separated by a colon (`G...:17509749319012223907`).
+  * If neither of the above conditions are true, the `sub` value should be Stellar account (`G...`).
```
leighmcculloch, line 205: 💡 Some grouping of the formats might make this easier to pass and digest.
```suggestion
* `sub` (the principal that is the subject of the JWT, [RFC7519, Section 4.1.2](https://tools.ietf.org/html/rfc7519#section-4.1.2)) — there are several possible formats:
  * If the **Client Account** is a muxed account (`M...`), the `sub` value should be the muxed account (`M...`).
  * If the **Client Account** is a stellar account (`G...`):
     * And, a memo was attached to the challenge transaction, the `sub` should be the stellar account appended with the memo, separated by a colon (`G...:17509749319012223907`).
     * Otherwise, the `sub` value should be Stellar account (`G...`).
```

ecosystem/sep-0010.md
```diff
@@ -309,9 +309,11 @@ This setup allows the **Server** to verify that the challenge returned by the **
...
+The `memo` parameter supported in `GET <WEB_AUTH_ENDPOINT>` API calls can be used to communicate to the **Server** that the client intends to authenticate a Stellar account with an additional claim that the account is shared and that the user should be identified using the `account` and `memo` parameters, instead of only using `account`. The value of the `memo` passed will ultimately be added to the decoded JWT's `sub` field, separated from the account address by a colon (`:`).
+
+Note that conceptually, a Stellar account and memo (of type `id`) is equivalent to a muxed account (`M...`) defined by [CAP-0027](../core/cap-0027.md). If a SEP-10 implementation supports the use of memos to identify users of shared accounts, it is highly recommended to also support the muxed account address format. **Muxed accounts will become the primary (and eventually exclusive) method for identifying a user of a shared, omnibus, or pooled account.**
```
leighmcculloch, line 316: 💡 I think we should reserve bolding for entities in SEP-10. This sentence is important, but not necessarily more important than other text in the document. 
```suggestion
Note that conceptually, a Stellar account and memo (of type `id`) is equivalent to a muxed account (`M...`) defined by [CAP-0027](../core/cap-0027.md). If a SEP-10 implementation supports the use of memos to identify users of shared accounts, it is highly recommended to also support the muxed account address format. Muxed accounts will become the primary (and eventually exclusive) method for identifying a user of a shared, omnibus, or pooled account.
```

leighmcculloch, line 316: 💡 In a few places we link to CAP-27, but I think SEP-10 would benefit from a section about muxed accounts similar to the memos section. We could link to CAP-27 from within that, and probably SEP-23 too which defines the M address, but could describe inside the SEP what a muxed account means for SEP-10. Wdyt?

leighmcculloch (approved):
I think this looks good. I've left a few suggestions (💡), but they are only suggestions.

ecosystem/sep-0010.md
JakeUrban: Ah I forgot SEP-23 defined the M address format. I can swap links for CAP-27 for SEP-23 but I think we should probably combine the existing Memo section with the muxed account information since their use in SEP-10 is the same.

leighmcculloch: I think we should link to both. SEP-23 is no longer a proposal for M addresses but a general specification for strkeys.

accordeiro (commented):
This looks good to me. There are a few details that I'd like to get ironed out before fully approving (see my comments below), but I think overall this is the right direction and set of changes.

ecosystem/sep-0006.md
```diff
@@ -59,11 +59,19 @@ The JWT should be included in all requests as request header:
...
+
+### Shared, Omnibus, or Pooled Accounts
+
+Client applications can use a single Stellar account to hold multiple users' funds. To distinguish users that use the same Stellar account, the decoded SEP-10 JWT's `sub` value may contain a memo value after the Stellar account (`G...:2810101841641761712`) OR the `sub` value will be a [Muxed Account (`M...`)](../core/cap-0027.md). The anchor should use the `sub` attribute in their data model to identify unique users.
+
+This document will refer to these accounts as shared accounts. See the [SEP-10 Memos](sep-0010.md#memos) and [Muxed Accounts](sep-0010.md#muxed-accounts) sections for more information.
+
+Note that Stellar accounts are either shared or they are not. This means anchors should ensure that a Stellar account previously authenticated with a memo should not be authenticated later without a memo. Conversely, an account that was previously authenticated without a memo should not be later authenticated as a shared account.
```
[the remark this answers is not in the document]
JakeUrban: I think this can get a bit confusing. What does it mean if an account that was previously indicated as shared authenticates without a memo? I don't see a use case for allowing this case, even if we treat the account without a memo as a distinct identity.

If we do come across a use case, we can update the SEP to be more prescriptive, but I'd vote to leave it general for now.

ecosystem/sep-0012.md
```diff
@@ -83,23 +88,25 @@ This endpoint allows clients to:
...
-`memo` | string | (optional) a [properly formatted memo](https://developers.stellar.org/docs/glossary/transactions/#memo) that uniquely identifies a customer. This value is generated by the client making the request. This parameter and `memo_type` are identical to the `PUT` request parameters of the same name.
-`memo_type` | string | (optional) type of `memo`. One of `text`, `id` or `hash`. If `hash`, `memo` should be base64-encoded.
+`account` | `G...` or `M...` string | (**deprecated**, optional) This field should match the `sub` value of the decoded SEP-10 JWT.
+`memo` | string | (optional) the client-generated [memo](https://developers.stellar.org/docs/glossary/transactions/#memo) that uniquely identifies the customer. If a memo is present in the decoded SEP-10 JWT's `sub` value, it must match this parameter value. If a muxed account is used as the JWT's `sub` value, requests containing a memo should be rejected. See the [Shared Accounts](#shared-omnibus-or-pooled-accounts) section for more information.
```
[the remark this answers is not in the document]
JakeUrban: I think its better to reject it, rather than assume the client made a mistake. If a muxed account is authenticated and a memo parameter is sent in the request, conceptually the client trying to create a sub account of a sub account, which isn't permitted.

ecosystem/sep-0006.md
```diff
@@ -388,9 +396,9 @@ Name | Type | Description
...
-`memo` | string | (optional) A wallet will send this to uniquely identify a user if the wallet has multiple users sharing one Stellar account. The anchor can use this along with `account` to look up the user's KYC info.
-`memo_type` | string | (optional) Type of `memo`. One of `text`, `id` or `hash`.
+`account` | `G...` or `M...` string | (optional) The Stellar or muxed account the client will use as the source of the withdrawal payment to the anchor. If SEP-10 authentiation is not used, the anchor can use `account` to look up the user's KYC information. Note that the account specified in this request could differ from the account authenticated via SEP-10.
+`memo` | string | (optional) This field should only be used if SEP-10 authentication is not. It was originally intended to distinguish users of the same Stellar account. However if SEP-10 is supported, the anchor should use the `sub` value included in the decoded SEP-10 JWT instead. See the [Shared Account Authentication](#shared-omnibus-or-pooled-accounts) section for more information.
```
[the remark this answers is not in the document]
JakeUrban: The anchor should associate the account/memo/muxed account included in the `sub` field with the KYC information collected, but the stellar account used as the source of funds withdrawn can be different than the `sub` value.

If SEP-10 is not supported though, then the `account` parameter cannot be different than the source of funds withdrawn. This is more of an edge case since any service that collects KYC needs to implement SEP-10 to ensure the KYC data provided can only be modified by the initial provider. 

Ideally, we make a breaking change and require SEP-10 for SEP-6 implementations so we don't have to make this distinction and can remove the potential for confusion.

ecosystem/sep-0006.md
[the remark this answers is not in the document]
accordeiro: OK, sounds good to me.

ecosystem/sep-0006.md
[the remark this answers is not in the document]
accordeiro: Oh I see – I didn't realize this was mostly focused on the fact SEP-10 is not a requirement for SEP-6. So it seems like our options to simplify this are either:

a) as you suggested, we make SEP-10 a requirement for SEP-6 (which is a breaking change); or
b) we change the entire SEP to require that `account == source of withdrawn funds`, which is also breaking.

I'm not comfortable with making a breaking change at this point, so let's keep this line as-is.

ecosystem/sep-0012.md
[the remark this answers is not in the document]
accordeiro: I generally agree that we shouldn't accept conflicting information, but in the interest of following [Postel's law](https://devopedia.org/postel-s-law) I think we could be flexible in the case where information is equivalent, i.e. when `memo == id_used_to_generate_muxed_address`. Our SDKs allow us to easily compare one with the other. What do you think?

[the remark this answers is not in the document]
JakeUrban: I can get behind that, I'll update the definition 👍

ecosystem/sep-0010.md
```diff
@@ -24,6 +26,7 @@ It involves the following components:
     - The `SIGNING_KEY` from the **Home Domain** is the **Server Account**.
 - A **Client**: the software used by the holder of the account being authenticated by the **Server**.
     - The account being authenticated is the **Client Account**.
+    - A Stellar account (`G...`) or a muxed account (`M...`).
```
leighmcculloch, line 29: 💡 I think this few lines could do with a little tweaking. In the sub-points we discuss the account mostly, not the client, so I think the Client Account deserves it's own top-level point.

I notice we're also missing G+memo from the list of possible values of the Client Account. I think that's important because a G+memo is not authenticating the G alone.

```suggestion
- A **Client Account**: the account being authenticated.
    - A Stellar account (`G...`) or a muxed account (`M...`).
    - If a Stellar account (`G...`), may be accompanied by a memo to scope the authentication to a user or sub-account of the account.
- A **Client**: the software used by the holder of the **Client Account** being authenticated by the **Server**.
```

ecosystem/sep-0010.md
```diff
@@ -300,6 +309,16 @@ This setup allows the **Server** to verify that the challenge returned by the **
...
+
+The `memo` parameter supported in `GET <WEB_AUTH_ENDPOINT>` API calls is used to communicate to the **Server** that the client intends to authenticate a Stellar account with an additional claim that the account is shared and that the user should be identified using the `account` and `memo` parameters, instead of only using `account`. The value of the `memo` passed will ultimately be added to the decoded JWT's `sub` field, separated from the account address by a colon (`:`).
+
+#### Muxed Accounts
+
+Note that conceptually, a Stellar account and memo (of type `id`) is equivalent to a muxed account (`M...`) defined in the protocol by [CAP-0027](../core/cap-0027.md) and standardized for use in the ecosystem in [SEP-0023](sep-0023.md). If a SEP-10 implementation supports the use of memos to identify users of shared accounts, it is highly recommended to also support the muxed account address format. Muxed accounts will become the primary method for identifying a user of a shared, omnibus, or pooled account.
```
leighmcculloch, line 320: 🔠
```suggestion
Conceptually, a Stellar account and memo (of type `id`) is equivalent to a muxed account (`M...`) defined in the protocol by [CAP-0027](../core/cap-0027.md) and standardized for use in the ecosystem in [SEP-0023](sep-0023.md). If a SEP-10 implementation supports the use of memos to identify users of shared accounts, it is highly recommended to also support the muxed account address format. Muxed accounts will become the primary method for identifying a user of a shared, omnibus, or pooled account.
```

leighmcculloch (approved):
I reviewed the changes since my last review when I approved, and I re-reviewed all the changes to SEP-10.

I've left one suggestion (💡) and one word change (🔠) but both are minor details.

This is great work and I think we've landed in the best place we can given the legacy constraints.

accordeiro (dismissed):
Looks great from my perspective, thank you @JakeUrban!

leighmcculloch (approved):
🎉

PR #1037, opened 2021-08-11
CAP-38: Clarify what assetA < assetB means

jonjove:
There has been a lot of confusion around what `assetA < assetB` means, so clearly CAP-38 was not communicating this clearly enough. I have sufficiently expanded the explanation and given it a prominent place in the semantics section.

### DIFF
--- a/core/cap-0038.md
+++ b/core/cap-0038.md
@@ -717,7 +717,11 @@ bool lessThan<T>(u, v)
         // less than v
         return false
     else if isContainer(T)
-        for i in 0...length(T)
+        // Handle variable length correctly
+        if length(u) != length(v)
+            return length(u) < length(v)
+
+        for i in 0...length(u)
             if lessThan(u[i], v[i])
                 return true
             else if lessThan(v[i], u[i])
--- a/core/cap-0038.md
+++ b/core/cap-0038.md
@@ -686,15 +686,68 @@ index 75f39eb4..f590ad4b 100644
 
 ### Semantics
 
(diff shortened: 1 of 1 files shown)


core/cap-0038.md
```diff
@@ -686,15 +686,68 @@ index 75f39eb4..f590ad4b 100644
...
+            if lessThan(u[i], v[i])
+                return true
+            else if lessThan(v[i], u[i])
+                return false
+            // This element was equal so compare the next one
+        // If we reached this point, all elements were equal so clearly u is not
+        // less than v
+        return false
+    else // T is an integer type
+        return u < v
```
paulbellamy, line 730: what?

fwiw, I implemented this in go as roughly:
```go
func aLessThanB(a, b int) bool {
	if a.Type < b.Type {
		return true
	}

	if a.Code < b.Code {
		return true
	}

	if a.Issuer < b.Issuer {
		return true
	}

	return false
}
```

jonjove: I'll try to clarify this.

I know you described the above as "roughly" so it might not apply to your real implementation, but there is an error in this rough implementation. This doesn't correctly handle the case where `a.Type > b.Type` (and likewise for `a.Code > b.Code`). The proceeding compared value must be equal before considering the next one.

jonjove: The definition in the CAP is going to have more complexity than that simple implementation, because I want it to continue to be accurate even if `Asset` changes.

PR #1038, opened 2021-08-12
Dm/cap21v2

stanford-scs:
Address issues raised at the protocol working group.

### DIFF
--- a/core/cap-0021.md
+++ b/core/cap-0021.md
@@ -57,39 +57,32 @@ transaction to execute when the `sourceAccount` is within some range.
 
 ## Specification
 
-`AccountEntry`'s `ext` field is extended to keep track of `seqLedger`
-and `seqTime`--the ledger number and time at which the sequence number
-was set to its present value.  These values are updated for the
-`sourceAccount` of every executed transaction, and also for the
-`sourceAccount` of every successfully executed `BumpSequenceOp`
-operation (regardless of whether the `BumpSequenceOp` actually
-increased the sequence number).
+`AccountEntryExtensionV2`'s `ext` field is extended to keep track of
+`seqLedger` and `seqTime`--the ledger number and time at which the
+sequence number was set to its present value.  These values are
+updated for the `sourceAccount` of every executed transaction, and
+also for the `sourceAccount` of every successfully executed
+`BumpSequenceOp` operation (regardless of whether the `BumpSequenceOp`
+actually increased the sequence number).
 
-If an account has an `AccountEntryExtensionV1` instead of an
-`AccountEntryExtensionV3`, because it hasn't been upgraded yet, then
-it behaves as if `seqLedger` and `seqTime` are both 0.
-
-Note that unlike the `v1` and `v2` extensions, we do not add an extra
-union at the end, because further changes can be accomplished by
-re-versioning `ext`.  To minimize source-code changes required in the
-face of such future extensions, the version 3 arm is called `cur`, and
-can be renamed to `v3` when version 4 is added.  At that point a
-single function can be used to translate `v3` to the new `cur`, rather
-than smearing the updated logic throughout the code.
-
-[Note: This point about extension structure is orthogonal to the goals
-of this CAP, which can just as easily dangle AccountEntryExtensionV3
-off the end of AccountEntryExtensionV2 if the protocol working group
-so decides.]
+If an account does not have an `AccountEntryExtensionV3` because it
+hasn't been upgraded yet, then it behaves as if `seqLedger` and
+`seqTime` are both 0.
 
 ```c++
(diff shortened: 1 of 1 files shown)


core/cap-0021.md
```diff
@@ -286,27 +261,41 @@ T2 have the same sequence number but T1 has twice the
...
+fields.
+
+Once a block is externalized by the consensus algorithm, the block is
+applied.  Before executing any operations, fees are charged to the
+source account for all transactions.  Once fees have been deducted
+from all accounts, transactions are one-by-one validated a second time
+then executed.  It is possible for a previously valid transaction to
+fail the second validation, for instance if a `BUMP_SEQUENCE`
+operation made the sequence number invalid.  Whenever a transaction
+fails validation during execution, the `sourceAccount` loses the fee.
```
leighmcculloch, line 284: This section states that transactions can continue to fail during execution if the preconditions fail. I thought you were going to change it so that they couldn't fail during execution? I think this is a problem for payment channels otherwise, but I need to think about it some more.

core/cap-0021.md
```diff
@@ -286,27 +261,41 @@ T2 have the same sequence number but T1 has twice the
...
+
+Because `TxConditionsV2` specifies multiple pre-conditions, there may
+be multiple reasons why a transaction is invalid.  If any of the
+reasons is permanent---for instance the `maxTime` or `maxLedger` has
+been exceeded--then the transaction is rejected with
 `TransactionResultCode` `txTOO_LATE`.  Otherwise, if all the
 unsatisfied preconditions are based on values that could later be
 valid, then the transaction is rejected with `txTOO_EARLY`.
```
leighmcculloch, line 292: It is worth noting that the only precondition that causes too late is the ledger bounds or time bounds preconditions. The vast majority will cause too early. The extra signers will cause bad auth. We should probably specify here that extra signers will cause bad auth not too early/late.

leighmcculloch (commented):
I left a couple comments inline.

core/cap-0021.md
leighmcculloch: This might not be a problem. After more thought I couldn't find a sequence of txs in the channel design that would be a problem.

core/cap-0021.md
```diff
@@ -286,27 +264,45 @@ T2 have the same sequence number but T1 has twice the
...
+be multiple reasons why a transaction is invalid.  If any of the
+reasons is permanent---namely the `maxTime` or `maxLedger` has been
+exceeded--then the transaction is rejected with
 `TransactionResultCode` `txTOO_LATE`.  Otherwise, if all the
 unsatisfied preconditions are based on values that could later be
-valid, then the transaction is rejected with `txTOO_EARLY`.
+valid, then the transaction is rejected with `txTOO_EARLY`.  If
+neither of the above conditions holds (`txTOO_EARLY` and `txTOO_LATE`
+do not apply), but on of the `extraSigners` is unsatisfied, then the
+transaction fails with `txBAD_AUTH`.
```
leighmcculloch, line 299: @MonsieurNicolas's [comment](https://groups.google.com/g/stellar-dev/c/N8vzP2Mi89U/m/ve1_VIi3AQAJ) on the mailing list about the `txTOO_EARLY` and `txTOO_LATE` caused me to notice that there's another result code we're missing. `minSeqNum` should be defined to cause `txBAD_SEQ`, not `txTOO_EARLY`, so we should mention that here too.

core/cap-0021.md
```diff
@@ -155,7 +137,7 @@ struct LedgerBounds {
     uint32 maxLedger;
 };
 
-struct GeneralPreconditions {
+struct TxConditionsV2 {
```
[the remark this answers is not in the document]
leighmcculloch: @stanford-scs I attempted to think of a better name as you requested. I think I'd still prefer and go with `PreconditionsV2`. Here's why:
- The `V2` in this types name comes from the fact that it is the `2` item in the `PreconditionsType`, and so case `2` in the `Preconditions` union. So, this type seems intentionally coupled to the `Preconditions` union.
- Assigning a name other than `PreconditionsV2` would indicate that the type plays some other role, when the only role it plays is to be a definition of the `2` case for the `Preconditions` union.
- We could define the type anonymously inside the `Preconditions` union and we'd lose nothing in semantics, which suggests again that its only role is to be a definition for the `2` case. I don't propose we make it anonymous.

leighmcculloch (commented):
Noticed one more thing about the result codes that is missing, and left a comment on the renaming of `GeneralPreconditions`.

leighmcculloch (approved):
I think the two things mentioned in my review above still need addressing, but otherwise this looks good. We should discuss the transaction failure issue in the protocol meeting today to see if we need to do anything differently there.

PR #1041, opened 2021-08-16
SEP-6, SEP-31: add explicit support for source and destination accounts different than authenticated accounts

### LINKED ISSUE #1039: SEP 6, 24, & 31: Allow source and destination account of funds to differ from the SEP-10 client account
SEP-10 enables clients (or users) to use the same credentials to authenticate with any number of independent services. Put simply, SEP-10 is the 'Login with Google' of the Stellar ecosystem. The difference is that instead of your Google username and password to login, you use a transaction containing your Stellar account signed by your secret key.

However, SEP-10 does not directly contribute to the security of funds held by or transferred to and from the authenticated Stellar account. Rather, SEP-10 authentication only ensures the KYC and transaction information provided to the anchor cannot be altered by a malicious actor.

Despite this, there is an implicit assumption in SEP-6, 24, & 31 that the account authenticated via SEP-10 is either the source or destination account of funds sent on Stellar. There is no reason that must be the case, since a user who initiated a transaction should be able to alter the source or destination of funds if they use the login credentials for the account that initiated the transaction.

We should update SEP-6, 24, and 31 to clarify that using a source account or specifying a destination account different from the account authenticated via SEP-10 is acceptable.

This change primarily benefits users of custodial services, who cannot sign challenge transactions with the secret key of their account because they do not have access to it. With this change, a straightforward solution would be to self-custody a Stellar account and secret key used in SEP-10 authentication, while using the Stellar account managed by a third party as the source and destination of outgoing and incoming transactions, respectively.

JakeUrban:
resolves #1039 

With the exception of adding a required `account` field to SEP-6's `deposit-exchange` endpoint, the changes here are largely patch-level changes that clarify the ability to send and receive funds to and from accounts that were not used in SEP-10 authentication.

Its possible that anchors assume deposits are always made to the account authenticated via SEP-10. Hopefully these anchors reject requests that specify `account` parameters that differ from the authenticated account. If they do not, anchors would mistakenly send funds to the account authenticated via SEP-10 instead of account provided in the deposit request. However, this should not be a major issue since the client has access to the authenticated account and can send the funds to the correct account.

### DIFF
--- a/ecosystem/sep-0006.md
+++ b/ecosystem/sep-0006.md
@@ -6,8 +6,8 @@ Title: Deposit and Withdrawal API
 Author: SDF
 Status: Active (Interactive components are deprecated in favor of SEP-24)
 Created: 2017-10-30
-Updated: 2021-08-09
-Version 3.6.0
+Updated: 2021-08-16
+Version 3.7.0
 ```
 
 ## Simple Summary
@@ -190,7 +190,7 @@ Request Parameters:
 Name | Type | Description
 -----|------|------------
 `asset_code` | string | The code of the on-chain asset the user wants to get from the Anchor after doing an off-chain deposit. The value passed must match one of the codes listed in the [/info](#info) response's deposit object.
-`account` | `G...` string | The stellar account ID of the user that wants to deposit. This is where the asset token will be sent.
+`account` | `G...` string | The stellar account ID of the user that wants to deposit. This is where the asset token will be sent. Note that the account specified in this request could differ from the account authenticated via SEP-10.
 `memo_type` | string | (optional) Type of memo that anchor should attach to the Stellar payment transaction, one of `text`, `id` or `hash`.
 `memo` | string | (optional) Value of memo to attach to transaction, for `hash` this should be base64-encoded.
 `email_address` | string | (optional) Email address of depositor. If desired, an anchor can use this to send email updates to the user about the deposit.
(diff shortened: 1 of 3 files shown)


ecosystem/sep-0006.md
```diff
@@ -902,7 +903,7 @@ Request parameters:
 Name | Type | Description
 -----|------|------------
 `asset_code` | string | The code of the asset of interest. E.g. BTC, ETH, USD, INR, etc.
-`account` | string | The stellar account ID involved in the transactions.
+`account` | string | The stellar account ID involved in the transactions. If the service requires SEP-10 authentication, this parameter match the authenticated account.
```
leighmcculloch, line 906: There's a word missing in this sentence, "this parameter match the authenticated account." Should it be "may not match" or "will match"?

ecosystem/sep-0006.md
```diff
@@ -190,7 +190,7 @@ Request Parameters:
 Name | Type | Description
 -----|------|------------
 `asset_code` | string | The code of the on-chain asset the user wants to get from the Anchor after doing an off-chain deposit. The value passed must match one of the codes listed in the [/info](#info) response's deposit object.
-`account` | `G...` string | The stellar account ID of the user that wants to deposit. This is where the asset token will be sent.
+`account` | `G...` string | The stellar account ID of the user that wants to deposit. This is where the asset token will be sent. Note that the account specified in this request could differ from the account authenticated via SEP-10.
```
leighmcculloch, line 193: Is there text already somewhere in the SEPs that explain why the accounts can be different?

leighmcculloch (commented):
I think this looks good. Two questions inline.

I also edited the title of the PR to specify the SEPs being changed.

ecosystem/sep-0006.md
JakeUrban: Ah thanks for pointing that out. It should say "must match". Clients should only be able to see records of transactions they initiated.

ecosystem/sep-0006.md
JakeUrban: No there isn't, but I could add a subsection under `Authentication`?

ecosystem/sep-0006.md
leighmcculloch: Makes sense. 👍🏻

ecosystem/sep-0006.md
leighmcculloch: I think it could be helpful, although maybe it is also obvious. I defer to you and will approve now and I can reapprove if you want to add a section.

leighmcculloch (dismissed):

JakeUrban:
@leighmcculloch added the subsection, can you take a look and re-approve if it looks good?

ecosystem/sep-0006.md
```diff
@@ -61,6 +61,12 @@ Authorization: Bearer <JWT>
 
 It is highly recommended to require authentication for all endpoints other than `/info` (and optionally `/fee`).
 
+### Source and Destination Accounts
+
+Note that both the source account of a withdrawal payment and the destination account of a deposit can be different than the account authenticated via SEP-10. 
+
+This is particularly useful for clients who use a third-party custody provider for their users' funds held on Stellar. Using this flexibility, a client can use a Stellar public and private key managed using their own infrastructure to authenticate with anchor services while using the Stellar account managed by a custodian as the source and destination of users' transactions.
```
leighmcculloch, line 68: I think we should drop this paragraph as it is a little too prescriptive and makes it sound like this is the only use case. I can think of a variety of cases where someone might have different authenticating and source/destination accounts. For example, an account holder might have multiple Stellar accounts and want to deposit and withdraw from any of them, but they authenticate with one account because they only need to have one account with the anchor.

leighmcculloch (commented):
I think this PR is good, except I think we should drop one paragraph in the new section. Let me know what you think.

ecosystem/sep-0006.md
JakeUrban: Thats a good use case, I hadn't thought of that. But I agree it is very prescriptive so I'll remove it.

leighmcculloch (approved):

PR #1043, opened 2021-08-17
SEP-6 & 24: Add `features` object to GET /info

### LINKED ISSUE #1030: SEP-6 & 24: Add `account_creation_supported` flag to `/info` endpoints

SEP-6 and 24 currently suggest that the anchor should fund nonexistent accounts requesting deposits, but this is a business decision the anchor must have the ability to enable or not. 

An `account_creation_supported` flag or similar added to the anchor's `GET /info` endpoint response. Wallet applications would be able to detect such a flag and notify the user if their account does not yet exist.

It may make sense to add a more generate `feature_flags` object to the `/info` endpoints to allow for additional feature flags in the future.

cc @yuriescl @tomerweller @meiyib

JakeUrban:
resolves #1030

Adds a `feature_flags` object to `GET /info`'s response body containing boolean values indicating whether or not specific features are support by the anchor. Currently the only two values and defaults are:

- `account_creation`: `true`
- `claimable_balances`: `false`

As ecosystem adoption for claimable balances increases, the feature flag's default may be updated to `true`.

This is helpful for clients who would adjust their behavior or UX based on the anchor's supported features. For example, a wallet could either require that a user create an account prior to requesting a deposit if the account didn't exist and `account_creation` was `false`, or require the user to create a trustline prior to requesting a deposit if `claimable_balances` was `false`.

### DIFF
--- a/ecosystem/sep-0006.md
+++ b/ecosystem/sep-0006.md
@@ -6,8 +6,8 @@ Title: Deposit and Withdrawal API
 Author: SDF
 Status: Active (Interactive components are deprecated in favor of SEP-24)
 Created: 2017-10-30
-Updated: 2021-08-16
-Version 3.7.0
+Updated: 2021-08-17
+Version 3.8.0
 ```
 
 ## Simple Summary
@@ -281,7 +281,9 @@ Mexican peso (MXN) response example:
 ### Special Cases
 #### Stellar account does not exist
 
-If the given Stellar `account` does not exist, on receipt of the deposit, the anchor should use the `CreateAccount` operation to create the account with at least enough XLM for the minimum reserve and a trust line to the requested asset (2.01 XLM is recommended). The Anchor can add this minimal funding amount to the service fee, but this requires calculating the worth of the minimum funding amount in units of the requested asset.
+If the given Stellar `account` does not exist, on receipt of the deposit, the anchor should use the `CreateAccount` operation to create the account with at least enough XLM for the minimum reserve and a trust line to the requested asset (2.01 XLM is recommended). To indicate that account creation is not supported, set the `account_creation` attribute within `GET /info`'s `feature_flags` object to `false`, otherwise clients will assume account creation is supported.
+
+The Anchor can add this minimal funding amount to the service fee, but this requires calculating the worth of the minimum funding amount in units of the requested asset.
 
 Since the anchor doesn't have the user account's secret key, the user must create a trust line to the anchor's asset before the anchor can send the requested asset tokens to the user's account. The anchor should listen for the user to establish this trust line. Once the trust line is there, the anchor should send the requested asset tokens to the account in Stellar to complete the deposit.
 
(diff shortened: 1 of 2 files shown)


ecosystem/sep-0006.md
```diff
@@ -793,6 +796,10 @@ The response should be a JSON object like:
   "transaction": {
     "enabled": false,
     "authentication_required": true
+  },
+  "feature_flags": {
```
leighmcculloch, line 800: We could also name this `features`.

ecosystem/sep-0006.md
```diff
@@ -844,6 +851,17 @@ An anchor should also indicate in the `/info` response if they support the `fee`
...
+#### Feature Flags
+
+The `feature_flags` object contains boolean values indicating whether or not specific features are supported by the anchor. If the object or specific feature is not present in the response, the default value described below may be assumed. This information enables wallets to adjust their behavior based on the feature set supported by the anchor.
+
+Name | Default | Description
+-----|---------|------------
+`account_creation` | `true` | Whether or not the anchor supports creating accounts for users requesting deposits.
+`claimable_balances` | `false` | Whether or not the anchor supports sending deposit funds as claimable balances. This is relevant for users of Stellar accounts without a trustline to the requested asset.
+
+The default values for the features listed above have been selected based on the ecosystem's current support. It is highly recommened to support all features enumerated for the best user experience.
```
leighmcculloch, line 863: I don't think we need to say it's recommended, which has the spec offer opinion, rather we can focus on highlighting how to get the best use experience. 

```suggestion
The default values for the features listed above have been selected based on the ecosystem's current support. Supporting all features provided the best user experience.
```

ecosystem/sep-0006.md
```diff
@@ -315,6 +317,7 @@ Using this feature, anchors will no longer have to wait until the user's Stellar
 
 **Anchors**: To support claimable balances anchors must
 
+* Set the `claimable_balances` attribute within `GET /info`'s `feature_flags` object to `true`
```
leighmcculloch, line 320: Is there any negative affect on the situation where anchors have already implemented claimable balances but not the feature flags object?

I'm wondering if a client could be surprised, thinking that claimable balances would not be used because the default is false. 

Would it work to say that a missing features flag object means the client doesn't know what they support?

leighmcculloch (approved):
Looks great. Some questions inline, but nothing really blocking.

ecosystem/sep-0006.md
JakeUrban: I think thats better 👍

ecosystem/sep-0006.md
JakeUrban: Today, all wallets have to assume anchors do not support claimable balances, so I don't think this will be a problem.

If a wallet _knows_ an anchor supports claimable balances despite not yet updating their `GET /info` response, it can leave it's implementation unchanged -- passing `claimable_balance_supported: true` in the `GET /deposit` request and waiting for the claimable balance to be created. Or, it could implement the `features` detection while also adding an allowlist of sorts simultaneously. Its not perfect but I think the process is manageable.

leighmcculloch (approved):
👍

PR #1050, opened 2021-09-07
Modify the wording of PATCH /transactions/:id description

lijamie98:
1. Changed "Receiver Anchor" to "Receiving Anchor" to make the document consistent.
2. The word "Receiving Anchor requests" sounds like the receiving anchor is polling information from the sending anchor. But it should be the sending anchor "patching" to receiving anchor. Change the wording to avoid the confusion.

### DIFF
--- a/ecosystem/sep-0031.md
+++ b/ecosystem/sep-0031.md
@@ -7,8 +7,8 @@ Title: Cross-Border Payments API
 Author: SDF
 Status: Active
 Created: 2020-04-07
-Updated: 2021-08-16
-Version 1.3.1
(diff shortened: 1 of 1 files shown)


JakeUrban (changes requested):
This looks good to me, can you update the `Updated At` and `Version` info at the top of the document? Changing language like this is considered a "patch" level increment.

lijamie98:
Updated accordingly.
1. Version changed to 1.3.2
2. Updated at 2021-09-07

JakeUrban (approved):

PR #1059, opened 2021-09-28
SEP-38: add status codes

JakeUrban:
We forgot to document the status codes that should be returned in success responses for SEP-38 requests.


leighmcculloch (approved):
Thanks for catching this and filling the gap. 👏🏻

PR #1062, opened 2021-10-01
SEP-6: clarify GET /fee endpoint usage for Stellar assets

### LINKED ISSUE #1061: SEP-6: fees for transactions using SEP-38 quotes
## Problem Description

SEP-6's `GET /fee` endpoint states the following:

> When using the `/deposit-exchange` or `/withdraw-exchange` endpoints fees are included in the conversion price provided with SEP-38 GET /price or SEP-38 POST /quote.

However, SEP-6's `GET /transaction(s)` endpoints include the `amount_fee` and optional `amount_fee_asset` attributes, implying that fixed fees can in fact be charged in addition to any margin included in the exchange rate between the assets used in the transaction.

## Possible Solutions

There are two ways we can resolve this contradiction in our protocol, each with their pros and cons.

### Solution 1: Remove `amount_fee_asset` from `GET /transactions(s)`

This is the simplest solution, although it technically is a breaking change. The statement made in the `GET /fee` endpoint description would remain as-is and the information contradicting that statement would be removed.

However, I don't think this is the right approach. Businesses should be able to decide how to charge fees for their services, and requiring a subset of transactions they facilitate to use a different fee mechanism imposes unnecessary restrictions for the sake of avoiding a protocol adjustment.

For example, lets say a cryptocurrency exchange charges flat and/or percentage fees for buying XLM with fiat USD as well as selling XLM for USD. Now lets say that exchange wants to make their services interoperable with the various Stellar ecosystem wallets applications. SEP-6 doesn't support fixed or percentage fees for transactions involving different assets, so the exchange is either out of luck or has to adjust their fee mechanisms for Stellar.

### Solution 2: Support fees for transactions using quotes

This solution allows businesses to charge fees in any manner they choose. However, our `GET /fee` endpoint, and the fee-related info in `GET /info` endpoint, currently assumes fees are charged in units of the Stellar asset. There are a few ways we can resolve this.

#### Option 1: Adjust the request and response attributes to `GET /fee` (and `GET /info`)

This is difficult to design. My first thought was we could do something like the following examples.

---------

```
GET /fee?asset_code=USDC&offchain_asset=iso4217:BRL&operation=deposit&amount=100&amount_asset=iso4217:BRL

{
    "fee": "2.00",
    "fee_asset": "stellar:USDC:GA5ZSEJYB37JRC5AVCIA5MOP4RHTM335X2KGX3IHOJAPP5RE34K4KZVN"
}
```

The request above means: I want to deposit 100 BRL for some unknown amount of USDC. The fee is 2 USDC.

```
GET /fee?asset_code=USDC&offchain_asset=iso4217:BRL&operation=withdrawal&amount=100&amount_asset=iso4217:BRL

{
    "fee": "5.00",
    "fee_asset": "iso4217:BRL"
}
```

The request above means: I want to withdraw some unknown amount of USDC and receive 100 BRL. The fee is 5 BRL.

--------

This obviously is not ideal because the request info is incomplete: it doesn't provide the amount of the other asset. However adding a parameter for this would effectively allow the client to determine the exchange rate for a hypothetical transfer, which is nonsensical. Its also not clear if the anchor would even be able to give accurate fees without the amount of the other asset.

An adjustment could be to let the anchor decide the amount of the other asset and include it in the response. However, this requires indicative quote logic in the `GET /fee` endpoint, similar to SEP-38's `GET /price(s)`. Do we really want to require anchors to do this?

#### Option 2: Deprecate the `GET /fee` endpoint

I'd like to go big here and question the real-world value of the `GET /fee` endpoint. The use case for it is as follows:

> This endpoint is important to allow an anchor to accurately report fees to a user even when the fee schedule is complex. If a fee can be fully expressed with the `fee_fixed` and `fee_percent` fields in the `/info` response, then an anchor should not implement this endpoint.

What does it mean for a fee to be "complex"? It means there are many factors that go into its calculation. However, the `GET /fee` endpoint doesn't actually allow for many input factors. It has the following parameters:

- amount
- operation (deposit or withdraw)
- asset_code (stellar asset)
- type (custom value defined in `GET /info`)

The `type` parameter is really the key parameter here, however because the values are custom (i.e. defined in each anchor's `GET /info` endpoint) its effectively not interoperable and therefore not very useable.

**Instead of continuing to try to standardize a method for calculating  fees with complex structures, I vote to leave this feature behind and recognize that it does a poor job at what it aims to do.**

JakeUrban:
resolves #1061

Adds a disclaimer to the `GET /fee` endpoint for a lack of multi-asset transaction support. The fee endpoint will only support reporting fees in terms of Stellar assets for amounts of the same Stellar asset.

Also corrects the description of `fee_fixed` and `fee_percent` for deposit assets. Fees cannot be expressed in terms of the off-chain asset in these cases because the fee values would vary per off-chain asset.

### DIFF
--- a/ecosystem/sep-0006.md
+++ b/ecosystem/sep-0006.md
@@ -43,7 +43,7 @@ To support this protocol an anchor acts as a server and implements the specified
 * [`GET /deposit-exchange`](#deposit-exchange): optional
 * [`GET /withdraw-exchange`](#withdraw-exchange): optional
 * [`GET /info`](#info): required
-* [`GET /fee`](#fee): **deprecated**, optional (only needed for complex fee structures)
+* [`GET /fee`](#fee-deprecated): **deprecated**, optional (only needed for complex fee structures)
 * [`GET /transactions`](#transaction-history): required
 * [`GET /transaction`](#single-historical-transaction): required
 
@@ -173,7 +173,7 @@ SEP-6 lays out many options for how deposit and withdrawal can work. These are r
 
 * Provide a full-featured implementation of [`/info`](#info).
 * Decide which endpoints, if any, need to be [authenticated](#authentication), and declare that properly in the `/info` endpoint.
-* Pick your approach to [fees](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0006.md#fee). We recommend using `/info` to express fees as it provides a better user experience (the user can see the fee structure in the wallet early in the process).
+* Pick your approach to [fees](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0006.md#fee-deprecated). We recommend using `/info` to express fees as it provides a better user experience (the user can see the fee structure in the wallet early in the process).
 * **For both deposit and withdrawal**:
     * Include the `id` field in your response so the wallet can check up on the status of the transaction if it wants.
     * We recommend you use [SEP-10 authentication](#authentication) for all endpoints other than `/info`. Anchors may want to allow clients to access the `/fee` endpoint as well.
--- a/ecosystem/sep-0006.md
+++ b/ecosystem/sep-0006.md
@@ -880,7 +880,7 @@ The fee endpoint allows an anchor to report the fee that would be charged when u
 
 This endpoint is important to allow an anchor to accurately report fees to a user even when the fee schedule is complex. If a fee can be fully expressed with the `fee_fixed` and `fee_percent` fields in the `/info` response, then an anchor should not implement this endpoint.
 
-This endpoint only reports fees expressed in units of Stellar assets. Fetching fee amounts for transactions using both on & off-chain assets (using either `/deposit-exchange` and `/withdraw-exchange`)is not supported unless fees are only dependent on the amount of the Stellar asset transacted.
+This endpoint only reports fees expressed in units of Stellar assets. Fetching fee amounts for transactions using both on & off-chain assets (using either `/deposit-exchange` and `/withdraw-exchange`) is not supported unless fees are only dependent on the amount of the Stellar asset transacted.
 
 ```
 GET TRANSFER_SERVER/fee
--- a/ecosystem/sep-0006.md
+++ b/ecosystem/sep-0006.md
@@ -6,8 +6,8 @@ Title: Deposit and Withdrawal API
 Author: SDF
 Status: Active (Interactive components are deprecated in favor of SEP-24)
 Created: 2017-10-30
-Updated: 2021-08-17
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0006.md
```diff
@@ -795,7 +789,8 @@ The response should be a JSON object like:
     }
   },
   "fee": {
-    "enabled": false
+    "enabled": false,
+    "description": "Fees vary based on the the assets transacted and method by which funds are delivered to or collected by the anchor."
```
leighmcculloch, line 793: I would expect to see some more detailed information, such as a % for fee, or something. Do you think that is realistic how people will use the description? I worry by using a non-informative example here we will encourage similar non-information descriptions.

JakeUrban: I can make the description better here but I think some descriptions will inevitably be like this if they are sufficiently complex.

leighmcculloch: I agree, that'll probably happen, but I think our example here should aim to contain the sorts of things we hope implementers will include.

leighmcculloch (approved):
Looks good, one suggestion, otherwise 👍🏻.

leighmcculloch (approved):

ecosystem/sep-0006.md
```diff
@@ -874,10 +878,10 @@ The default values for the features listed above have been selected based on the
 
 The fee endpoint allows an anchor to report the fee that would be charged when using the `/deposit` or `/withdraw` endpoints.
 
-When using the `/deposit-exchange` or `/withdraw-exchange` endpoints fees are included in the conversion price provided with [`SEP-38 GET /price`](sep-0038.md#get-price) or [`SEP-38 POST /quote`](sep-0038.md#post-quote).
-
 This endpoint is important to allow an anchor to accurately report fees to a user even when the fee schedule is complex. If a fee can be fully expressed with the `fee_fixed` and `fee_percent` fields in the `/info` response, then an anchor should not implement this endpoint.
 
+This endpoint only reports fees expressed in units of Stellar assets. Fetching fee amounts for transactions using both on & off-chain assets (using either `/deposit-exchange` and `/withdraw-exchange`)is not supported unless fees are only dependent on the amount of the Stellar asset transacted.
```
leighmcculloch, line 883:
```suggestion
This endpoint only reports fees expressed in units of Stellar assets. Fetching fee amounts for transactions using both on & off-chain assets (using either `/deposit-exchange` and `/withdraw-exchange`) is not supported unless fees are only dependent on the amount of the Stellar asset transacted.
```

leighmcculloch (approved):

PR #1064, opened 2021-10-05
SEP-24: Add support for transacting non-equivalent assets

JakeUrban:
Adds support for exchanging non-equivalent assets (i.e. using quotes) in SEP-24. Note SEP-38 is not required, unlike in SEP-6 and SEP-31, because all quote-related info can be communicated via the interactive flow.

Note the change made to the `GET /fee` endpoint description. This is a temporary change while the decision around deprecating the `GET /fee` endpoint altogether is finalized (see #1061).

### DIFF
--- a/ecosystem/sep-0024.md
+++ b/ecosystem/sep-0024.md
@@ -6,8 +6,8 @@ Title: Hosted Deposit and Withdrawal
 Author: SDF
 Status: Active
 Created: 2019-09-18
-Updated: 2021-08-17
-Version 1.4.0
+Updated: 2021-10-05
+Version 1.5.0
 ```
 
 ## Simple Summary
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0024.md
```diff
@@ -6,8 +6,8 @@ Title: Hosted Deposit and Withdrawal
 Author: SDF
 Status: Active
 Created: 2019-09-18
-Updated: 2021-08-17
-Version 1.4.0
+Updated: 2021-10-05
+Version 1.5.0
```
lijamie98, line 10: I am not sure if 1.4 -> 1.5 indicates breaking change? 

Since these fields are optional, they should not be breaking.

ecosystem/sep-0024.md
```diff
@@ -65,6 +65,14 @@ Note that Stellar accounts are either shared or they are not. This means anchors
 
 Note that both the source account of a withdrawal payment and the destination account of a deposit can be different than the account authenticated via SEP-10.
 
+## Asset Exchanges
+
+This protocol was originally designed to provide conversions between off-chain and on-chain equivalent assets – for instance `BRL <> BRLT`, `USD <> USDC`, `NGN <> NGNT`, etc. – but it has been updated to make exchanges between off-chain and on-chain non-equivalent assets like `BRL <> USDC`.
```
lijamie98, line 70: Making exchanges between off-chain and on-chain non-equivalent assets seem to overlap with SEP-38. I am not sure if we should update sep-24 or update both sep-24 and sep-38 together.

ecosystem/sep-0024.md
JakeUrban: Only major version increments are considered "breaking" 👍

leighmcculloch (approved):

ecosystem/sep-0024.md
JakeUrban: SEP-38 doesn't have any information on how the quotes returned from its endpoints are used in other protocols, so I don't think we need any adjustments there.

However, there are functionality gaps between SEP-6 and SEP-24 in terms of how they support non-equivalent assets. For example, you cannot provide a SEP-38 quote ID to SEP-24's endpoints like you can with SEP-6. Client applications also cannot fetch `GET /quote` information for the exchange of assets within a SEP-24 transaction, although they can infer most of it from the info returned in `GET /transaction(s)`.

All this is to say that we could further integrate SEP-38 into SEP-24, but the changes made here are the bare-bones needed to facilitate this use case. cc @leighmcculloch

lijamie98 (approved):

PR #1072, opened 2021-10-19
SEP-11: Add canonical string representation of the ASSET_TYPE_POOL_SHARE asset type

leighmcculloch:
### What
Add a string representation of the `ASSET_TYPE_POOL_SHARE` asset type in the form of `<LiquidityPoolID>:lp`, where LiquidityPoolID is the `PoolID` in lower case hex form, and `lp` is a hardcoded string identifying that the prior form is a pool share.

### Why
SEP-11 defines a canonical string form for all asset types for Protocol 17 and earlier, however Protocol 18 adds a new asset type and SEP-11 does not define how that asset type should be rendered as a string.

This is problematic for tools implementing the full breadth of SEP-11, such as stc, who rely on concise text formats for the XDR types. Tools of that type no longer have a concise string representation for all assets.

This is problematic for SDKs because SDKs used to be able to provide a canonical string representation for all asset types. In the SDKs this manifested as the ability to take any string and get a string representation, and then the ability to take that string representation and reform the asset. Since Protocol 18 was rolled out to SDKs this is no longer the case. 

Introducing a new canonical string representation is not straight-forward because asset strings do not contain any data that acts as an absolute discriminant.

The format proposed for the pool share exploits the fact that the existing asset types have gaps in values that would be otherwise invalid. The format proposed, where-by a fixed value is used post a colon and the ID is used pre the colon, was proposed by @stanford-scs. The form is backwards compatible with existing asset string representations since it never collides with the native asset type which is limited to 12 characters not including the colon, and it never collides with issued credit assets since they require a valid strkey encoded key as defined in SEP-23.

The suffix `:lp` was selected because `ASSET_TYPE_POOL_SHARE` assets are always liquidity pool IDs, and as far as we expect will always be. The string `lp` does not collide with issuer account IDs used for credit assets.

The version has been updated to v1.1.0 assuming that the prior version of the SEP was v1.0.0 even though it was not explicitly noted in the pragma.

### DIFF
--- a/ecosystem/sep-0011.md
+++ b/ecosystem/sep-0011.md
@@ -199,7 +199,7 @@ The `Asset`, and `TrustLineAsset` union types are rendered as a string in one of
 
   - `LiquidityPoolID:lp` for liquidity pool shares, i.e. assets of type `ASSET_TYPE_POOL_SHARE`.
 
-    "LiquidityPoolID" is the hex encoded `PoolID`.
+    "LiquidityPoolID" is the hex encoded `PoolID` using a lowercase alphabet.
 
     "lp" is the string `"lp"`.
 


ecosystem/sep-0011.md
```diff
@@ -172,11 +174,17 @@ A few aggregate values are special-cased:
 
 The `Asset` type is rendered as a string in one of two forms:
```
ire-and-curses, line 175:
```suggestion
The `Asset` type is rendered as a string in one of three forms:
```

ecosystem/sep-0011.md
```diff
@@ -172,11 +174,17 @@ A few aggregate values are special-cased:
...
   - `native` (or any string up to 12 characters not containing an unescaped colon) for the native asset.
-  - `Code:IssuerAccountID` for issued assets.
+  - `Code:IssuerAccountID` for issued assets, i.e. assets of type
+    `ASSET_TYPE_CREDIT_ALPHANUM4` and `ASSET_TYPE_CREDIT_ALPHANUM12`.
 
     "Code" must consist of printable ASCII characters (octets 0x21 through 0x7e).
 
     The sequence `\x` introduces a hex escape sequence, e.g., `\x00` to introduce a 0-valued byte.  Otherwise, `\` escapes the next character, so `\\` is required to introduce a backslash. Stellar disallows assets of type `ASSET_TYPE_CREDIT_ALPHANUM12` that have fewer than 5 bytes, but such assets can be represented in binary XDR, and so in txrep are rendered with trailing `\x00` (escaped NUL bytes) to as to make the length 5---e.g., the 12-byte asset code `ABC` is rendered `ABC\x00\x00`. Note that stellar-core disallows non ASCII bytes in `AssetCode` fields, so the primary use of this feature is to construct or examine invalid transaction test cases.
+  - `PoolID:POOL` for pool shares, i.e.  `ASSET_TYPE_POOL_SHARE`.
```
ire-and-curses, line 183:
```suggestion
  - `PoolID:POOL` for liquidity pool shares, i.e. assets of type `ASSET_TYPE_POOL_SHARE`.
```

leighmcculloch: That does beg the question, maybe the suffix should be `:POOLSHARE`. @jonjove Are there likely to be other forms of POOL assets?

leighmcculloch: > the XDR type is actually `PoolID` and the asset type is formally `POOL_SHARE`

Refs:
- https://github.com/stellar/stellar-core/blob/dc5f5a392098b82bd9453a2aa4259e7af600ad9d/src/xdr/Stellar-ledger-entries.x#L17
- https://github.com/stellar/stellar-core/blob/dc5f5a392098b82bd9453a2aa4259e7af600ad9d/src/xdr/Stellar-ledger-entries.x#L240-L241

ire-and-curses: I suggested "liquidity" because we had the exact same discussion for the Horizon API, and we concluded that being explicit here was better. Although the XDR naming is more general, it refers to liquidity pools.

leighmcculloch: >XDR naming is more general

I agree, the right thing to do in Horizon or applications is probably to be more explicit. Or possible we should have been more explicit in the XDR in the first place.

SEP-11 is a string representation of the XDR though, so I think it should mirror the XDR. SEP-11 should bring as little as possible to the XDR.

leighmcculloch: >But it's worth mentioning that typedefs don't correspond to different types just different names, so if we had LiquidityPoolID and OtherPoolID both map to the same type (Hash) then they would be compatible anyway.

The property you describe is true of XDR as its defined, and I'm guessing our C++ implementation, but it's not practically true in all programming languages, so that property doesn't flow downstream.

Regardless, naming, and communication, are more important to me than type identity, or convert-ability. For future definitions I think there is more value in the consistency of names, e.g. `ASSET_TYPE_LIQUIDITY_POOL_SHARE`.

Given that they are always liquidity pools, I've changed the format to a `:LP` suffix, which as it happens was @stanford-scs's original suggestion.

ire-and-curses (approved):

ecosystem/sep-0011.md
```diff
@@ -170,13 +172,19 @@ A few aggregate values are special-cased:
...
-  - `Code:IssuerAccountID` for issued assets.
+  - `Code:IssuerAccountID` for issued assets, i.e. assets of type
+    `ASSET_TYPE_CREDIT_ALPHANUM4` and `ASSET_TYPE_CREDIT_ALPHANUM12`.
 
     "Code" must consist of printable ASCII characters (octets 0x21 through 0x7e).
 
     The sequence `\x` introduces a hex escape sequence, e.g., `\x00` to introduce a 0-valued byte.  Otherwise, `\` escapes the next character, so `\\` is required to introduce a backslash. Stellar disallows assets of type `ASSET_TYPE_CREDIT_ALPHANUM12` that have fewer than 5 bytes, but such assets can be represented in binary XDR, and so in txrep are rendered with trailing `\x00` (escaped NUL bytes) to as to make the length 5---e.g., the 12-byte asset code `ABC` is rendered `ABC\x00\x00`. Note that stellar-core disallows non ASCII bytes in `AssetCode` fields, so the primary use of this feature is to construct or examine invalid transaction test cases.
+  - `LiquidityPoolID:LP` for liquidity pool shares, i.e. assets of type `ASSET_TYPE_POOL_SHARE`.
+
+    "LiquidityPoolID" is the hex encoded `PoolID`.
```
Shaptic, line 185: Should we use `LP:$ID`, instead, to parallel the way `issuer` in `code:issuer` is the main differentiator / unique component?

ecosystem/sep-0011.md
```diff
@@ -170,13 +172,19 @@ A few aggregate values are special-cased:
 
 ##### Asset
 
-The `Asset` type is rendered as a string in one of two forms:
+The `Asset` type is rendered as a string in one of three forms:
   - `native` (or any string up to 12 characters not containing an unescaped colon) for the native asset.
```
Shaptic, line 176: lol I don't think a single SDK follows this parenthetical :sweat_smile:

ecosystem/sep-0011.md
leighmcculloch: I also like the prefix format `LP:...`, but unfortunately there's an opportunity for ambiguity since `LP` could be a valid asset code. It makes it harder to programmatically determine if the text after the colon is an issuer account ID or a liquidity pool. This problem is exasperated if the network adds new strkey types that can be used as the issuer account ID and that overlap with the valid character sets liquidity pool IDs.

leighmcculloch:
> Also, liquidity pool assets aren't `Asset` objects in the XDR sense.

@Shaptic Good point. Thanks for catching this.

I've updated SEP-11 in c20fd28 to explicitly refer to each of the types, however I think the SEP still only needs to provide one format definition that applies to all of them uniformly. SDK implementations that provide canonical string encoding and decoding based on SEP-11 should error when attempting to decode any asset type unsupported by the specific asset type.

ecosystem/sep-0011.md
```diff
@@ -168,16 +170,24 @@ Enums are represented by the bare keyword of the value.  They can also be specif
 
 A few aggregate values are special-cased:
 
-##### Asset
+##### Asset, TrustLineAsset, ChangeTrustAsset
```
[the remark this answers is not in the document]
leighmcculloch: Could you elaborate on this?

> We can't go from str -> XDR for the `ChangeTrustAsset` case

What do we need other than the code and the issuer account ID? Noting that, for the credit asset case, we can select whether it's a 4 or 12 character code using the logic defined on line 183.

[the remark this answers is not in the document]
leighmcculloch: Great point. Thanks for catching this. I've pushed a changes that define encoding of AlphaNum4/12 separately so that for `ChangeTrustAsset` we string encode the fields holding the AlphaNum4 and AlphaNum12 appropriately. In regards to how this impact the SDKs, I think just as you mentioned, any function like `ChangeTrustAsset.fromCanonicalString(...)` cannot support liquidity pool shares. Folks need to construct that manually on their own.

ecosystem/sep-0011.md
```diff
@@ -170,19 +170,34 @@ Enums are represented by the bare keyword of the value.  They can also be specif
...
+
+  - `native` (or any string up to 12 characters not containing an unescaped colon) for the native asset, i.e. `ASSET_TYPE_NATIVE`.
 
-    The sequence `\x` introduces a hex escape sequence, e.g., `\x00` to introduce a 0-valued byte.  Otherwise, `\` escapes the next character, so `\\` is required to introduce a backslash. Stellar disallows assets of type `ASSET_TYPE_CREDIT_ALPHANUM12` that have fewer than 5 bytes, but such assets can be represented in binary XDR, and so in txrep are rendered with trailing `\x00` (escaped NUL bytes) to as to make the length 5---e.g., the 12-byte asset code `ABC` is rendered `ABC\x00\x00`. Note that stellar-core disallows non ASCII bytes in `AssetCode` fields, so the primary use of this feature is to construct or examine invalid transaction test cases.
+  - `Code:IssuerAccountID` as defined for `AlphaNum4` or `AlphaNum12` for issued assets, i.e. `ASSET_TYPE_CREDIT_ALPHANUM4` and `ASSET_TYPE_CREDIT_ALPHANUM12`.
 
-  - `LiquidityPoolID:LP` for liquidity pool shares, i.e. assets of type `ASSET_TYPE_POOL_SHARE`.
+  - `LiquidityPoolID::LP` for liquidity pool shares, i.e. assets of type `ASSET_TYPE_POOL_SHARE`.
```
Shaptic, line 200: I see now that we're doing an intentional hard break from the `code:issuer` format; can you elaborate?

ecosystem/sep-0011.md
```diff
@@ -170,19 +170,34 @@ Enums are represented by the bare keyword of the value.  They can also be specif
...
+Stellar disallows assets of type `ASSET_TYPE_CREDIT_ALPHANUM12` that have fewer
+than 5 bytes, but such assets can be represented in binary XDR, and so in txrep
+an `AlphaNum12` value with fewer than 5 bytes in its `assetCode` is rendered
+with trailing `\x00` (escaped NUL bytes) to as to make the length 5---e.g., the
+12-byte asset code `ABC` is rendered `ABC\x00\x00`. Note that stellar-core
+disallows non ASCII bytes in `AssetCode` fields, so the primary use of this
+feature is to construct or examine invalid transaction test cases.
 
-  - `Code:IssuerAccountID` for issued assets, i.e. assets of type `ASSET_TYPE_CREDIT_ALPHANUM4` and `ASSET_TYPE_CREDIT_ALPHANUM12`.
+##### Asset, TrustLineAsset
```
Shaptic, line 192: Sorry for all of the back-and-forth, but since this is a formal spec we've gotta nail the details. Are we saying here that `ChangeTrustAsset` will not have a SEP-11 form at all, even if it contains a native/credit4/credit12 asset? Seems fine, but we should be explicit. Alternatively, a subsection for the LPs where we say "the following does not apply to ChangeTrustAsset".

Shaptic: > This section only details exceptions the standard format

This is the knowledge I was missing! Thanks for cluing me in, and definitely appreciate the detailed explanation :100: all of my concerns have melted away :ok_hand:

Shaptic (dismissed):

ecosystem/sep-0011.md
leighmcculloch: I've changed it back to a single colon.

leighmcculloch: @stanford-scs You also suggested previously introducing an underscore at the beginning of the type identifier. That is clearer and less ambiguous to me than lower case.

e.g. `f81c...d24e:_LP`

leighmcculloch: I've updated the SEP to `:lp` because that's what @stanford-scs has used in stc. @ire-and-curses @Shaptic Does that work for you?

PR #1080, opened 2021-11-01
CAP-xx: Concurrent Transactions

leighmcculloch:
### What
Add a proposal to support concurrent transactions in that primary users.

### Why
See the motivation section of the CAP.

### Notes
@stanford-scs Is listed as an author because this proposal is based on and similar to an idea he shared with me a little while ago. He hasn't reviewed this yet.


jonjove (approved):
Approving to merge, as this is a draft.

PR #1082, opened 2021-11-03
SEP-24: remove unused withdrawal response

### LINKED ISSUE #864: SEP-24: Add no-information-needed response to deposit or remove it from withdraw
SEP-24 offers a no additional information needed response for `POST /transactions/withdraw/interactive`, but does not offer a similar endpoint for `POST /transactions/deposit/interactive`.

It is not clear why this is the case. Anchors could conceivably not need any additional information for a deposit in the same way it is possible for withdraw, that is: the user has been KYC previously in connection with the same Stellar account being used in the current transaction request.

So, we have two choices:

1. Adding a deposit no-additional-info-needed response. The response body would probably look something like this:
```json
{
    "id": "a8e11fcb-9ae5-410a-86d6-a6a0bf1fe40f",
    "how": "a human-readable message outlining how the user should send funds to the anchor's off-chain account. This is how SEP-6 does it",
    "fee_fixed": "optional number as a string",
    "fee_percent": "optional number between 0 and 100 as a string",
    "max_amount": "optional number as a string",
    "min_amount": "optional number as a string"
}
```

2. Removing the withdraw no-additional-info-needed response. This would technically be a breaking change, although it is possible this response is not used by any anchor. Polaris doesn't support it. And it is very possible that wallet applications don't either. If we go this route, anchors would always return an interactive info need response, but the `url` would just be a confirmation/informational page when no additional information is required.

JakeUrban:
resolves #864 

Removes the no-additional-info-needed response from the withdrawal endpoint, making it consistent with the responses offered for the deposit endpoint.

I can see a future where wallets concerned about UX want the ability to skip the interactive flow entirely, but we can add responses back for both endpoints when that use case arises.

### DIFF
--- a/ecosystem/sep-0024.md
+++ b/ecosystem/sep-0024.md
@@ -6,8 +6,8 @@ Title: Hosted Deposit and Withdrawal
 Author: SDF
 Status: Active
 Created: 2019-09-18
-Updated: 2021-10-27
-Version 1.6.0
+Updated: 2021-11-03
+Version 2.0.0
 ```
 
 ## Simple Summary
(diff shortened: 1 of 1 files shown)


leighmcculloch (commented):
❔ Do you think it is worth cutting v2 for this? Or just worth deprecating at this point and cutting v2 in the future when we've collected other breaking changes we want to make?

JakeUrban:
Since our protocols are usually changed because a new use case has recently been identified, I don't see us collecting breaking changes and holding off on actually merging them in order to make fewer major version updates. I figure breaking changes will be made at the rate at which they're necessary in order to support the ecosystem (while also considering the cost for existing implementations).

That being said there is no urgency with this change, so we could simply deprecate it with the intention of removing when the next breaking change comes along that actually has some urgency.

leighmcculloch:
Ok, makes sense. 

One last question, how do we confirm no one wants this? Do we already have any input that suggests this is valuable to someone? I assume no if we're here.

Otherwise lgtm.

leighmcculloch (approved):

PR #1087, opened 2021-11-09
SEP-38: relax the requirements for specifying delivery methods

JakeUrban:
The `buy_delivery_method` and `sell_delivery_method` parameters on the `GET /price`, `GET /prices`, and `POST /quote` endpoints are used to provide the anchor with additional information that can be used to adjust the price and expiration provided in the response.

However, client have reasons to make requests to the `GET /price(s)` endpoints other than actually exchanging the assets, namely:

- Displaying the available pairs
- Displaying an estimated exchange rates prior to the user initiating a transaction

For the above use cases, especially the first, its not strictly necessary to specify how the user will deliver or pick up the off-chain asset to the anchor. The anchor can decide to show its best rates when not specified.

For `POST /quote`, the delivery method parameters are certainly relevant and should be required if the anchor needs them. However, its possible that the anchor doesn't need them, and we specification should allow for that possibility.

### DIFF
--- a/ecosystem/sep-0038.md
+++ b/ecosystem/sep-0038.md
@@ -366,7 +366,9 @@ In the reverse scenario, the anchor will submit a Stellar transaction to deliver
 
 #### Request
 
-The client must provide either `sell_amount` or `buy_amount`, but not both.
+The client must provide either `sell_amount` or `buy_amount`, but not both. 
+
+Unless the list included in the `GET /info` response is empty or missing for the associated off-chain asset, the client must also provide either `sell_delivery_method` or `buy_delivery_method`, but not both.
 
 Name | Type | Description
 -----|------|------------
@@ -375,8 +377,8 @@ Name | Type | Description
 `sell_amount` | string | Same as the definition of `sell_amount` in `GET /price`.
 `buy_amount` | string | The same definition of `buy_amount` in `GET /price`.
 `expire_after` | [UTC ISO 8601](https://www.iso.org/iso-8601-date-and-time-format.html) string | (optional) The client's desired `expires_at` date and time for the quote. Anchors should return `400 Bad Request` if the an expiration on or after the requested value cannot be provided.
-`sell_delivery_method` | string | (optional) One of the `name` values specified by the `sell_delivery_methods` array for the associated asset returned from [`GET /info`](#info). Must be provided if the user is delivering an off-chain asset to the anchor.
-`buy_delivery_method` | string | (optional) One of the `name` values specified by the `buy_delivery_methods` array for the associated asset returned from [`GET /info`](#info). Must be provided if the user intends to receive an off-chain asset from the anchor.
+`sell_delivery_method` | string | (optional) One of the `name` values specified by the `sell_delivery_methods` array for the associated asset returned from [`GET /info`](#info). Must be provided if the user is delivering an off-chain asset to the anchor unless no methods are specified.
+`buy_delivery_method` | string | (optional) One of the `name` values specified by the `buy_delivery_methods` array for the associated asset returned from [`GET /info`](#info). Must be provided if the user intends to receive an off-chain asset from the anchor unless no methods are specified.
 `country_code` | string | (optional) The [ISO 3166-1 alpha-3](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-3) code of the user's current address. Should be provided if there are two or more country codes available for the desired asset in [`GET /info`](#get-info).
 
 #### Response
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0038.md
```diff
@@ -130,8 +130,8 @@ Name | Type | Description
...
-`buy_delivery_methods` | array | (optional) Only for non-Stellar assets. An array of objects describing the methods a client can use to buy/retrieve funds from the anchor. The method of delivery may affect the expiration and/or price provided in a [`POST /quote`](#post-quote) response.
+`sell_delivery_methods` | array | (optional) Only for non-Stellar assets. An array of objects describing the methods a client can use to sell/deliver funds to the anchor. The method of delivery may affect the expiration and/or price provided in a [`POST /quote`](#post-quote) response. If the delivery method is not necessary for providing accurate quotes and expirations, the server can omit this attribute.
+`buy_delivery_methods` | array | (optional) Only for non-Stellar assets. An array of objects describing the methods a client can use to buy/retrieve funds from the anchor. The method of delivery may affect the expiration and/or price provided in a [`POST /quote`](#post-quote) response. If the delivery method is not necessary for providing accurate quotes and expirations, the server can omit this attribute.
```
marcelosalloum, line 134: 💡  I still think it's valuable to require the `/info` endpoint to display all available delivery methods, even if there's only one option.

ecosystem/sep-0038.md
```diff
@@ -378,8 +377,8 @@ Name | Type | Description
...
-`sell_delivery_method` | string | (optional) One of the `name` values specified by the `sell_delivery_methods` array for the associated asset returned from [`GET /info`](#info). Must be provided if the user is delivering an off-chain asset to the anchor.
-`buy_delivery_method` | string | (optional) One of the `name` values specified by the `buy_delivery_methods` array for the associated asset returned from [`GET /info`](#info). Must be provided if the user intends to receive an off-chain asset from the anchor.
+`sell_delivery_method` | string | (optional) One of the `name` values specified by the `sell_delivery_methods` array for the associated asset returned from [`GET /info`](#info). Must be provided if the user is delivering an off-chain asset to the anchor unless no methods are specified in the `GET /info` response.
+`buy_delivery_method` | string | (optional) One of the `name` values specified by the `buy_delivery_methods` array for the associated asset returned from [`GET /info`](#info). Must be provided if the user intends to receive an off-chain asset from the anchor unless no methods are specified in the `GET /info` response.
```
marcelosalloum, line 381: 💡  I believe this could also be omitted if there's only one option:

```suggestion
`sell_delivery_method` | string | (optional) One of the `name` values specified by the `sell_delivery_methods` array for the associated asset returned from [`GET /info`](#info). Must be provided if the user is delivering an off-chain asset to the anchor unless no more than one method is specified in the `GET /info` response.
`buy_delivery_method` | string | (optional) One of the `name` values specified by the `buy_delivery_methods` array for the associated asset returned from [`GET /info`](#info). Must be provided if the user intends to receive an off-chain asset from the anchor unless no more than one method is specified in the `GET /info` response.
```

marcelosalloum (commented):
Great improvement! I've left a couple of suggestions (💡 ) below:

ecosystem/sep-0038.md
JakeUrban: Hmm thats a good point, it seems there are two reasons why delivery methods should be listed:
- To show what delivery methods are supported / available
- To provide the client with the `name`'s so they can be used in future SEP-38 requests

I think we should keep the current text because the 1st reason can be satisfied in the transaction flow in which the quote is used. For example, I could omit the delivery methods from SEP-38's `GET /info` but I could still display the delivery options in the SEP-24 interactive flow. And both SEP-6 and SEP-31 have a [fields](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0006.md#fields) object in their own `GET /info` responses that could be used to document the possible delivery methods as well.

ecosystem/sep-0038.md
JakeUrban: Thats a good point, if there's only 1 option then the delivery method is implied.

ecosystem/sep-0038.md
marcelosalloum: Could SEP-38 be used independently from the other SEPs? SEP-38 doesn't mention anything about SEPs 6, 24, or 31 and I believe it's better that SEPs contain all the information needed within themselves, rather than rely on other SEPs. WDYT?

JakeUrban: > Could SEP-38 be used independently from the other SEPs? 

It could, although if the purpose of SEP-38 is to provide quotes to be used in transactions, and delivery method info isn't needed to calculate quotes, I think listing supported delivery methods should be the responsibility of the application facilitating the transaction.

marcelosalloum (approved):
LGTM!

PR #1088, opened 2021-11-17
SEP-10: Fix "Verifying the Client Domain" hyperlink

CassioMG:
This PR updates the "Verifying Client Application Identity" copy to read "Verifying the Client Domain" instead (like it's appearing in all the other places in the doc) and fix the hyperlink.


leighmcculloch (commented):
Thanks for fixing this @CassioMG!!

So that we can merge this, could you update the pragma section right at the top of the file so that the version is `3.3.1` and the updated date is today's date? Thanks!

CassioMG:
> Thanks for fixing this @CassioMG!!
> 
> So that we can merge this, could you update the pragma section right at the top of the file so that the version is `3.3.1` and the updated date is today's date? Thanks!

@leighmcculloch totally forgot about it, updated! - thanks

leighmcculloch (approved):
🎉

Issue #1046, opened 2021-08-23
Meta: SEP Versioning


leighmcculloch:
Over the last couple years a few folks have pointed out that SEP versioning needs some work.

Unlike CAPs, SEPs are not a point-in-time proposal. SEPs are protocols and specifications that are improved over time. This means that changes are made to them over long periods of time, sometimes bug fixes, sometimes improvements or new features, and sometimes breaking changes where necessary. 

All SEPs have a `Version` meta field in their preamble. The [SEP README](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/README.md#sep-versioning) provides instructions on how versions should be assigned during the draft stage (i.e. `v0.X.Y`) and in the Active and other stages (i.e. `vMAJOR.MINOR.PATCH`).

That is the extent of the support structures for versioning SEPs. SEPs do not typically provide a way for implementers to support multiple incompatible versions at the same time if a breaking change has occurred, or to communicate with another party about what version they support.

This issue is a place where we can discuss these and other problems relating to SEP versionins, and discuss ways we could improve.

cc @rice2000 @JakeUrban @tomerweller @accordeiro @marcelosalloum @yuriescl

yuriescl:
Thanks for creating this discussion. Indeed it's difficult to manage so many changes in SEPs while keeping Anchor and Wallet software compatible across time.
We need some kind of version and dependency system to allow SDKs, Anchors and Wallets to smoothly adapt to SEP changes.
About the versioning, I got some ideas:
- Turn each SEP into a directory to allow version subdirectories:
  `stellar-protocol/ecosystem/sep-0024/1.4.0/sep-0024.md`  (this structure is just an example)
- Mention the compatible SEPs in each SEP .md file. For example, in `sep-0024.md`, add a section `SEP Compatibility` and detail the compatibility between this and the other SEPs.
- Add very detailed changelogs and instructions between SEP versions. This makes it easy for Anchors, Wallets, and SDK developers to understand and follow the changes. At the moment we have diffs but maybe we can have something more elaborate.

leighmcculloch:
> * Turn each SEP into a directory to allow version subdirectories:
>   `stellar-protocol/ecosystem/sep-0024/1.4.0/sep-0024.md`

[...]

PR #1110, opened 2022-01-13
SEP: Add changelog section to SEP template

leighmcculloch:
### What
Add a very lightweight changelog section to the SEP template.

### Why
This idea was suggested by @marcelosalloum in https://github.com/stellar/stellar-protocol/issues/1046#issuecomment-995861028.

SEP documents evolve over time but there is no location where a reader can quickly read in prose a description of what has changed from one version to the next. While the Git history and PR descriptions for changes made to the SEP provide insights it requires a lot of effort for a reader to sift through all that information to find what the final incremental changes are that have been made.

Requiring a changelog entry will also help us evaluate the necessity and clarity of a change since the changelog is persisted forever we must choose wisely the changes we make.

The changelog @marcelosalloum had proposed was a larger document with sections under each version. After some experimentation I am proposing a more lightweight changelog. The reason being is that version changes in SEPs are often quite different to software releases in that we don't usually queue up changes then issue a release. Instead we issue a release and update a version with every small change we make to a SEP. For this reason I think we need only a single line to briefly state what each versions change is. Since the changelog will grow over time this will also keep it succinct and easy to parse and read.

The changelog @marcelosalloum had proposed was a separate markdown document but I've included the changelog in the SEP itself. I think it is very helpful today that most things for a SEP live inside the one SEP document. If the changelog is at the end it is unlikely to get in the way of the document content. It is easier to discover as well if it is embedded.

### DIFF
--- a/sep-template.md
+++ b/sep-template.md
@@ -46,8 +46,8 @@ All SEPs should contain a changelog that succinctly lists its versions and what 
 
 Example:
 
-- `v0.1.0`: Initial draft. [#1000](https://github.com/stellar/stellar-protocol/pulls/1000)
-- `v0.2.0`: Redesigned the API. [#1004](https://github.com/stellar/stellar-protocol/pulls/1004)
-- `v1.0.0`: Updated status to Active. [#1234](https://github.com/stellar/stellar-protocol/pulls/1234)
-- `v1.1.0`: Add new paremeters to the X endpoint. [#3498](https://github.com/stellar/stellar-protocol/pulls/3498)
 - `v1.1.1`: Fix misleading statements about the P parameter. [#3499](https://github.com/stellar/stellar-protocol/pulls/3499)
+- `v1.1.0`: Add new paremeters to the X endpoint. [#3498](https://github.com/stellar/stellar-protocol/pulls/3498)
+- `v1.0.0`: Updated status to Active. [#1234](https://github.com/stellar/stellar-protocol/pulls/1234)
+- `v0.2.0`: Redesigned the API. [#1004](https://github.com/stellar/stellar-protocol/pulls/1004)
+- `v0.1.0`: Initial draft. [#1000](https://github.com/stellar/stellar-protocol/pulls/1000)
--- a/sep-template.md
+++ b/sep-template.md
@@ -42,7 +42,7 @@ All SEPs should carefully consider areas where security may be a concern, and do
 accordingly. If a change does not have security implications, write "N/A".
 
 ## Changelog
-All SEPs should contain a changelog that succinctly lists its versions and what change was introduced in each version. Information included should contain a succinct by complete description of what has changed. Succinct is important so that over time a large changelog is easily digested.
+All SEPs should contain a changelog that succinctly lists its versions and what changes were introduced in each version. A complete description of what has changed should be included, but it should be as succinct as possible. Succinct is important so that over time a large changelog is easily digested.
 
 Example:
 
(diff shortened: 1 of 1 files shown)


sep-template.md
```diff
@@ -40,3 +40,14 @@ objections or concerns raised during discussion.
 ## Security Concerns
 All SEPs should carefully consider areas where security may be a concern, and document them
 accordingly. If a change does not have security implications, write "N/A".
+
+## Changelog
+All SEPs should contain a changelog that succinctly lists its versions and what change was introduced in each version. Information included should contain a succinct by complete description of what has changed. Succinct is important so that over time a large changelog is easily digested.
+
+Example:
+
+- `v0.1.0`: Initial draft.
```
JakeUrban, line 49: How about also linking to the commit or PR that incremented the document to the associated version? This would allow each SEP to effectively have a git commit history embedded in the file, which I think is valuable because looking at the stellar-protocol commit history has commits to many other documents.

JakeUrban (approved):

sep-template.md
leighmcculloch: Added to the example in 4210cb279c11a95aee6a9bdf7025d7806e85d34a.

JakeUrban (dismissed):

sep-template.md
```diff
@@ -40,3 +40,14 @@ objections or concerns raised during discussion.
...
+## Changelog
+All SEPs should contain a changelog that succinctly lists its versions and what change was introduced in each version. Information included should contain a succinct by complete description of what has changed. Succinct is important so that over time a large changelog is easily digested.
+
+Example:
+
+- `v0.1.0`: Initial draft. [#1000](https://github.com/stellar/stellar-protocol/pulls/1000)
+- `v0.2.0`: Redesigned the API. [#1004](https://github.com/stellar/stellar-protocol/pulls/1004)
+- `v1.0.0`: Updated status to Active. [#1234](https://github.com/stellar/stellar-protocol/pulls/1234)
+- `v1.1.0`: Add new paremeters to the X endpoint. [#3498](https://github.com/stellar/stellar-protocol/pulls/3498)
+- `v1.1.1`: Fix misleading statements about the P parameter. [#3499](https://github.com/stellar/stellar-protocol/pulls/3499)
```
marcelosalloum, line 53: 🤔  I'd prefer ordering changes from newest to oldest, WDYT? It's the kind of order found in [npm](https://www.npmjs.com/package/stellar-sdk) and other platforms.

marcelosalloum (dismissed):
LGTM! I've left a suggestion/thought below.

sep-template.md
leighmcculloch: Good idea. Changed in 9546e9a613ec3503cc5893be4cf0d3b1de3a8d4e.

JakeUrban (approved):

marcelosalloum (dismissed):

Shaptic (approved):
Nice! Small question: what's the guidance on if the version deltas occur within the same PR (for example, big changes prior to "Active" status _cough_ like in the SEP-39 giga-PR _cough_)?

JakeUrban (approved):

leighmcculloch:
>what's the guidance on if the version deltas occur within the same PR (for example, big changes prior to "Active" status cough like in the SEP-39 giga-PR cough)?

@Shaptic I wouldn't update the version until the PR merges.

marcelosalloum (approved):

Shaptic (approved):

PR #1111, opened 2022-02-02
SEP Proposal: A non-fungible token issuance and metadata specification

Shaptic:
This is an effort to create a specification for describing NFTs. The spec leans heavily on both the [EIP-721 standard](https://eips.ethereum.org/EIPS/eip-721) (which offers some basic metadata) and the existing Litemint model (which offers a lot more).

In summary, there are three main parts:

 - defining a standard NFT issuance model
 - defining the NFTs themselves, which involves:
   * defining how to reference their metadata from an account, and
   * defining the metadata format itself

There's much more to do, and I've yet to elaborate on the _Design Rationale_ portion, but I think it's a good starting point for discussion.

Tagging some minimum interested parties, based on the last discussion: 

@leighmcculloch @tyvdh @tomerweller 
@FredericRezeau @grempe

### DIFF
--- a/ecosystem/sep-0040.md
+++ b/ecosystem/sep-0040.md
@@ -96,12 +96,14 @@ The metadata must be valid JSON and must contain the following **required** fiel
 
 | Field       | Description |
 | ----------- | ----------- |
-| name | a short name or title for the NFT |
-| description | a longer, more-detailed description of the NFT |
-| type | a [Media Type](https://www.iana.org/assignments/media-types/media-types.xhtml) describing the NFT's primary format (pointed to by `url`) |
-| uri  | a [URI](https://datatracker.ietf.org/doc/html/rfc3986) pointing to the actual NFT contents |
-| issuer | a Stellar public key ([`G...`](https://stellar.org/protocol/sep-23)) corresponding to the account which issued this NFT |
-| code | the NFT's asset code |
+| `name` | a short name or title for the NFT |
+| `description` | a longer, more-detailed description of the NFT |
+| `type` | a [Media Type](https://www.iana.org/assignments/media-types/media-types.xhtml) describing the NFT's primary format (pointed to by `url`) |
+| `uri` **or** `url`  | a [URI](https://datatracker.ietf.org/doc/html/rfc3986) pointing to the actual NFT contents |
+| `issuer` | a Stellar public key ([`G...`](https://stellar.org/protocol/sep-23)) corresponding to the account which issued this NFT |
+| `code` | the NFT's asset code |
+
+We allow either `uri` or `url` to maintain compatibility with existing ecosystem conventions. If _both_ are present, prefer the `uri`.
 
 For example,
 
@@ -110,11 +112,13 @@ For example,
   "name": "A demonstration of NFT metadata",
   "description": "This is a description of the cool NFT used in this demo.",
   "type": "image/png",
-  "uri": "https://cloudflare-ipfs.com/ipfs/QmXoypizjW3WknFiJnKLwHCnL72vedxjQkDDP1mXWo6uco/I/m/RickAstleyNeverGonnaGiveYouUp7InchSingleCover.jpg",
+  "url": "https://cloudflare-ipfs.com/ipfs/QmXoypizjW3WknFiJnKLwHCnL72vedxjQkDDP1mXWo6uco/I/m/RickAstleyNeverGonnaGiveYouUp7InchSingleCover.jpg",
   "issuer": "GALAXYVOIDAOPZTDLHILAJQKCVVFMD4IKLXLSZV5YHO7VY74IWZILUTO",
   "code": "DEMOASSET"
 }
 ```
+
+Means that the `DEMOASSET:GALAXYVOIDAOPZTDLHILAJQKCVVFMD4IKLXLSZV5YHO7VY74IWZILUTO` asset represents [this image](https://cloudflare-ipfs.com/ipfs/QmXoypizjW3WknFiJnKLwHCnL72vedxjQkDDP1mXWo6uco/I/m/RickAstleyNeverGonnaGiveYouUp7InchSingleCover.jpg).
 
 There are other fields which are **optional** but standardized:
 
@@ -168,7 +172,7 @@ The issuance model follows Stellar's [standard model for issuing assets](https:/
 ### On the metadata specification
 While `ipfs://cid` is a type of URI, we allow `type: ipfshash` for compatibility with existing ecosystem conventions as well as to highlight the popularity of IPFS as a storage mechanism in the NFT ecosystem.
 
-The field `uri` was chosen instead of `url` in order to emphasize that data does not have to live on the traditional Web. Since URLs are a subset of URIs, this should not pose a problem to Web-centric platforms.
+The field `uri` was chosen in preference to `url` in order to emphasize that data does not have to live on the traditional Web. Since URLs are a subset of URIs, this should not pose a problem to Web-centric platforms, and the specification still allows `url` for ecosystem compatibility.
 
 
 ## Security Concerns
@@ -180,4 +184,4 @@ Those processing an NFT should take care to ensure that the `issuer` field match
 
 
 ## Changelog
-- `v0.1.0`: Initial draft ([#1000](https://github.com/stellar/stellar-protocol/pulls/1000)).
+- `v0.1.0`: Initial draft ([#1111](https://github.com/stellar/stellar-protocol/pulls/1111)).
--- a/ecosystem/sep-0040.md
+++ b/ecosystem/sep-0040.md
@@ -25,7 +25,7 @@ Non-fungible tokens (NFTs) can describe many types of assets, from digital colle
 
 
 ## Specification
-We begin with the [issuance model](#issuance-model) and then cover the NFT [metadata specification](#nft-metadata-specification).
+The protocol defines an [issuance model](#issuance-model) that defines how NFT assets are issued on Stellar, and a [metadata specification](#nft-metadata-specification) that defines how data about the NFT is captured in a way that Stellar guarantees is immutable.
 
 ### Issuance Model
 Given an arbitrary digital asset, here's how issuance typically works:
--- a/ecosystem/sep-0040.md
+++ b/ecosystem/sep-0040.md
@@ -23,21 +23,39 @@ A standard interface for NFTs that makes NFTs issued by any ecosystem applicatio
 ## Abstract
 Non-fungible tokens (NFTs) can describe many types of assets, from digital collectibles to physical items. This protocol defines a standard series of steps that NFT creators should use to issue NFTs on the Stellar network, defines metadata to describe an NFT, and specifies how the metadata is linked to the NFT asset on Stellar.
 
-
-## Specification
-The protocol defines an [issuance model](#issuance-model) that defines how NFT assets are issued on Stellar, and a [metadata specification](#nft-metadata-specification) that defines how data about the NFT is captured in a way that Stellar guarantees is immutable.
-
-### Issuance Model
-Given an arbitrary digital asset, here's how issuance typically works:
+The issuance model involves four steps:
 
   1. An account is created specifically to represent the NFT in the Stellar ledger.
   2. The account configures itself with metadata describing the NFT.
   3. The account issues a new Stellar asset representing ownership over said NFT.
   4. Any account holding that asset lays claim to the NFT.
 
-Ownership of the token directly conveys ownership of a unit of the NFT, however that may be defined. Other scenarios such as locking down the issuing account to avoid supply increases, ownership transfers via payments, etc. are built on top of this basic model and are thus extraneous to this SEP.
+The NFT model involves two levels of indirection in order to minimize on-chain storage:
 
-Note also that this SEP does not address concerns around immutability. While there is an optional checksum component (addressed later), the account itself is free to modify the data at-will if it is not locked.
+    +---------------+      +----------+      +---------+
+    |  Stellar NFT  | ---> | metadata | ---> |   NFT   |
+    |  account data |      |   JSON   |      | content |
+    +---------------+      +----------+      +---------+
+
+
+## Specification
+This SEP defines an [issuance model](#issuance-model) that defines how NFT assets are issued on Stellar and a [metadata specification](#nft-metadata-specification) that defines how data about the NFT is captured. The metadata is stored in a way such that the Stellar network guarantee its authenticity.
+
+### Issuance Model
+Standard issuance of NFTs must take the following steps:
+
+  1. Create an account to represent the NFT: `CreateAccountOp(A, <reserve>)`
+  2. Configure the NFT's metadata pointer, as described [below](#nft-account-description): `ManageDataOp("type", "...")`
+  3. Form an asset describing your NFT: `Code:Issuer` (where issuer is `A`)
+  4. Create a distribution account: `B = CreateAccountOp()` and `ChangeTrust("Code:Issuer", <limit>)`
+  5. Issue the tokens: `PaymentOp(B, "Code:Issuer", <amount>)`
+  6. (Optional) Lock the issuing account: `SetOptionsOp(MasterWeight: 0)`
+
+Note that without this final step, there are concerns around both immutability and supply: an account may modify its NFT metadata pointer or issue more tokens unless it's locked down. Thus, Step 6 is recommended.
+
+Ownership of the token directly conveys ownership of a unit of the NFT, however that may be defined. Other scenarios such as ownership transfers via payments are built on top of this basic issuance model and are thus extraneous to this SEP.
+
+Other issuance models may exist and may be compatible with the following specification, but this is the standard issuance model on Stellar.
 
 ### NFT Specification
 This specification is partially inspired by Ethereum's [EIP-721 specification](https://eips.ethereum.org/EIPS/eip-721#specification) on NFTs; however, it is generalized to arbitrary asset types (i.e. not necessarily images).
--- a/ecosystem/sep-0040.md
+++ b/ecosystem/sep-0040.md
@@ -41,41 +41,6 @@ The NFT model involves two levels of indirection in order to minimize on-chain s
 ## Specification
 This SEP defines an [issuance model](#issuance-model) that defines how NFT assets are issued on Stellar and a [metadata specification](#nft-metadata-specification) that defines how data about the NFT is captured. The metadata is stored in a way such that the Stellar network guarantee its authenticity.
 
-### Issuance Model
-Standard issuance of NFTs involves the following components:
-
-  - An **issuing account**: This is an account which represents the NFT itself. It is responsible for minting the supply of tokens as well as storing information about how to find the NFT asset. We'll refer to it as `I`.
-  - A **token**: This is an asset issued by the issuing account. Its _asset code_ should be descriptive. We'll refer to this asset token as `T`.
-  - A **distribution account**: This is an account which will actually distribute the tokens. Its funded by `I` and initially holds the entire supply of `T`. We'll refer to it as `D`.
-
-The parties must take the following steps:
-
-  1. Someone creates the issuing account `I` to represent the NFT.
-  2. `I` stores a reference to the NFT's [metadata](#nft-account-description) pointer.
-  4. `I` creates a distribution account, `D`.
-  5. `D` creates a trustline to `T`.
-  6. `I` issues the full supply of tokens `T` to `D`.
-  7. (Optional) `I` locks itself out of its account.
-  8. `D` distributes `T` to the world.
-
-This corresponds to the following series of (pseudocode) operations:
-
-```
-    I = CreateAccountOp(...)
-    ManageDataOp(source: I, key: "type", value: "...")
-    T = "AssetCode:I"
-    D = CreateAccountOp(source: I, ...)
-    ChangeTrustOp(source: D, asset: T, limit: <supply>)
-    PaymentOp(source: I, destination: D, asset: T, amount: <supply>)
-    SetOptionsOp(source: I, masterWeight: 0)
-```
-
-Note that Step 7 _permanently_ locks the issuing account which makes it immutable, meaning that the issuer can neither modify the NFT metadata pointer it stored nor issue more tokens. For fixed supply tokens and tokens intended to be immutable, Step 7 is recommended.
-
-Ownership of the token directly conveys ownership of a unit of the NFT, however that may be defined. Other scenarios such as ownership transfers via payments are built on top of this basic issuance model and are thus extraneous to this SEP.
-
-Other issuance models may exist and may be compatible with the following specification, but this is the standard issuance model on Stellar.
-
 ### NFT Specification
 This specification is partially inspired by Ethereum's [EIP-721 specification](https://eips.ethereum.org/EIPS/eip-721#specification) on NFTs; however, it is generalized to arbitrary asset types (i.e. not necessarily images).
 
@@ -185,6 +150,35 @@ Looking this up on IPFS [via CloudFlare](https://cloudflare-ipfs.com/ipfs/QmaydG
 ```
 
 We can see that this account [really does](https://horizon.stellar.org/operations/164563600389443587) own and issue this asset, meaning the NFT is genuine.
+
+### Recommended Issuance Model
+Standard issuance of NFTs involves the following components:
+
+  - An **issuing account**: This is an account which represents the NFT itself. It is responsible for minting the supply of tokens as well as storing information about how to find the NFT asset. We'll refer to it as `I`.
+  - A **token**: This is an asset issued by the issuing account. Its _asset code_ should be descriptive. We'll refer to this asset token as `T`.
+
+The parties must take the following steps:
+
+  1. Someone creates the issuing account `I` to represent the NFT.
+  2. `I` stores a reference to the NFT's [metadata](#nft-account-description) pointer.
+  3. `I` distributes the token to the world.
+  4. (Optional) `I` locks itself out of its account.
+
+This corresponds to the following series of (pseudocode) operations:
+
+```
+    I = CreateAccountOp(...)
+    T = "AssetCode:I"
+    ManageDataOp(source: I, key: "type", value: "...")
+    PaymentOp(source: I, asset: T, ...)
+    SetOptionsOp(source: I, masterWeight: 0)
+```
+
+Note that the last step _permanently_ locks the issuing account which makes it immutable, meaning that the issuer can neither modify the NFT metadata pointer it stored nor issue more tokens. For fixed supply tokens and tokens intended to be immutable, this step is recommended.
+
+Ownership of the token directly conveys ownership of a unit of the NFT, however that may be defined. Other scenarios such as ownership transfers via payments are built on top of this basic issuance model and are thus extraneous to this SEP.
+
+Other issuance models may exist and may be compatible with the following specification, but this is the standard issuance model on Stellar.
 
 
 ## Design Rationale
--- a/ecosystem/sep-0040.md
+++ b/ecosystem/sep-0040.md
@@ -91,7 +91,7 @@ Fields should be stored as simple key-value pairs within an account's data entri
 ##### IPFS Storage
 If the `type` is `ipfshash`, the `ipfshash` key must point to an [IPFS Content IDentifier](https://docs.ipfs.io/concepts/content-addressing/) corresponding to the location of the metadata.
 
-If you are the issuer, make sure you understand the implications of using IPFS. You must use an IPFS lookup service to actually resolve to the metadata. For example, [this URL](https://cloudflare-ipfs.com/ipfs/QmaydGEshbL7HMQkD6FZPcMeEt6o15aC4p4P82VxQ7amjy) resolves to the Meridian 2021 Hat NFT via the Cloudflare IPFS service. If you're the issuer, you must also ensure that IPFS actually stores your data (by running your own node or entrusting a provider).
+If you are the issuer, make sure you understand the implications of using IPFS. You must use an IPFS lookup service to actually resolve to the metadata. For example, [this URL](https://cloudflare-ipfs.com/ipfs/QmaydGEshbL7HMQkD6FZPcMeEt6o15aC4p4P82VxQ7amjy) resolves to the Meridian 2021 Hat NFT via the Cloudflare IPFS service. If you're the issuer, you must also ensure that the resource is addressable and accessible on the IPFS network by running your own IPFS node or entrusting a provider to do so.
 
 ##### URL Storage
 If the `type` is `uri`, the corresponding `uri` key must point to a [valid URI](https://datatracker.ietf.org/doc/html/rfc3986) containing the NFT's metadata. 
--- a/ecosystem/sep-0040.md
+++ b/ecosystem/sep-0040.md
@@ -44,29 +44,31 @@ This SEP defines an [issuance model](#issuance-model) that defines how NFT asset
 ### Issuance Model
 Standard issuance of NFTs involves the following components:
 
-  - An **issuing account**: This is an account which represents the NFT itself. It is responsible for minting the supply of tokens as well as storing information about how to find the NFT asset. We'll refer to it as `A`.
+  - An **issuing account**: This is an account which represents the NFT itself. It is responsible for minting the supply of tokens as well as storing information about how to find the NFT asset. We'll refer to it as `I`.
   - A **token**: This is an asset issued by the issuing account. Its _asset code_ should be descriptive. We'll refer to this asset token as `T`.
-  - A **distribution account**: This is an account which will actually distribute the tokens. Its funded by `A` and initially holds the entire supply of `T`. We'll refer to it as `B`
+  - A **distribution account**: This is an account which will actually distribute the tokens. Its funded by `I` and initially holds the entire supply of `T`. We'll refer to it as `D`.
 
 The parties must take the following steps:
 
-  1. Someone creates the issuing account `A` to represent the NFT.
-  2. `A` stores a reference to the NFT's [metadata](#nft-account-description) pointer.
-  4. `A` creates a distribution account, `B`.
-  5. `B` creates a trustline to `T`.
-  6. `A` issues the full supply of tokens `T` to `B`.
-  7. (Optional) `A` locks itself out of its account.
-  8. `B` distributes `T` to the world.
+  1. Someone creates the issuing account `I` to represent the NFT.
+  2. `I` stores a reference to the NFT's [metadata](#nft-account-description) pointer.
+  4. `I` creates a distribution account, `D`.
+  5. `D` creates a trustline to `T`.
+  6. `I` issues the full supply of tokens `T` to `D`.
+  7. (Optional) `I` locks itself out of its account.
+  8. `D` distributes `T` to the world.
 
 This corresponds to the following series of (pseudocode) operations:
 
-    A = CreateAccountOp(...)
-    ManageDataOp(source: A, key: "type", value: "...")
-    T = "AssetCode:A"
-    B = CreateAccountOp(source: A, ...)
-    ChangeTrustOp(source: B, asset: T, limit: <supply>)
-    PaymentOp(source: A, destination: B, asset: T, amount: <supply>)
-    SetOptionsOp(source: A, masterWeight: 0)
+```
+    I = CreateAccountOp(...)
+    ManageDataOp(source: I, key: "type", value: "...")
+    T = "AssetCode:I"
+    D = CreateAccountOp(source: I, ...)
+    ChangeTrustOp(source: D, asset: T, limit: <supply>)
+    PaymentOp(source: I, destination: D, asset: T, amount: <supply>)
+    SetOptionsOp(source: I, masterWeight: 0)
+```
 
 Note that Step 7 _permanently_ locks the issuing account which makes it immutable, meaning that the issuer can neither modify the NFT metadata pointer it stored nor issue more tokens. For fixed supply tokens and tokens intended to be immutable, Step 7 is recommended.
 
--- a/ecosystem/sep-0040.md
+++ b/ecosystem/sep-0040.md
@@ -70,14 +70,20 @@ There are two **required** fields on the account itself:
 | Field       | Purpose |
 | ----------- | ------- |
 | `type`      | Describes how the "first hop" of metadata is defined, standard values include `ipfshash` and `uri` |
-| `hash`      | The SHA3 hash of the metadata contents |
+| `hash`      | The [SHA3](https://csrc.nist.gov/publications/detail/fips/202/final) hash of the metadata contents |
 
+
+##### `type`
 The two possible values for the `type` field describe two standard ways to store the metadata, and the details will be in a data entry corresponding to this value. In other words, `type: ipfshash` means that you should look under the `ipfshash` key data entry which further defines how to find the metadata.
 
 All other values for the `type` are considered non-standard until they are approved for inclusion within this SEP.
 
+##### `hash`
 The `hash` is provided to ensure integrity of the metadata that the account is pointing to. This prevents the data at the destination from being modified without explicit knowledge and approval by the issuer.
 
+If it is excluded, the NFT **must not be considered genuine**, _unless_ the `type` is `ipfshash`, since the IPFS CID includes an integrity check on the data it points to. 
+
+##### Storage
 Fields should be stored as simple key-value pairs within an account's data entries. The standard limitations of `ManageDataOp` apply: field names and their values can only be 64 bytes each. Most metadata attributes should be stored at the NFT reference location itself.
 
 **Backward Compatibility:** Existing NFT platforms essentially default to `type: ipfshash`. While this standard aims to free up the ecosystem from dependence on IPFS, it should also work with existing conventions that developed organically. To do so, **if you see an account with a lone `ipfshash` data entry, you may treat it as equivalent to an NFT account with `type: ipfshash`**.
--- a/ecosystem/sep-0040.md
+++ b/ecosystem/sep-0040.md
@@ -81,7 +81,7 @@ This specification is partially inspired by Ethereum's [EIP-721 specification](h
 
 In order to minimize the costs of creating an NFT (i.e. on-chain storage), the issuing account only stores a reference to the NFT metadata which is stored off-chain.
 
-#### NFT Account Description
+#### NFT Account Configuration
 An issuing account should specify its `homedomain` (via `SetOptionsOp`). It should also configure the token in its domain's `stellar.toml` file (see [SEP-1](https://stellar.org/sep-1#currency-documentation)) if appropriate. This ensures compatibility with the existing ecosystem (for example, rendering it in wallets).
 
 There are two ways to configure an account for an NFT and both involve storing data entries within the account. These are simple key-value pairs configured with `ManageDataOp`. Its standard limitations apply: field names and their values can only be 64 bytes each.
@@ -92,9 +92,9 @@ There are two **required** fields on the account itself:
 | Field  | Purpose |
 | ------ | ------- |
 | `url`  | A [URL][url] pointing to the "first hop" of metadata about the NFT |
-| `hash` | A [SHA-256](https://datatracker.ietf.org/doc/html/rfc4634#section-4.1) digest of the metadata referenced by the `url` |
+| `hash` | A raw [SHA-256][sha256] digest of the metadata referenced by the `url` |
 
-The `hash` is provided to ensure integrity of the metadata that the account is pointing to. This prevents the data at the destination from being modified without explicit knowledge and approval by the issuer.
+The `hash` is provided to ensure integrity of the metadata that the account is pointing to. This prevents the data at the destination from being modified without explicit knowledge and approval by the issuer. This hash should be computed on the raw metadata file rather than on a "loaded" JSON object in order to ensure consistent hashing. The hash must be stored in its raw, binary form.
 
 **Warning**: If the hash does not match the metadata content, you should carefully consider the NFT's authenticity.
 
@@ -117,14 +117,15 @@ This section describes the metadata itself rather than how to find it.
 
 The metadata must be valid JSON and must contain the following **required** fields:
 
-| Field       | Description |
-| ----------- | ----------- |
-| `name` | a short name or title for the NFT |
+| Field         | Description |
+| ------------- | ----------- |
+| `name`        | a short name or title for the NFT |
 | `description` | a longer, more-detailed description of the NFT |
-| `type` | a [Media Type][mediatype] describing the NFT's primary format (pointed to by `url`) |
-| `url`  | a [URL][url] pointing to the actual NFT contents |
-| `issuer` | a Stellar public key ([`G...`](https://stellar.org/protocol/sep-23)) corresponding to the account which issued this NFT |
-| `code` | the NFT's asset code |
+| `type`        | a [Media Type][mediatype] describing the NFT's primary format (pointed to by `url`) |
+| `url`         | a [URL][url] pointing to the actual NFT contents |
+| `hash`        | the hex-encoded [SHA-256][sha256] digest of the content pointed to by `url` |
+| `issuer`      | a Stellar public key ([`G...`](https://stellar.org/protocol/sep-23)) corresponding to the account which issued this NFT |
+| `code`        | the NFT's asset code |
 
 For example,
 
@@ -134,6 +135,7 @@ For example,
   "description": "This is a description of the cool NFT used in this demo.",
   "type": "image/png",
   "url": "https://cloudflare-ipfs.com/ipfs/QmXoypizjW3WknFiJnKLwHCnL72vedxjQkDDP1mXWo6uco/I/m/RickAstleyNeverGonnaGiveYouUp7InchSingleCover.jpg",
+  "hash": "2837de01bc746f7840aee6ebedcd60d07b4b0781453205ffd376e42e2f1a30fd",
   "issuer": "GALAXYVOIDAOPZTDLHILAJQKCVVFMD4IKLXLSZV5YHO7VY74IWZILUTO",
   "code": "DEMOASSET"
 }
@@ -210,3 +212,4 @@ Those processing an NFT should take care to ensure that the `issuer` field match
 [cid]: https://docs.ipfs.io/concepts/content-addressing/
 [url]: https://www.ietf.org/rfc/rfc1738.txt
 [mediatype]: https://www.iana.org/assignments/media-types/media-types.xhtml
+[sha256]: https://datatracker.ietf.org/doc/html/rfc4634#section-4.1
--- a/ecosystem/sep-0040.md
+++ b/ecosystem/sep-0040.md
@@ -65,15 +65,18 @@ In order to minimize the costs of on-chain storage, we have multiple levels of i
 #### NFT Account Description
 A proper NFT account should specify its `homedomain`. It should also configure the asset in its domain's `stellar.toml` file (see [SEP-1](https://stellar.org/sep-1#currency-documentation)) if appropriate.
 
-There's only one required field on the account itself:
+There are two **required** fields on the account itself:
 
 | Field       | Purpose |
 | ----------- | ------- |
 | `type`      | Describes how the "first hop" of metadata is defined, standard values include `ipfshash` and `uri` |
+| `hash`      | The SHA3 hash of the metadata contents |
 
 The two possible values for the `type` field describe two standard ways to store the metadata, and the details will be in a data entry corresponding to this value. In other words, `type: ipfshash` means that you should look under the `ipfshash` key data entry which further defines how to find the metadata.
 
 All other values for the `type` are considered non-standard until they are approved for inclusion within this SEP.
+
+The `hash` is provided to ensure integrity of the metadata that the account is pointing to. This prevents the data at the destination from being modified without explicit knowledge and approval by the issuer.
 
 Fields should be stored as simple key-value pairs within an account's data entries. The standard limitations of `ManageDataOp` apply: field names and their values can only be 64 bytes each. Most metadata attributes should be stored at the NFT reference location itself.
 
@@ -167,9 +170,11 @@ We can see that this account [really does](https://horizon.stellar.org/operation
 ## Design Rationale
 
 ### On the issuance model
-The issuance model follows Stellar's [standard model for issuing assets](https://developers.stellar.org/docs/issuing-assets/how-to-issue-an-asset/): there's an _issuing account_ and a _distribution account_. The benefits of this model are described in the linked document, but the most important one is this: it makes it easier to track supply. In the world of NFTs, it's really important to track how many of a particular token are in circulation, and 
+The issuance model follows Stellar's [standard model for issuing assets](https://developers.stellar.org/docs/issuing-assets/how-to-issue-an-asset/): there's an _issuing account_ and a _distribution account_. The benefits of this model are described in the linked document, but the most important one is this: it makes it easier to track supply. In the world of NFTs, it's really important to track how many of a particular token are in circulation, and using a distribution account achieves this goal.
 
 ### On the metadata specification
+Since not all issuers will use `ipfshash`, which already provides integrity by including the hash within the CID, we need the `hash` field to further NFT authenticity.
+
 While `ipfs://cid` is a type of URI, we allow `type: ipfshash` for compatibility with existing ecosystem conventions as well as to highlight the popularity of IPFS as a storage mechanism in the NFT ecosystem.
 
 The field `uri` was chosen in preference to `url` in order to emphasize that data does not have to live on the traditional Web. Since URLs are a subset of URIs, this should not pose a problem to Web-centric platforms, and the specification still allows `url` for ecosystem compatibility.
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0040.md
```diff
@@ -0,0 +1,160 @@
...
+Version: 0.1.0
+Discussion: TBD
+```
+
+## Simple Summary
+This SEP defines a standard interface to describe non-fungible tokens (NFTs) on the Stellar network, including a distribution mechanism and metadata fields. It draws inspiration from the [EIP-721 standard](https://eips.ethereum.org/EIPS/eip-721).
+
+
+## Motivation
+A standard interface for NFTs allows ecosystem applications (such as wallets or marketplaces) to create, view, or otherwise work with any NFT on Stellar.
```
leighmcculloch, line 20: This section could be more explicit in that it is motivated by interoperability, for an NFT created in one application, to be viewable, tradable, useable, anywhere Stellar is used.
```suggestion
A standard interface for NFTs that makes NFTs issued by any ecosystem application (such as a wallet or marketplace) accessible and interoperable with any other ecosystem application supporting NFTs.
```

ecosystem/sep-0040.md
```diff
@@ -0,0 +1,160 @@
...
+
+
+## Abstract
+Non-fungible tokens (NFTs) can describe many types of assets, from digital collectibles to physical items. We should also define a standard issuance model to explain how NFTs can and should be issued on the network. In order to maintain compatibility across the Stellar ecosystem and beyond, we need a standard way to describe and define them that is agnostic of what the NFT actually describes.
```
leighmcculloch, line 24: The `We should define` is somewhat confusing, should we define it, or is it defined by the protocol? It looks like the protocol defines it, so I think we can say it defines it. Suggested rewording:
```suggestion
Non-fungible tokens (NFTs) can describe many types of assets, from digital collectibles to physical items. This protocol defines a standard set of steps that NFT issuers should use to issue NFTs on the Stellar network, defines metadata that describes an NFT, and specifies how the metadata is linked to the NFT asset on Stellar.
```

ecosystem/sep-0040.md
```diff
@@ -0,0 +1,160 @@
...
+
+##### IPFS Storage
+If the `type` is `ipfshash`, the `ipfshash` key must point to an [IPFS Content IDentifier](https://docs.ipfs.io/concepts/content-addressing/) corresponding to the location of the metadata.
+
+If you are the issuer, make sure you understand the implications of using IPFS. You must use an IPFS lookup service to actually resolve to the metadata. For example, [this URL](https://cloudflare-ipfs.com/ipfs/QmaydGEshbL7HMQkD6FZPcMeEt6o15aC4p4P82VxQ7amjy) resolves to the Meridian 2021 Hat NFT via the Cloudflare IPFS service. If you're the issuer, you must also ensure that IPFS actually stores your data (by running your own node or entrusting a provider).
+
+##### Stellar Storage
+If the `type` is `stellar`, the `stellar` key must point to a Stellar account ID. That account will store the metadata within its data entries. Refer to the unofficial [SEP-39](https://github.com/stellar/stellar-protocol/pull/1090) specification for details on how to read an account's data entries for the NFT metadata.
```
leighmcculloch, line 70: There isn't anything such as an official or unofficial SEP in my opinion, if that SEP is SEP-39 we should merge it, or if not drop this section until SEP-39 is merged.
```suggestion
If the `type` is `stellar`, the `stellar` key must point to a Stellar account ID. That account will store the metadata within its data entries. Refer to [SEP-39](SEP-0039.md) specification for details on how to read an account's data entries for the NFT metadata.
```

ecosystem/sep-0040.md
```diff
@@ -0,0 +1,160 @@
...
+
+
+## Security Concerns
+While there are no direct security concerns to the network introduced by this SEP, the trust model of NFTs deserves some clarification.
+
+Digital assets are infinitely and perfectly reproduceable by virtue of being digital. Thus, the goal of an NFT should **not** be to represent ownership or possession of a particular digital asset, but rather to represent the existence of a particular _relationship_ with the issuer.
+
+Those processing an NFT should take care to ensure that the `issuer` field matches the account referencing the NFT, and that the `code` is in fact an asset issued by said account. Without these two links, you cannot make any conclusions about the NFT.
+
+There is another trust issue worth nothing: **supply**. This metadata field _should not be trusted_, and should instead be cross-referenced with the account's distribution operations. Unless an account is completely locked, nothing inherently prevents it from ignoring this `supply` field and issuing new tokens.
```
leighmcculloch, line 155: Why include `supply` in the metadata if it requires verification from another source? Is it intended to capture the supply at the time of mint?

ecosystem/sep-0040.md
```diff
@@ -0,0 +1,160 @@
...
+## Motivation
+A standard interface for NFTs allows ecosystem applications (such as wallets or marketplaces) to create, view, or otherwise work with any NFT on Stellar.
+
+
+## Abstract
+Non-fungible tokens (NFTs) can describe many types of assets, from digital collectibles to physical items. We should also define a standard issuance model to explain how NFTs can and should be issued on the network. In order to maintain compatibility across the Stellar ecosystem and beyond, we need a standard way to describe and define them that is agnostic of what the NFT actually describes.
+
+
+## Specification
+We begin with the [issuance model](#issuance-model) and then cover the NFT [metadata specification](#nft-metadata-specification).
```
leighmcculloch, line 28: It can be helpful for readers if the protocol describes the protocol and doesn't need to tell a story about how the protocol comes to be, so we probably don't need the `We begin` prose. @tomerweller raised a similar comment to this on another protocol we wrote recently with some reasoning that is maybe more elegant than I'm suggesting, so if he chimes in here he might have some thing more thoughtful to say about it. For example:
```suggestion
The protocol defines an [issuance model](#issuance-model) that defines how NFT assets are issued on Stellar, and a [metadata specification](#nft-metadata-specification) that defines how data about the NFT is captured in a way that Stellar guarantees is immutable.
```

I also think that this specification introduction is maybe superfluous and you could drop it if the abstract say substantially the same thing.

ecosystem/sep-0040.md
```diff
@@ -0,0 +1,160 @@
...
+Track: Standard
+Status: Draft
+Created: 2021-01-31
+Updated: 2021-01-31
+Version: 0.1.0
+Discussion: TBD
+```
+
+## Simple Summary
+This SEP defines a standard interface to describe non-fungible tokens (NFTs) on the Stellar network, including a distribution mechanism and metadata fields. It draws inspiration from the [EIP-721 standard](https://eips.ethereum.org/EIPS/eip-721).
```
leighmcculloch, line 16: Suggest using the same terms here as you do later on, so that they are clearly relatable as the same concepts.
```suggestion
This SEP defines a standard interface for non-fungible tokens (NFTs) on the Stellar network, including a issuance model and metadata fields. It draws inspiration from the [EIP-721 standard](https://eips.ethereum.org/EIPS/eip-721).
```

ecosystem/sep-0040.md
```diff
@@ -0,0 +1,160 @@
...
+
+## Abstract
+Non-fungible tokens (NFTs) can describe many types of assets, from digital collectibles to physical items. We should also define a standard issuance model to explain how NFTs can and should be issued on the network. In order to maintain compatibility across the Stellar ecosystem and beyond, we need a standard way to describe and define them that is agnostic of what the NFT actually describes.
+
+
+## Specification
+We begin with the [issuance model](#issuance-model) and then cover the NFT [metadata specification](#nft-metadata-specification).
+
+### Issuance Model
+Given an arbitrary digital asset, here's how issuance typically works:
```
leighmcculloch, line 31:
```suggestion
Issuance of NFTs should take the following steps:
```

ecosystem/sep-0040.md
```diff
@@ -0,0 +1,160 @@
...
+
+  1. An account is created specifically to represent the NFT in the Stellar ledger.
+  2. The account configures itself with metadata describing the NFT.
+  3. The account issues a new Stellar asset representing ownership over said NFT.
+  4. Any account holding that asset lays claim to the NFT.
```
leighmcculloch, line 36: These steps feel too high-level for the specification. These steps could go into the abstract possibly, or you could keep it here, but if the protocol is specifying how to issue the asset, I think it should do so explicitly, by stating what transactions/operations need submitting, and what values to use, etc.

leighmcculloch (commented):
👏🏻 Very excited to see this coming together. I've left some suggestions throughout. Let me know if you'd like to chat synchronously as well about any.

ecosystem/sep-0040.md
Shaptic: I think the `should` there was an editing mishap. I do like your reword, though: it's a good thesis sentence.

ecosystem/sep-0040.md
Shaptic: It's included for compatibility with Litemint, see [here](https://cloudflare-ipfs.com/ipfs/QmaydGEshbL7HMQkD6FZPcMeEt6o15aC4p4P82VxQ7amjy). However, I was hesitant to add it, so I'm fine leaving it off of the official spec.

ecosystem/sep-0040.md
Shaptic: Yep, I think this is a good point, I just wanted it to flow a little more naturally (be "prose-y").

leighmcculloch (commented):
The above comments are just suggestions, I think this is good though and could be merged and iterated on once the SEP number is set in the meta. The matter of the SEP number is awkward. It should be SEP-39 as that is the next SEP number to be assigned, however an unmerged SEP started referring to itself as SEP-39 and this PR has SEP-40 in the title. SEP numbers should only be assigned at merge, and not before so that there are not gaps.

ecosystem/sep-0040.md
```diff
@@ -0,0 +1,160 @@
...
+
+**Backward Compatibility:** Existing NFT platforms essentially default to `type: ipfshash`. While this standard aims to free up the ecosystem from dependence on IPFS, it should also work with existing conventions that developed organically. To do so, **if you see an account with a lone `ipfshash` data entry, you may treat it as an NFT account with `type: ipfshash`**.
+
+##### IPFS Storage
+If the `type` is `ipfshash`, the `ipfshash` key must point to an [IPFS Content IDentifier](https://docs.ipfs.io/concepts/content-addressing/) corresponding to the location of the metadata.
+
+If you are the issuer, make sure you understand the implications of using IPFS. You must use an IPFS lookup service to actually resolve to the metadata. For example, [this URL](https://cloudflare-ipfs.com/ipfs/QmaydGEshbL7HMQkD6FZPcMeEt6o15aC4p4P82VxQ7amjy) resolves to the Meridian 2021 Hat NFT via the Cloudflare IPFS service. If you're the issuer, you must also ensure that IPFS actually stores your data (by running your own node or entrusting a provider).
+
+##### Stellar Storage
```
tomerweller, line 69: Can we remove stellar storage altogether? Storing binary data on-chain should not be encouraged. Maybe we can add an FAQ section explaining why it's not included?

ecosystem/sep-0040.md
```diff
@@ -0,0 +1,160 @@
...
+#### NFT Account Description
+A proper NFT account should specify its `homedomain`. It should also configure the asset in its domain's `stellar.toml` file (see [SEP-1](https://stellar.org/sep-1#currency-documentation)) if appropriate.
+
+There's only one required field on the account itself:
+
+| Field       | Purpose |
+| ----------- | ------- |
+| `type`      | Describes how the "first hop" of metadata is defined, standard values include `ipfshash` and `stellar` |
+
+The two possible values for the `type` field describe two standard ways to store the metadata, and the details will be in a data entry corresponding to this value. In other words, `type: ipfshash` means that you should look under the `ipfshash` key data entry which further defines how to find the metadata.
```
tomerweller, line 56: What about regular non-ipfs urls? these are pretty common in the eth nft land.  This is important if people will ever try to build nft bridges between eth and stellar.

ecosystem/sep-0040.md
leighmcculloch: It'd simplify the spec to leave it out. Any field included should be very well defined and intentional imo. Services like litemint can always define extra fields or we can add it later.

@FredericRezeau what is supply? Is it required by litemint?

tomerweller (approved):
LGTM - minor comments. Ship it.

ecosystem/sep-0040.md
leighmcculloch: Or what final state of expected if it's unimportant which operations are used when.

ecosystem/sep-0040.md
Shaptic: While I hate the idea of supporting storing assets on a centralized web server, you're right that we should probably include that option here. In fact, both this and the `ipfshash` field could be combined under a single `uri` data entry, but that'd be incompatible with Litemint's current model. I'll add a possible `type: uri` value.

ecosystem/sep-0040.md
Shaptic: There was a key motivation to adding this section: adding more than one `type`. With the (pending) addition of `uri` in response to your [comment below](https://github.com/stellar/stellar-protocol/pull/1111#discussion_r797231832), this makes sense to drop in light of your response.

I still think keeping things on-chain is way cooler than other NFT solutions, but I understand why it's not something we want to encourage.

ecosystem/sep-0040.md
Shaptic: Good idea! I've expanded the issuance model (incl. requiring a distribution account) in 37df0a5; wdyt?

ecosystem/sep-0040.md
Shaptic: I dropped references to `supply` in 87d6a3bb8367d7ff5285772cd9a6f381ff44dc44, but will leave this open if Frederic wants to chime in.

Shaptic:
Thanks for your comments, both! The latest push addresses your concerns, but also contains an important change re: the issuance model. I think we should require distribution accounts to make NFT supply more transparent. Let me know what you think about it, @leighmcculloch, since you noted that that section could use some more detail.

@tomerweller GitHub dismissed your review as stale, for some reason, so please give it another :heavy_check_mark: if you don't have other comments.

ecosystem/sep-0040.md
```diff
@@ -0,0 +1,198 @@
...
+This section describes the metadata itself rather than how to find it. 
+
+The metadata must be valid JSON and must contain the following **required** fields:
+
+| Field       | Description |
+| ----------- | ----------- |
+| `name` | a short name or title for the NFT |
+| `description` | a longer, more-detailed description of the NFT |
+| `type` | a [Media Type](https://www.iana.org/assignments/media-types/media-types.xhtml) describing the NFT's primary format (pointed to by `url`) |
+| `uri` **or** `url`  | a [URI](https://datatracker.ietf.org/doc/html/rfc3986) pointing to the actual NFT contents |
```
leighmcculloch, line 111: Could we pick one field name, or do we need both for consistency with other standards?

leighmcculloch (commented):
👏🏻 Few more thoughts after the edits:

ecosystem/sep-0040.md
```diff
@@ -0,0 +1,198 @@
...
+
+There are two **required** fields on the account itself:
+
+| Field       | Purpose |
+| ----------- | ------- |
+| `type`      | Describes how the "first hop" of metadata is defined, standard values include `ipfshash` and `uri` |
+| `hash`      | The [SHA3](https://csrc.nist.gov/publications/detail/fips/202/final) hash of the metadata contents |
+
+
+##### `type`
```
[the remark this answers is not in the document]
leighmcculloch: Actually, the big advantage to not using something like an nft namespace is the interoperability with tokens that maybe aren't conceptually NFTs but are otherwise compatible. This could also be seen as a disadvantage, but I think it's an advantage.

ecosystem/sep-0040.md
Shaptic: Well I want it to be `uri`, since that's broader, but Litemint already uses `url`, so I shoe-horned both in here. I'm not happy about it, though. We could say that the name is `url` but it can be any URI??

ecosystem/sep-0040.md
```diff
@@ -0,0 +1,198 @@
...
+
+**Backward Compatibility:** Existing NFT platforms essentially default to `type: ipfshash`. While this standard aims to free up the ecosystem from dependence on IPFS, it should also work with existing conventions that developed organically. To do so, **if you see an account with a lone `ipfshash` data entry, you may treat it as equivalent to an NFT account with `type: ipfshash`**.
+
+##### IPFS Storage
+If the `type` is `ipfshash`, the `ipfshash` key must point to an [IPFS Content IDentifier](https://docs.ipfs.io/concepts/content-addressing/) corresponding to the location of the metadata.
+
+If you are the issuer, make sure you understand the implications of using IPFS. You must use an IPFS lookup service to actually resolve to the metadata. For example, [this URL](https://cloudflare-ipfs.com/ipfs/QmaydGEshbL7HMQkD6FZPcMeEt6o15aC4p4P82VxQ7amjy) resolves to the Meridian 2021 Hat NFT via the Cloudflare IPFS service. If you're the issuer, you must also ensure that IPFS actually stores your data (by running your own node or entrusting a provider).
```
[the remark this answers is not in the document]
Shaptic: Really nice clarification, thanks!

ecosystem/sep-0040.md
leighmcculloch: A URI is just an identifier, a URL is a locator. Given that the intention here is to define a location that the NFT is stored, I think URL is fine and easily understood by a larger audience anyway.

There is probably not a huge amount of utility in saying it can be an identifier that isn't a locator, so I don't think URI is worth it.

leighmcculloch: To be clear `ipfs://...` is a URL.

ecosystem/sep-0040.md
```diff
@@ -0,0 +1,198 @@
...
+
+#### NFT Account Description
+A proper NFT account should specify its `homedomain`. It should also configure the asset in its domain's `stellar.toml` file (see [SEP-1](https://stellar.org/sep-1#currency-documentation)) if appropriate.
+
+There are two **required** fields on the account itself:
+
+| Field       | Purpose |
+| ----------- | ------- |
+| `type`      | Describes how the "first hop" of metadata is defined, standard values include `ipfshash` and `uri` |
+| `hash`      | The [SHA3](https://csrc.nist.gov/publications/detail/fips/202/final) hash of the metadata contents |
```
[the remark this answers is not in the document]
grempe: I came here to chime in with the same thought. SHA3-256 is not an "upgraded" version of SHA2-256, it is considered an alternative should a break in the SHA2 family be found.

I would make the suggestion that perhaps there should be a 'hashType' field in addition to the 'hash' field. This would allow any arbitrary hash function type to be used (e.g. `sha`, `sha2-256`, `sha2-512`, `sha3-256`, `sha3-512`, `blake3`). If this approach is taken an enum of accepted types could be consulted to determine if what is being submitted is an acceptable type. I would continue to allow SHA1, even though it is not considered to be as secure, only to handle the case where content can be stored in a public Git repository and pointed to by the hash of its commit.

I would also suggest specifying the hash encoding as part of the spec. Hexadecimal encoding is the more popular way to represent hashes, but Base64 is also common and more compact. You should choose one and validate it on submission.

Cheers.

ecosystem/sep-0040.md
grempe: URL. This is the way.

https://danielmiessler.com/study/difference-between-uri-url/

ecosystem/sep-0040.md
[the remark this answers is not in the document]
leighmcculloch: Storing hashType assumes a single hash type and so I think we store different hashes types in unique fields and applications can simply look for their fields for the hash types they support. i.e. store the sha256 in a `sha256` field. If someone decided to store a sha-3 hash too, they can instead or in addition store a `sha3` field.

We don't need to encode the hashes, but should be explicit and clear that the hash is stored in it's binary form in the data value. Applications can render the hash in whatever encoding they find most appropriate.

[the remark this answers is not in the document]
grempe: I would take these comments a step further and *require* that a hash of the content be provided in all cases. It has been a *major* shortcoming of existing NFT specs to allow linking to content that is not committed to by a hash of the content. It is so easy to hash the bytes of any content that it should never be skipped.

This can be the case even if IPFS is chosen as the `url`. An IPFS CID is not a file content hash and is harder for people to verify independently.

See:

https://docs.ipfs.io/concepts/hashing/#content-identifiers-are-not-file-hashes

By including a `hash` and `hashType` you make it much easier for client libs to verify the integrity of the file content using commonly found hashing libraries or stdlib functions. The IPFS hash is application specific code and its method of verification may change over time and should be considered a bonus integrity check on top of the content hash.

ecosystem/sep-0040.md
```diff
@@ -0,0 +1,198 @@
...
+{
+  "name": "A demonstration of NFT metadata",
+  "description": "This is a description of the cool NFT used in this demo.",
+  "type": "image/png",
+  "url": "https://cloudflare-ipfs.com/ipfs/QmXoypizjW3WknFiJnKLwHCnL72vedxjQkDDP1mXWo6uco/I/m/RickAstleyNeverGonnaGiveYouUp7InchSingleCover.jpg",
+  "issuer": "GALAXYVOIDAOPZTDLHILAJQKCVVFMD4IKLXLSZV5YHO7VY74IWZILUTO",
+  "code": "DEMOASSET"
+}
+```
+
```
grempe, line 129: Why not support the concept of a single NFT pointing to multiple pieces of content as an array of content items? This would allow a single NFT to point to a metadata file that in turn has pointers to multiple resources (e.g. a video, a thumbnail of the video, a certificate document).

This could be represented by a `data` top-level property that always contains an Array of at least one object. each object in the array could have additional attributes.

For example,

```json
{
  "name": "A demonstration of NFT metadata",
  "description": "This is a description of the cool NFT used in this demo.",
  "data": [
    {
     "description": " This is a cool image",
     "type": "image/png",
     "url": "https://filename.jpg",
     "hash": "deadbeef01",
     "hashType": "sha2-256",
    },
    {
     "description": " This is a cool video",
     "type": "video/mpeg",
     "url": "https://filename.mp3",
     "hash": "deadbeef02",
     "hashType": "sha2-512",
    }
  ],
  "issuer": "GALAXYVOIDAOPZTDLHILAJQKCVVFMD4IKLXLSZV5YHO7VY74IWZILUTO",
  "code": "DEMOASSET"
}
```

This would allow new types of multi-part assets to be referenced in a single NFT.

Something that will have to be addressed in any case is also how to canonicalize the contents of the metadata file so that it can be hashed consistently by both the creator and the verifier. JSON objects are not guaranteed to have their contents in any particular order. A standard method for ordering the metadata file so that it can be consistently hashed by all parties must be defined. Otherwise, it may not be possible to verify the on-chain metadata content hash.

See for example:

https://github.com/SkeLLLa/node-object-hash#readme

ecosystem/sep-0040.md
Shaptic: That's a good point, I think I mixed myself up double-checking my statements against too many RFCs.

ecosystem/sep-0040.md
```diff
@@ -0,0 +1,198 @@
...
+##### `type`
+The two possible values for the `type` field describe two standard ways to store the metadata, and the details will be in a data entry corresponding to this value. In other words, `type: ipfshash` means that you should look under the `ipfshash` key data entry which further defines how to find the metadata.
+
+All other values for the `type` are considered non-standard until they are approved for inclusion within this SEP.
+
+##### `hash`
+The `hash` is provided to ensure integrity of the metadata that the account is pointing to. This prevents the data at the destination from being modified without explicit knowledge and approval by the issuer.
+
+If it is excluded, the NFT **must not be considered genuine**, _unless_ the `type` is `ipfshash`, since the IPFS CID includes an integrity check on the data it points to.
```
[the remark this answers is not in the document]
Shaptic: If the hash doesn't match then you have absolutely have no guarantees about the authenticity of the metadata. I'll reword this :+1:

ecosystem/sep-0040.md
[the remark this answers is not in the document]
Shaptic: Sha3 was supposed to (a) future-proof and (b) align with Ethereum, but I appreciate the benefit of sticking to what we already use.

ecosystem/sep-0040.md
```diff
@@ -0,0 +1,198 @@
...
+In order to minimize the costs of on-chain storage, we have multiple levels of indirection of NFT metadata. The only thing the NFT account stores on-chain is a description of where to find the metadata.
+
+#### NFT Account Description
+A proper NFT account should specify its `homedomain`. It should also configure the asset in its domain's `stellar.toml` file (see [SEP-1](https://stellar.org/sep-1#currency-documentation)) if appropriate.
+
+There are two **required** fields on the account itself:
+
+| Field       | Purpose |
+| ----------- | ------- |
+| `type`      | Describes how the "first hop" of metadata is defined, standard values include `ipfshash` and `uri` |
```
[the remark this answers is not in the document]
Shaptic: Good point. Since `ipfshash` is a backwards-compatibility thing, `url` is a catch-all that should future-proof things.

ecosystem/sep-0040.md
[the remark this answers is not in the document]
grempe: Side note. Ethereum uses Kechack256 which is not the same as SHA3-256. Different internal padding results in different hash outputs. So using SHA3 would not end up aligning with Ethereum in any case.

https://ethereum.stackexchange.com/questions/30369/difference-between-keccak256-and-sha3#30373

ecosystem/sep-0040.md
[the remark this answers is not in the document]
MonsieurNicolas: We should really discourage using many subentries here to avoid further ledger bloat

ecosystem/sep-0040.md
```diff
@@ -0,0 +1,198 @@
...
+This SEP defines an [issuance model](#issuance-model) that defines how NFT assets are issued on Stellar and a [metadata specification](#nft-metadata-specification) that defines how data about the NFT is captured. The metadata is stored in a way such that the Stellar network guarantee its authenticity.
+
+### Issuance Model
+Standard issuance of NFTs must take the following steps:
+
+  1. Create an account to represent the NFT: `CreateAccountOp(A, <reserve>)`
+  2. Configure the NFT's metadata pointer, as described [below](#nft-account-description): `ManageDataOp("type", "...")`
+  3. Form an asset describing your NFT: `Code:Issuer` (where issuer is `A`)
+  4. Create a distribution account: `B = CreateAccountOp()` and `ChangeTrust("Code:Issuer", <limit>)`
+  5. Issue the tokens: `PaymentOp(B, "Code:Issuer", <amount>)`
```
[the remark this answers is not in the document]
Shaptic: I amended it to put the pseudocode in a separate block, wdyt?

ecosystem/sep-0040.md
```diff
@@ -0,0 +1,198 @@
...
+  6. (Optional) Lock the issuing account: `SetOptionsOp(MasterWeight: 0)`
+
+Note that without this final step, there are concerns around both immutability and supply: an account may modify its NFT metadata pointer or issue more tokens unless it's locked down. Thus, Step 6 is recommended.
```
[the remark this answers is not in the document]
Shaptic: Great point re: impact; addressed.

ecosystem/sep-0040.md
```diff
@@ -0,0 +1,198 @@
...
+## Specification
+This SEP defines an [issuance model](#issuance-model) that defines how NFT assets are issued on Stellar and a [metadata specification](#nft-metadata-specification) that defines how data about the NFT is captured. The metadata is stored in a way such that the Stellar network guarantee its authenticity.
+
+### Issuance Model
+Standard issuance of NFTs must take the following steps:
+
+  1. Create an account to represent the NFT: `CreateAccountOp(A, <reserve>)`
+  2. Configure the NFT's metadata pointer, as described [below](#nft-account-description): `ManageDataOp("type", "...")`
+  3. Form an asset describing your NFT: `Code:Issuer` (where issuer is `A`)
+  4. Create a distribution account: `B = CreateAccountOp()` and `ChangeTrust("Code:Issuer", <limit>)`
```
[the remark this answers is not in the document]
Shaptic: Another great idea, done.

ecosystem/sep-0040.md
Shaptic: The multi-part suggestion is really nice, thanks! We want to maximize compatibility, though, so we may have to resort to "either a `data` array or the same fields on the root object" which I hate but aligns with existing NFTs.

Regarding JSON ordering: The hash should be computed on the raw metadata file rather than the "loaded" JSON blob in order to avoid this. It's an important subtlety, so I'll note it.

ecosystem/sep-0040.md
```diff
@@ -0,0 +1,198 @@
...
+
+Other issuance models may exist and may be compatible with the following specification, but this is the standard issuance model on Stellar.
+
+### NFT Specification
+This specification is partially inspired by Ethereum's [EIP-721 specification](https://eips.ethereum.org/EIPS/eip-721#specification) on NFTs; however, it is generalized to arbitrary asset types (i.e. not necessarily images).
+
+In order to minimize the costs of on-chain storage, we have multiple levels of indirection of NFT metadata. The only thing the NFT account stores on-chain is a description of where to find the metadata.
+
+#### NFT Account Description
+A proper NFT account should specify its `homedomain`. It should also configure the asset in its domain's `stellar.toml` file (see [SEP-1](https://stellar.org/sep-1#currency-documentation)) if appropriate.
```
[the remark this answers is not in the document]
Shaptic: There are two parts: we already have a way to make an asset "verified" (via the toml) and also "render" it in wallets (logos and such). NFTs should be compatible with those existing methods, but also have their own spec for _new_ tools.

ecosystem/sep-0040.md
leighmcculloch: We could support the URL/ipfshash being to a directory, and inside the directory each asset code could get its own .json, so that an account can issue multiple asset codes but share the issuing account. This should work with most URL protocols.

ecosystem/sep-0040.md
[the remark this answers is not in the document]
Shaptic: Blegh, I hate that GitHub doesn't show me new comments as they come in. My comment was only in reply to leigh's first comment...

ecosystem/sep-0040.md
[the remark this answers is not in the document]
Shaptic: Agreed, we're trying to heavily limit them and keep most data off-chain.

ecosystem/sep-0040.md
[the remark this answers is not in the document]
Shaptic: Good thoughts here, thanks both! So we'll have an enumeration of possible keys that correlate to hashes? I don't really like that, and would prefer a type field. It makes it easier to look it up, no? Otherwise you have clients doing strange for-loops looking for a hash entry.

Regarding IPFS, this may be a product of my own ignorance, but I think CIDs still give you integrity, right? You know that the content at the CID is exactly the content put there by the issuer since if it changed, the CID would also change. There's no _need_ to verify that hashes match, since it's baked into the CID, right? To be clear, I do still think the hash is valuable for integrity (and especially in the non-IPFS case), I'm just trying to clarify my own understanding.

ecosystem/sep-0040.md
```diff
@@ -0,0 +1,212 @@
...
+
+```json
+{
+  "name": "A demonstration of NFT metadata",
+  "description": "This is a description of the cool NFT used in this demo.",
+  "type": "image/png",
+  "url": "https://cloudflare-ipfs.com/ipfs/QmXoypizjW3WknFiJnKLwHCnL72vedxjQkDDP1mXWo6uco/I/m/RickAstleyNeverGonnaGiveYouUp7InchSingleCover.jpg",
+  "issuer": "GALAXYVOIDAOPZTDLHILAJQKCVVFMD4IKLXLSZV5YHO7VY74IWZILUTO",
+  "code": "DEMOASSET"
+}
```
leighmcculloch, line 139: Could we add the hash of the asset in the metadata?

Shaptic: Good call, but I'll echo what I said [above](): I still don't think we should have an enumeration-based approach for hashing. We should either have a type (so `type: sha256` and then `sha256: <hash>`), a standard hashing algorithm under `hash:`, or some short whitelist of hashes (like `sha256: <hash>`, `blake3: <hash>`, etc.)

grempe: > Could we add the hash of the asset in the metadata?

If I understand what you are proposing @leighmcculloch, I think you would lose the major benefit of having a cryptographic commitment to the content on-chain. The hash that commits to the metadata file must be on-chain. A malicious hoster of the metadata file could change the content of the metadata (including an embedded hash) to entirely new content.  This is similar to the problem that exists today with ERC721 were images hosted and linked to from an NFT could be swapped out or removed with the holder of the NFT not being able to detect that change.

> I still don't think we should have an enumeration-based approach for hashing. 

I would concur with this opinion.  I think there is a lot to be said for the simplicity of a `hash` field that contains arbitrary hash bytes, and a `type` or `hashType` field that contains the description of what the `hash` bytes represent. This is the scheme we are taking at Truestamp.

With this scheme there is always a consistent property name to find what you are looking for. It was mentioned that having a JSON key per hash type allows for multiple hashes to be associated with an NFT, but is this a real use-case? Under which circumstances would a developer need to store both a SHA2-256 and BLAKE3 hash of the same metadata file?

This same scheme can then be applied to having a `data` array at the top level of the metadata file that contains an array of objects representing at least one, but potentially many, image, video, document, etc objects that the metadata points to.  The metadata should also then have a `hash` and `hashType` field for each of those content objects that commits to the hash of the contents of each file. The same `hashType` does not have to be used for each content object in the `data` array.

This affords a simplicity for the developer who can always just look for a single `hash` and `hashType` value at both the on-chain and metadata levels to verify that the metadata file, and all content objects it links to, are probably unchanged..

FredericRezeau:
Hi everyone,

I am just coming across this proposal, hopefully not too late as I wanted to share some major concerns, impracticality and issues, I see with it:

> Other issuance models may exist and may be compatible with the following specification, but this is the standard issuance model on Stellar.

This statement illustrates my first concern perfectly. Since Stellar assets are not created, transferred, and traded via contract there is no rationale for a rigid and extraneous standard for issuance. We have assets with **properties**, not a new class of assets that would require an arbitrary issuance model adding friction to an otherwise and so far, efficient, and open process that can fit all use cases. i.e. supply is an asset property already and token creators and issuers on Stellar should naturally be able to use it, in the first place, to create limited supply items without any other steps or risking being non-standard. 

Similarly, we already know, de-facto, when an asset is nonfungible and/or permanently limited in supply because its properties are intrinsic. Again, unlike for Ethereum, no binding contract requiring APIs or issuance framework is needed to infer non-fungibility or supply for assets on Stellar so this extraneous standardization is limiting for no sensible reason. Any such asset is valid as no repudiation is possible on their properties regardless.

> The recommended metadata pointer method is via url since we don't want to enforce reliance on IPFS

I would be advocating for leaving storage off any specs to not limit users on Stellar to what we think would work today, clearly limiting usage for tomorrow. To give you a real-world example, multiple assets per issuer is already a use case we had to cover to reduce issuance costs for a Brazilian company seeking to use Stellar to issue 10,000 NFTs last year (generative art with ability to mint on Ethereum). Where would they fit in these specs despite having perfectly valid, de facto, non-fungible tokens?

> If you are the issuer and decide to use IPFS, make sure you understand its implications. You must ensure that the resource is addressable and accessible on the IPFS network by running your own IPFS node or entrusting a provider to do so

Content distribution does not need to rely on IPFS nodes and there are many use cases (OTC/commissions, unlockable content, confidential data... ) where another medium than a public IPFS node would be used for delivering such content while using a CID stored on the asset to prove its authenticity. Besides, CID addressing already contains a hash so there is redundancy with the SHA-256 data entry (could save 0.5 XLM/NFT) and that's another point to make the case _for not_ standardizing storage. As we see, any specified method could quickly become redundant/obsolete/unusable in current/future use cases.

NFT use cases are evolving fast, embracing more diverse functions for global digital good businesses (metaverse, gaming, music, publishing and so on) and issuers should remain as free as possible to use the issuance model (e.g. multiple assets per account, no distribution accounts, I can hardly image forcing creators on Stellar generating and operating yet another account...) and storage method (e.g. cross chain compatibility for generative art...) that fit their needs and resources. Implementation details for their chosen process, decentralized and immutable storage does not need to be standardized for wallets and ecosystem services to be able to deliver the end-user experience because we can leverage the existing standard (SEP-1) for describing assets on the network for interoperability (in this regard, it would be much more useful, and not limiting, to specify an extension to SEP-1 for adding some dedicated fields as we go along to improve NFT creators and user experience).

I realize this may be an unpopular standpoint here however, I really hope we do not see such rigid specifications for issuance and storage on Stellar as it would clearly be limiting our freedom to onboard, execute and adapt in this space for the future.

Hope this helps.

All the best,
Fred

leighmcculloch:
>Implementation details for their chosen process, decentralized and immutable storage does not need to be standardized for wallets and ecosystem services to be able to deliver the end-user experience because we can leverage the existing standard (SEP-1) for describing assets on the network for interoperability (in this regard, it would be much more useful, and not limiting, to specify an extension to SEP-1 for adding some dedicated fields as we go along to improve NFT creators and user experience).

+1 This makes a lot of sense to me.

If SEP-1 is updated to add fields to describe an NFT, do you think there would be no utility in wallets being able to interpret and understand what data is stored on chain? I would have thought yes, but I don't really know. Do you see them as being more application specific and therefore not portable? Or is the idea they'll be portable but no standard will define them because realistically any standard here would stifle innovation?

On a separate note, do you think simple NFTs could exist without any data entries if they could be described completely by SEP-1? This is something I've been wondering. Even simple IPFS stored NFTs could possibly leverage SEP-1 if the home domain field of an account could store the IPFS hash of a directory containing a `.well-known/stellar.toml`.

FredericRezeau:
> Do you see them as being more application specific and therefore not portable? Or is the idea they'll be portable but no standard will define them because realistically any standard here would stifle innovation?

Yes, if the process fails to account for current and future requirements it stifles innovation as you said but also, it adds friction and hinders adoption (by analogy, it is like Stellar requiring fiat issuers to use a specific issuance and collateralization framework).

For example, as I understand, this standard would send 12,000 NFTs on Litemint to the legacy realm while only providing a more costly alternative for IPFS storage (2 data entries with redundant hash) to a perfectly valid issuance process. It would also require individual creators to use a distribution account preventing them from _minting directly_ on their wallet (a more natural process to them).

> If SEP-1 is updated to add fields to describe an NFT, do you think there would be no utility in wallets being able to interpret and understand what data is stored on chain?

Yes, regardless of the internal process, I see SEP-1 as the key for interoperability needs on Stellar. Currently the only requirement for an asset to be listed on Litemint is to have a valid TOML, so we can list items like Stellar Quest badges with on-chain data, or IPFS ones seamlessly and it seems the best way to go. With only a few added fields, a wallet could provide a complete experience without making any assumption about issuance and storage (e.g. media types, unit fields...).

>  Even simple IPFS stored NFTs could possibly leverage SEP-1 if the home domain field of an account could store the IPFS hash of a directory containing a .well-known/stellar.toml.

Interesting idea, there could be some issues with resolving such TOMLs (rate limits and slowness on IPFS gateways) and also more instances of copyright related issues as without domains it would become easier to create scams, yet it could fit some use cases.

Shaptic:
Thanks for your comments Fred! Deeply appreciated, and definitely not too late. I'll drop some inline replies.

> > Other issuance models may exist and may be compatible with the following specification, but this is the standard issuance model on Stellar.
> 
> This statement illustrates my first concern perfectly.

You can see in one of the later commits (797f2196, albeit one I forgot to push) that I realized the same thing. The issuance model is (a) not that important and (b) outside of the scope of this SEP if anything. The main reason for its inclusion was to provide context around how NFTs work on Stellar in general; I can definitely get behind dropping any sort of "standard" here.

> > The recommended metadata pointer method is via url since we don't want to enforce reliance on IPFS
> 
> I would be advocating for leaving storage off any specs to not limit users on Stellar to what we think would work today, clearly limiting usage for tomorrow. To give you a real-world example, multiple assets per issuer is already a use case we had to cover to reduce issuance costs for a Brazilian company seeking to use Stellar to issue 10,000 NFTs last year (generative art with ability to mint on Ethereum). Where would they fit in these specs despite having perfectly valid, de facto, non-fungible tokens?
> 
> > If you are the issuer and decide to use IPFS, make sure you understand its implications. You must ensure that the resource is addressable and accessible on the IPFS network by running your own IPFS node or entrusting a provider to do so
> 
> Content distribution does not need to rely on IPFS nodes and there are many use cases (OTC/commissions, unlockable content, confidential data... ) where another medium than a public IPFS node would be used for delivering such content while using a CID stored on the asset to prove its authenticity. Besides, CID addressing already contains a hash so there is redundancy with the SHA-256 data entry (could save 0.5 XLM/NFT) and that's another point to make the case _for not_ standardizing storage. As we see, any specified method could quickly become redundant/obsolete/unusable in current/future use cases.

I agree that IPFS isn't the only model, and neither should it get preferential treatment (hence the qualifier at the beginning of your quote). It was included for compatibility with existing conventions (e.g. Litemint's `ipfshash`).

Regarding the hash, I think it's viable to make it optional for a "self-certifying" URL like an IPFS CID. However, I _strongly_ believe that **integrity should remain on-chain**, as this solves a lot of trust issues around NFTs in general.

> Implementation details for their chosen process, decentralized and immutable storage does not need to be standardized for wallets and ecosystem services to be able to deliver the end-user experience because we can leverage the existing standard (SEP-1) for describing assets on the network for interoperability (in this regard, it would be much more useful, and not limiting, to specify an extension to SEP-1 for adding some dedicated fields as we go along to improve NFT creators and user experience).


In response to @leighmcculloch:
> On a separate note, do you think simple NFTs could exist without any data entries if they could be described completely by SEP-1?

Shaptic:
>  Currently the only requirement for an asset to be listed on Litemint is to have a valid TOML, so we can list items like Stellar Quest badges with on-chain data, or IPFS ones seamlessly and it seems the best way to go.

This is really interesting to me. First off, that's awesome. Second off, though, I'm really curious how the assets then actually render?

In general (I posted the last comment early on accident), I would love to see SEP-1 extended for NFTs (I think this is a really natural extension given the existing `[[CURRENCIES]]` standard). However, while it seems like that'd be the most extensible option, I don't see how it solves things like integrity and interoperability. At the end of the day, we still need required fields that represent something we all agree on (like `url` or `image`). Whether that lives at a metadata pointer on IPFS or within a TOML, I don't have a strong opinion, but we do still need to decide on what those fields are.

Shaptic:
> For example, as I understand, this standard would send 12,000 NFTs on Litemint to the legacy realm while only providing a more costly alternative for IPFS storage (2 data entries with redundant hash) to a perfectly valid issuance process.

Yeah, generative art is something we'd discussed offline. In accordance with this proposal (current discussion notwithstanding), we'd have the "first hop" of metadata describe the art, and a "second hop" describe each individual "generation".

ecosystem/sep-0040.md
Shaptic: > If I understand what you are proposing

I interpreted this as Leigh wanting an _additional_ hash of the file contents :thinking: Otherwise, I completely agree with you re: integrity on-chain. You could have the issuer sign the metadata and put that in the data itself, but that seems overly convoluted. We just need integrity, since the chain itself provides authenticity.

> I think there is a lot to be said for the simplicity of a hash field that contains arbitrary hash bytes, and a type or hashType field that contains the description of what the hash bytes represent. 

Is there any reason why we wouldn't just stick to a particular hashing algorithm? Every data entry saved is a win in our book.

I really liked the `data` array approach, but after Fred's comments (see below in the issue), it's a little unclear to me where and how we're going to describe NFTs in general now.

leighmcculloch: I was referring to adding a `sha256` field to the metadata JSON that contained a hash of the asset `url` file, assuming that the url is stored somewhere that isn't intrinsically verifiable like IPFS is.

leighmcculloch: I agree this discussion is mostly not relevant if we are exploring a focus on SEP-1.

FredericRezeau:
Many thanks for your answer.

> I agree that IPFS isn't the only model, and neither should it get preferential treatment

Essentially, by accepting all non-fungible tokens that implement SEP-1 on Litemint, we have built an open and future-forward NFT marketplace. SEP-40 would result in the opposite, closing the door to many projects, including existing ones.

So to me, the issue is not about any method being imposed preferentially, it is that **none of these methods should be imposed at all**. I think these belong to recommendations/best practices, not standard.

Concretely, SEP-40 is already demoting thousands of valid and cost-effective NFTs as legacy while promoting a much rigid alternative which excludes other valid existing use cases as well. By enforcing specific fields, SEP-40 can only cover a limited set of use cases efficiently, which is the opposite of what we tried to achieve so far.

> Whether that lives at a metadata pointer on IPFS or within a TOML, I don't have a strong opinion, but we do still need to decide on what those fields are.

In my view, there can never be an overlap of purpose between the stellar.toml (transient file) and the actual storage of the NFT data. The stellar.toml should only be used to maximize tokens interoperability, whether to hide implementation details for collateral vaults of fiat assets or the immutable storage for non-fungible assets.

> Second off, though, I'm really curious how the assets then actually render?

NFTs issued on Stellar can already _decently_ be presented to ecosystem services seamlessly by exposing immutable data (IPFS, off/on-chain...) via TOML leveraging the `image`, `name`, `desc` of `CURRENCIES`.

SEP-1 could be further extended to include generic content (e.g. based on [iana.org Media Types](https://www.iana.org/assignments/media-types/media-types.xhtml)) and preferred fractional unit for friendly display of e.g. stroop based assets.

> However, I strongly believe that integrity should remain on-chain, as this solves a lot of trust issues around NFTs in general.

A CID is not just a _self-certifying_ URL, it contains a (multi)hash of the data being addressed and therefore it is functionally the same as storing URL + SHA-256.

FredericRezeau:
To be more concrete and help understand the issues, I compiled a list of some **existing use cases** which illustrate the diversity of projects on Stellar **which would not work with this SEP**.

Also, while listing these projects I realized that the metadata digest is only `the hex-encoded SHA-256 digest of the content pointed to by url`, so if the metadata contains any other URL fields (e.g. video and audio), the content at these URLs would not be included in the digest (unless using a CID), this would be a significant flaw, especially in the first example I listed below with mixed content NFTs.

| Litemint Link   |      SEP Compliance Breach      |      Project/Use Case      |  Project Website |
|:------:|:-------------:|:------:|:-------------:
| [Mixed Content NFT (Audio,Video,Image)](https://litemint.com/items/GDYOJHBHBZFMYJL7BVCP2TYOXMMGJQB7B47WHIKTJDEXFZMA6JPHJFZD/SANOREPLY001) |  Multiple types | [This NFT](https://litemint.com/items/GDYOJHBHBZFMYJL7BVCP2TYOXMMGJQB7B47WHIKTJDEXFZMA6JPHJFZD/SANOREPLY001) includes an audio file, a digital image and animated video art by artist Shark Anthony. The current SEP would only allow one `type`. Many artists have released interdisciplinary work on Litemint. There is also a case for artists collaboration, here is [an example](https://litemint.com/items/GBLDU3OMBUHNQI3EGILEYR5S6BXKE2QL7XCUPCBEY2LSONF7BJDDKWXN/ALart) with audio by Jartz, an oil paint + animated video art by Arina Afonina. | [Shark Anthony Collection](https://litemint.com/explore?t=%23nftmusic&c=nftmusic&n=sharkant.litemint.store) |
| [Alart Collection](https://litemint.com/explore?q=GARC6SOBTN6GVLZUFJ3ETW6APZ35TEJG4Q5JIJ3DYDS6YAV55NTBIMNI) |  Hi-res art hosted offline by artist, unlocked supply | Newcomer to NFT, the artist initially leaned toward a more traditional approach to selling her digital art (external hosting with supply control). Litemint could accommodate all needs by not imposing a specific model. She is currently one of the top-selling artist on Litemint. | [alart.store](https://alart.store/) |
| [Antiblock Cakes](https://litemint.com/items/GCNIYUR56YY7MJ5VE4DQ6FRACZRPDIJNDOGZHRQJZ5XGAPZTYKLYK6I7/T00001) |  Multiple assets per issuer | External project by [ABRAJI](https://twitter.com/abraji) (Brazilian Association of Investigative Journalism) and [Congresso em Foco](https://twitter.com/congressoemfoco) to provide censorship resistant solution with crypto tokens (10,000 units). Low costs were the most important criteria in selecting their blockchain and since we could reduce their costs on Stellar by several orders of magnitude issuing multiple assets per account they opted for Stellar over Ethereum/_Classic_. Art is generative and they wanted to keep the ability for cross-chain minting with Ethereum. | [bolosantiblock.com.br](https://bolosantiblock.com.br/) |
| [litemint.io Cards](https://litemint.com/explore?q=litemint.io&t=litemint.io) |  Multiple assets per issuer, mutable content | Just like cryptokitties and other game projects, Stellar is used for decentralized ownership verification allowing players to freely exchange and trade their game cards and in-game currency. Yet, due to the needs to evolve the game balance and design for keeping our gameplay competitive, both the art and item properties are kept mutable (i.e. No data entry nor hash required, the value is not in item data immutability but in the gameplay experience).  | [litemint.io](https://litemint.io/) |
| [Stellar Quest Collection](https://litemint.com/explore?n=quest.stellar.org) |  On-chain Storage | I am aware of a separate spec for on-chain data storage but wanted to emphasize that regardless of the storage used, any asset providing a valid TOML can be listed on Litemint. Stellar quest items are no exception. Litemint displays the full collection, allowing users to trade these items, discover prices just like with any other asset. Importantly, should these items use a different/non-compliant on-chain storage, Litemint would still be able to list them.  | [quest.stellar.org](https://quest.stellar.org/) |
| [Cross-platform with Off-chain Unlockable Content](https://litemint.com/items/GCKR5E2J2GYFSDDG353NFOS4JFRKVMX2NDXENIK7DDCW6Y5BQDR2IY22/MT055) |  Cross-platform off-chain unlockable content | The ability to list items on different platforms on Stellar (interoperable via TOML) allows artists to use features that would otherwise not be available to them. This item is also [listed on StellarNFT](https://stellarnft.com/assets/MT055-GCKR5E2J2GYFSDDG353NFOS4JFRKVMX2NDXENIK7DDCW6Y5BQDR2IY22) which provides a way for artists to add off-chain unlockable content.  | [Moon Twinkie 055](https://litemint.com/items/GCKR5E2J2GYFSDDG353NFOS4JFRKVMX2NDXENIK7DDCW6Y5BQDR2IY22/MT055) |
| [Lusi Collection](https://litemint.com/explore?n=nft.lumenswap.io&t=Lusi%20Collection) |  Multiple assets per issuer, data entries naming | Another example of an external collection fully available on Litemint. The collection stores art on IPFS however they use a different naming convention and are issuing several items per issuers. Litemint can ignore the storage details and still provide a full experience. | [nft.lumenswap.io](https://nft.lumenswap.io/) |

By _not_ standardizing storage models, Stellar could remain open to all these use cases or any combination of them and allow future ones. In my view, the focus should be on providing the right tools for interoperability via TOML (as mentioned before, media types, fractionable units) and also refactoring the INT32 limit which affects pricing stroop items on the SDEX and forces us to use [cryptic orders to bypass this limitation](https://blog.litemint.com/nft-sdex-pricing/). These are more pressing issues imho and would go a long way in making Stellar a much friendlier environment for everyone and every project building in this space.

All the best,
Fred

Shaptic:
Fred, thanks so much for the use-case breakdown. It put something crucial into perspective for me, which is how diverse Litemint is in how many assets there are living outside of the `[username].litemint.store` garden.

It's hard to juggle everything we're discussing in long-form text, but I'm going to try my best to define a separation of concerns/replies here.


### On SEP Incompatibility
First, I just want to address the SEP incompatibilities you brought up. While some are true, others aren't:

  - Multi-type assets are supported just fine. There is a _primary_ `type` field, but other representations are optionally allowed.
  - You can store high-res art off-chain just fine, and since we've dropped requiring an issuance model, unlimited supply isn't a problem, either. (Though, I don't think unlimited supply of a token constitutes an NFT... Alart just seems like digital art with proof of purchase.)
  - Multiple assets per issuer was definitely a problem we discussed in this very issue; there just wasn't consensus yet.
  - Mutability: I don't think this is an NFT. The "non-[fungible](https://www.thefreedictionary.com/fungible)" aspect is violated: the older version of the card _is not interchangeable_ with the newer one. I don't know what to call this (a blockchain-backed asset?), but it's not an NFT. 

Hopefully that clarifies the SEP as well as my own perspective. Regardless, **all of this is moot**. Fundamentally, I think moving to SEP-1 is a great idea (with caveats below), so I'll get into that next.


### Litemint's approach
To start, I want to make sure I understand your perspective on the issue. Otherwise, any discussion built upon that misunderstanding falls apart. This is your core thesis:

> By accepting all non-fungible tokens that implement SEP-1 on Litemint, we have built an open and future-forward NFT marketplace.

Since nothing about SEP-1 makes a currency an NFT, what you mean is: **Litemint accepts listing any [SEP-1 currency](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0001.md#currency-documentation) as an NFT.** Please correct me if this is a misinterpretation, but I'll built the rest of my reply based on that understanding. From your perspective, we should abandon the idea of mandating a particular metadata format for NFTs (as defined in this SEP) and lean on SEP-1. I think this is brilliant and **I agree**.

Anyone who wants to list an NFT just needs to host a [SEP-1](https://stellar.org/protocol/sep-1) stellar.toml file. I love the simplicity and the interoperability. However, it begs a few questions:

  - Which fields under the [currency specification](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0001.md#currency-documentation) should we require for a "currency" to really mean an "NFT"?

  - How do we ensure integrity?

#### 1: Mandated Fields
I think we should require a particular subset of the available fields for a currency for it to be considered an NFT. An NFT needs to answer three questions:

  1. What is it?
  2. How many are there?
  3. Who created it?

For (3), I propose the following required fields: `code`, `issuer`, and `name`.

For (2), I propose requiring **one** of: `fixed_number`, `max_number`, or `is_unlimited`

For (1), I propose requiring either `image` (for backwards compatibility) or a **new** field `url`. 

This URL defines what the token represents. It could be the same as `image`, it could be an IPFS CID of some metadata, it could point to a random text file on the Internet, it could point to a directory of differing asset types, etc. The point is, it should be a generic way to represent what token ownership represents.

#### 2: Integrity
I'd really like to understand why we keep brushing this under the rug. Integrity is a crucial weak point of NFTs on Ethereum. This is an opportunity for us to improve on that! Yes, IPFS CIDs provide integrity, but not everyone will use it and our goal is interoperability.

I believe there **must** be a `hash` entry on the issuing account that ensures the integrity of the NFT contents, regardless of whether its held on-chain, on IPFS, or otherwise hosted by a third party. (Complexities like multi-asset NFTs are irrelevant: the issuer must _somehow_ store a reproducible hash of what they're "selling".)

In fact, I would even go so far as to propose that the **transaction memo** for an NFT purchase must contain a hash of the account and/or NFT at the point of payment. This means changes to the NFT after-the-fact are extremely clear.

Thanks for reading, and I look forward to your thoughts!

Shaptic:
Since (I think/hope?) we're approaching _local_ consensus, what do y'all think about throwing this discussion on the mailing list to reach the larger NFTs-on-Stellar ecosystem?

FredericRezeau:
Thank you for the clarification.

This is not _exactly_ my perspective :) but I am glad that we reached the same conclusion in that this SEP is not needed.

To clarify, Litemint accepting all non-fungible tokens that implement SEP-1 does not mean that any SEP-1 currency is a non-fungible token on Litemint, this would be a syllogism. Essentially, Litemint is treating all assets with the least possible distinction because the definitive purpose is leveraging digital good markets within decentralized environments, and within such horizontal context, creating distinctive property-based processes, just falls short and hinders the variety and/or ability to intersect use cases. This is also the reason I do not see a violation with supporting such intersecting and rich use cases because these projects are just as much of a reality that this very technology is capable of powering indistinctively.

This should hopefully clarify Litemint’ s approach and help with my answers below:

> Which fields under the currency specification should we require for a "currency" to really mean an "NFT"?

In my opinion and by extension there is no need for **required fields** to create a distinction between fungible and non fungible tokens. By analogy, feature detection wins over browser detection so, likewise, I would rather see ecosystem wallets detecting media fields and render appropriately. Whether an asset is non-fungible would be irrelevant at that point.

> How do we ensure integrity?

Couple points there:

First, getting back to your initial comment `However, I strongly believe that integrity should remain on-chain, as this solves a lot of trust issues around NFTs in general.`, trust would not be addressed by integrity check, especially on Stellar where it is virtually costless for anyone to just mint an asset with any data—digest will match. And for all instances of scam/theft/trust issues we had on Litemint marketplace, at no time data integrity was the issue. Trust requires a process for due diligence including copyright, licensing and this is why it is hard to implement (e.g. we run a KYC verification service to provide some helpers. A lot more is needed for users to fully assess trust).

Secondly, I agree that a hash should be used when needed (collision/uniqueness/alteration). However, in reference to my previous analogy with fiat anchors which are not required to provide attestation of reserves, audit reports and/or forced into a specific collateralization process, data integrity is also implementation specific. So it would be inconsistent to provide issuers with the ability to abstract their data storage implementation behind a TOML for interoperability while requiring them to use a possibly incompatible or redundant process for data integrity.

And yes, glad to take this discussion to the mailing list.

Many thanks.

Shaptic:
This standard proposal has been abandoned in preference for an _informational_ SEP that describes best practices and provides some SEP-1 extensions to better facilitate digital goods; see #1140.

PR #1114, opened 2022-02-04
SEP-2,5,6,10,11,12: Add PHP SDK and Flutter SDK implementations

christian-rogobete:


leighmcculloch (approved):
We normally require bumping the version number of SEPs, but given that there is zero content of the actual SEP content being changed, I think we can probably forego that here. It may even be misleading to bump it in this case. So I think this is good as is.

JakeUrban:
Agreed.

PR #1115, opened 2022-02-05
SEP: Add clarification to when versions should be updated

leighmcculloch:
### What
Add notes about when versions should be updated and exception cases.

### Why
We've had the rule for a while versions always get updated, and it is working well. There are some exception cases where it doesn't make sense to do it since these exception cases have nothing to do with the content of the SEP.


marcelosalloum (approved):
LGTM!

PR #1128, opened 2022-02-15
Transfer SEPS: deprecate `refunded` boolean, add `refunds` object to transaction records

### LINKED ISSUE #1122: SEP-24: Withdraw transaction refund handling.
It seems Plan-E is the clear consensus. 

#### Plan E: Add extra field `refund`

```
GET /transaction?id=xxx
{
    "id": "xxx",
    "kind": "withdrawal",
    "amount_in": 110, # amount of original transaction.
    ...,
    "status": "completed",
    // original stellar transaction ID
    stellar_transaction_id: "b9d0b2292c4e09e8eb22d036171491e87b8d2086bf8b265874c8d182cb9c9020",
    "refunds": {
        "type": "partial",
        "amount_refunded": "10",
        "transactions": [
            {
                "amount": 10,  # amount refunded in stellar network
                "stellar_transaction_id": "b9d0b2292c4e09e8eb22d036171491e87b8d2086bf8b265874c8d182cb9c9020"
            },
            {
                "amount": 15, # amount refunded in external network, if applicable.
                "external_transaction_id": "54321ab047a193c6fda1c47f5962cbcca8708d79b87089ababd57532c21c5402"
            }
        ]
    }
}

```
---

In the withdraw process, the user may transfer the USDC to the anchor. However, the anchor may need to refund the transaction due to technical or business reasons. 

In that scenario, there are two Stellar network transactions but only one SEP-24 transaction. The wallet won't be able to see the "anchor-initiated" refund transactions. 

The wallet should be able to see both the withdraw transaction and the "reverse" transaction.

The goal of this ticket is to choose one of the three alternatives (others are welcome) to propose a change for the refund process. 

There are three plans:

#### Plan A: Create a new value: `refund` ore `reverse` of the `kind` field.

In this plan, a new possible value, `refund` will be added to SEP-24. A new transaction schema will be added.

##### Pros: 
- New schema does not impact the existing `withdraw` and `deposit` transactions without overloading the meanings of the existing fields.
- First class citizen. Easier to maintain in the long run
- The wallet will be able to link multiple `refund` transactions with the existing `withdraw` or `deposit` transaction.

#### Cons:
- Impact to the existing wallet to interpret the `refund` kind.

The impact may be limited because for existing wallets working with existing anchor, the new `refund` kind has no impact. For wallets working with this new anchor, we can quickly implement. 

#### Plan B: Create a separate `deposit` transaction
The anchor can create a separate `deposit` transaction with the same amount of the `withdraw transaction. We may overload the `refunded` field to indicate a refund transaction. 

or use the `memo` field to indicate if this is a refund. 

##### Pros:
- Least impact to existing wallets.

##### Cons:
- The user may see `deposit` transaction not initiated from the wallet.
- The wallet won't be able to link both transactions together.

#### Plan C: Create a separate `deposit` transaction with `memo` 
The anchor may create a separate `deposit` transaction with a `memo` that links to the withdraw transaction.

##### Pros:
- Least impact to wallets.

##### Cons:
- Overloading the memo field may have impact with other functions.
- It is yet another protocol to between the anchor and the wallet to interpret the memo to link both transactions.

#### Plan D: Add extra field in the existing `transaction` schema
Add an extra field such as `sub_transactions` or `refund_transactions` to the Transaction schema to subsequent transactions related to the original one.

Example:
```
GET /transaction?id=xxx

{
    "id": "xxx",
    "kind": "withdrawal",
    "amount_in": 110,
    ...,
    status: "completed",
    // original stellar transaction ID
    stellar_transaction_id: "b9d0b2292c4e09e8eb22d036171491e87b8d2086bf8b265874c8d182cb9c9020",
    sub_transactions: [
        {
            "type": "refund",
            "total_refunded": "10",
            // refund transaction ID
            "transaction_id": "54321ab047a193c6fda1c47f5962cbcca8708d79b87089ababd57532c21c5402"
        },
       {
            "type": "future_type...",
           ...,
        }
       ]
}
```

##### Pros
- Non-breaking
- Can attach multiple subsequent transactions to the original one.

##### Cons
- TBD (cannot think of one at this moment)

#### Plan E: Add extra field `refund`
```
GET /transaction?id=xxx

{
    "id": "xxx",
    "kind": "withdrawal",
    "amount_in": 110, # amount of original transaction.
    ...,
    "status": "completed",
    // original stellar transaction ID
    stellar_transaction_id: "b9d0b2292c4e09e8eb22d036171491e87b8d2086bf8b265874c8d182cb9c9020",
    "refunds": {
        "type": "partial",
        "amount_refunded": "25",
        "transactions": [
            {
                "amount": 10,  # amount refunded in stellar network
                "stellar_transaction_id": "b9d0b2292c4e09e8eb22d036171491e87b8d2086bf8b265874c8d182cb9c9020"
            },
            {
                "amount": 15, # amount refunded in external network, if applicable.
                "external_transaction_id": "54321ab047a193c6fda1c47f5962cbcca8708d79b87089ababd57532c21c5402"
            }
        ]
    }
}

```

JakeUrban:
resolves #1122 

Deprecates the `refunded` attribute included in objects returned from `GET /transaction(s)` endpoints for SEP-6, 24, & 31 in favor of adding a `refund` object.

This refund object contains information about the type of refund, total amount refunded, and the individual payments made back to the sender / user. 

Note that SEP-31 refunds are assumed to be made via Stellar (since funds are always delivered to receiving anchors via Stellar).

### DIFF
--- a/ecosystem/sep-0006.md
+++ b/ecosystem/sep-0006.md
@@ -987,8 +987,8 @@ Name | Type | Description
 `stellar_transaction_id` | string | (optional) transaction_id on Stellar network of the transfer that either completed the deposit or started the withdrawal.
 `external_transaction_id` | string | (optional) ID of transaction on external network that either started the deposit or completed the withdrawal.
 `message` | string | (optional) Human readable explanation of transaction status, if needed.
-`refunded` | boolean | (**deprecated**, optional) This field is deprecated in favor of the `refund` object. Should be true if the transaction was refunded. Not including this field means the transaction was not refunded.
-`refund` | object | (optional) An object describing any on or off-chain refund associated with this transaction. The schema for this object is defined in the [Refund Object Schema](#refund-object-schema) section below.
+`refunded` | boolean | (**deprecated**, optional) This field is deprecated in favor of the `refunds` object. Should be true if the transaction was refunded. Not including this field means the transaction was not refunded.
+`refunds` | object | (optional) An object describing any on or off-chain refund associated with this transaction. The schema for this object is defined in the [Refund Object Schema](#refund-object-schema) section below.
 `required_info_message` | string | (optional) A human-readable message indicating any errors that require updated information from the user.
 `required_info_updates` | object | (optional) A set of fields that require update from the user described in the same format as [/info](#info).  This field is only relevant when `status` is `pending_transaction_info_update`.
 `claimable_balance_id` | string | (optional) ID of the Claimable Balance used to send the asset initially requested. Only relevant for deposit transactions.
@@ -1016,14 +1016,14 @@ Name | Type | Description
 Name | Type | Description
 -----|------|------------
 `type` | string | One of `partial` or `full`
-`total_refunded` | string | The total amount refunded to the user, in units of `amount_in_asset`. If `type` is `full`, this amount must match `amount_in`.
+`amount_refunded` | string | The total amount refunded to the user, in units of `amount_in_asset`. If `type` is `full`, this amount must match `amount_in`.
 `payments` | array | A list of objects containing information on the individual payments made back to the user. The schema for these objects is defined in the section below.
 
 #### Refund Transactions Object Schema
 
 Name | Type | Description
 -----|------|------------
-`refund_payment_id` | string | The payment ID that can be used to identify the refund payment. This is either a Stellar transaction hash or an off-chain payment identifier, such as a reference number provided to the user when the refund was initiated.
+`id` | string | The payment ID that can be used to identify the refund payment. This is either a Stellar transaction hash or an off-chain payment identifier, such as a reference number provided to the user when the refund was initiated.
 `id_type` | string | `stellar` or `external`.
 `amount` | string | The amount sent back to the user for the payment identified by `refund_payment_id`.
 
@@ -1068,12 +1068,12 @@ Example response:
       "completed_at": "2017-03-20T17:09:58Z",
       "stellar_transaction_id": "17a670bc424ff5ce3b386dbfaae9990b66a2a37b4fbe51547e8794962a3f9e6a",
       "external_transaction_id": "2dd16cb409513026fbe7defc0c6f826c2d2c65c3da993f747d09bf7dafd31093",
-      "refund": {
+      "refunds": {
         "type": "partial",
-        "total_refunded": "10",
+        "amount_refunded": "10",
         "payments": [
             {
-                "refund_payment_id": "b9d0b2292c4e09e8eb22d036171491e87b8d2086bf8b265874c8d182cb9c9020",
+                "id": "b9d0b2292c4e09e8eb22d036171491e87b8d2086bf8b265874c8d182cb9c9020",
                 "id_type": "stellar",
                 "amount": "10"
             }
--- a/ecosystem/sep-0006.md
+++ b/ecosystem/sep-0006.md
@@ -987,7 +987,7 @@ Name | Type | Description
 `stellar_transaction_id` | string | (optional) transaction_id on Stellar network of the transfer that either completed the deposit or started the withdrawal.
 `external_transaction_id` | string | (optional) ID of transaction on external network that either started the deposit or completed the withdrawal.
 `message` | string | (optional) Human readable explanation of transaction status, if needed.
-`refunded` | boolean | (**deprecated**, optional) This field is deprecated in favor of the `refunds` object. Should be true if the transaction was refunded. Not including this field means the transaction was not refunded.
+`refunded` | boolean | (**deprecated**, optional) This field is deprecated in favor of the `refunds` object. Should be true if the transaction was refunded in full. Partial refunds can still be expressed using the `refunds` object.
 `refunds` | object | (optional) An object describing any on or off-chain refund associated with this transaction. The schema for this object is defined in the [Refunds Object Schema](#refunds-object-schema) section below.
 `required_info_message` | string | (optional) A human-readable message indicating any errors that require updated information from the user.
 `required_info_updates` | object | (optional) A set of fields that require update from the user described in the same format as [/info](#info).  This field is only relevant when `status` is `pending_transaction_info_update`.
@@ -1015,7 +1015,6 @@ Name | Type | Description
 
 Name | Type | Description
 -----|------|------------
-`type` | string | One of `partial` or `full`
 `amount_refunded` | string | The total amount refunded to the user, in units of `amount_in_asset`. If `type` is `full`, this amount must match `amount_in`.
 `payments` | array | A list of objects containing information on the individual payments made back to the user. The schema for these objects is defined in the section below.
 
--- a/ecosystem/sep-0006.md
+++ b/ecosystem/sep-0006.md
@@ -988,7 +988,7 @@ Name | Type | Description
 `external_transaction_id` | string | (optional) ID of transaction on external network that either started the deposit or completed the withdrawal.
 `message` | string | (optional) Human readable explanation of transaction status, if needed.
 `refunded` | boolean | (**deprecated**, optional) This field is deprecated in favor of the `refunds` object. Should be true if the transaction was refunded. Not including this field means the transaction was not refunded.
-`refunds` | object | (optional) An object describing any on or off-chain refund associated with this transaction. The schema for this object is defined in the [Refund Object Schema](#refund-object-schema) section below.
+`refunds` | object | (optional) An object describing any on or off-chain refund associated with this transaction. The schema for this object is defined in the [Refunds Object Schema](#refunds-object-schema) section below.
 `required_info_message` | string | (optional) A human-readable message indicating any errors that require updated information from the user.
 `required_info_updates` | object | (optional) A set of fields that require update from the user described in the same format as [/info](#info).  This field is only relevant when `status` is `pending_transaction_info_update`.
 `claimable_balance_id` | string | (optional) ID of the Claimable Balance used to send the asset initially requested. Only relevant for deposit transactions.
@@ -1011,7 +1011,7 @@ Name | Type | Description
 * `too_large` -- deposit/withdrawal size exceeded `max_amount`.
 * `error` -- catch-all for any error not enumerated above.
 
-### Refund Object Schema
+### Refunds Object Schema
 
 Name | Type | Description
 -----|------|------------
@@ -1025,7 +1025,7 @@ Name | Type | Description
 -----|------|------------
 `id` | string | The payment ID that can be used to identify the refund payment. This is either a Stellar transaction hash or an off-chain payment identifier, such as a reference number provided to the user when the refund was initiated.
 `id_type` | string | `stellar` or `external`.
-`amount` | string | The amount sent back to the user for the payment identified by `refund_payment_id`.
+`amount` | string | The amount sent back to the user for the payment identified by `id`.
 
 Example response:
 
--- a/ecosystem/sep-0006.md
+++ b/ecosystem/sep-0006.md
@@ -1079,7 +1079,7 @@ Example response:
       "kind": "withdrawal",
       "status": "completed",
       "amount_in": "510",
-      "amount_out": "495",
+      "amount_out": "490",
       "amount_fee": "5",
       "started_at": "2017-03-20T17:00:02Z",
       "completed_at": "2017-03-20T17:09:58Z",
@@ -1087,13 +1087,13 @@ Example response:
       "external_transaction_id": "2dd16cb409513026fbe7defc0c6f826c2d2c65c3da993f747d09bf7dafd31093",
       "refunds": {
         "amount_refunded": "10",
-        "amount_fee": "0",
+        "amount_fee": "5",
         "payments": [
           {
             "id": "b9d0b2292c4e09e8eb22d036171491e87b8d2086bf8b265874c8d182cb9c9020",
             "id_type": "stellar",
             "amount": "10",
-            "fee": "0"
+            "fee": "5"
           }
         ]
       }
--- a/ecosystem/sep-0006.md
+++ b/ecosystem/sep-0006.md
@@ -1017,7 +1017,7 @@ Name | Type | Description
 -----|------|------------
 `amount_refunded` | string | The total amount refunded to the user, in units of `amount_in_asset`. If `type` is `full`, this amount must match `amount_in`.
 `amount_fee` | string | The total amount charged in fees for processing all refund payments. The sum of all `fee` values in the `payments` object list should equal this value.
-`payments` | array | A list of objects containing information on the individual payments made back to the user. The schema for these objects is defined in the section below.
+`payments` | array | A list of objects containing information on the individual payments made back to the user as refunds. The schema for these objects is defined in the section below.
 
 #### Refund Payment Object Schema
 
--- a/ecosystem/sep-0006.md
+++ b/ecosystem/sep-0006.md
@@ -1016,6 +1016,7 @@ Name | Type | Description
 Name | Type | Description
 -----|------|------------
 `amount_refunded` | string | The total amount refunded to the user, in units of `amount_in_asset`. If `type` is `full`, this amount must match `amount_in`.
+`total_fee` | string | The total amount charged in fees for processing all refund payments. The sum of all `fee` values in the `payments` object list should equal this value.
 `payments` | array | A list of objects containing information on the individual payments made back to the user. The schema for these objects is defined in the section below.
 
 #### Refund Transactions Object Schema
@@ -1024,7 +1025,8 @@ Name | Type | Description
 -----|------|------------
 `id` | string | The payment ID that can be used to identify the refund payment. This is either a Stellar transaction hash or an off-chain payment identifier, such as a reference number provided to the user when the refund was initiated.
 `id_type` | string | `stellar` or `external`.
-`amount` | string | The amount sent back to the user for the payment identified by `id`.
+`amount` | string | The amount sent back to the user for the payment identified by `id`, in units of `amount_in_asset`.
+`fee` | string | The amount charged as a fee for processing the refund, in units of `amount_in_asset`.
 
 Example response:
 
--- a/ecosystem/sep-0006.md
+++ b/ecosystem/sep-0006.md
@@ -1238,3 +1238,7 @@ If the information was malformed, or if the sender tried to update data that isn
 * PHP SDK: https://github.com/Soneso/stellar-php-sdk/blob/main/examples/sep-0006-transfer.md
 
 [SEP-38]: sep-0038.md
+
+## Changelog
+
+* `v3.10.0`: deprecate refunded boolean, add refund object to transaction records ([#1128](https://github.com/stellar/stellar-protocol/pull/1128))
--- a/ecosystem/sep-0006.md
+++ b/ecosystem/sep-0006.md
@@ -1070,14 +1070,15 @@ Example response:
       "stellar_transaction_id": "17a670bc424ff5ce3b386dbfaae9990b66a2a37b4fbe51547e8794962a3f9e6a",
       "external_transaction_id": "2dd16cb409513026fbe7defc0c6f826c2d2c65c3da993f747d09bf7dafd31093",
       "refunds": {
-        "type": "partial",
         "amount_refunded": "10",
+        "total_fee": "0",
         "payments": [
-            {
-                "id": "b9d0b2292c4e09e8eb22d036171491e87b8d2086bf8b265874c8d182cb9c9020",
-                "id_type": "stellar",
-                "amount": "10"
-            }
+          {
+            "id": "b9d0b2292c4e09e8eb22d036171491e87b8d2086bf8b265874c8d182cb9c9020",
+            "id_type": "stellar",
+            "amount": "10",
+            "fee": "0"
+          }
         ]
       }
     },
--- a/ecosystem/sep-0006.md
+++ b/ecosystem/sep-0006.md
@@ -1084,7 +1084,10 @@ Example response:
       "started_at": "2017-03-20T17:00:02Z",
       "completed_at": "2017-03-20T17:09:58Z",
       "stellar_transaction_id": "17a670bc424ff5ce3b386dbfaae9990b66a2a37b4fbe51547e8794962a3f9e6a",
-      "external_transaction_id": "2dd16cb409513026fbe7defc0c6f826c2d2c65c3da993f747d09bf7dafd31093",
+      "external_transaction_id": "1238234",
+      "withdraw_anchor_account": "GBANAGOAXH5ONSBI2I6I5LHP2TCRHWMZIAMGUQH2TNKQNCOGJ7GC3ZOL",
+      "withdraw_memo": "186384",
+      "withdraw_memo_type": "id",
       "refunds": {
         "amount_refunded": "10",
         "amount_fee": "5",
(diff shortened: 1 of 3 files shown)


ecosystem/sep-0006.md
```diff
@@ -1010,6 +1011,22 @@ Name | Type | Description
...
+-----|------|------------
+`type` | string | One of `partial` or `full`
+`total_refunded` | string | The total amount refunded to the user, in units of `amount_in_asset`. If `type` is `full`, this amount must match `amount_in`.
+`payments` | array | A list of objects containing information on the individual payments made back to the user. The schema for these objects is defined in the section below.
+
+#### Refund Transactions Object Schema
+
+Name | Type | Description
+-----|------|------------
+`refund_payment_id` | string | The payment ID that can be used to identify the refund payment. This is either a Stellar transaction hash or an off-chain payment identifier, such as a reference number provided to the user when the refund was initiated.
```
leighmcculloch, line 1026: This isn't going to be a unique ID since there's nothing stopping an off-chain system from having overlapping IDs with Stellar transactions. Is that okay?

ecosystem/sep-0006.md
```diff
@@ -1010,6 +1011,22 @@ Name | Type | Description
...
+`id_type` | string | `stellar` or `external`.
```
leighmcculloch, line 1027: By name spacing this field it can help to associate it.
```suggestion
`refund_payment_id_type` | string | `stellar` or `external`.
```

ecosystem/sep-0006.md
```diff
@@ -1010,6 +1011,22 @@ Name | Type | Description
 * `too_large` -- deposit/withdrawal size exceeded `max_amount`.
 * `error` -- catch-all for any error not enumerated above.
 
+### Refund Object Schema
+
+Name | Type | Description
+-----|------|------------
+`type` | string | One of `partial` or `full`
+`total_refunded` | string | The total amount refunded to the user, in units of `amount_in_asset`. If `type` is `full`, this amount must match `amount_in`.
```
leighmcculloch, line 1019: I suggest we use the term amount here so that it is not ambiguous to this being an amount or a count.
```suggestion
`amount_refunded` | string | The total amount refunded to the user, in units of `amount_in_asset`. If `type` is `full`, this amount must match `amount_in`.
```

ecosystem/sep-0006.md
```diff
@@ -987,7 +987,8 @@ Name | Type | Description
 `stellar_transaction_id` | string | (optional) transaction_id on Stellar network of the transfer that either completed the deposit or started the withdrawal.
 `external_transaction_id` | string | (optional) ID of transaction on external network that either started the deposit or completed the withdrawal.
 `message` | string | (optional) Human readable explanation of transaction status, if needed.
-`refunded` | boolean | (optional) Should be true if the transaction was refunded. Not including this field means the transaction was not refunded.
+`refunded` | boolean | (**deprecated**, optional) This field is deprecated in favor of the `refund` object. Should be true if the transaction was refunded. Not including this field means the transaction was not refunded.
```
leighmcculloch, line 990: Even if we are deprecating this field, I think we need to define for this field if it is true or false for partial refunds. Looking at some other APIs for inspiration, like Stripe's Charge API, I'd be inclined to say it should be false until the transaction is fully refunded.

On that note, it is unclear why we would deprecate this field. Is there significant value in us doing so? Since we are going to support the field indefinitely it might make the API less confusing to leave it not-deprecated.

ecosystem/sep-0006.md
```diff
@@ -987,7 +987,8 @@ Name | Type | Description
...
+`refund` | object | (optional) An object describing any on or off-chain refund associated with this transaction. The schema for this object is defined in the [Refund Object Schema](#refund-object-schema) section below.
```
leighmcculloch, line 991: I think we should name this field `refunds` because it is going to provide us with information about related refunds, and their aggregate affect. The name `refund` implies to me that it contains a single refund.

ecosystem/sep-0006.md
```diff
@@ -1010,6 +1011,22 @@ Name | Type | Description
 * `too_large` -- deposit/withdrawal size exceeded `max_amount`.
 * `error` -- catch-all for any error not enumerated above.
 
+### Refund Object Schema
+
+Name | Type | Description
+-----|------|------------
+`type` | string | One of `partial` or `full`
```
leighmcculloch, line 1018: I think we can drop this field, since it overlaps with the existing top-level `refunded` field. The top-level refunded field could indicate if the transaction has been fully refunded or not, and the presence of any refunds in the list will indicate that the transaction is partially refunded.

ecosystem/sep-0006.md
leighmcculloch, line 1026: I think, given that we have an associated `_type` field, that this field could be simply named `id`. This may be less confusing since we're saying less about what the identifier is.

leighmcculloch (commented):
👏🏻 Thanks for progressing this discussion to a PR. I have started a few conversations inline.

ecosystem/sep-0006.md
JakeUrban: If I understand this scenario, you're saying that an off-chain system could use the Stellar hash as their own ID? Therefore the ID would be both `stellar` and `external`?

ecosystem/sep-0006.md
JakeUrban: Thats fine by me 👍

ecosystem/sep-0006.md
JakeUrban: I'd rather change `refund_payment_id` to `id` and keep the `id_type` attribute as-is.

ecosystem/sep-0006.md
JakeUrban: >  I'd be inclined to say it should be false until the transaction is fully refunded.

That makes sense to me.

> it is unclear why we would deprecate this field. Is there significant value in us doing so?

I'm don't think the value is significant, but I think it makes more sense to have all refund-related info in a single object.

ecosystem/sep-0006.md
JakeUrban: I'm not a huge fan of this because I can see a client misinterpreting `refunded: false` to mean that there were no refund payments whatsoever. 

The SEP currently doesn't define the concept of a partial refund, so any client that checks the `refunded` attribute today almost certainly interprets `refunded: false` to mean no value was refunded. From that perspective, we'd be redefining what `refunded: false` means, which I think is more problematic than deprecating the field.

leighmcculloch: The type field may be more descriptive, but since we have another field that already communicates the same thing it feels like we're improving something that still carries the same cost of continuing to use the thing we already have.

leighmcculloch (dismissed):
Looks good to me. I think we should get weigh in from a variety of folks before merging.

ecosystem/sep-0006.md
marcelosalloum: Clients should be able to figure out if this is a total or partial refund by comparing `refunds.amount_refunded` with `amount_in` like you described in the field `amount_refunded` below.

ecosystem/sep-0006.md
```diff
@@ -1010,6 +1011,22 @@ Name | Type | Description
...
+`payments` | array | A list of objects containing information on the individual payments made back to the user. The schema for these objects is defined in the section below.
+
+#### Refund Transactions Object Schema
+
+Name | Type | Description
+-----|------|------------
+`id` | string | The payment ID that can be used to identify the refund payment. This is either a Stellar transaction hash or an off-chain payment identifier, such as a reference number provided to the user when the refund was initiated.
+`id_type` | string | `stellar` or `external`.
+`amount` | string | The amount sent back to the user for the payment identified by `id`.
+
```
marcelosalloum, line 1029: It's common for exchanges to charge a refund fee in case something goes wrong with the transaction. Shouldn't we include some kind of `fee` field as well?

marcelosalloum: For Stellar network payments, the anchor could add 2 transactions, one for the refund and the other for the refund fee, but that may not be practical for bank transfers.

ecosystem/sep-0006.md
marcelosalloum: Cool! In that case, can we update the parameter description to explicitly say it's `true` for a full refund?

marcelosalloum (commented):
It's looking good! Could you add a changelog section to these SEPs as (recently) suggested in the SEP template? https://github.com/stellar/stellar-protocol/blob/master/sep-template.md#changelog.

No need to include previous versions there, tracking this and future changes is great to start with.

ecosystem/sep-0006.md
JakeUrban: Oh, thats a good point, I'll remove it.

ecosystem/sep-0006.md
JakeUrban: Sure, I can add a `fee` attribute. Shouldn't we have a `total_fee` attribute or something similar alongside `amount_refunded` then?

leighmcculloch: Is the idea with the fee attribute that this would represented an amount that the user no longer has a right to, but was not included in the refund transaction? How would this affect the amount refunded?

For example:
- Original deposit $100.
- Anchor needs to refund $25, charges $5 per refund.
- Amount refunded = $25.
- Refunds includes a single refund with id and amount.
- What value does the `refunds.payments[0].amount` field hold? Is it $25, or $20?

JakeUrban: My understanding is that it would look like this:

* Original deposit $100
* Anchor needs to refund $25, and charges $5 per refund

So:
```json
{
  ...
  "amount_in": "100",
  "amount_out": "65",
  "amount_fee": "5",
  "refunded": false,
  "refunds": {
    "amount_refunded": "25",
    "total_fee": "5",
    "payments": [
      {
        "id": "<stellar hash>"
        "id_type": "stellar",
        "amount": "25",
        "fee": "5"
      }
    ]
  }
}
```

marcelosalloum: That makes sense to me too, since it's consistent with the amount_in description:



`amount_in` | string | (optional) Amount received by anchor at start of transaction as a string with up to 7 decimals. Excludes any fees charged before the anchor received the funds.
-- | -- | --

leighmcculloch: Got it, so is it correct that this formula should balance to zero?

```
amount_in
– amount_out
– amount_fee
+ refunds.amount_refunded
– refunds.total_fee (this should be named amount_fee I think)
```

And:
```
refunds.amount_refunded = sum(refunds.payments[].amount
refunds.total_fee = sum(refunds.payments[].fee)
```

JakeUrban: Yes thats right. In fact I think those formulas are worth including in the specifications.

leighmcculloch: >Amount received by anchor at start of transaction as a string with up to 7 decimals. Excludes any fees charged before the anchor received the funds.

The excludes statement is ambiguous to me. When I read it I can't tell if it means the amount_in has already had fees excluded, i.e. I'm depositing $100, but amount_in will be $95 because it excludes the fee I paid.

ecosystem/sep-0024.md
```diff
@@ -734,6 +735,23 @@ Name | Type | Description
...
 
+### Refunds Object Schema
+
+Name | Type | Description
+-----|------|------------
+`amount_refunded` | string | The total amount refunded to the user, in units of `amount_in_asset`. If `type` is `full`, this amount must match `amount_in`.
+`amount_fee` | string | The total amount charged in fees for processing all refund payments. The sum of all `fee` values in the `payments` object list should equal this value.
+`payments` | array | A list of objects containing information on the individual payments made back to the user. The schema for these objects is defined in the section below.
+
+#### Refund Transactions Object Schema
```
CassioMG, line 746: Should it say "Refund Payments Object Schema" instead to refer to `payments` ?

ecosystem/sep-0006.md
```diff
@@ -1010,6 +1011,39 @@ Name | Type | Description
 * `too_large` -- deposit/withdrawal size exceeded `max_amount`.
 * `error` -- catch-all for any error not enumerated above.
 
+### Refunds Object Schema
+
+Name | Type | Description
+-----|------|------------
+`amount_refunded` | string | The total amount refunded to the user, in units of `amount_in_asset`. If `type` is `full`, this amount must match `amount_in`.
+`amount_fee` | string | The total amount charged in fees for processing all refund payments. The sum of all `fee` values in the `payments` object list should equal this value.
+`payments` | array | A list of objects containing information on the individual payments made back to the user. The schema for these objects is defined in the section below.
```
leighmcculloch, line 1020:
```suggestion
`payments` | array | A list of objects containing information on the individual payments made back to the user as refunds. The schema for these objects is defined in the section below.
```

ecosystem/sep-0006.md
```diff
@@ -1010,6 +1011,39 @@ Name | Type | Description
...
+
+#### Refund Payment Object Schema
+
+Name | Type | Description
+-----|------|------------
+`id` | string | The payment ID that can be used to identify the refund payment. This is either a Stellar transaction hash or an off-chain payment identifier, such as a reference number provided to the user when the refund was initiated.
```
leighmcculloch, line 1026:
```suggestion
`id` | string | The payment ID that can be used to identify the refund payment. This is either a Stellar transaction hash or an off-chain payment identifier, such as a reference number provided to the user when the refund was initiated. This id is not guaranteed to be unique.
```

ecosystem/sep-0006.md
```diff
@@ -987,7 +987,8 @@ Name | Type | Description
 `stellar_transaction_id` | string | (optional) transaction_id on Stellar network of the transfer that either completed the deposit or started the withdrawal.
 `external_transaction_id` | string | (optional) ID of transaction on external network that either started the deposit or completed the withdrawal.
 `message` | string | (optional) Human readable explanation of transaction status, if needed.
-`refunded` | boolean | (optional) Should be true if the transaction was refunded. Not including this field means the transaction was not refunded.
+`refunded` | boolean | (**deprecated**, optional) This field is deprecated in favor of the `refunds` object. Should be true if the transaction was refunded in full. Partial refunds can still be expressed using the `refunds` object.
```
leighmcculloch, line 990: Marking this field as deprecated is still ambiguous to me. What are our plans with the field? For backwards compatibility we are still likely to support it indefinitely since removing support seems like it would have very little value. Do we need to deprecate it?

leighmcculloch, line 990: Suggested edits:
```suggestion
`refunded` | boolean | (**deprecated**, optional) This field is deprecated in favor of the `refunds` object. True if the transaction was refunded in full. False if the transaction was partially refunded or not refunded. For more details about any refunds, see the `refunds` object.
```

ecosystem/sep-0006.md
```diff
@@ -1209,3 +1255,7 @@ If the information was malformed, or if the sender tried to update data that isn
 * PHP SDK: https://github.com/Soneso/stellar-php-sdk/blob/main/examples/sep-0006-transfer.md
 
 [SEP-38]: sep-0038.md
+
+## Changelog
+
+* `v3.10.0`: deprecate refunded boolean, add refund object to transaction records ([#1128](https://github.com/stellar/stellar-protocol/pull/1128))
```
leighmcculloch, line 1261: Could we use capitals and typical grammar?
```suggestion
* `v3.10.0`: Deprecate refunded boolean. Add refund object to transaction records. ([#1128](https://github.com/stellar/stellar-protocol/pull/1128))
```

ecosystem/sep-0006.md
```diff
@@ -1010,6 +1011,39 @@ Name | Type | Description
...
+```
+amount_out = amount_in - amount_fee - refunds.amount_refunded - refunds.amount_fee
+```
+
+```
+refunds.amount_refunded = sum(refunds.payments[].amount)
+```
+
+```
+refunds.total_fee = sum(refunds.payments[].fee)
```
CassioMG, line 1044:
```suggestion
refunds.amount_fee = sum(refunds.payments[].fee)
```

ecosystem/sep-0006.md
```diff
@@ -1044,13 +1078,25 @@ Example response:
...
-      "amount_fee": "3",
+      "amount_fee": "5",
       "started_at": "2017-03-20T17:00:02Z",
       "completed_at": "2017-03-20T17:09:58Z",
       "stellar_transaction_id": "17a670bc424ff5ce3b386dbfaae9990b66a2a37b4fbe51547e8794962a3f9e6a",
-      "external_transaction_id": "2dd16cb409513026fbe7defc0c6f826c2d2c65c3da993f747d09bf7dafd31093"
+      "external_transaction_id": "2dd16cb409513026fbe7defc0c6f826c2d2c65c3da993f747d09bf7dafd31093",
+      "refunds": {
+        "amount_refunded": "10",
+        "amount_fee": "0",
```
CassioMG, line 1090: Could we add another example with fee values > 0 to help with the understanding?

ecosystem/sep-0024.md
```diff
@@ -734,6 +735,38 @@ Name | Type | Description
...
+```
+amount_out = amount_in - amount_fee - refunds.amount_refunded - refunds.amount_fee
+```
+
+```
+refunds.amount_refunded = sum(refunds.payments[].amount)
+```
+
+```
+refunds.total_fee = sum(refunds.payments[].fee)
```
CassioMG, line 768:
```suggestion
refunds.amount_fee = sum(refunds.payments[].fee)
```

ecosystem/sep-0031.md
```diff
@@ -464,6 +463,38 @@ Name | Type | Description
...
+```
+amount_out = amount_in - amount_fee - refunds.amount_refunded - refunds.amount_fee
+```
+
+```
+refunds.amount_refunded = sum(refunds.payments[].amount)
+```
+
+```
+refunds.total_fee = sum(refunds.payments[].fee)
```
CassioMG, line 495:
```suggestion
refunds.amount_fee = sum(refunds.payments[].fee)
```

ecosystem/sep-0006.md
```diff
@@ -1010,6 +1011,39 @@ Name | Type | Description
...
 * `error` -- catch-all for any error not enumerated above.
 
+### Refunds Object Schema
+
+Name | Type | Description
+-----|------|------------
+`amount_refunded` | string | The total amount refunded to the user, in units of `amount_in_asset`. If `type` is `full`, this amount must match `amount_in`.
+`amount_fee` | string | The total amount charged in fees for processing all refund payments. The sum of all `fee` values in the `payments` object list should equal this value.
+`payments` | array | A list of objects containing information on the individual payments made back to the user as refunds. The schema for these objects is defined in the section below.
+
```
lijamie98, line 1021: Are we assuming the same `asset_code` of the original transaction?

ecosystem/sep-0024.md
```diff
@@ -758,13 +791,25 @@ Example response:
       "status": "completed",
       "amount_in": "500",
       "amount_out": "495",
```
lijamie98, line 793: This should be 485. 
`amount_out = amount_in - amount_fee - refunds.amount_refunded - refunds.amount_fee`
485 = 500 - 5 - 10 - 0

ecosystem/sep-0006.md
```diff
@@ -1044,13 +1078,25 @@ Example response:
       "id": "82fhs729f63dh0v4",
       "kind": "withdrawal",
       "status": "completed",
-      "amount_in": "500",
+      "amount_in": "510",
       "amount_out": "495",
```
lijamie98, line 1082: This should be 485.
amount_out = amount_in - amount_fee - refunds.amount_refunded - refunds.amount_fee
485 = 500 - 5 - 10 - 0

ecosystem/sep-0006.md
JakeUrban: We're assuming all refund amounts are in units of `amount_in_asset`

ecosystem/sep-0006.md
```diff
@@ -1044,13 +1078,25 @@ Example response:
...
       "started_at": "2017-03-20T17:00:02Z",
       "completed_at": "2017-03-20T17:09:58Z",
       "stellar_transaction_id": "17a670bc424ff5ce3b386dbfaae9990b66a2a37b4fbe51547e8794962a3f9e6a",
-      "external_transaction_id": "2dd16cb409513026fbe7defc0c6f826c2d2c65c3da993f747d09bf7dafd31093"
+      "external_transaction_id": "2dd16cb409513026fbe7defc0c6f826c2d2c65c3da993f747d09bf7dafd31093",
+      "refunds": {
+        "amount_refunded": "10",
+        "amount_fee": "0",
+        "payments": [
+          {
```
lijamie98, line 1092: Can we add an example of `"id_type":  "external"`?

ecosystem/sep-0024.md
```diff
@@ -758,13 +791,25 @@ Example response:
...
       "completed_at": "2017-03-20T17:09:58Z",
       "more_info_url": "https://youranchor.com/tx/242523523",
       "stellar_transaction_id": "17a670bc424ff5ce3b386dbfaae9990b66a2a37b4fbe51547e8794962a3f9e6a",
       "external_transaction_id": "2dd16cb409513026fbe7defc0c6f826c2d2c65c3da993f747d09bf7dafd31093",
-      "claimable_balance_id": "00000000c2d8c89264288dbde8488364fd3fd30850fd4e7fbf6d1e9809702558afa4fdea"
+      "claimable_balance_id": "00000000c2d8c89264288dbde8488364fd3fd30850fd4e7fbf6d1e9809702558afa4fdea",
+      "refunds": {
+        "amount_refunded": "10",
+        "amount_fee": "0",
+        "payments": [
```
lijamie98, line 804: Can we add an example of "id_type": "external"?

ecosystem/sep-0031.md
```diff
@@ -504,6 +543,34 @@ Example response:
...
+      "amount_fee": "5",
+      "started_at": "2017-03-20T17:05:32Z",
+      "stellar_transaction_id": "b9d0b2292c4e09e8eb22d036171491e87b8d2086bf8b265874c8d182cb9c9020",
+      "stellar_account_id": "GBRPYHIL2CI3FNQ4BXLFMNDLFJUNPU2HY3ZMFSHONUCEOASW7QC7OX2H",
+      "stellar_memo": "123456789",
+      "stellar_memo_type": "id",
+      "refunds": {
+        "amount_refunded": "10",
+        "amount_fee": "0",
+        "payments": [
```
lijamie98, line 562: Can we add an example of "id_type": "external"?

ecosystem/sep-0006.md
leighmcculloch: Can we make this explicit in the doc?

ecosystem/sep-0006.md
JakeUrban: I just updated the example to use a non-zero refund fee.

ecosystem/sep-0031.md
JakeUrban: There is no `id_type` attribute (and no `external` attribute value) for SEP-31 transactions, since the receiving anchor always receives payments via Stellar.

ecosystem/sep-0006.md
```diff
@@ -1010,6 +1011,39 @@ Name | Type | Description
 * `too_large` -- deposit/withdrawal size exceeded `max_amount`.
 * `error` -- catch-all for any error not enumerated above.
 
+### Refunds Object Schema
+
+Name | Type | Description
+-----|------|------------
+`amount_refunded` | string | The total amount refunded to the user, in units of `amount_in_asset`. If `type` is `full`, this amount must match `amount_in`.
```
CassioMG, line 1018: Where is this `type` definition? Should we remove it?

CassioMG:
> @JakeUrban FYI the PR title still says "refund" instead of "refunds"

Same for the PR description

ecosystem/sep-0006.md
```diff
@@ -1010,6 +1011,39 @@ Name | Type | Description
...
+`id_type` | string | `stellar` or `external`.
+`amount` | string | The amount sent back to the user for the payment identified by `id`, in units of `amount_in_asset`.
+`fee` | string | The amount charged as a fee for processing the refund, in units of `amount_in_asset`.
+
+### Amount Formulas
+
+The following should hold true for all transaction records, assuming `amount_in_asset` and `amount_out_asset` are the same. If they are different, the following should still hold true after converting all amounts to units of one of the assets.
+
+```
+amount_out = amount_in - amount_fee - refunds.amount_refunded - refunds.amount_fee
```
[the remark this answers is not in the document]
JakeUrban: Yes, if the anchor charges fees for processing a refund.

accordeiro (approved):

PR #1129, opened 2022-02-15
CAP-40: Make some text clearer

leighmcculloch:
### What
Removed the discussion about how a transaction hash is produced from the semantics, and moved it to the design rationale.

### Why
Feedback from @Shaptic highlighted that the section in the semantics section about how the transaction hash was constructed was confusing as it was presented like a new process that takes place, when in fact this is just the regular process all transaction hashes go through.

I moved it to the design rationale because that content relates more to how the changes will be used, and it is not really semantics of the change itself.

### DIFF
--- a/core/cap-0040.md
+++ b/core/cap-0040.md
@@ -117,23 +117,27 @@ of 64 bytes and an ed25519 public key. A signature for the signer is the result
 of signing the payload with the private key that the public key is derived.
 
 The ed25519 signed payload signer is usable everywhere existing signers may be
-used, including in `extraSigners<1>` added in [CAP-21].
+used, including in the `extraSigners` transaction precondition added in
+[CAP-21].
 
 #### Signature
 
-The signature of a ed25519 signed payload signer is the raw ed25519 signature of
-the signer's payload using the private key that derives the signer's ed25519
+The signature of an ed25519 signed payload signer is the raw ed25519 signature
+of the signer's payload using the private key that derives the signer's ed25519
 public key.
 
-Unlike other signatures in the Stellar protocol, the payload is not combined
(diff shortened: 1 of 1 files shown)


Shaptic (approved):
Goes above & beyond my feedback about readability/clarity :clap:

core/cap-0040.md
```diff
@@ -163,10 +167,17 @@ This proposal provides a primitive that makes it possible to construct a set
...
-This proposal makes this possible when the signer is used with the
-`extraSigners` precondition introduced in [CAP-21], and does so without
-introducing any new state to be stored on the ledger, allowing the signing
-dependency to exist without creating ledger entries.
+This proposal makes this possible since a Stellar transaction hash can be
+specified as the payload of an ed25519 signed payload signer. A Stellar
+transaction hash is the SHA-256 hash of the `TransactionSignaturePayload`
+containing the Stellar transaction. A signature for the signer will also be an
+ed25519 signature of the other Stellar transaction that could be attached to the
+other transaction.
```
Shaptic, line 175: > A signature for the signer will also be an ed25519 signature of the other Stellar transaction that could be attached to the other transaction.

I _think_ I follow this, but "other transaction" seems overloaded here? Is this saying that a payload signature here could be identical to another transaction's "normal" signature?

leighmcculloch: Ah, thanks. Addressed.

leighmcculloch: The signatures won't be identical, but one can be ported to the other without any secret information.

leighmcculloch: They aren't identical because the hint will be different.

Shaptic (approved):
Nice! Thanks for addressing this so promptly.

PR #1134, opened 2022-02-21
Correction to the argument supporting the liquidity pool's trading rule

gramseyer:
The argument as written does not fully justify the protocol's choice of liquidity pool trading rule (there are quite a few formulae that would satisfy the earlier argument).  Namely, it shows that using the specified formula, the liquidity pool trades in the expected directions, but does not justify trade amounts.

I also switched notation of the form a / b / c to a / (b * c), which is (in my view) easier to read.

### DIFF
--- a/core/cap-0045.md
+++ b/core/cap-0045.md
@@ -355,9 +355,8 @@ Y' = (pX + Y) / (2 - F)
     < (Y + (1 - F)Y) / (2 - F)
     = Y
 ```
-,
 
-so the liquidity pool, as expected, sells `Y` and purchases `X`.
+Thus, the liquidity pool, as expected, sells `Y` and purchases `X`.
 
 Next, suppose that `(1/p) < (1 - F) X/Y`.
 Then the liquidity pool sells `X` and buys `Y` until
@@ -380,9 +379,8 @@ Y' = (pX + Y) (1 - F) / (2 - F)
     > (Y + (1 - F)Y) / (2 - F)
     = Y
 ```
-,
 
-so the liquidity pool, as expected, sells `X` and purchases `Y`.
+Thus, the liquidity pool, as expected, sells `X` and purchases `Y`.
 
 ### Demand from an order
 Tatonnement does not handle discontinuities well. The demand from an order is
--- a/core/cap-0045.md
+++ b/core/cap-0045.md
@@ -354,7 +354,8 @@ and
 Y' = (pX + Y) / (2 - F)
     < (Y + (1 - F)Y) / (2 - F)
     = Y
-```,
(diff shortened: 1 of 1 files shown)


core/cap-0045.md
```diff
@@ -334,34 +334,17 @@ Y' = Y .
...
-   = X
-Y' = (pX + Y) (1 - F) / (2 - F)
-   = pX (1 - F) / (2 - F) + Y (1 - F) / (2 - F)
-   > X (1 - F) / (2 - F) / ((1 - F) X / Y) + Y (1 - F) / (2 - F)
-   = Y / (2 - F) + Y (1 - F) / (2 - F)
-   = Y
-```
-as expected based on the description that the liquidity pool sells `X` and buys
-`Y`.
+First, suppose that `p (1 - F) > Y/X`.
```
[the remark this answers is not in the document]
gramseyer: I mixed up the order of the cases in the last section, so I fixed that (`p (1 - F) > Y/X` <=> `1/p < (1 - F) X/y`)

Either notation is fine with me, but you're right that we should standardize on one.  I usually think about this as in "if p(x) rises, then p rises and thus X is sold" so my notes used the first expression (which is how this slipped into the commit)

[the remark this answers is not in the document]
gramseyer: I re-added a shorter proof of the selection I deleted, to demonstrate that the pool actually makes trades in the direction that it should.

core/cap-0045.md
```diff
@@ -334,34 +334,37 @@ Y' = Y .
...
-as expected based on the description that the liquidity pool sells `X` and buys
-`Y`.
+First, suppose that `p < (1 - F) Y/X`.  
+Then the liquidity pool sells `Y` and buys `X` until
+`p = (1 - F) Y'/X'`.  Combining this equation with the budget constraint gives
+`pX + Y = Y' + (1 - F)Y'`.
+Rearranging gives `Y' = (pX + Y) / (2 - F)` and `X' = (pX + Y) (1 - F) / (p * (2 - F))`.
+
+Observe that in this case, 
+`X' = (pX + Y) (1 - F) / (p * (2 - F))
```
jonjove, line 344: Can we change these 4 inline code segments to be out-of-line so they are easy to read? Based on how you wrote them, I feel like that is what you meant to do anyway.

jonjove: By "these 4" I mean the 4 in the two "Observe that..." blocks.

gramseyer: Makes sense.  Yeah, that's what I intended to do.

jonjove (approved):

PR #1140, opened 2022-02-25
SEP-39: NFT minting and interoperability recommendations

Shaptic:
This is a follow-up on #1111, amended to instead create a unified place for best practices for "NFTs" on Stellar rather than strict standards. It also extends SEP-1 to allow a `[[CURRENCIES]]` entry to represent _any_ value rather than be limited to a currency with an optional image representation.


Shaptic:
cc @FredericRezeau and @grempe for visibility from the last thread

ecosystem/sep-TBD.md
```diff
@@ -0,0 +1,154 @@
...
+
+The minting best practices are merely one way to create NFTs on Stellar and should not be taken as a standard. The interoperability guidelines, however, are important in representing your NFT via [SEP-1][sep1] and maximizing its compatibility with the greater ecosystem.
+
+### Minting NFTs
+This section presents a guide on _best practices_ (rather than a _standard_) for minting some types of NFTs. It's based heavily on [this community guide][litemint]. Your NFT may or may not need certain elements of this guide, or you may need additional components on top of it. These diversions are noted where appropriate.
+
+At a high level, your NFT is represented by an asset on the Stellar network. Owning the asset _represents_ owning the NFT. This makes your good a "first-class citizen" on the network: it immediately benefits from all of Stellar's native features for assets like [path payments](https://developers.stellar.org/docs/start/list-of-operations/#path-payment-strict-send) and the [decentralized exchange](https://developers.stellar.org/docs/glossary/decentralized-exchange/).
```
nickgilbert, line 40: > This makes your good a "first-class citizen"

Using good here instead of "asset" or just "it" is a bit confusing to me.

Shaptic: Nice catch: the doc used to discuss "digital goods" generically but there were concerns about nomenclature, so the language was being amended to say NFT. This is a remnant of that change! Fixed in fe8c199.

nickgilbert (commented):
I think we should add language that encourages users to minimize using on-chain data to the minimum necessary to accomplish their use case.  I would like to acknowledge the need to balance security and data integrity with the consumption of network resources, I think its worth reminding people that consuming those resources has the potential to drive up overall costs to operate and use the network which can impact the networks ability to support the broader mission of access and inclusion by pricing out users with less means. 

We already mention examples on other chains of uses and standards for NFTs, it might also be worth while to call out in the design rationale or motivation section the negative impacts NFTs have had at times on those chains capacity and fees and remind users how they can help mitigate those negative outcomes.

FredericRezeau:
Many thanks, this looks great, I just have 2 additional comments/suggestions.

> `the display_decimals field helps render the right quantity of your NFT, since a single unit of it is likely to be represented by one stroop (i.e. "0.0000001") but you'd want to see "1"`

The display decimals field cannot be used for the purpose of displaying 0.0000001 as 1 (display_decimals=7 simply instruct clients to display amounts with 7 decimals so a stroop would still be shown as 0.0000001). To be able to represent 1 stroop as `1`, clients would need a dedicated **denomination** field (Like Ether has with Wei). It would be great if a **denomination** field could be added to this SEP-1 extension so clients could represent these small amounts with greater usability. e.g. `denomination=stroop` would allow clients to represent amounts as 1. or even for more flexibility using scientific notation e.g. `denomination=1e7`

> `the data entry can only reference the CID describing a single NFT, but you may want the account to issue many different NFTs`

If the `ipfshash` data entry is not found, Litemint also lookup the entry named `ipfshash-ASSET_CODE` allowing issuers to reference multiple assets per issuing accounts. Maybe, this could be added for suggestion as a way to add multi-assets support with data entries.

ecosystem/sep-TBD.md
```diff
@@ -0,0 +1,154 @@
...
+
+#### Issuing your NFT
+To implement non-fractional assets it is necessary to use Stellar's indivisible unit: the stroop. This is the smallest quantity in which a Stellar asset can be sent, received, or traded, and it's equal to one ten-millionth or 0.0000001 of a Stellar lumen (XLM). Therefore, to issue one NFT, you would send 0.0000001 of your asset.
+
+**Note**: Such small amounts may cause issues on the decentralized exchange because of Stellar Core's internal representation of order prices. There are ways to bypass these limitations, such as the one [described by Litemint](https://blog.litemint.com/nft-sdex-pricing/).
+
+#### Ensuring NFT immutability
+If you plan never to increase the supply or modify your NFT, it is best practice to "commit" to this by locking the issuing account. Freezing the account representing your NFT is key to providing immutability for the above steps:
```
leighmcculloch, line 72:
```suggestion
If you plan never to increase the supply or modify your NFT, it is best practice to commit to this by locking the issuing account. Freezing the account representing your NFT is key to providing immutability for the above steps:
```

ecosystem/sep-TBD.md
```diff
@@ -0,0 +1,154 @@
...
+
+### Representing NFTs
+This section provides interoperability guidance around _representing_ NFTs. 
+
+Since NFTs are represented by Stellar assets and are thus "first-class citizens" in the ecosystem, they should leverage existing interoperability layers. Setting up a [SEP-1 `stellar.toml`][sep1] file provides immediate interoperability with all services and wallets on the Stellar ecosystem. It is highly recommended that all assets you issue on Stellar, NFTs or not, follow the [SEP-1][sep1] standard to provide a valid `stellar.toml` file. This grants your NFT a degree of legitimacy, because most Stellar ecosystem services and wallets tend to discard TOML-less assets and/or flag them as spam. There is detailed documentation about how to do that on [this page](https://developers.stellar.org/docs/issuing-assets/publishing-asset-info/).
+
+#### Using SEP-1
+Many of the [SEP-1][sep1] are highly relevant to NFTs. You should include as many of them as is appropriate for your use case:
```
leighmcculloch, line 89:
```suggestion
Many of the [SEP-1][sep1] fields are highly relevant to NFTs. You should include as many of them as is appropriate for your use case:
```

ecosystem/sep-TBD.md
```diff
@@ -0,0 +1,154 @@
...
+issuer="GAKZGD5BFXZ7P7P45WHTM6DODMOEWWJUSCAIPGYIVIPYQSVR6MM6YR7M"
+code="DEMOASSET"
+name="A demonstration of NFT metadata"
+desc="This is a description of the cool NFT used in this demo."
+image="https://cloudflare-ipfs.com/ipfs/QmXoypizjW3WknFiJnKLwHCnL72vedxjQkDDP1mXWo6uco/I/m/RickAstleyNeverGonnaGiveYouUp7InchSingleCover.jpg"
+fixed_number=1
+display_decimals=7
+```
+
+Once you have a `stellar.toml` file under your domain, you should also configure the issuing account's `homedomain` to point to it via the [`SetOptions` operation][set-options]. This unifies the three separate components: the account issuing the NFT, the metadata describing the NFT, and the NFT itself.
```
leighmcculloch, line 115: We should mention here that the home domain can't be changed if the account is frozen.

ecosystem/sep-TBD.md
```diff
@@ -0,0 +1,154 @@
...
+
+#### SEP-1 Extensions
+Since NFTs can represent anything, but the SEP-1 currency specification does not have a generic way to refer to an "anything." There is an `image` key, but it is generally used for something like a currency logo.
+
+Thus, this SEP adds additional supported fields to the SEP-1 currency specification to faciliate this need:
+
+| Field Name | Description and purpose |
+|------------|-------------------------|
+| `url`      | A [valid URL][url] pointing to a NFT that this asset represents |
+| `urlhash`  | A [SHA-256][sha2] hash of the data pointed to by `url` |
```
leighmcculloch, line 125: Is anyone using `urlhash` today in SEP-1?

ecosystem/sep-TBD.md
```diff
@@ -0,0 +1,154 @@
...
+
+
+## Design Rationale
+A key component of a flourishing NFTs marketplace is _interoperability_: that drives all of the design decisions in this informational SEP. The [first section](#minting-nfts) is a set of best practices for a _particular_ set of needs in creating NFTs. There is no "one size fits all" way to do this, so creators should feel free to deviate from the recommendations to fulfill their needs. The [second section](#representing-nfts) adds some extremely flexible fields to SEP-1 to accomodate the fact that NFTs can be anything.
+
+It's worth noting that the idea of "ownership" described throughout the SEP—ownership of an asset and its relationship to owning the respective NFT—is a little diluted: neither the network nor a standard can make any claims about _legitimacy_ of the NFT itself. Owning a unit of the Stellar asset representing your NFT is a way to establish a _relationship_ between buyer and seller that is _linked_ to the NFT, nothing more.
```
leighmcculloch, line 133:
```suggestion
It's worth noting that the idea of ownership described throughout the SEP—ownership of an asset and its relationship to owning the respective NFT—is a little diluted: neither the network nor a standard can make any claims about _legitimacy_ of the NFT itself. Owning a unit of the Stellar asset representing your NFT is a way to establish a _relationship_ between buyer and seller that is _linked_ to the NFT, nothing more.
```

leighmcculloch (commented):
Some minor comments and suggested tweaks, but overall this does a decent job I think of capturing existing practices. I'm not sure about one thing where I ask a question.

ecosystem/sep-TBD.md
Shaptic: These are both _new_ fields introduced in this SEP to extend SEP-1.

ecosystem/sep-TBD.md
```diff
@@ -0,0 +1,154 @@
...
+ORG_URL="<https://ink.litemint.store>"
+
+[[CURRENCIES]]
+issuer="GAKZGD5BFXZ7P7P45WHTM6DODMOEWWJUSCAIPGYIVIPYQSVR6MM6YR7M"
+code="DEMOASSET"
+name="A demonstration of NFT metadata"
+desc="This is a description of the cool NFT used in this demo."
+image="https://cloudflare-ipfs.com/ipfs/QmXoypizjW3WknFiJnKLwHCnL72vedxjQkDDP1mXWo6uco/I/m/RickAstleyNeverGonnaGiveYouUp7InchSingleCover.jpg"
+fixed_number=1
+display_decimals=7
```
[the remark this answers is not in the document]
leighmcculloch: SEP-1's `anchor_asset_type` field, and related fields, probably already fit the bill:
https://github.com/stellar/stellar-protocol/blob/623a2d3b0221d1a2896d3a731f3f255a1b97dca1/ecosystem/sep-0001.md#L131-L133

It doesn't exactly match because anchoring specifies that the anchor will give you asset referenced in exchange for your token, where-as with NFTs you hold the token and can retrieve the asset at the same time without giving up your token.

It would be ashame to add another field when `anchor_asset_type` already support things like "real estate".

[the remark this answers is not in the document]
Shaptic: But this is just a type rather than a value; we still need a generic way to refer to the asset itself :thinking:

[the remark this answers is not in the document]
Shaptic: What about a `representation=<url>` field pointing to a URL? If the issue is that `url=<url>` is too generic

Shaptic:
@FredericRezeau good call about `display_decimals`; that was a misread on my part. I also added the section about `ipfshash-<asset code>`, thanks for that! :+1:

ecosystem/sep-TBD.md
Shaptic: The quotes add emphasis, imo. What's the rationale to drop them?

nickgilbert (approved):
LGTM

tomerweller (dismissed):
lgtm. 🚢 it

ecosystem/sep-TBD.md
leighmcculloch: I avoid emphasis in SEPs because it suggests that one statement is more important than another. Usually in the SEPs everything is equally important, if something isn't important we probably shouldn't include it. Quotes are also an interesting way to emphasize because most of the time on a single word I read them like air quotes, which loosens meaning of the word.

I generally agree with this statement:

>The strength of your words should make any formatting unnecessary, but if you really want to emphasize something, use boldface or italics.

Ref: https://leffcommunications.com/2017/03/10/how-to-use-quotation-marks/

(The exception with emphasis is that we've found in some SEPs where there are a significant number of components emphasising with bold component names can help navigate long and confusing flows. Kudos to @JakeUrban for coming up with that.)

ecosystem/sep-TBD.md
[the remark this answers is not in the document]
leighmcculloch: Representation is pretty generic too. Is this URL intended to be the actual resource, like the final file, or will it be the JSON referenced by the ipfshash?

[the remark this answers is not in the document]
leighmcculloch: Maybe `url` is fine given that we can't easily pin it down to something else.

ecosystem/sep-TBD.md
```diff
@@ -0,0 +1,155 @@
+
+## Preamble
+
+```
+SEP: To Be Assigned
```
leighmcculloch, line 5: @Shaptic Can you assign this 0039, and rename the file to match, and then add the file to the list of SEPs in the ecosystem folder's README.md?

leighmcculloch (changes requested):
Looks like there are approvals and consensus. Before this is merged, an you address the administrative things below, then I think it's good to go?

ecosystem/sep-TBD.md
Shaptic: Changes in https://github.com/stellar/stellar-protocol/pull/1140/commits/ea1dfc6de64ad04b655d7320774006cbedd1c0cb and subsequent rename in https://github.com/stellar/stellar-protocol/pull/1140/commits/1d367aefcf6b47021f9a8b17e6c369275d1c1ebc!

ecosystem/sep-0039.md
```diff
@@ -0,0 +1,155 @@
+
+## Preamble
+
+```
+SEP: 0039
+Title: Interoperability Recommendations for NFTs
+Authors: Tomer Weller <tomer@stellar.org>, Frederic Rezeau <frederic.rezeau@gmail.com>
+Track: Informational
+Status: Active
```
leighmcculloch, line 9: I think we should leave this in draft status for some period of time to let this develop a little? Or did we already discuss status somewhere and progressing it?

ecosystem/sep-0039.md
```diff
@@ -0,0 +1,155 @@
+
+## Preamble
+
+```
+SEP: 0039
+Title: Interoperability Recommendations for NFTs
+Authors: Tomer Weller <tomer@stellar.org>, Frederic Rezeau <frederic.rezeau@gmail.com>
```
leighmcculloch, line 7: More recently we've been using folks GitHub handles rather than emails, for privacy. Unless the authors prefer their emails be included here, could you change it to follows?
```suggestion
Authors: Tomer Weller <@tomerweller>, Frederic Rezeau <@FredericRezeau>
```

leighmcculloch, line 7: Was @tomerweller an author of this? I didn't think he was. Maybe this is a copy-paste mistake from another SEP. @Shaptic If you were wanting to not list yourself personally, you could place "SDF" here.

Shaptic: Yeah this was a product of some language discussion we had offline; putting SDF is much better!

ecosystem/sep-0039.md
Shaptic: I thought that was the point of this PR and all of the prior iterations around this topic :thinking: maybe FCP?

Shaptic: Amended in https://github.com/stellar/stellar-protocol/pull/1140/commits/d8b3386b80bb12e110ee06b5d85b70edd6b148f2

leighmcculloch (approved):
👏🏻

ecosystem/sep-TBD.md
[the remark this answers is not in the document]
Shaptic: Just to respond: it should be the actual resource

ecosystem/sep-TBD.md
Shaptic: >  Quotes are also an interesting way to emphasize because most of the time on a single word I read them like air quotes, which loosens meaning of the word.

You hit the nail on the head here: the concept of `ownership` is loosened here from what people are used to (I used the wrong word "emphasis" earlier), and that loosening is qualified by the remainder of the paragraph.

Regardless, I don't feel strongly, was just curious about the rationale. Appreciate the link :+1:

PR #1141, opened 2022-03-03
CAP-21: Rename TxConditionsV2 to PreconditionsV2

leighmcculloch:
### What
Rename `TxConditionsV2` to `PreconditionsV2`.

### Why
In the XDR changes of CAP-21 two new concepts are introduced, preconditions and txconditions. The inclusion of two concepts is unnecessary as there is really only one new concept being introduced, preconditions. Introducing new concepts when we don't need them makes it harder to understand an interface, and there's value in developing a [ubiquitous language](https://martinfowler.com/bliki/UbiquitousLanguage.html) for the concepts in the XDR so that the ubiquitous language bleeds into all downstream systems.

cc @MonsieurNicolas @stanford-scs @sisuresh @jonjove

### DIFF
--- a/core/cap-0021.md
+++ b/core/cap-0021.md
@@ -143,7 +143,7 @@ struct LedgerBounds {
     uint32 maxLedger;
 };
 
-struct TxConditionsV2 {
+struct PreconditionsV2 {
     TimeBounds *timeBounds;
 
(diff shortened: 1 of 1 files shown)


sisuresh (approved):
LGTM

PR #1143, opened 2022-03-04
SEP-1: Allow externalization of currency metadata to other URLs.

Shaptic:
This is a follow-up to the discussion on the [mailing list](https://groups.google.com/g/stellar-dev/c/-S7FIXJAi2A) to extend SEP-1 by introducing a new top-level `CURRENCY_TOMLS` array in stellar.toml which must be a list of URLs pointing to `[[CURRENCIES]]` table entries.

### DIFF
--- a/ecosystem/sep-0001.md
+++ b/ecosystem/sep-0001.md
@@ -153,13 +153,15 @@ Note: `fixed_number`, `max_number`, and `is_unlimited` are mutually exclusive is
 
 
 ### External Currency Specification
-There are two ways to "outsource" currency entries to external files: via the `toml` key and via the top-level `CURRENCY_TOMLS` list.
+There are two ways to reference currency entries in separate files:
+- The `CURRENCIES` `toml` key.
+- The top-level `CURRENCY_TOMLS` list.
 
-#### Via the `toml` key
+#### `CURRENCIES` `toml` Key
 `stellar.toml` can link out to a separate TOML file for each currency by specifying `toml="https://DOMAIN/.well-known/CURRENCY.toml"` as the currency's only field.
 
-#### Via the top-level list
-We also allow a top-level `CURRENCY_TOMLS` field, which must be an array of URLs to TOML files that follow the above [currency specification](#currency-documentation).
+#### `CURRENCY_TOMLS` List
+A top-level `CURRENCY_TOMLS` field may be specified as an array of URLs to TOML files that contain fields as defined in the [currency specification](#currency-documentation). These additional fields are interpreted as if their contents exist in the `stellar.toml`.
 
 Example usage:
 
@@ -171,9 +173,7 @@ CURRENCY_TOMLS = [
 ]
 ```
 
-Then, the `https://example.com/CURRENCIES_3.toml`-like files would contain ``[[CURRENCIES]]` entries as defined above.
-
-This field must be completely compatible with having existing `[[CURRENCIES]]` table entries within the `stellar.toml` file. The externalized TOML files must **only** include `[[CURRENCIES]]` entries. All other fields should be considered invalid.
+This files referenced must contain only `[[CURRENCIES]]` table entries.
 
 
 ### Validator Information
--- a/ecosystem/sep-0001.md
+++ b/ecosystem/sep-0001.md
@@ -6,8 +6,8 @@ Title: Stellar Info File
 Author: stellar.org
 Status: Active
 Created: 2017-10-30
-Updated: 2021-05-17
-Version: 2.4.1
+Updated: 2022-03-07
+Version: 2.5.0
 ```
 
 ## Simple Summary
@@ -153,15 +153,23 @@ Note: `fixed_number`, `max_number`, and `is_unlimited` are mutually exclusive is
 
 
 ### External Currency Specification
-There are two ways to reference currency entries in separate files:
-- The `CURRENCIES` `toml` key.
-- The top-level `CURRENCY_TOMLS` list.
+There are two (compatible) ways to reference currency entries in separate files:
+
+  1. The `CURRENCIES` `toml` key.
+  2. The top-level `CURRENCY_TOMLS` list.
 
 #### `CURRENCIES` `toml` Key
-`stellar.toml` can link out to a separate TOML file for each currency by specifying `toml="https://DOMAIN/.well-known/CURRENCY.toml"` as the currency's only field.
+`stellar.toml` can link out to a separate TOML file for each currency by specifying the URL in the `toml` key as the currency's only field.
+
+Example usage:
+
+```toml
+[[CURRENCIES]]
+toml = "https://example.com/SPECIFIC_CURRENCY.toml"
+```
 
 #### `CURRENCY_TOMLS` List
-A top-level `CURRENCY_TOMLS` field may be specified as an array of URLs to TOML files that contain fields as defined in the [currency specification](#currency-documentation). These additional fields are interpreted as if their contents exist in the `stellar.toml`.
+A top-level `CURRENCY_TOMLS` field may be specified as an array of URLs to TOML files that contain [currency entries](#currency-documentation). These additional fields are interpreted as if their contents exist in the `stellar.toml`. These files referenced must contain only `[[CURRENCIES]]` table entries.
 
 Example usage:
 
@@ -169,11 +177,24 @@ Example usage:
 CURRENCY_TOMLS = [
   "https://example.com/.well-known/CURRENCIES_1.toml",
   "https://example.com/assets/CURRENCIES_2.toml",
-  "https://example.com/CURRENCIES_3.toml"
+  "https://anotherexample.com/CURRENCIES_3.toml"
 ]
 ```
 
-This files referenced must contain only `[[CURRENCIES]]` table entries.
+Then, within `CURRENCIES_1.toml`, you could have:
+
+```toml
+[[CURRENCIES]]
+code="USD"
+issuer="GCZJM35NKGVK47BB4SPBDV25477PZYIYPVVG453LPYFNXLS3FGHDXOCM"
+
+[[CURRENCIES]]
+code="BTC"
+issuer="GAOO3LWBC4XF6VWRP5ESJ6IBHAISVJMSBTALHOQM2EZG7Q477UWA6L7U"
+
+[[CURRENCIES]]
+toml = "https://example.com/SPECIFIC_CURRENCY.toml"
+```
 
 
 ### Validator Information
--- a/ecosystem/sep-0001.md
+++ b/ecosystem/sep-0001.md
@@ -158,7 +158,7 @@ The `stellar.toml` file may store some or all of its `CURRENCIES` in separate fi
   1. A `CURRENCIES` entry's `toml` key.
   2. A top-level `CURRENCIES_TOMLS` list of URLs.
 
-Both of these methods can be used together.
+Both of these methods can be used together. In both cases, the external files must adhere to the same size restrictions as the `stellar.toml` itself (see [above](#specification): 100 KB each.
 
 #### `CURRENCIES` `toml` Key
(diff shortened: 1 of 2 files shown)


leighmcculloch:
I think given this new proposal is substantially different to SEP-14, it should be a new SEP.

leighmcculloch:
Or given the new proposal is just a modification of SEP-1, we could minimally propose a change to SEP-1.

ecosystem/sep-0001.md
```diff
@@ -151,7 +151,30 @@ Replace `$PUBLIC_KEY` with the account's public key, `$CODE` with your asset cod
 
 Note: `fixed_number`, `max_number`, and `is_unlimited` are mutually exclusive issuance policies. Include exactly one of those fields.
 
-Alternately, `stellar.toml` can link out to a separate TOML file for each currency by specifying `toml="https://DOMAIN/.well-known/CURRENCY.toml"` as the currency's only field.
+
+### External Currency Specification
+There are two ways to "outsource" currency entries to external files: via the `toml` key and via the top-level `CURRENCY_TOMLS` list.
```
leighmcculloch, line 156: Suggested formatting:
```suggestion
There are two ways to reference currency entries in separate files:
- The `CURRENCIES` `toml` key.
- The top-level `CURRENCY_TOMLS` list.
```

ecosystem/sep-0001.md
```diff
@@ -151,7 +151,30 @@ Replace `$PUBLIC_KEY` with the account's public key, `$CODE` with your asset cod
...
+
+#### Via the `toml` key
```
leighmcculloch, line 158: We should capitalize the headings to match the other headings.
```suggestion
#### `CURRENCIES` `toml` Key
```

ecosystem/sep-0001.md
```diff
@@ -151,7 +151,30 @@ Replace `$PUBLIC_KEY` with the account's public key, `$CODE` with your asset cod
...
+`stellar.toml` can link out to a separate TOML file for each currency by specifying `toml="https://DOMAIN/.well-known/CURRENCY.toml"` as the currency's only field.
+
+#### Via the top-level list
```
leighmcculloch, line 161:
```suggestion
#### `CURRENCY_TOMLS` List
```

leighmcculloch (commented):
This is great. I've made some suggestions inline.

@tomerweller Regarding https://github.com/stellar/stellar-protocol/pull/1143#issuecomment-1058784111, do you have thoughts about how this change would be best introduced to the ecosystem? I think the change belongs to SEP-1 as it specifies the stellar.toml file. The SEP is an active SEP which means we can change it. The broad impact of this file is almost breaking in some ways, but technically it isn't.

@Shaptic Could you update the full example at the end to demonstrate both ways to reference external files? Could you also update the version and the updated date in the pragma?

ecosystem/sep-0001.md
```diff
@@ -151,7 +151,30 @@ Replace `$PUBLIC_KEY` with the account's public key, `$CODE` with your asset cod
 
 Note: `fixed_number`, `max_number`, and `is_unlimited` are mutually exclusive issuance policies. Include exactly one of those fields.
 
-Alternately, `stellar.toml` can link out to a separate TOML file for each currency by specifying `toml="https://DOMAIN/.well-known/CURRENCY.toml"` as the currency's only field.
+
+### External Currency Specification
+There are two ways to "outsource" currency entries to external files: via the `toml` key and via the top-level `CURRENCY_TOMLS` list.
+
+#### Via the `toml` key
+`stellar.toml` can link out to a separate TOML file for each currency by specifying `toml="https://DOMAIN/.well-known/CURRENCY.toml"` as the currency's only field.
```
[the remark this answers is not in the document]
Shaptic: Good call; done in 2768bfc :+1:

ecosystem/sep-0001.md
```diff
@@ -151,7 +151,51 @@ Replace `$PUBLIC_KEY` with the account's public key, `$CODE` with your asset cod
 
 Note: `fixed_number`, `max_number`, and `is_unlimited` are mutually exclusive issuance policies. Include exactly one of those fields.
 
-Alternately, `stellar.toml` can link out to a separate TOML file for each currency by specifying `toml="https://DOMAIN/.well-known/CURRENCY.toml"` as the currency's only field.
+
+### External Currency Specification
+There are two (compatible) ways to reference currency entries in separate files:
```
leighmcculloch, line 156: It's unclear to me what compatible means here, regardless it seems superfluous to me.

@Shaptic Was there some meaning you were intending to communicate?

```suggestion
There are two ways to reference currency entries in separate files:
```

ecosystem/sep-0001.md
```diff
@@ -151,7 +151,51 @@ Replace `$PUBLIC_KEY` with the account's public key, `$CODE` with your asset cod
...
+toml = "https://example.com/SPECIFIC_CURRENCY.toml"
+```
+
+#### `CURRENCY_TOMLS` List
+A top-level `CURRENCY_TOMLS` field may be specified as an array of URLs to TOML files that contain [currency entries](#currency-documentation). These additional fields are interpreted as if their contents exist in the `stellar.toml`. These files referenced must contain only `[[CURRENCIES]]` table entries.
+
+Example usage:
+
+```toml
+CURRENCY_TOMLS = [
```
leighmcculloch, line 177: I think we should name the field `CURRENCIES_TOMLS` for an exact one-to-one correlation with the section that is allowed in the field. I think this was changed to `CURRENCY_TOMLS` to avoid a double plural, however this is a list of TOMLs, that each will contain multiple currencies, so I think the double plural here is not particularly problematic.

leighmcculloch (commented):
@Shaptic I left one suggestion, but regardless I think this is in a good place to start getting feedback from a wider audience.

---

Before this is merged, I think we need to identify a list of specific individuals who should be asked to review it. 

@tomerweller @JakeUrban @marcelosalloum @lydiat Could you please review this change?

If we have some general consensus then I think we should post it to the mailing list for feedback, and discuss if there is any other targeted feedback we should seek.

Anyone please post here if you think we need more or less process around this change.

ecosystem/sep-0001.md
Shaptic: I'm trying to say that the two methods are compatible with each other, i.e. they're not mutually exclusive. Maybe `(mutually compatible)` is a better way to convey that?

ecosystem/sep-0001.md
Shaptic: Good point, it is a set of `TOMLS` each of which point to `CURRENCIES`. I hate the mouthful of a name (I'd love something more descriptive like `CURRENCY_METADATA_URLS` or something, [alas](https://martinfowler.com/bliki/TwoHardThings.html)), but it does do the job.

ecosystem/sep-0001.md
leighmcculloch: If this is important I would say it in a sentence rathe than bottling it into a word or two. i.e.:
```suggestion
`stellar.toml`s may store all or some of their `CURRENCIES` in separate files. There are two methods to reference `CURRENCIES` entries that are stored in separate files. Both methods can be used on their own, or together.
```

JakeUrban (commented):
Do the changes here allow for a single entity (platform / issuer / domain / etc) to define an infinite number of assets?

Maybe I'm misunderstanding the proposed solution, but if the root `stellar.toml` file has to list a URL for each page of currency entries, isn't there still a limitation based on the number of URLs you can fit into a single toml file?

I suggest we list a single `CURRENCIES_LIST` attribute on `stellar.toml` whose value is the URL for the first page of currencies. This page could then have a `NEXT_PAGE` pointing to the next page of currencies.

Shaptic:
> Do the changes here allow for a single entity (platform / issuer / domain / etc) to define an infinite number of assets?

Yes. You place them in however many TOML files you want as part of `CURRENCIES_TOMLS`. 

> If the root stellar.toml file has to list a URL for each page of currency entries, isn't there still a limitation based on the number of URLs you can fit into a single toml file?

You can put them all into one if you'd like; there are no size limitations on the external files. We wanted to avoid pagination as part of the spec to keep things simple: the list-based approach lets clients split their assets into however many files they want.

leighmcculloch:
>there are no size limitations on the external files

Hmm, there probably should be a size limit. Machines cannot load infinitely large files and depending on the TOML implementation may not be able to stream in and out contents to reduce memory usage.

>Do the changes here allow for a single entity (platform / issuer / domain / etc) to define an infinite number of assets?

I think this highlights a lacking with the solution we have so far. You can define an infinitely large number of assets in the TOML files but the larger the set gets, the more inaccessible the set gets for a vast majority of clients, many of which are mobile devices with reduce memory capabilities.

That leads us to another problem. By introducing the pagination scheme only without any any method to randomly access a single asset, there is no way to go to the `CURRENCIES` definition for a single asset, which is likely what most wallets do when looking up asset information for an asset held. Wallets can work around this today because an asset set is unlikely to be too large to enumerate. This proposal will make it possible for the asset set to be too large to enumerate.

Shaptic:
I honestly don't see size/pagination as a problem for us to solve: if your deployed currencies file is 10MB, you're putting strain on your users and it's likely that your assets don't have as much reach, so it's in your best interest to break up your files in a way that makes your currencies accessible. Furthermore, you can already have ~14k assets in your base stellar.toml if you use the `toml = "<url>"` approach (napkin math), so I feel like we're generally trying to solve problems that don't exist yet.

If we *do* want to solve this in the SEP, there's a simple solution: require that each external TOML (a) follow stellar.toml size limitations of 100KB and (b) contain mutually exclusive entries. The latter is essentially pagination, and it leaves the implementation/splitting up to the author. But again, I don't think it's necessary unless we have demonstrable issues with this.

ecosystem/sep-0001.md
Shaptic: Good idea, done!

orbitlens:
@leighmcculloch rightly pointed out that blowing up the size of `stellar.toml` leads to unpredictable resources usage on the client side. The 100KB limit is there for a reason. 

> if your deployed currencies file is 10MB, you're putting strain on your users and it's likely that your assets don't have as much reach, so it's in your best interest to break up your files in a way that makes your currencies accessible

This suggestion implies that all asset issuers are benevolent and intelligent. Which is not the case.

Using the proposed syntax, anybody will be able to execute a resource exhaustion attack that can affect any active user and almost any Stellar wallet. It is as simple as a) create a few thousands of garbage recurrent TOML files and b) send claimable balances to all Stellar accounts with recent on-chain activity. Once a user opens a wallet on a claimable balances page, the client will start downloading gigabytes of garbage. 

And I'm not even talking here about various cases when asset developers may accidentally mess with TOML code generator leading to spontaneous problems for the clients. Please make no mistake, 100% of issuers with more than 100 assets will use automatic TOML generators because managing hundreds of assets manually is a completely stupid and prone to human errors task. I also totally agree with @JakeUrban regarding the paramount importance of the fast asset lookup ability, that has always been my primary argument.

All these considerations have been discussed in Slack and on protocol meetings more than three years ago, resulting in a SEP-14 proposal. This relatively simple proposal in a combination with a few tiny Core protocol modifications would have made Stellar an ideal platform for NFTs and on-chain stocks/commodities trading platforms back there, in 2019. And now we have lost all the potential of the leading NFT platform because we cannot even agree on a dynamic metadata SEP despite of 3 rounds of discussions that happened in 2018-2019 years.

SEP-1 by its nature provides a vast playground for various resource exhaustion and spoofing attacks because it doesn't lay restrictions on the external resources. For example, scam tokens often copy-paste asset logos and SEP-24 endpoint urls from trustworthy issuers, like UltraStellar or Circle to trick our asset ranking algorithm. Some asset issuers utilized huge images for asset logos, sometimes more than 10 MiB. Let's not aggravate the problem by providing even more opportunities.

PR #1144, opened 2022-03-08
SEP-38: make SEP-10 token optional for some endpoints

### LINKED ISSUE #1086: SEP-38: Allow unauthenticated requests to non-`/quote` endpoints

Anchors may want to allow unauthenticated clients to make requests to the `GET /prices` and `GET /price` endpoints. 

For example, client applications may want to show rates from multiple liquidity providers to a user, or simply display the exchange rates of an anchor's pairs before the user has indicated they want to start a transaction.

Make SEP-10 authentication optional for the `GET /price` and `GET /prices` endpoints.

Leaving SEP-10 required for `GET /price` and `GET /prices`.

marcelosalloum:
### What

Make SEP-10 token optional for the endpoints:
* `GET /info`
* `GET /price`
* `GET /prices`

### Why

Client applications may want to show rates from multiple liquidity providers to a user, or simply display the exchange rates of an anchor's pairs before the user has indicated they want to start a transaction.

Closes #1086.

### DIFF
--- a/ecosystem/sep-0038.md
+++ b/ecosystem/sep-0038.md
@@ -32,7 +32,7 @@ An anchor must define the location of their `ANCHOR_QUOTE_SERVER` in their [`ste
 
 This protocol's endpoints use authentication in the form of a [SEP-10](sep-0010.md) JSON Web Token (JWT) in the `Authorization` header of the request.
 
-While it is mandatory for [`POST /quote`](#post-quote) and [`GET /quote/:id`](#get-quote), it is optional for the remaining endpoints. When optional, the JWT token can be used to personalize the respective responses.
+Authentication is mandatory for [`POST /quote`](#post-quote) and [`GET /quote/:id`](#get-quote), but optional for the remaining endpoints. When optional, if authentication is provided it can be used to personalize the respective responses.
 
 ```
 Authorization: Bearer <jwt>
--- a/ecosystem/sep-0038.md
+++ b/ecosystem/sep-0038.md
@@ -30,7 +30,9 @@ An anchor must define the location of their `ANCHOR_QUOTE_SERVER` in their [`ste
 
 ### Authentication
 
-All endpoints require authentication in the form of a [SEP-10](sep-0010.md) JSON Web Token (JWT) in the `Authorization` header of the request. 
+This protocol's endpoints use authentication in the form of a [SEP-10](sep-0010.md) JSON Web Token (JWT) in the `Authorization` header of the request.
+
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0038.md
```diff
@@ -30,7 +30,9 @@ An anchor must define the location of their `ANCHOR_QUOTE_SERVER` in their [`ste
 
 ### Authentication
 
-All endpoints require authentication in the form of a [SEP-10](sep-0010.md) JSON Web Token (JWT) in the `Authorization` header of the request. 
+This protocol's endpoints use authentication in the form of a [SEP-10](sep-0010.md) JSON Web Token (JWT) in the `Authorization` header of the request.
+
+While it is mandatory for [`POST /quote`](#post-quote) and [`GET /quote/:id`](#get-quote), it is optional for the remaining endpoints. When optional, the JWT token can be used to personalize the respective responses.
```
leighmcculloch, line 35: Small wording suggestion, otherwise looks great.
```suggestion
Authentication is mandatory for [`POST /quote`](#post-quote) and [`GET /quote/:id`](#get-quote), but optional for the remaining endpoints. When optional, if authentication is provided it can be used to personalize the respective responses.
```

leighmcculloch, line 35: I think the document would benefit from explicitly stating the authentication in the section of each endpoint. We do this in SEP-30 and it makes it ultra clear in the section for each endpoint the types of authentication that's available or required. See this for an example:

https://github.com/stellar/stellar-protocol/blob/799105e14c4cf2b963870760ffade1132928da50/ecosystem/sep-0030.md#L498-L499

leighmcculloch (commented):
This looks great, I think this change is good and very backwards compatible.

I make a couple suggestions below.

Could you also update the version by bumping the minor version number, and the updated field in the preamble?

JakeUrban (dismissed):
I agree with Leigh's points, but otherwise LGTM! Thanks for taking this on.

ecosystem/sep-0038.md
marcelosalloum: https://github.com/stellar/stellar-protocol/pull/1144/commits/26a8534f440318987301ab0e3abcd08f36e8908c

marcelosalloum: Done in https://github.com/stellar/stellar-protocol/pull/1144/commits/d943c889c06a25c51df79bdc48a665187ca86912.

leighmcculloch (approved):

PR #1145, opened 2022-03-08
SEP-31: Add mermaid sequence diagram

JakeUrban:
Adds a "Diagrams" section to SEP-31 and starts with a happy-path sequence diagram.

We've discussed whether or not to include diagrams in our SEPs several times in the past and have historically decided not to do so due to the additional burden of maintaining another document. However, now that Github has added support for mermaid syntax, we can keep the SEP as the single source of truth and edit diagrams directly as markdown.

### DIFF
--- a/ecosystem/sep-0031.md
+++ b/ecosystem/sep-0031.md
@@ -36,6 +36,73 @@ Typically, the Sending and Receiving Clients reside in different regulatory juri
 
 Alice in Nigeria wants to send money to Bob in Europe. Alice signs up with NigeriaPay to make this payment to send money directly into Bob’s bank account. Bob doesn’t need to do anything, or know anything about this payment, besides letting Alice know what his bank account information is. Alice only needs to deal with her anchor (NigeriaPay). Alice passes this information and her money to NigeriaPay, and NigeriaPay sends those funds to the EuroPay Anchor service, and Europay deposits those funds into Bob’s bank account.
 
+## Diagrams
+
+### Successful Transaction with Firm Quote
+
+```mermaid
+sequenceDiagram
+    title: SEP-31 Transaction Flow
+    participant Sender
+    participant Sending Anchor
+    participant Stellar
+    participant Receiving Anchor
+    participant Recipient
+    Sender->>Sending Anchor: initiates send to <br> Recipient's country
+    Sending Anchor-->>Sending Anchor: maps country to Receiving Anchor
+
+    Sending Anchor->>+Receiving Anchor: GET /.well-known/stellar.toml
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0031.md
```diff
@@ -631,4 +701,5 @@ If the information was malformed, or if the sender tried to update data that isn
 
 ## Changelog
 
-* `v1.5.0`: Deprecate refunded boolean. Add refund object to transaction records. ([#1128](https://github.com/stellar/stellar-protocol/pull/1128))
\ No newline at end of file
+* `v1.5.0`: Deprecate refunded boolean. Add refund object to transaction records. ([#1128](https://github.com/stellar/stellar-protocol/pull/1128))
+* `v1.5.1`: Add a sequence diagram for successful transactions using firm quotes. ([#1145](https://github.com/stellar/stellar-protocol/pull/1145))
```
leighmcculloch, line 705: Can you put this at the top of the change log so the change log is ordered descending? Most recent changes are the most interesting to folks.

ecosystem/sep-0031.md
```diff
@@ -36,6 +36,76 @@ Typically, the Sending and Receiving Clients reside in different regulatory juri
...
 
+## Diagrams
+
+The diagrams provided offer a detailed view of various flows that are possible using this protocol. Note that these diagrams are opinionated. Variations of the these flows may also be sufficient.
+
+### Completing a Transaction with a Firm Quote
+
+This diagram demonstrates the interactions between the entities involved in a successful transaction. Specifically it uses the [Receiving Anchor Asset Conversion](#receiving-anchor-asset-conversion) strategy defined described later in the document.
+
+```mermaid
```
leighmcculloch, line 47: This diagram looks great. 👏🏻 

One thing I noticed is that it renders with a massive blank area at the bottom of the diagram. Do you know why this is? Is this a @github bug, or something to do with how we've defined the diagram do you think?

Here's a screenshot of what I see:
Screen Shot 2022-03-08 at 1 12 31 PM

JakeUrban: I see the same thing, but I don't see anything wrong with the diagram syntax. Let me google a bit to see if this is a known issue.

JakeUrban: https://github.com/mermaid-js/mermaid/issues/1984

leighmcculloch (approved):
This looks great, great idea.

I don't think we need to block merging this on the rendering issue. It is likely a GitHub bug. It could be worthwhile creating a discussion thread about this bug at https://github.com/github/feedback/discussions/categories/general-feedback or sending a message to GitHub Support.

PR #1149, opened 2022-03-16
Add on-stellar asset for receiving client.

lijamie98:
In the following scenario, the receiving client may receive on-stellar asset. 

Sending client --off-stellar --> Sending Anchor
Sending Anchor --on-stellar --> Receiving Anchor
Receiving Anchor--on-stellar --> Receiving Client

In this use case, the sending customer does not have a Stellar account and Receiving Client has a Stellar account. We should add on-stellar asset to the receiving client.

### DIFF
--- a/ecosystem/sep-0031.md
+++ b/ecosystem/sep-0031.md
@@ -700,6 +700,6 @@ If the information was malformed, or if the sender tried to update data that isn
 ```
 
 ## Changelog
-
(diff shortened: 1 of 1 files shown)


JakeUrban (approved):
We probably need to provide some more examples that demonstrate when the recipient would receive funds on-chain, but we can add those when we have a concrete use case.

JakeUrban (changes requested):
Oh sorry, can you add this PR to the changelog at the bottom?

JakeUrban (approved):

lijamie98:
> Oh sorry, can you add this PR to the changelog at the bottom?

Added to the changelog at the bottom.

PR #1152, opened 2022-03-18
Clarify some points, add explanation for liquidity pool demand

jayz22:


core/cap-0045.md
```diff
@@ -311,91 +320,77 @@ then we should acknowledge that no prices at all is probably just as useful.
...
-```
+The liquidity pool starts at equilibrium point `E`, with reserve amounts `X` and `Y` respectively. 
 
-We will now show that the liquidity pool does take the actions described above.
-First, suppose that `p < (1 - F) Y/X`.  
-Then the liquidity pool sells `Y` and buys `X` until
-`p = (1 - F) Y'/X'`.  Combining this equation with the budget constraint gives
-`pX + Y = Y' + (1 - F)Y'`.
-Rearranging gives `Y' = (pX + Y) / (2 - F)` and `X' = (pX + Y) (1 - F) / (p * (2 - F))`.
+The spot price of asset `X` is the amount of asset `Y` exchanged for infinitesimal quantity of `X`:
```
jonjove, line 328: You should probably define the constant product invariant before this.

Also, I know we were using this language when discussing but we should be precise in the proposal. Something like, "The spot price is the ratio of amount of asset `Y` sold to the amount of asset `X` bought, as the amount of `X` bought is reduced toward 0."

core/cap-0045.md
```diff
@@ -311,91 +320,77 @@ then we should acknowledge that no prices at all is probably just as useful.
...
 
-Observe that in this case, 
+<div align="center"><img style="background: white;" src="https://render.githubusercontent.com/render/math?math=p%20%3D%20%5Clim_%7B%5Cdelta%7Bx%7D-%3E0%7D%5Cfrac%7B%5Cdelta%7By%7D%7D%7B%5Cdelta%7Bx%7D%7D"></div>
```
jonjove, line 330: The arrow in the subscript of the limit isn't rendering correctly. If this is LaTeX then you would want to use `\to`. Note, this happens everywhere you used this notation although I'll only comment here.

core/cap-0045.md
```diff
@@ -311,91 +320,77 @@ then we should acknowledge that no prices at all is probably just as useful.
...
 
-```
-X' = (pX + Y) (1 - F) / (p * (2 - F))
-    > (pX (1 - F) + pX) / (p * (2 - F))
-    = X
-```
+Consider the scenario where the market decides on a higher price for `X` than what is manifested in the current equilibrium `E`, the liquidity pool will buy `X` and sell `Y` until a new equilibrium is established (call it `E’`).
```
jonjove, line 332: Not a "higher price for `X`". Rather, a "higher price for `X` relative to `Y`" (only price ratios matter).

But more importantly, shouldn't this say lower price? Compare against line 350, "as long as the external market price of `X` is less than or equal to the ratio".

core/cap-0045.md
```diff
@@ -311,91 +320,77 @@ then we should acknowledge that no prices at all is probably just as useful.
...
+<div align="center"><img style="background: white;" src="https://render.githubusercontent.com/render/math?math=Y'%3D%5Cfrac%7BpX%2BY%7D%7B2-F%7D"></div>
+
+Observe that in this case, <img style="transform: translateY(0.1em); background: white;" src="https://render.githubusercontent.com/render/math?math=X'%20%3D%20%5Cfrac%7B(pX%2BY)(1-F)%7D%7Bp%5Ccdot(2-F)%7D%20%3E%20%5Cfrac%7BpX(1-F)%2BpX%7D%7Bp%5Ccdot(2-F)%7D%20%3D%20X">
+
+
+and <img style="transform: translateY(0.1em); background: white;" src="https://render.githubusercontent.com/render/math?math=Y'%3D%5Cfrac%7BpX%2BY%7D%7B2-F%7D%3C%5Cfrac%7B(1-F)Y%2BY%7D%7B2-F%7D%3DY">
+
+Thus, the liquidity pool, as expected, sells `Y` and purchases `X`.
+
+By the exact same notion, when <img style="transform: translateY(0.1em); background: white;" src="https://render.githubusercontent.com/render/math?math=%5Cfrac%7B1%7D%7Bp%7D%3C%5Cfrac%7B(1-F)X%7D%7BY%7D">, the demand for `X` and `Y` are:
```
jonjove, line 373: You should clarify that this is for the other scenario.

core/cap-0045.md
```diff
@@ -311,91 +320,77 @@ then we should acknowledge that no prices at all is probably just as useful.
...
+
+Observe in this case, <img style="transform: translateY(0.1em); background: white;" src="https://render.githubusercontent.com/render/math?math=X'%3D%5Cfrac%7BpX%2BY%7D%7B(2-F)%5Ccdot%20p%7D%3C%5Cfrac%7BpX%2BpX(1-F)%7D%7B(2-F)%5Ccdot%20p%7D%3DX"> and <img style="transform: translateY(0.1em); background: white;" src="https://render.githubusercontent.com/render/math?math=Y'%3D%5Cfrac%7B(pX%2BY)(1-F)%7D%7B2-F%7D%3E%5Cfrac%7B(Y%2B(1-F)Y)%7D%7B2-F%7D%20%3DY">
 
 Thus, the liquidity pool, as expected, sells `X` and purchases `Y`.
 
+When the price is in between, i.e. <img style="transform: translateY(0.1em); background: white;" src="https://render.githubusercontent.com/render/math?math=%5Cfrac%7B(1-F)Y%7D%7BX%7D%20%5Cle%20p%20%5Cle%20%5Cfrac%7BY%7D%7B(1-F)X%7D">, no trade will happen and both reserves stay the same.
+
 ### Demand from an order
 Tatonnement does not handle discontinuities well. The demand from an order is
 intrinsically discontinuous, so a natural solution is to smooth the demand. This
```
jonjove, line 385: There was some confusion about what should be smoothed. I think you should clarify that the price-weighted demand should be linearly interpolated.

core/cap-0045.md
```diff
@@ -391,8 +391,41 @@ price <img style="transform: translateY(0.1em); background: white;" src="https:/
 Specifically, <img style="transform: translateY(0.1em); background: white;" src="https://render.githubusercontent.com/render/math?math=%5Cfrac%7Bp_%7Bfull%7D%7D%7Bp_%7Blim%7D%7D%20%3D%20%5Cfrac%7B1%7D%7B(1-2%5E%7B-%5Csigma%7D)%7D">.
 
 It doesn't matter whether you linearly interpolate the asset sold or the asset bought. The two quantities are linearly related according to the budget
-constraint `pX + Y = pX' + Y'`, so linear interpolation of one implies linear
-interpolation of the other.
+constraint <img style="transform: translateY(0.1em); background: white;" src="https://render.githubusercontent.com/render/math?math=p_x%20x%20%2B%20p_y%20y%20%3D%20p_x%20x'%20%2B%20p_y%20y'">, so linear interpolation of one asset implies linear interpolation of the other.
+
+To derive the demand results in the previous section, consider an offer selling asset `X` (we also use `X` to denote the amount of `X` being offered) for asset `Y` at limit price <img style="transform: translateY(0.1em); background: white;" src="https://render.githubusercontent.com/render/math?math=p_%7Blim%7D%3DP">. At any point, the budget must be conserved (subject to a fee `F`):
```
jonjove, line 397: There are no fees for orders in this proposal. If you think there should be (I know we discussed this) then you should discuss it on the mailing list.

core/cap-0045.md
```diff
@@ -391,8 +391,41 @@ price <img style="transform: translateY(0.1em); background: white;" src="https:/
...
+
+<div align="center"><img style="background: white;" src="https://render.githubusercontent.com/render/math?math=y%5Ccdot%20p_%7By%7D(1-F)%20%3D%20p_%7Bx%7D%5Ccdot%20x"></div>
+
+When <img style="transform: translateY(0.1em); background: white;" src="https://render.githubusercontent.com/render/math?math=p_x%20%5Cle%20p_%7Blim%7D">, no trade will happen, demand for `Y` is 0, and demand for `X` is the full amount of X valued at <img style="transform: translateY(0.1em); background: white;" src="https://render.githubusercontent.com/render/math?math=p_%7Bx%7DX">.
```
jonjove, line 401: This is wrong, you're using both `p(X)` and `p(Y)` here so you always need to take ratios. Same applies in all of these statements, and the graph.

core/cap-0045.md
```diff
@@ -391,8 +391,41 @@ price <img style="transform: translateY(0.1em); background: white;" src="https:/
...
+
+<div align="center"><img style="background: white;" src="https://render.githubusercontent.com/render/math?math=Y'%20%3D%20%5Cfrac%7B(p-P)X%7D%7BS%7D"></div>
+
+<div align="center"><img style="background: white;" src="https://render.githubusercontent.com/render/math?math=X'%20%3D%20X%20-%20%5Cfrac%7B(p-P)X%7D%7BS%5Ccdot%20p%7D"></div>
+
+Same results as before.
```
jonjove, line 429: Nitpick: This is a bit of a strange ending, could probably be removed.

core/cap-0045.md
jayz22: I updated. Thanks for point it out.

core/cap-0045.md
jayz22: I clarified it. Also, I thought the full illustration with math should've remove any confusion.

core/cap-0045.md
jayz22: Good catch, thanks and I've updated.

core/cap-0045.md
jonjove: I think I was unclear about defining the constant product invariant. In your definition of the spot price you're using delta-x, delta-y but they haven't been defined yet. I meant that you should specify the equation https://render.githubusercontent.com/render/math?math=(X%20%2B%20(1-F)%5Cdelta%7Bx%7D)(Y-%5Cdelta%7By%7D)%5Cge%20XY for the invariant (where you introduce delta-x, delta-y) before this. 

As for why my language is more precise than what's here: the amount of asset `Y` exchanged for an infinitesimal quantity of `X` is also infinitesimal. We can treat delta-y as a function of delta-x. Then if you take the limit to 0 using this _written description_, the limit should be 0. Obviously, that's not what you intend and it's not what happens in the _mathematical description_. You want to the amount of `Y` per `X` as you take the limit to 0.

core/cap-0045.md
jonjove: I guess my point is people will think that this means we intend for there to be a fee, or intend the fee to be a future extension. If we don't ever intend that, there's no reason to handle the generality. If we do intend it, then yes let's keep it.

core/cap-0045.md
jayz22: I see. The spot price definition is a general definition and exists outside of liquidity pool, that's why I defined it before the constant-product invariant. But I do see what you mean, the delta-x and delta-y were not clearly defined. 
I've rewritten the first part a bit to address this, hopefully makes it clear now.

core/cap-0045.md
jayz22: I don't intend to propose a fee without a clear reason of doing so (other than possible numerical benefit of having a smaller smooth-factor, which should be pretty minor). 
In that case, I have removed all `F` components from the derivation (except the very first one, to state the general form), and made it clear from the beginning there will be no fee. It also shortens the derivation a bit which is nice.

jayz22:
Thanks @jonjove Jon, happy to iterate on it. 
Once you feel ready to merge, I will squash all commits into one.

jonjove (approved):

PR #1153, opened 2022-03-21
CAP-40: Add rationale for how hint is derived

leighmcculloch:
### What
Add design rationale paragraph about how hint is derived for the signed payload signer.

### Why
@Shaptic pointed out that this isn't explained in the document, but there was specific intent to the hints design.


Shaptic (approved):
:clap:

PR #1160, opened 2022-03-31
Update mddiffcheck.yml

ire-and-curses:
Fix for Go 1.18


JakeUrban (approved):
LGTM, but are we able to bypass the existing check so we can merge?

JakeUrban:
Looks like we are, thanks @ire-and-curses

PR #1164, opened 2022-04-02
Add the state diagram of a SEP-31 transaction.

lijamie98:


JakeUrban (approved):
Awesome addition to the specification 👏 

Using a PNG is fine for now, but I think it would be best if we ultimately used a [mermaid state diagram](https://mermaid-js.github.io/mermaid/#/stateDiagram) instead. This would allow us to edit the rendered image using standard syntax directing in this document, similar to our sequence diagram.

PR #1167, opened 2022-04-05
SEP-10: Fix typo in challenge & token uri example

theclabs:
Removed the extra slashes " / " since <WEB_AUTH_ENDPOINT> shouldn't have them or may not and we ran into an integration problem due to misinterpretation
Thanks!

### DIFF
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -81,7 +81,7 @@ The flow achieves several things:
 
 ## Authentication Endpoint
 
-The organization with a **Home Domain** indicates that it supports authentication via this protocol by specifying `WEB_AUTH_ENDPOINT` in their [`stellar.toml`](sep-0001.md) file. This is how a wallet knows where to find the **Server**. A **Server** is required to implement the following behavior for the web authentication endpoint:
+The organization with a **Home Domain** indicates that it supports authentication via this protocol by specifying `WEB_AUTH_ENDPOINT` in their [`stellar.toml`](sep-0001.md) file. Ideally wallets take the `WEB_AUTH_ENDPOINT` and use it without making any alterations. This is how a wallet knows where to find the **Server**. A **Server** is required to implement the following behavior for the web authentication endpoint:
 
 * [`GET <WEB_AUTH_ENDPOINT>`](#challenge): request a challenge (step 1)
 * [`POST <WEB_AUTH_ENDPOINT>`](#token): exchange a signed challenge for session JWT (step 2)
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0010.md
```diff
@@ -118,7 +118,7 @@ Name      | Type          | Description
 Example:
 
 ```
-GET https://auth.example.com/?account=GCIBUCGPOHWMMMFPFTDWBSVHQRT4DIBJ7AD6BZJYDITBK2LCVBYW7HUQ
+GET https://auth.example.com?account=GCIBUCGPOHWMMMFPFTDWBSVHQRT4DIBJ7AD6BZJYDITBK2LCVBYW7HUQ
```
leighmcculloch, line 121: Servers should handle either with or without the slash, and most web servers do not usually distinguish. Clients are likely to include or not include the slash, depending on the HTTP client.

Take the Go http client for example. The slash is always included for root paths, and then may or may not be included for sub-root paths:
https://go.dev/play/p/srqwfsW-Jdw

leighmcculloch (commented):
>we ran into an integration problem due to misinterpretation

@theclabs Could you elaborate to the integration problem that occurred? Servers and clients should be resilient to the slash being included or not. Ideally clients take the `WEB_AUTH_ENDPOINT` and append the query string without making any alterations, so if the `WEB_AUTH_ENDPOINT` ends in a slash, its there, if not, it's not.

ecosystem/sep-0010.md
theclabs: I understand this, however, an alternative example to avoid confusion could be

GET <WEB_AUTH_ENDPOINT>?account=GCIBUCGPOHWMMMFPFTDWBSVHQRT4DIBJ7AD6BZJYDITBK2LCVBYW7HUQ

now i'm checking why polaris doesn't support ending slash in server/auth

theclabs:
> > we ran into an integration problem due to misinterpretation
> 
> @theclabs Could you elaborate to the integration problem that occurred? Servers and clients should be resilient to the slash being included or not. Ideally clients take the `WEB_AUTH_ENDPOINT` and append the query string without making any alterations, so if the `WEB_AUTH_ENDPOINT` ends in a slash, its there, if not, it's not.

The problem originated that in the example provided, a "/" is intentionally added to the example [see](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0010.md#request), then the integrator team intentionally added this slash to its implementation, for which the polaris server is rejecting the request to obtain challenge.

Polaris Logs
```
server | 2022-04-05 14:50:18,008 - WARNING: Not Found: /auth/
server | 2022-04-05 14:50:18,010 - WARNING: "GET /auth/?account=GDSB7QMSKTNZQW2I3 . . . R3MXH2ETURG2 HTTP/1.1" 404 179
```

ecosystem/sep-0010.md
leighmcculloch: >i'm checking why polaris doesn't support ending slash in server/auth

cc @JakeUrban

JakeUrban:
Thanks @theclabs for identifying this issue. I've updated Polaris' URL pattern matching logic to support requests with and without a trailing slash. The change will be released in Polaris v2.2.2.

ecosystem/sep-0010.md
theclabs: Thanks for mentioning the polaris team.

theclabs:
Feel free to accept or reject the proposal to clarify the correct or optimal use of the <WEB_AUTH_ENDPOINT> field.

leighmcculloch:
Thanks for bringing the inconsistency, esp with Polaris, to our attention. I'll close this now, since `/`s typically show up on root paths anyway so they have to be supported in any case.

PR #1180, opened 2022-04-24
Meta: Require consistent newline at EOF in docs

leighmcculloch:
### What
Require consistent newlines at EOF in docs.

### Why
Mildly annoying when editing docs and getting EOF line changes due to inconsistencies. Not really a big problem, but trivial to prevent. Most files in the repo have a newline at EOF, so this requires it.


MonsieurNicolas:
Is the problem just EOF or also EOL? I remember we talked about it in the context of mddiffcheck

leighmcculloch:
EOL is already handled by `Check CAP Line Endings`. Arguably that is the way more annoying thing. This is just mildly annoying, it's just happened a few times to me lately.

PR #1185, opened 2022-04-29
SEP-24: Change `completed_at` field to optional

### LINKED ISSUE #1184: `completed_at` in SEP24 is inconsistently defined from SEP6 and SEP31
`completed_at` was defined as Optional fields in [SEP-6](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0006.md#transaction-history) and [SEP-31](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0031.md#get-transaction). However, the field is defined as ` Assigned null for in-progress transactions.`  in [SEP-24](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0024.md#shared-fields-for-both-deposits-and-withdrawals).

Suggest to make the field consistent among all protocols.

lijamie98:
Closes #1184 
`completed_at` was defined as Optional fields in [SEP-6](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0006.md#transaction-history) and [SEP-31](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0031.md#get-transaction). However, the field is defined as Assigned null for in-progress transactions. in [SEP-24](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0024.md#shared-fields-for-both-deposits-and-withdrawals).

Suggest to make the field consistent among all protocols.

### DIFF
--- a/ecosystem/sep-0024.md
+++ b/ecosystem/sep-0024.md
@@ -6,8 +6,8 @@ Title: Hosted Deposit and Withdrawal
 Author: SDF
 Status: Active
 Created: 2019-09-18
-Updated: 2021-12-01
-Version 2.2.0
+Updated: 2022-05-03
+Version 2.2.1
 ```
 
 ## Simple Summary
@@ -689,7 +689,7 @@ Name | Type | Description
 `amount_fee` | string | Amount of fee charged by anchor.
 `amount_fee_asset` | string | (optional) The asset in which fees are calculated in. Must be present if the deposit/withdraw was made using non-equivalent assets. The value must be in [SEP-38 Asset Identification Format](sep-0038.md#asset-identification-format). See the [Asset Exchanges](#asset-exchanges) section for more information.
 `started_at` | UTC ISO 8601 string | Start date and time of transaction.
-`completed_at` | UTC ISO 8601 string | Completion date and time of transaction.  Assigned null for in-progress transactions.
+`completed_at` | UTC ISO 8601 string | (optional) Completion date and time of transaction.
 `stellar_transaction_id` | string | transaction_id on Stellar network of the transfer that either completed the deposit or started the withdrawal.
 `external_transaction_id` | string | (optional) ID of transaction on external network that either started the deposit or completed the withdrawal.
 `message` | string | (optional) Human readable explanation of transaction status, if needed.
@@ -935,5 +935,5 @@ There is a small set of changes when upgrading from SEP-6 to SEP-24.
 * Solar wallet: https://solarwallet.io
 
 ## Changelog
-
+* `v2.2.1`: Make `completed_at` field optional. ([#1185](https://github.com/stellar/stellar-protocol/pull/1185))
 * `v2.2.0`: Deprecate refunded boolean. Add refund object to transaction records. ([#1128](https://github.com/stellar/stellar-protocol/pull/1128))


ecosystem/sep-0024.md
```diff
@@ -689,7 +689,7 @@ Name | Type | Description
 `amount_fee` | string | Amount of fee charged by anchor.
 `amount_fee_asset` | string | (optional) The asset in which fees are calculated in. Must be present if the deposit/withdraw was made using non-equivalent assets. The value must be in [SEP-38 Asset Identification Format](sep-0038.md#asset-identification-format). See the [Asset Exchanges](#asset-exchanges) section for more information.
 `started_at` | UTC ISO 8601 string | Start date and time of transaction.
-`completed_at` | UTC ISO 8601 string | Completion date and time of transaction.  Assigned null for in-progress transactions.
+`completed_at` | UTC ISO 8601 string | (optional) Completion date and time of transaction.
```
leighmcculloch, line 692: What does this change mean in terms of the API response. Will optional mean that the response field will be omitted? There might have been a reason we made this explicitly null. I don't remember if this field specifically related to this, but the Vibrant team provided feedback at one point that having omitted fields is more difficult to debug and code against than null fields, and so we started preferring null fields.

leighmcculloch (commented):
See comment inline.

ecosystem/sep-0024.md
lijamie98: @leighmcculloch Yes, this means optional fields may be omitted. When we were looking at the example JSON API responses, optional fields seems to imply omission. 

@CassioMG Can you also provide feedback of making SEP24 consistent with SEP-6 and SEP-31 which mark `completed_at` as optional? Is Vibrant treating the `completed_at` field in SEP-24 different from SEP-6 and SEP-31?

JakeUrban (commented):
This looks good to me, can we make sure to update the meta info at the top and the changelog at the bottom @lijamie98 ?

I agree with @leighmcculloch that ideally we use `null` values in JSON responses instead of omitting fields, but unfortunately it looks like our example responses in the SEPs already omit fields labeled as optional. 

I would think that robust client applications are implemented without assuming the presense of a key vs. it having a `null` value.

leighmcculloch:
> I would think that robust client applications are implemented without assuming the presense of a key vs. it having a `null` value.

For context, I believe the reason null is preferred is not due to clients not being robust. It is for helping developers understand APIs as well as handling and catching typos in client code. If a fields are omitted, the shape of the response changes from one request to the next which is difficult to reason about. Also, if a field is omitted it appears as "undefined" in JS apps. If a field has a typo in its name on the client side, its value appears as "undefined". It increases the difficulty of identifying these problems.

Regardless, makes sense to keep consistent with the rest of the doc and make it ommitted.

lijamie98:
> This looks good to me, can we make sure to update the meta info at the top and the changelog at the bottom @lijamie98 ?
> 
> I agree with @leighmcculloch that ideally we use `null` values in JSON responses instead of omitting fields, but unfortunately it looks like our example responses in the SEPs already omit fields labeled as optional.
> 
> I would think that robust client applications are implemented without assuming the presense of a key vs. it having a `null` value.

@JakeUrban Thanks for the reminder. The version and dates are updated.

lijamie98:
Can someone approve this PR?

ecosystem/sep-0024.md
CassioMG: @lijamie98 it's completely fine to omit this field, we are not making any comparisons with `null`

CassioMG (dismissed):
LGTM

ecosystem/sep-0024.md
CassioMG: Also +1 for making it consistent throughout all SEPs, thanks for fixing this

lijamie98:
> Could you update the changelog which is at the end of the document?

Added and updated.

leighmcculloch (approved):

PR #1195, opened 2022-05-06
SEP-24: Added refunded transaction status

lijamie98:
Since the `refunded` flag is being deprecated, MGI requested the `refunded` status.

### DIFF
--- a/ecosystem/sep-0024.md
+++ b/ecosystem/sep-0024.md
@@ -694,7 +694,7 @@ Name | Type | Description
 `stellar_transaction_id` | string | transaction_id on Stellar network of the transfer that either completed the deposit or started the withdrawal.
 `external_transaction_id` | string | (optional) ID of transaction on external network that either started the deposit or completed the withdrawal.
 `message` | string | (optional) Human readable explanation of transaction status, if needed.
-`refunded` | boolean | (**deprecated**, optional) This field is deprecated in favor of the `refunds` object. True if the transaction was refunded in full. False if the transaction was partially refunded or not refunded. For more details about any refunds, see the `refunds` object.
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0024.md
```diff
@@ -6,8 +6,8 @@ Title: Hosted Deposit and Withdrawal
 Author: SDF
 Status: Active
 Created: 2019-09-18
-Updated: 2022-05-05
-Version 2.3.0
+Updated: 2022-05-06
+Version 2.3.1
```
leighmcculloch, line 10: Since this is a new feature, the addition of a new status, this should be a minor version bump. Reserve patch version bumps for bug fixes.
```suggestion
Version 2.4.0
```

leighmcculloch (commented):
One change request, otherwise looks good to me. cc @JakeUrban could you also 👀?

ecosystem/sep-0024.md
lijamie98: Updated and pushed.

CassioMG (dismissed):
LGTM

JakeUrban (commented):
🤔  what is the relationship between `refunded` and `completed_at`? Can we also add a note to the deprecated `refunded` attribute that there is a transaction status for this now?

lijamie98:
> 🤔 what is the relationship between `refunded` and `completed_at`? Can we also add a note to the deprecated `refunded` attribute that there is a transaction status for this now?

Updated the `refunded: boolean` field. 

I think there are a couple of options:
1. `completed_at` sticks with the status `completed`.
2. List the statuses that are considered 'terminal' state, such as: `completed`, `error`, `refunded`, `no_market`, `too_small` and `too_large`
3. List the statuses that are considered `terminal` and non-errored states, such as `completed`, `refunded`.

Please feel free to comment if there are more sensible options.

JakeUrban:
I think the third option makes the most sense. Can we update the description of `completed_at`? Then I'll ✅

lijamie98:
> I think the third option makes the most sense. Can we update the description of `completed_at`? Then I'll ✅
updated.

JakeUrban (approved):

PR #1197, opened 2022-05-09
CAP-0040: Add security concern

jonjove:


Shaptic (dismissed):
I would argue people should never sign anything unless they know (slash can validate) what it represents, but this is still accurate.

core/cap-0040.md
```diff
@@ -347,7 +347,11 @@ None yet.
 
 ## Security Concerns
 
-None known.
+It is possible for an attacker to convince someone to unknowingly sign a hash
+which corresponds to a `TransactionSignaturePayload` for an account they
+control. This is only possible when the payload is 32 bytes long. To avoid these
+attacks, people should never sign a 32 byte payload unless they know the
+preimage.
```
leighmcculloch, line 354: Good catch, we should mention this. I think the warning is overly specific because attacks of this kind can occur with any length payload. The payload could be the signature of a hash for use in a contract on another network.
```suggestion
People and applications should never sign a transaction they do not completely
understand. There are many cases where signing a transaction not completely
understood can pose a security concern for the signer because the signer could
be authorizing something they would not approve of if they did understand the
transaction. This proposal introduces a new case with the Signed Payload Signer.
If a transaction requires the signature of a Signed Payload Signer and a signer
signs it without understanding the preimage of the payload, the signer could be
authorizing another Stellar transaction if the payload is a Stellar transaciton
hash, or the signer could be authorizing some other operation on another network
or application.
```

leighmcculloch (approved):

core/cap-0040.md
jonjove: This is definitely better, but I think you should expand on "completely". Specifically, you have to try to interpret the preimage in every relevant context. For example, it is theoretically possible that the preimage of the hash has different interpretations on different networks (eg. it might be a valid transaction on one network but not another, or it could be a valid transaction on both).

core/cap-0040.md
```diff
@@ -347,7 +347,16 @@ None yet.
...
+If a transaction requires the signature of a Signed Payload Signer and a signer
+signs it without understanding the preimage of the payload, the signer could be
+authorizing another Stellar transaction if the payload is a Stellar transaciton
+hash, or the signer could be authorizing some other operation on another network
+or application. People and applications should never sign a transaction they do
+not completely understand. There are many cases where signing a transaction not
+completely understood can pose a security concern for the signer because the
+signer could be authorizing something they would not approve of if they did
+understand the transaction. This proposal introduces a new case with the Signed
+Payload Signer.
```
leighmcculloch, line 359:
```suggestion
If a transaction requires the signature of a Signed Payload Signer and a signer
signs it without understanding the preimage of the payload, the signer could be
authorizing another Stellar transaction if the payload is a Stellar transaciton
hash, or the signer could be authorizing some other operation on another network
or application. People and applications should never sign a hash that they do
not completely understand the consequences of signing, in all the contexts that
their key provides authentication or authorization. People and applications
should never sign a transaction they do not completely understand. There are
many cases where signing a transaction not completely understood can pose a
security concern for the signer because the signer could be authorizing
something they would not approve of if they did understand the transaction. This
proposal introduces a new case with the Signed Payload Signer.
```

core/cap-0040.md
leighmcculloch: See https://github.com/stellar/stellar-protocol/pull/1197/commits/151ac193f259fdb1736aa8b4a1f4043d061ae910. How does that sound?

leighmcculloch (approved):

PR #1203, opened 2022-05-13
SEP-6 + SEP-31: clarify that before sending the payment, the sending party should make a GET request to retrieve the expected `amount_in`

marcelosalloum:
### What

Clarify that before sending the payment, the sending party should make a GET request to retrieve the expected `amount_in`.

Also added a few polishes to the text.

### Why

Some anchors could use additive fees in the transactions, which would make the amount proposed by the Wallet diverge from the one approved by the Anchor.

### DIFF
--- a/ecosystem/sep-0006.md
+++ b/ecosystem/sep-0006.md
@@ -105,8 +105,8 @@ sequenceDiagram
 
 ## API Endpoints
 
-* [`GET /deposit`](#deposit): required
-* [`GET /withdraw`](#withdraw): required
+* [`GET /deposit`](#deposit-2): required
+* [`GET /withdraw`](#withdraw-2): required
 * [`GET /deposit-exchange`](#deposit-exchange): optional
 * [`GET /withdraw-exchange`](#withdraw-exchange): optional
 * [`GET /info`](#info): required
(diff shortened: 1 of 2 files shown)


ecosystem/sep-0006.md
```diff
@@ -225,6 +229,7 @@ SEP-6 lays out many options for how deposit and withdrawal can work. These are r
 * **Make a request to `/deposit` or `/withdraw`.**
     * As a conservative measure, pass in any optional fields (including amount) that the wallet has on-hand, such as `email_address` and `account`.
     * In the `/withdraw` request, have the user enter the crypto account or bank account where they'd like their withdrawal to end up and provide that to the anchor via `dest` and `dest_extra`.
+    * Before making the payment, make a request to `[SEP-6]/transaction?id=<transaction_id>` to get the amount_in details. Although unusual and not recommended, some anchors may use additive fees, which would produce an `amount_in` different from the wallet's proposed amount.
```
JakeUrban, line 232: I think we need to approach additive fees as a valid use case.

```suggestion
    * Before making the payment, make a request to `[SEP-6]/transaction?id=<transaction_id>` to get the amount_in details. Some anchors may use additive fees, which would produce an `amount_in` different from the wallet's proposed amount.
```

JakeUrban (dismissed):
Approved with the suggestion

JakeUrban (approved):
LGTM!

PR #1214, opened 2022-05-24
SEP-31: add callback URL registration endpoint

### LINKED ISSUE #1209: SEP-31: add a transaction status callback mechanism

Sending Anchors with high transaction throughput must poll Receiving Anchors' `GET /transactions/:id` endpoint for many transactions concurrently in order to detect status changes in a timely manner. This is resource-intensive for both the Sending and Receiving Anchor.

A mechanism by which Sending Anchors can specify URLs for the Receiving Anchor to make requests to when a transaction's status changes.

One idea is to add a `callback_url` argument to `POST /transactions` and `PATCH /transactions/:id`. This would allow clients to define a URL with the initial creation of a transaction using `POST` but also allow clients to update the URL used using `PATCH`.

JakeUrban:
resolves #1209 

Adds a `PUT /transactions/:id/callback` endpoint for Sending Anchor to request callback requests to a specified URL. This removes the need for Sending Anchors to poll the statuses of in-progress transactions, reducing the chatter between services, especially for anchors with high transaction throughput.

I considered adding this functionality into the existing `POST /transactions` and `PATCH /transactions` requests as an optional parameter, but figured that it was likely for Receiving Anchors who did not yet implement this feature to simply ignore the `callback_url` request attribute, leading Sending Anchors to believe their callback URL was accepted but never receiving the callback requests.

Note: originally we decided not to add this functionality in #636. I disagree with my former self there and don't see why we wouldn't add this capability for the benefit of Sending Anchors that wanted it. Receiving Anchors can chose not to implement this until a Sending Anchors requests it. cc @gaclaudiu

### DIFF
--- a/ecosystem/sep-0031.md
+++ b/ecosystem/sep-0031.md
@@ -190,7 +190,7 @@ This requires the Receiving Anchor to implement the [SEP-38 Anchor RFQ API](sep-
 1. The Sending Anchor makes a `POST /transactions` request to create a transaction record with the Receiving Anchor. This request contains the `id`s returned by the `PUT /customer` requests, the `transaction.fields` values collected from the Sending Client, as well as the transaction information. The response will include a `id` that will be used to check the transaction data and status.
 1. Optionally, the Sending Anchor makes a `PUT /transactions/:id/callback` request to register a URL that the Receiving Anchor will make requests to when the transaction's status changes.
 1. The Sending Anchor makes a `GET /transactions/:id` request to ensure the status is `pending_sender`. If it is not, the Sending Anchor must wait until it receives a callback request or poll the `GET /transactions/:id` endpoint until the status becomes `pending_sender`.
-1. The Sending Anchor submits the payment transaction to Stellar using the information provided in the `POST /transactions` response.
+1. The Sending Anchor submits the payment transaction to Stellar using the information provided in the `GET /transactions/:id` response.
 1. If a callback was not registered, the Sending Anchor makes `GET /transactions/:id` requests until the transaction's `status` is `completed`, `error`, `pending_customer_info_update`, or `pending_transaction_info_update`. Otherwise, the Sending Anchor will receive status updates from the Receiving Anchor at the registered callback URL.
 1. If `completed`, the Sending Anchor should notify the Sending Client that funds have been delivered to the Receiving Client.
 1. If `error`, the Receiving Anchor should be contacted to resolve the situation.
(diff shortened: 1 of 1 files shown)


leighmcculloch (approved):
This new endpont is subject to the same authenticaton requirements of SEP-10? LGTM!

JakeUrban:
> This new endpont is subject to the same authenticaton requirements of SEP-10?

Yes it is, but I think thats clear from the authentication section. 

I'll wait for another member of the anchor eng team to approve before merging.

marcelosalloum (commented):
There seem to be a few details missing in this change:
* Add this endpoint to the [API Endpoints](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0031.md#api-endpoints) section.
* Mention the usage of this endpoint in the "detailed sending/receiving anchor flows"

JakeUrban:
Thanks for the comments @marcelosalloum ! I definitely missed those flows.

ecosystem/sep-0031.md
```diff
@@ -187,9 +188,10 @@ This requires the Receiving Anchor to implement the [SEP-38 Anchor RFQ API](sep-
...
-1. The Sending Anchor makes `GET /transactions/:id` requests until the transaction's `status` is `pending_sender`.
-1. The Sending Anchor submits the payment transaction to Stellar using the information provided in the `GET /transactions/:id` response.
-1. The Sending Anchor makes `GET /transactions/:id` requests until the transaction's `status` is `completed`, `error`, `pending_customer_info_update`, or `pending_transaction_info_update`.
+1. Optionally, the Sending Anchor makes a `PUT /transactions/:id/callback` request to register a URL that the Receiving Anchor will make requests to when the transaction's status changes.
+1. The Sending Anchor makes a `GET /transactions/:id` request to ensure the status is `pending_sender`. If it is not, the Sending Anchor must wait until it receives a callback request or poll the `GET /transactions/:id` endpoint until the status becomes `pending_sender`.
+1. The Sending Anchor submits the payment transaction to Stellar using the information provided in the `POST /transactions` response.
```
marcelosalloum, line 193: This wording contradicts what we agreed for PR [#1203](https://github.com/stellar/stellar-protocol/pull/1203/files#diff-6abf4f55bde3bfb0f1b81e2437e70d45ea84c64213c3d09715bd901e3ec2a101R191). 

The wallet should actually double-check the amount_in from the GET /transactions/:id before sending the payment.

marcelosalloum (commented):
Not so sure about one line of your change:

marcelosalloum (approved):

PR #1223, opened 2022-06-02
SEP-0039: Changing NFT metadata's `url` field to use an HTTP gateway

ElliotFriend:
After some discussion within the Stellar Quest community, relating to a
newly released side quest all about NFTs, we noticed the SEP-0039 example
metadata uses the `ipfs://` protocol scheme for the metadata field of `url`.
It seems like it would be more compatible with existing ecosystem practices
to use an HTTP(S) gateway in this field, rather than the direct IPFS url.

This PR would implement exactly that suggestion in the example JSON within
this document. I've used my buttflare URL that is already used later on in
the document, under the 'Representing NFTs' section.

Signed-off-by: Elliot Voris <elliot@voris.me>

### DIFF
--- a/ecosystem/sep-0039.md
+++ b/ecosystem/sep-0039.md
@@ -50,7 +50,7 @@ Since your NFT can be anything, another best practice is storing metadata that d
 {
   "name": "A demonstration of NFT metadata",
   "description": "This is a description of the cool NFT used for an informational SEP demo.",
-  "url": "ipfs://QmXoypizjW3WknFiJnKLwHCnL72vedxjQkDDP1mXWo6uco/I/m/RickAstleyNeverGonnaGiveYouUp7InchSingleCover.jpg",
+  "url": "https://cloudflare-ipfs.com/ipfs/QmXoypizjW3WknFiJnKLwHCnL72vedxjQkDDP1mXWo6uco/I/m/RickAstleyNeverGonnaGiveYouUp7InchSingleCover.jpg",
   "issuer": "GALAXYVOIDAOPZTDLHILAJQKCVVFMD4IKLXLSZV5YHO7VY74IWZILUTO",
   "code": "DEMOASSET"
 }


leighmcculloch:
> it would be more compatible with existing ecosystem practices

Which applications do not support `ipfs://` URLs today, and do you think we have an opportunity to improve them?

For applications that do not recognize the `ipfs://` URLs, they could simply replace the prefix with a gateway URL. This provides some freedom to the client, since the client can choose which gateway to use, or the client can use its own node to fetch the file using the IPFS CID.

For example:

```js
const url = "ipfs://QmXoypizjW3WknFiJnKLwHCnL72vedxjQkDDP1mXWo6uco/I/m/RickAstleyNeverGonnaGiveYouUp7InchSingleCover.jpg";
const httpUrl = url.replace(/^ipfs:\/\//, "https://cloudflare-ipfs.com/ipfs/");
console.log(httpUrl);
// "https://cloudflare-ipfs.com/ipfs/QmXoypizjW3WknFiJnKLwHCnL72vedxjQkDDP1mXWo6uco/I/m/RickAstleyNeverGonnaGiveYouUp7InchSingleCover.jpg"
```

ElliotFriend:
I was looking mostly at what litemint has done for their metadata, and mimicking some of that. They have both `ipfs://` and `https://` urls in different fields, although I'm not sure what led to those choices:

```json
{
  "some": "thing here",
  "image": "ipfs://QmaByzRRTjSK14N3EiQqfwxde6GRQKEBdQWZpzwyJigDhd",
  "url": "https://ipfs.io/ipfs/QmaByzRRTjSK14N3EiQqfwxde6GRQKEBdQWZpzwyJigDhd",
  "another": "thing here",
  "etc": "etc"
}
```

I can certainly see the added flexibility to utilize whichever HTTPS gateway, though.

I'm gonna tag @tyvdh here, since he's got a lot more insights than I do about the pros, cons, etc. I think he'll be able to better answer the questions you brought up.

tyvdh:
`fetch` doesn't which is arguably the most popular API for retrieving resources. Also very few browsers, Chrome included, would actually render a resource from an ipfs:// protocol. imo we should require `url` to be a valid resource endpoint for the SEP example metadata and if people want to load it with ipfs:// and assume that frontends or backends will adjust as needed fine, but as a defacto default it should be a traditional valid protocol and url.

alejomendoza:
> > it would be more compatible with existing ecosystem practices
> 
> Which applications do not support `ipfs://` URLs today? Do you think we have an opportunity to improve them?
> 
> For applications that do not recognize the `ipfs://` URLs, they could replace the prefix with a gateway URL. This provides some freedom to the client, since the client can choose which gateway to use, or the client can use its own node to fetch the file using the IPFS CID.
> 
> For example:
> 
> ```js
> const url = "ipfs://QmXoypizjW3WknFiJnKLwHCnL72vedxjQkDDP1mXWo6uco/I/m/RickAstleyNeverGonnaGiveYouUp7InchSingleCover.jpg";
> const httpUrl = url.replace(/^ipfs:\/\//, "https://cloudflare-ipfs.com/ipfs/");
> console.log(httpUrl);
> // "https://cloudflare-ipfs.com/ipfs/QmXoypizjW3WknFiJnKLwHCnL72vedxjQkDDP1mXWo6uco/I/m/RickAstleyNeverGonnaGiveYouUp7InchSingleCover.jpg"
> ```

Leaving Dapps pick their gateway seems like the right solution, you don't want to end up with missing resources because of a gateway's downtime, cloudflare ipfs gateway is not as reliable as you think, several times we had to switch to ipfs.io and the other way around.

tyvdh:
> [ERC721](https://ethereum.stackexchange.com/questions/112927/linking-ipfs-in-erc-721) best practices, and opensea metadata [standards](https://docs.opensea.io/docs/metadata-standards).

Looks like according to [Opensea](https://docs.opensea.io/docs/metadata-standards#ipfs-and-arweave-uris) the answer to
> Is the idea that the protocol signals the platform so if it were some other decentralized storage solution like Arweave or Storj, etc. each would have it's own protocol prefix?

is yes

leighmcculloch:
> `url` in my mind denotes a, well, url. 

The terms URL and URI are not as well defined as we might assume. [RFC3305](https://datatracker.ietf.org/doc/html/rfc3305#section-2.3) even goes into some detail about this confusion and notes that URL is an informal concept, and that attempts to clear up the ambiguity have largely gone unsuccessful.

In the IPFS ecosystem the `ipfs://` scheme is described as a URL: https://docs.ipfs.io/how-to/address-ipfs-on-web/#native-urls. It is also described as a URI because URLs are URIs. I think describing `ipfs://QmXoypizjW3WknFiJnKLwHCnL72vedxjQkDDP1mXWo6uco/I/m/RickAstleyNeverGonnaGiveYouUp7InchSingleCover.jpg` as a URL this is accurate because it defines the location of a Rick Astley file, rooted at a content root given as the CID. Just because mainstream browsers do not support the scheme doesn't make it not a URL.

This is consistent with OpenSea's recommendations as you pointed out:

>If you use IPFS to host your metadata, your URL should be in the format `ipfs://<hash>`.

Ref: https://docs.opensea.io/docs/metadata-standards#ipfs-and-arweave-uris

And from other documentation that forms good best practices:

>If you want to use gateway URLs for convenience or interoperability, you can always generate them in your application's presentation layer.

Ref: https://docs.ipfs.io/how-to/best-practices-for-nft-data/#metadata

Using `ipfs://` URLs in the metadata does not prevent applications using gateways. Doing so is a simple string replace on the `ipfs://` prefix: https://github.com/stellar/stellar-protocol/pull/1223#issuecomment-1144401318.

Shaptic:
At the time of the proposal, we intentionally did not limit ourselves to HTTP(S). You might want to store your resource as a magnet link, on an FTP server, etc. Implementors should do what they see fit!

PR #1231, opened 2022-06-08
CAP-51: add several hostfns, add explaination of BigInt Convertion, SCStatus return.

jayz22:
- [ ]  Keep the Hash and public key as SCO types, as they are fairly ubiquitous and specially-purposed. But provide the following functions for interoperability:
    - [x]  Make them union types
    - [ ]  Convert between SCO_ACCOUNT_ID and SCO_PUBLIC_KEY (will be done later in whichever CAP `ACCOUNT_ID` is defined)
- [x]  Add vec functions for slicing
- [x]  ScBigInt signess:
    - [x]  add an enum {POSITIVE, ZERO, NEGATIVE} to denote its sign (instead of just a POSITIVE).
    - [x]  Update its description
    - [x]  Update the description of ScBigInt to be implementation independent
- [x]  Map: provide host functions (similar to lower-bound, upper-bound) to iterate through a map, as well as knowing when to end.
- [x]  Binary:
    - [x]  Memcpy
    - [x]  Interop: SCO_BINARY and SCO_HASH
    - [x]  Interop: SCO_BINARY and PUBLIC_KEY
- [x]  Add a small section on the status type, that they are not supposed to be stored, and explain the ambiguity issue.
- [x]  Variadic functions for contract calling.

### DIFF
--- a/core/cap-0051.md
+++ b/core/cap-0051.md
@@ -6,7 +6,7 @@ Title: Smart Contract Host Functions
 Working Group:
     Owner: Jay Geng <@jay>
     Authors: TBD
-    Consulted: Graydon Hoare <@graydon>, Leigh McCulloch <@leighmcculloch>, Nicolas Barry <@MonsieurNicolas>
+    Consulted: Graydon Hoare <@graydon>, Leigh McCulloch <@leighmcculloch>, Nicolas Barry <@MonsieurNicolas>, Siddharth Suresh <@sisuresh>
 Status: Draft
 Created: 2022-05-20
 Discussion: TBD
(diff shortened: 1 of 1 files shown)


core/cap-0051.md
```diff
@@ -153,6 +165,10 @@ func $call2(param $contract i64) (param $func i64) (param $arg1 i64) (param $arg
 func $call3(param $contract i64) (param $func i64) (param $arg1 i64) (param $arg2 i64) (param $char i64) (result i64)
 
 func $call4(param $contract i64) (param $func i64) (param $arg1 i64) (param $arg2 i64) (param $arg3 i64) (param $arg4 i64) (result i64)
+
+/// Calls another function $func in a contract $contract with variadic arguments stored in a vector $args_vec.
+/// If the call is successful, it forwards the result of the called function. Otherwise, it traps.
+func $call(param $contract i64) (param $func i64) (param $args_vec i64) (result i64)
```
leighmcculloch, line 171: Do we need call1 to 4 if we have this?

leighmcculloch (approved):

core/cap-0051.md
jayz22: Probably not, but I'll do some experimentation before confirming and removing them.

PR #1240, opened 2022-06-15
CAP-42 - updated `txSet` validation rules.

dmkozh:
The `txSet` components have to be non-empty or otherwise it's possible to build tx sets that contain the same transactions


MonsieurNicolas (approved):
thank you!

PR #1244, opened 2022-06-21
SEP-12: Add callback signature

C0x41lch0x41:
Proposed changes for the callback status:
* Add **timestamp** field to ensure freshness of the call
* Add `X-Stellar-Signature` HTTP header: Base64 signature of the body. The signature is computed using the Anchor Stellar private key and verified using the corresponding public key

### DIFF
--- a/ecosystem/sep-0012.md
+++ b/ecosystem/sep-0012.md
@@ -476,7 +476,7 @@ Anchor has no information on the customer | `404 Not Found`
 Allow the wallet to provide a callback URL to the anchor. The provided callback URL will replace (and supercede) any previously-set callback URL for this account.
 
 Whenever the user's `status` field changes, the anchor will issue a POST request to the callback URL. 
-The payload of the POST request will be the same as the response of [`GET /customer`](#customer-get) with an additional field `timestamp` set to the current timestamp at the time the Anchor issue the POST request. This is used to assure the freshness of the request and to prevent this request to be replay in the future. Additionally the Anchor will provide a base64 signature of the POST request payload in the HTTP Header `X-Stellar-Signature`.
+The payload of the POST request will be the same as the response of [`GET /customer`](#customer-get) with an additional field `timestamp` set to the current timestamp at the time the Anchor issue the POST request. This is used to assure the freshness of the request and to prevent this request to be replayed in the future. Additionally the Anchor will provide a base64 signature of the POST request payload in the HTTP Header `X-Stellar-Signature`.
 The signature is computed using the Stellar private key of the anchor used in the [SEP-10](sep-0010.md) authentication.
 
 * It's the wallet's responsibility to verify the signature using the corresponding Stellar public key provided in the Anchor [`stellar.toml`](sep-0001.md).
--- a/ecosystem/sep-0012.md
+++ b/ecosystem/sep-0012.md
@@ -477,9 +477,9 @@ Allow the wallet to provide a callback URL to the anchor. The provided callback 
 
 Whenever the user's `status` field changes, the anchor will issue a POST request to the callback URL. 
 The payload of the POST request will be the same as the response of [`GET /customer`](#customer-get) with an additional field `timestamp` set to the current timestamp at the time the Anchor issue the POST request. This is used to assure the freshness of the request and to prevent this request to be replayed in the future. Additionally the Anchor will provide a base64 signature of the POST request payload in the HTTP Header `X-Stellar-Signature`.
-The signature is computed using the Stellar private key of the anchor used in the [SEP-10](sep-0010.md) authentication.
+The signature is computed using the Stellar private key linked to the `SIGNING_KEY` field of the anchor's [`stellar.toml`](sep-0001.md).
 
-* It's the wallet's responsibility to verify the signature using the corresponding Stellar public key provided in the Anchor [`stellar.toml`](sep-0001.md).
+* It's the wallet's responsibility to verify the signature using the corresponding Stellar `SIGNING_KEY` field of the anchor's [`stellar.toml`](sep-0001.md).
 * It's the wallet's responsibility to verify the freshness of the request by comparing the `timestamp` in the request with the current timestamp at the time of the reception and discard every request above a threshold of few seconds.
 * It's the wallet's responsibility to send a working callback URL to the anchor.
 
--- a/ecosystem/sep-0012.md
+++ b/ecosystem/sep-0012.md
@@ -475,7 +475,8 @@ Anchor has no information on the customer | `404 Not Found`
 
 Allow the wallet to provide a callback URL to the anchor. The provided callback URL will replace (and supercede) any previously-set callback URL for this account.
 
-Whenever the user's `status` field changes, the anchor will issue a POST request to the callback URL. 
+Whenever the user's `status` field changes, the anchor will issue a POST request to the callback URL. The payload of the POST request will be the same as the response of [`GET /customer`](#customer-get).
+Anchors will submit POST requests until the user's status changes to `ACCEPTED` or `REJECTED`. If a wallet needs to watch a user's KYC status after that, it will need to set a callback again.
 
 ### PUT Request
 
@@ -505,24 +506,18 @@ Name | Type | Description
 POST [url from PUT request]
 ```
 
-Name | Type | Description
----- | ---- | -----------
-`id` | `string` | The ID of the registered customer.
-`account` | `G...` or `M...` string | Account of the regiestered customer
-`memo` | string | (optional) the client-generated [memo](https://developers.stellar.org/docs/glossary/transactions/#memo) that uniquely identifies the customer.
-`memo_type` | string | (**deprecated**, optional) type of `memo`. One of `text`, `id` or `hash`. Deprecated because memos should always be of type `id`, although anchors should continue to support this parameter for outdated clients. If `hash`, `memo` should be base64-encoded. If a memo is present in the decoded SEP-10 JWT's `sub` value, this parameter can be ignored. See the [Shared Accounts](#shared-omnibus-or-pooled-accounts) section for more information.
-`type` | `string` | (optional) the type of action the customer is being KYCd for. See the [Type Specification](#type-specification) below.
-`status` | `string` | One of the values described in [Provided Field Statuses](#provided-field-statuses)
+See [`GET /customer reponse`](#response) for the POST request fields.
+
 
 In order to validate the integrity and provenance of the request, the Anchor MUST include a signature in an additional HTTP Header `X-Stellar-Signature`.
-This Header MUST follow the specification: `X-Stellar-Signature: t=<timestamtp>, h=<host>, s=<base64 signature>` where:
+This Header MUST follow the specification: `X-Stellar-Signature: t=<timestamp>, s=<base64 signature>` where:
 * __timestamp__ is the current timestamp at the time the callback is sent. This is used to assure the freshness of the request and to prevent this request to be replayed in the future. 
-* __host__ is the host the callback is being sent to. The host will be part of the signature. This is used to prevent malicious wallet to relay the request to another wallet
-* __base64 signature__ is the base64 encoding of the request signature. We explain below how to compute and verify this signature. The signature is computed using the Stellar private key linked to the `SIGNING_KEY` field of the anchor's [`stellar.toml`](sep-0001.md).
+* __base64 signature__ is the base64 encoding of the request signature. We explain below how to compute and verify this signature. The signature is computed using the Stellar private key linked to the `SIGNING_KEY` field of the anchor's [`stellar.toml`](sep-0001.md). Note that the timestamp and the HTTP Header Host will be part of the signature to prevent replay and relay attacks.
 
 It is the wallet's responsability to:
 * Verify the signature using the corresponding Stellar `SIGNING_KEY` field of the anchor's [`stellar.toml`](sep-0001.md).
-* Verify the freshness of the request by comparing the `timestamp` in the request with the current timestamp at the time of the reception and discard every request above a threshold of few seconds.
+* Verify the freshness of the request by comparing the `timestamp` in the request with the current timestamp at the time of the reception and discard every request above a threshold of few seconds (1 or 2 minute(s) maximum).
+* Verify that the callback is not being relayed by validating that the HTTP Header __Host__ is equal to the Wallet Host.
 * Send a working callback URL to the anchor.
 
 ### VERIFY signature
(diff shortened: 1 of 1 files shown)


C0x41lch0x41:
@leighmcculloch @JakeUrban @howardtw please take a look thanks.

ecosystem/sep-0012.md
```diff
@@ -475,9 +475,107 @@ Anchor has no information on the customer | `404 Not Found`
...
+	if err != nil {
+		log.Fatal(err)
+	}
+	bodyBytes, err := io.ReadAll(resp.Body)
+	if err != nil {
+		log.Fatal(err)
+	}
+	log.Println(string(bodyBytes))
+}
+```
```
leighmcculloch, line 578: I think we'd benefit from pseudocode to keep this more succinct.

leighmcculloch (commented):
👏🏻 Nice. I've left some feedback inline. 

In addition, could you bump the minor component of the version at the top of the document, and add a short one sentence change log entry at the bottom? The change log won't already exist, but for inspiration look at: https://github.com/stellar/stellar-protocol/blob/e19ae81bd7808675c3b109a774f33810a8f49951/ecosystem/sep-0010.md#changelog.

ecosystem/sep-0012.md
```diff
@@ -475,9 +475,107 @@ Anchor has no information on the customer | `404 Not Found`
 
 Allow the wallet to provide a callback URL to the anchor. The provided callback URL will replace (and supercede) any previously-set callback URL for this account.
 
-Whenever the user's `status` field changes, the anchor will issue a POST request to the callback URL. The payload of the POST request will be the same as the response of [`GET /customer`](#customer-get).
+Whenever the user's `status` field changes, the anchor will issue a POST request to the callback URL. 
+The payload of the POST request will be the same as the response of [`GET /customer`](#customer-get) with an additional field `timestamp` set to the current timestamp at the time the Anchor issue the POST request. This is used to assure the freshness of the request and to prevent this request to be replayed in the future. Additionally the Anchor will provide a base64 signature of the POST request payload in the HTTP Header `X-Stellar-Signature`.
```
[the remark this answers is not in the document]
C0x41lch0x41: Yes that's why I added a timestamp but the relay is still possible if executed in a short timeframe.
I think I understand what you mean by signing the host header now. I thought about it the other way, in my mind the host was set to the KYC SEP-12 Server whereas what you suggest is to have a host value set to the destination wallet correct?
Something like `X-Stellar-Signature: t=<timestamp>, h=api.vibrantapp.com, s=<base64 signature>` ?
The SEP says that the callback needs to use the parameters of the [`Customer GET`](https://github.com/C0x41lch0x41/stellar-protocol/blob/patch-2/ecosystem/sep-0012.md#customer-get) endpoint. One of parameter is `account` (and optionally `memo`). It is optional since the endpoint uses SEP-10 JWT token but we could say this is mandatory here.
So that:
- `timestamp` prevents replay
- `host` prevents relay from one wallet backend to another
- `account` prevents ID clashing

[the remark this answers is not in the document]
leighmcculloch: If I'm understanding correctly the callback the HTTP host will be the wallet server, since the SEP-12 server will be sending a request to the wallet server.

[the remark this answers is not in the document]
C0x41lch0x41: Yes exactly. I was think on including it in the X-Stellar-Signature header as well but maybe that's not necessary.

C0x41lch0x41:
@leighmcculloch l included the changes we discussed. Let me know what you think.
Thanks!

ecosystem/sep-0012.md
```diff
@@ -475,13 +475,9 @@ Anchor has no information on the customer | `404 Not Found`
 
 Allow the wallet to provide a callback URL to the anchor. The provided callback URL will replace (and supercede) any previously-set callback URL for this account.
 
-Whenever the user's `status` field changes, the anchor will issue a POST request to the callback URL. The payload of the POST request will be the same as the response of [`GET /customer`](#customer-get).
+Whenever the user's `status` field changes, the anchor will issue a POST request to the callback URL.
```
JakeUrban, line 478: Are we removing the requirement to match the `GET /customer` response? The fields defined in the "Callback POST Request" don't match, and thats a breaking change. Can we keep continue to use the `GET /customer` response body as the payload and just add the header for verification?

ecosystem/sep-0012.md
```diff
@@ -503,6 +499,72 @@ Name | Type | Description
...
+  * The character `.`
+  * The wallet host to send the callback request to
+  * The character `.`
+  * The callback request body
+* Compute the signature using the Anchor private seed
+* Base64 encode the signature
+* Build the `X-Stellar-Signature` header:
+  * `X-Stellar-Signature: t=<current timestamp>, h=<wallet host>, s=<base64 encoded signature>`
+
+Anchors will submit POST requests until the user's status changes to `ACCEPTED` or `REJECTED`. If a wallet needs to watch a user's KYC status after that, it will need to set a callback again.
```
JakeUrban, line 566: Can we add this to the start of the request definition so it doesn't get missed?

JakeUrban (changes requested):

JakeUrban:
Thanks for providing this solution @C0x41lch0x41, I think we should replicate what we finalize here in SEP-31 and other SEPs that use a callback mechanism. The one issue I see is the breaking change made to the callback request body.

ecosystem/sep-0012.md
leighmcculloch: We should add this statement back, it doesn't need to be removed now that we've moved the new fields into the header.

ecosystem/sep-0012.md
```diff
@@ -475,13 +475,9 @@ Anchor has no information on the customer | `404 Not Found`
 
 Allow the wallet to provide a callback URL to the anchor. The provided callback URL will replace (and supercede) any previously-set callback URL for this account.
 
-Whenever the user's `status` field changes, the anchor will issue a POST request to the callback URL. The payload of the POST request will be the same as the response of [`GET /customer`](#customer-get).
+Whenever the user's `status` field changes, the anchor will issue a POST request to the callback URL. 
 
-It's the wallet's responsibility to send a working callback URL to the anchor.
-
-Anchors will submit POST requests until the user's status changes to `ACCEPTED` or `REJECTED`. If a wallet needs to watch a user's KYC status after that, it will need to set a callback again.
```
leighmcculloch, line 482: We probably don't need to move this discussion either. It could stay here.

ecosystem/sep-0012.md
JakeUrban: The request body defined in the request description doesn't match the `GET /customer` response though, that needs to change too.

ecosystem/sep-0012.md
```diff
@@ -503,6 +499,72 @@ Name | Type | Description
...
+* Verify the signature using the corresponding Stellar `SIGNING_KEY` field of the anchor's [`stellar.toml`](sep-0001.md).
+* Verify the freshness of the request by comparing the `timestamp` in the request with the current timestamp at the time of the reception and discard every request above a threshold of few seconds.
+* Send a working callback URL to the anchor.
+
+### VERIFY signature
+
+* Check that callback request has `X-Stellar-Signature` header
+* Parse the header and extract:
+  * Key `t`: __timestamp__
+  * Key `h`: __host__
```
leighmcculloch, line 533: Hmm, an issue with including the host here is the receiving server is unlikely to validate it. If no one validates it, or the presence of the field here doesn't change whether the request is valid to be received by this server or not, then it won't protect against relay.

leighmcculloch: I guess all that matters is the receiving host include their own value into their signature verification fn, and not take the value from here. That could be motivation for not including the value here since it might encourage poor practice of just using the value that is here.

ecosystem/sep-0012.md
```diff
@@ -503,6 +499,72 @@ Name | Type | Description
...
+  * Key `s`: __base64 signature__
+* Verify the request freshness: _current timestamp_ - __timestamp__ < few seconds (2-5)
```
leighmcculloch, line 535: >* Verify the request freshness: _current timestamp_ - __timestamp__ < few seconds (2-5)

We might need this period to be a bit longer to allow for time drift. I would go for 1 minute at least.

ecosystem/sep-0012.md
C0x41lch0x41: I described the fields to send in the POST request, they are the same than the `GET /customer` field with the only modification that `account` (and optionally `memo`) is mandatory. This is to make sure that the request cannot be replayed against a different Wallet that may have the same id.

ecosystem/sep-0012.md
C0x41lch0x41: Fair point. 
We can remove the host from the header and use the `Host` HTTP header.

ecosystem/sep-0012.md
leighmcculloch: This won't protect against replace of a different wallet, because the same account and memo ID might be being used by two wallets.

C0x41lch0x41: Yes that's why I added the host in the signature as well. I think the host + id is sufficient though. I'll make the modification.

C0x41lch0x41: @JakeUrban one thing about the parameters of POST request (callback). It still needs to have the `status` field in addition to the fields from `GET /customer` correct? The callback needs to report the KYC status of the user.

JakeUrban: Yes, it needs to have `status`, but thats included in the `GET /customer` response.

C0x41lch0x41: Right I was looking at the `/customer` request. Thanks!

C0x41lch0x41:
@JakeUrban @leighmcculloch pushed an update to integrate your comments. Let me know what you think. Thanks.

JakeUrban (dismissed):
This is great, LGTM!

ecosystem/sep-0012.md
```diff
@@ -503,6 +500,58 @@ Name | Type | Description
...
+### COMPUTE signature
+
+* Prepare the callback [request body](#response)
+* Prepare the payload to sign:
+  * Current timestamp (as a string)
+  * The character `.`
+  * The wallet Host to send the callback request to
+  * The character `.`
+  * The callback request body
+* Compute the signature using the Anchor private seed
```
howardtw, line 550: ❓ which encryption algorithm should the anchor use to generate the signature with the private key?

ecosystem/sep-0012.md
```diff
@@ -503,6 +500,58 @@ Name | Type | Description
...
+```
+POST [url from PUT request]
+```
+
+See [`GET /customer reponse`](#response) for the POST request fields.
+
+
+In order to validate the integrity and provenance of the request, the Anchor MUST include a signature in an additional HTTP Header `X-Stellar-Signature`.
+This Header MUST follow the specification: `X-Stellar-Signature: t=<timestamp>, s=<base64 signature>` where:
+* __timestamp__ is the current timestamp at the time the callback is sent. This is used to assure the freshness of the request and to prevent this request to be replayed in the future.
```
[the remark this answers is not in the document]
C0x41lch0x41: I think timestamp is always the number of seconds since epoch (1656026867). What other format do you think about?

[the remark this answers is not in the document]
howardtw: I wasn't thinking about a particular timestamp, but I think we could be clear here saying that it's the unix timestamp 😃

ecosystem/sep-0012.md
C0x41lch0x41: This is defined by the private seed. I think for stellar it can only be ed25519. This should be abstracted by the SDK.
For example in [Go](https://github.com/stellar/go/blob/b29a93b358b6669d951d4902c0bed6e1d67039fc/keypair/from_address.go#L56):
```go
import (
 b64 "encoding/base64"
 "github.com/stellar/go/keypair"
)

func sign(seed string, bufferToSign []byte) string {
  stellarKey := keypair.Parse(seed)
  signature := stellarKey.Sign(bufferToSign)
  return b64.StdEncoding.EncodeToString(signature)
}
```

ecosystem/sep-0012.md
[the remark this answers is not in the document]
C0x41lch0x41: Right, I added **Unix** to make sure. Thanks!

ecosystem/sep-0012.md
howardtw: Ah I mistakenly read it as compute the hash.

Can we add the final payload here to make it more clear on how the concatenation should be done?
```suggestion
* Sign the payload '<timestamp>.<host>.<body>' using the Anchor private key
```

howardtw (dismissed):

C0x41lch0x41:
> I think we should replicate what we finalize here in SEP-31 and other SEPs that use a callback mechanism.

@JakeUrban how do you want to this? Should we have a separate SEP that describes how callbacks should be handled and refer it in every SEPs using callbacks? Or we just port the same content in the relevant SEPs?

C0x41lch0x41:
@howardtw can I have a final review to merge it? Thanks!

JakeUrban:
@C0x41lch0x41 I would rather port the same content into the relevant SEP, if thats alright with you.

howardtw (approved):

howardtw:
Speaking of porting, SEP-24 also has the `on_change_callback` that requires this enhancement

C0x41lch0x41:
Sounds good I'll look into the other related SEPs and add the requirements.
Thanks all.
@JakeUrban do you mind merging this? (I don't have the access to do so).

leighmcculloch (commented):
One comment. Have we implemented this and prototyped it?

ecosystem/sep-0012.md
```diff
@@ -503,6 +500,58 @@ Name | Type | Description
...
+* Send a working callback URL to the anchor.
+
+### VERIFY signature
+
+* Check that callback request has `X-Stellar-Signature` header
+* Parse the header and extract:
+  * Key `t`: __timestamp__
+  * Key `s`: __base64 signature__
+* Verify the request freshness: _current timestamp_ - __timestamp__ < few seconds (1-2 minute(s) max)
+* Extract the HTTP Header __Host__, validate that it is equal to the Wallet Host
```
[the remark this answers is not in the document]
C0x41lch0x41: Yes that's a good point. I'll create another PR for that thanks.

C0x41lch0x41:
I did have a prototype in Go that is working.

ecosystem/sep-0012.md
```diff
@@ -503,6 +500,58 @@ Name | Type | Description
...
+  * The __body__
+* Verify the signature using the correct `SIGNING_KEY`
+
+### COMPUTE signature
+
+* Prepare the callback [request body](#response)
+* Prepare the payload to sign:
+  * Current timestamp (as a string)
+  * The character `.`
+  * The wallet Host to send the callback request to
```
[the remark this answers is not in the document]
C0x41lch0x41: Hi @alcalawil 
This is the host of the wallet to which the anchor will send the callback.

So for example if an anchor wants to send a callback to https://www.vibrantapp.com/kycstatus the host will be **www.vibrantapp.com**

Let me know if that makes sense.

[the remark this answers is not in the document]
alcalawil: Thanks. Also, I built this javascript function to generate the header, can you check if it is correct please?

```
function buildSignatureHeader(walletHost, reqBody) {
  const _reqBody = JSON.stringify(reqBody);
  const timestampInSeconds = Math.floor(Date.now() / 1000);
  const signature = Buffer.from(`${timestampInSeconds}.${walletHost}.${_reqBody}`).toString('base64');
  return `t=${timestampInSeconds}, s=${signature}`;
}
```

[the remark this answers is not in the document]
C0x41lch0x41: You need to sign the string: 
```javascript
Buffer.from(`${timestampInSeconds}.${walletHost}.${_reqBody}`)
```
with the Anchor private key. Then you need to base64 encode the signature.
You can use the [stellar javacript sdk](https://stellar.github.io/js-stellar-sdk/) to sign the buffer:
```javascript
const buffer = Buffer.from(`${timestampInSeconds}.${walletHost}.${_reqBody}`);
const keypair = StellarSdk.Keypair.fromSecret(privateKey);
const signature = keypair.sign(buffer).toString('base64');
return `t=${timestampInSeconds}, s=${signature}`;
```
Note that the private key should be handled as a secret. So you should not paste it in your code but load it from KMS/HashiCorp/etc or injected in an environment variable using Kubernetes secret for example.
I hope that helps.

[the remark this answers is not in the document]
alcalawil: Of course, I forgot the signature, thanks  🙏

PR #1252, opened 2022-06-30
Fix typo in cap 54

paulbellamy:


jonjove:
@paulbellamy Types that start with `Keyed` have the key specified, types that don't don't. Compare against `KeyedAccountAuthorization` and `Ed25519Authorization`. The `nonce` is used for replay prevention within the contract, so all `Authorization` types have a nonce.

For most functions, users will provide `KeyedAuthorization`. But admin functions take raw `Authorization` because the admin is actually recorded on-chain.

PR #1263, opened 2022-07-11
SEP-24: Add callback signature requirement

C0x41lch0x41:
Adding callback signature requirements from SEP12


C0x41lch0x41:
@JakeUrban please take a look thanks!

ecosystem/sep-0024.md
```diff
@@ -402,6 +446,8 @@ Because `callback` is used to notify the wallet that the interactive flow is com
 Wallet applications display content rendered by a third party anchor service. It is recommended to use the [`rel=noopener`](https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types/noopener) attribute on links or the `noopener` feature for [`Window.open()`](https://developer.mozilla.org/en-US/docs/Web/API/Window/open#window_functionality_features).
 
 Note that using `noopener` prohibits the use of `postMessage` callbacks. If `postMessage` callbacks are required for your implementation, it recommended to only open URLs from anchors that you, the wallet developer, trust.
+   
+To validate the provenance and integrity of the callback, anchors [must sign it and the UI must validate the signature](#callback-signature) before taking any action.
```
howardtw, line 450: For on_change_callbacks, the callbacks are sent after the interactive flows when there's an update to `statue` or `kyc_verified` props; for callbacks, it looks like the callbacks are also sent after the user finishes the interactive flow.

If I understand it correctly, we would want to take `UI` out from the sentence.
```suggestion
To validate the provenance and integrity of the callback, anchors [must sign it](#callback-signature) and wallet servers must validate the signature before taking any action.
```

howardtw (approved):
LGTM. Let's have another pair of eyes on it before merging it tho. @leighmcculloch can you take a look since @JakeUrban is out?

ecosystem/sep-0024.md
```diff
@@ -87,6 +87,50 @@ In order for browsers-based wallets to validate the CORS headers, as [specified
 
 This protocol involves the transfer of value, and so HTTPS is required for all endpoints for security.  Wallets and anchors should refuse to interact with any insecure HTTP endpoints.
 
+## Callback signature
```
leighmcculloch, line 90: What should implementers do, such as wallets, if they are currently integrated with a server that doesn't provide the callback? We should really be stating what a wallet should do in this case.

C0x41lch0x41: Do you mean what should wallets do if receiving a callback from an Anchor **that does not provide** a signature?

leighmcculloch: Yup. It might be important, given that this is an active SEP with many existing implementers, for us to note for clients that they should only expect a callback signature from anchors that send callback signatures.

The `/info` endpoint has a list of `features`, and we could include a boolean there for this so that clients can find out if the server supports it.

We should probably do the same thing with the other SEPs that are also adopting this.

cc @JakeUrban

C0x41lch0x41: I am not sure we should design the ability to not sign the callback as it may slow down the implementation. 
Since there is a real security risk of trusting callbacks without signature, I'd rather say that wallets should not trust a callback that is not signed. 
Now the wallet can still choose to deviate from this spec and have a period of time where the unsigned callback is trusted until the Anchor implement the changes.

leighmcculloch (approved):

ecosystem/sep-0024.md
```diff
@@ -380,8 +424,8 @@ The basic parameters are summarized in the table below.
...
-`on_change_callback` | string | (optional) [`postMessage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) or a URL that the anchor should `POST` a JSON message to when the `status` or `kyc_verified` properties change.
+`callback` | string | (optional) [`postMessage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) or a URL that the anchor should `POST` a JSON message to when the user successfully completes the interactive flow. The callback needs to be signed by the anchor and the signature needs to be verified by the wallet according to the [callback signature specification](#callback-signature).
+`on_change_callback` | string | (optional) [`postMessage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) or a URL that the anchor should `POST` a JSON message to when the `status` or `kyc_verified` properties change. The callback needs to be signed by the anchor and the signature needs to be verified by the wallet according to the [callback signature specification](#callback-signature).
```
JakeUrban, line 428: We need to clarify that the added statement only applies to URLs, not `postmessage` callbacks.

JakeUrban (commented):
One requested change, otherwise LGTM

ecosystem/sep-0024.md
JakeUrban: [postmessage](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) is a JavaScript construct, its not made with a HTTPS request, so the description of how the callback should be made doesn't fit the mechanism by which the message is sent in this case.

C0x41lch0x41: Yes sure we will need to come up with another way to compute the signature but don't you think that the messages should be signed somehow?

leighmcculloch: Are we overlapping with guarantees the browser provides signing the postMessage callbacks?

JakeUrban: I don't think its the same level of concern as the HTTPS callbacks, because the wallet verified the anchor's identity when it authenticated via SEP-10, which is required for requesting a webview URL.

So I would say the wallet has a higher level of confidence on the origin of the postmessage request because it comes from  the webview, not a random client on the web.

C0x41lch0x41: Yes I agree with you the risk is lower.

That being said when listening for events from `postMessage` all the messages are queued whatever the origin is.
For example with the code below:
```html
<!-- wallet.html -->
<html>
    <h1>Wallet</h1>
    <body>
        <script type="text/Javascript">
            function OpenPopupWindow(url, name, where)
            {
                myRef = window.open(url , name,'left=20,top=20,width=500,height=500,toolbar=1,resizable=0');
                myRef.focus()

            }
            window.addEventListener("message", (event) => {
                console.log(event.data);
            }, false);
        </script>

    <form> 
        <input type=button value="Anchor" onClick="OpenPopupWindow('anchor.html', 'Anchor');"> 
        <input type=button value="Evil" onClick="OpenPopupWindow('evil.html', 'Evil');"> 
    </form> 
    </body>
</html>

<!-- anchor.html -->
<html>
    <h1>Anchor</h1>
    <body>
        <p>ANCHOR</p>
        <script type="text/Javascript">
            window.opener.postMessage("ANCHOR", "*");
        </script>
    </body>
</html>

<!-- evil.html -->
<html>
    <h1>Evil</h1>
    <body>
        <p>EVIL</p>
        <script type="text/Javascript">
            window.opener.postMessage("ANCHOR", "*");
        </script>
    </body>
</html>
```

The wallet will not be able to distinguish calls from the Anchor and calls from Evil.
This can happen if the wallet uses iframe for ads and integrate 3rd party JS/CSS/HTML libraries.

A simpler solution would be to add a requirement for the wallet to check that `event.origin` is set to the anchor host when listening for events.

C0x41lch0x41: @JakeUrban what do you think? Maybe I can specify this does not apply to `postMessage` callbacks to merge this and open another discussion about it? Thanks!

JakeUrban: I'm up for discussing how to handle postmessage callbacks but I agree it would be ideal if we didn't block our solution for HTTP callbacks. Lets merge this and start another thread.

C0x41lch0x41:
@JakeUrban see my update mentioning postMessage callbacks are out of scope for the signature. Thanks!

JakeUrban (approved):

PR #1291, opened 2022-08-11
CAP-46: Updated and formalized the configuration settings proposal.

dmkozh:
The main difference is the proposal to store only the hash of the upgrade set in `SCPValue`. Otherwise, this expands on the details omitted from the original proposal.


dmkozh:
Up to avoid staleness

dmkozh:
Up to avoid staleness

dmkozh:
Updated this to be based on the CAP-46 and separated into a separate 'sub-CAP'. PTAL

sisuresh (approved):

PR #1294, opened 2022-08-12
SEP-31: make destination address & memo optional in POST /transactions response

### LINKED ISSUE #1290: SEP-31: make `POST /transactions` payment-related response attributes optional

SEP-31 provides receiving anchors with two options when a `POST /transactions` request is made to their service.

1. Reject the transaction
2. Accept the transaction & return a Stellar account & memo to use for the payment

However, some receiving anchors may need to accept the transaction without immediately providing a Stellar account & memo to use in the payment. For example, its possible that the receiving anchor needs to perform asynchronous AML or KYC checks on a per-transaction basis before they are willing to accept payents. 

We should make `stellar_account_id`, `stellar_memo`, & `stellar_memo_type` optional in the `POST /transactions` response, and update the flows documented in the protocol to use the transaction `id` returned to do one of the following:

1. Poll `GET /transactions/:id` until the status changes from `pending_receiver` to `pending_sender`. At this point `stellar_account_id`, `stellar_memo`, & `stellar_memo_type` should be populated in the response body.
2. Register a callback URL for changes in the transaction's status via `PUT /transactions/:id/callback`. The receiving anchor would make a callback to the sending anchor when the status changes from `pending_receiver` to `pending_sender`, and the callback payload should contain the `stellar_account_id`, `stellar_memo`, and `stellar_memo_type`

Don't change anything. This is a breaking change for sending anchors, since they likely have an assumption that the stellar account & memo is present in `POST /transactions` responses.

JakeUrban:
resolves #1290

See the issue description for details.

cc @gaclaudiu


JakeUrban:
@lijamie98, can you update the status diagram to show `pending_receiver` as a valid initial status?

ecosystem/sep-0031.md
```diff
@@ -442,14 +441,23 @@ Name | Type | Description
...
 Name | Type | Description
 -----|------|------------
 `id` | string | The persistent identifier to check the status of this payment transaction.
-`stellar_account_id` | string | The Stellar account to send payment to.
-`stellar_memo_type` | string | The type of memo to attach to the Stellar payment (`text`, `hash`, or `id`).
-`stellar_memo` | string | The memo to attach to the Stellar payment.
+`stellar_account_id` | string | (optional) The Stellar account to send payment to.
+`stellar_memo_type` | string | (optional) The type of memo to attach to the Stellar payment (`text`, `hash`, or `id`).
+`stellar_memo` | string | (optional) The memo to attach to the Stellar payment.
+`status` | string | (optional) The initial status of the transaction, either `pending_sender` or `pending_receiver`.
```
marcelosalloum, line 460: Since we're making a breaking change and going to v2.0.0, what do you think of changing the response body of `POST /transactions` to the same response body of `GET /transactions/:id`?

JakeUrban: I think that makes a lot of sense. We’ll have to be intentional about adding more guidance in the SEP and documentation though so it’s clear which attributes are useful immediately after transaction creation.

leighmcculloch, line 460: It's unfortunate that we have four optional parameters that are bound together, i.e. not individually optional, they're only optional together.

A status that isn't present is interesting, would it make sense do you think to add a status to represent this moment? It would mean the state machine always has a status, rather than this unknown status period.

JakeUrban: I added `status` as optional because existing implementations don't return it, not because there isn't a status value that represents the moment. `pending_receiver` would represent the moment if the address & memo fields are not returned, otherwise `pending_sender` would be appropriate.

If we go with @marcelosalloum's suggestion to return the same response body as the `GET /transactions/:id`, then `status` would be required.

So I think we just need to decide how big of a breaking change we want to make. Making previously required response attributes optional is less impactful to existing implementations than making previously absent fields required.

leighmcculloch: Ah, interesting. I wonder if breaking this change into two changes in the changelog will help people understand it clearer. i.e. add status as optional, then make other changes. It doesn't improve the resulting situation though.

JakeUrban: 🤔 I do think there are two breaking changes here

1. Make `stellar_account_id`,  `stellar_memo`, & `stellar_memo_type` optional in `POST /transactions` & `GET /transactions/:id` responses
2. Replace current `POST /transactions` response body with the response body of `GET /transactions/:id`

It just pains me to bump the major version twice. Making 1 breaking change per major version bump makes things clearer, but I see other pieces of software make many breaking changes per major version bump.

leighmcculloch: Ah yeah, we don't want to bump it twice. Breaking changes in software is one thing, in standards it creates more work, as how do clients interoperate with all of them.

JakeUrban: Ok.. then @marcelosalloum I think we should just go with breaking change 1 and not do 2. It would make the API cleaner and more RESTful but its not worth making current implementations out of date.

JakeUrban:
Ok, I removed the `status` attribute from the `POST /transactions` response. This minimizes the changes to the `POST /transactions` response.

ecosystem/sep-0031.md
```diff
@@ -884,6 +893,7 @@ It is important to note that the Receiving Anchor is not obligated, at least by
 
 ## Changelog
 
+* `v2.0.0`: Make `stellar_account_id`, `stellar_memo`, & `stellar_memo_type` optional in the `POST /transactions` response. Also adds the `status_message` attribute to `GET /transactions/:id`. ([#1294](https://github.com/stellar/stellar-protocol/pull/1294))
```
leighmcculloch, line 896:
```suggestion
* `v2.0.0`: Make `stellar_account_id`, `stellar_memo`, & `stellar_memo_type` optional in the `POST /transactions` response. Add the `status_message` attribute to `GET /transactions/:id`. ([#1294](https://github.com/stellar/stellar-protocol/pull/1294))
```

marcelosalloum (commented):
The change LGTM!

There's just a small typo in the state diagram that should be updated from ~`pre-payemnt`~ to `pre-payment`.

marcelosalloum (approved):
LGTM!

PR #1296, opened 2022-08-18
Clarify field format vs type in SEP-9

erika-sdf:


JakeUrban (commented):
This is great! Maybe we can be even more explicit and reference the formats like E164 in the formats column? I’d also be ok removing the format column entirely

erika-sdf:
Sure, I opted for adding the code name.

JakeUrban (approved):

marcelosalloum (approved):

PR #1302, opened 2022-08-29
SEP-12: make it explicit that the `account` and `memo` can be inferred from the JWT token

marcelosalloum:
### What

Make it explicit that the `account` and `memo` can be inferred from the JWT token.

### Why

Those fields were optional but it was not clear that the server should fill them in based on the JWT token.

### DIFF
--- a/ecosystem/sep-0012.md
+++ b/ecosystem/sep-0012.md
@@ -6,7 +6,7 @@ Title: KYC API
 Author: Interstellar
 Status: Active
 Created: 2018-09-11
-Updated: 2022-07-07
+Updated: 2022-08-29
 Version 1.9.1
 ```
 
@@ -667,5 +667,6 @@ All responses should return `200 OK`. If no files are found for the identifer us
 
 ## Changelog
 
+* `v1.10.0`: Clarify that the `account` and `memo` fields should be inferred from the decoded SEP-10 JWT's `sub` value even when not provided in the request body.
 * `v1.9.1`: Callback signature: using expected host instead of HTTP Header to validate signature
 * `v1.9.0`: Added signature requirement for the [`callback`](#customer-callback-put)
--- a/ecosystem/sep-0012.md
+++ b/ecosystem/sep-0012.md
@@ -667,5 +667,5 @@ All responses should return `200 OK`. If no files are found for the identifer us
 
 ## Changelog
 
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0012.md
```diff
@@ -6,7 +6,7 @@ Title: KYC API
 Author: Interstellar
 Status: Active
 Created: 2018-09-11
-Updated: 2022-07-07
+Updated: 2022-08-29
 Version 1.9.1
```
JakeUrban, line 10: Lets bump the patch version to 1.9.2

ecosystem/sep-0012.md
```diff
@@ -98,15 +98,15 @@ GET [KYC_SERVER]/customer?id=<customer-id>&type=<customer-type>
 Name | Type | Description
 -----|------|------------
 `id` | `string` | (optional) The ID of the customer as returned in the response of a previous `PUT` request. If the customer has not been registered, they do not yet have an `id`.
-`account` | `G...` or `M...` string | (**deprecated**, optional) This field should match the `sub` value of the decoded SEP-10 JWT.
+`account` | `G...` or `M...` string | (**deprecated**, optional) This field should match the `sub` value of the decoded SEP-10 JWT. When omitted, the server should use the `sub` value of the decoded SEP-10 JWT.
```
JakeUrban, line 101: Maybe we should replace the existing description entirely, so its clear that omitting the field is the preferred approach. This works though.

JakeUrban (approved):
LGTM, but I suggest rephrasing the `account` field's description.

ecosystem/sep-0012.md
marcelosalloum: Addressed in https://github.com/stellar/stellar-protocol/pull/1302/commits/fd40428111ccd66e77f36ebefcbdaf087d2e9ea8.

ecosystem/sep-0012.md
marcelosalloum: Addressed in https://github.com/stellar/stellar-protocol/pull/1302/commits/fd40428111ccd66e77f36ebefcbdaf087d2e9ea8.

JakeUrban (dismissed):
Awesome, this is much better than what we had 👍

JakeUrban (approved):

PR #1311, opened 2022-10-05
SEP-31: Updated `refunded` status and detailed receiving anchor flow

### LINKED ISSUE #1250: SEP-31: add `refunded` transaction status

The SEP-31 transaction object has a `refunds` sub-object to document payments sent back to the sender. However, it doesn't have a status value that indicates refund payments were made. So the receiving anchor is left with `error`, which isn't a good representation of the state of the transaction.

Add a `refunded` status value to the transaction object.

cc @nirojapanda @reecexlm

lijamie98:
closes #1250

### DIFF
--- a/ecosystem/sep-0031.md
+++ b/ecosystem/sep-0031.md
@@ -895,7 +895,7 @@ It is important to note that the Receiving Anchor is not obligated, at least by 
 
 ## Changelog
 
-* `v2.0.1`: Add the `refund` status and updated the Detailed Receiving Anchor Flow. ([#1294](https://github.com/stellar/stellar-protocol/pull/1294))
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0031.md
```diff
@@ -6,8 +6,8 @@ Title: Cross-Border Payments API
 Author: SDF
 Status: Active
 Created: 2020-04-07
-Updated: 2022-08-12
-Version 2.0.0
+Updated: 2022-10-05
+Version 2.0.1
```
JakeUrban, line 10: This is an additive change, can we update the minor version instead?

marcelosalloum (dismissed):
I've found a small error in the changelog below.

Feel free to merge it after fixing that.

marcelosalloum (approved):
LGTM!

PR #1315, opened 2022-11-02
[CAP-46-02] Updated contract lifecycle CAP.

dmkozh:
*Note*: This includes the configuration update commit that belongs to https://github.com/stellar/stellar-protocol/pull/1291, so the diff from CAP-46-09 commit should be ignored while reviewing this.

The main updates:

- Change `InvokeHostFunctionOp` to use XDR for contract creation args, as it's getting quite complex and doesn't ever need to be called from the contracts.
- Decouple WASM sources from the contract instances in order to deduplicate contracts that share their code.
- Change the contract-from-contract creation to use the code reference instead of the actual code
- Removed some outdated parts that don't allow well with the current implementation/vision

Some debatable decisions:

- This CAP still allows fusing WASM installation with contract instantiation. This could in theory be replaced with 2 invoke host fn operations (install, then create), though building these operations could be a bit tricky as the user would need to compute the code hash themselves.
- Contract initializers are not a part of the CAP. If we allow multiple host fn invocations, initializer call can be added to the list of operations (although the user would need to specify the contract ID, which could be a bit tricky). Unlike in the installation case, initialization is user-defined, so defining the initializer spec at the protocol level could be too rigid, especially considering auth.
- The CAP doesn't allow a contract source to refer another contract. On one hand, it would be easy to implement. On another hand, this doesn't seem too useful without the ability to modify the contract source references.
- I've removed the ability to deploy contracts from contracts with ID derived from `ed25519` key for now. This is not because I necessarily think we should never have those, but because I'd like to make sure it would align with the auth patterns in case if we change them.

### DIFF
--- a/core/cap-0046.md
+++ b/core/cap-0046.md
@@ -124,9 +124,30 @@ That calculates the following difference between the `src/protocol-curr` and
 `src/protocol-next` directories:
 
 ```diff mddiffcheck.ignore=true
+diff -ru '--exclude=*.h' '--exclude=.git*' '--exclude=*.md' src/protocol-curr/xdr/Stellar-contract.x src/protocol-next/xdr/Stellar-contract.x
+--- src/protocol-curr/xdr/Stellar-contract.x	2022-11-01 17:03:53.517489900 -0400
++++ src/protocol-next/xdr/Stellar-contract.x	2022-11-01 19:59:58.500101500 -0400
+@@ -250,14 +250,14 @@
+ 
+ enum SCContractCodeType
+ {
+-    SCCONTRACT_CODE_WASM = 0,
++    SCCONTRACT_CODE_WASM_REF = 0,
+     SCCONTRACT_CODE_TOKEN = 1
+ };
+ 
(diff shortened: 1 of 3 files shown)


core/cap-0046.md
```diff
@@ -451,35 +514,96 @@ diff -ru '--exclude=*.h' '--exclude=.git*' '--exclude=*.md' src/protocol-curr/xd
...
++                void;
++            case CONTRACT_ID_PUBLIC_KEY_ED25519:
++                struct
++                {
++                    uint256 key;
++                    Signature signature;
++                } ed25519KeyWithSignature;
++            } keySource;
++            uint256 salt;
++        } publicKey;
```
leighmcculloch, line 577: Fyi I have this lurking memory that some of the SDKs don't do well with many nested layers of anonymous types, and we might need to make each layer in this XDR type a named type. We can see how it plays out first, but definitely remember experiencing some pain with structures defined anonymously with more than 2 levels deep in the past.

sisuresh (approved):

PR #1320, opened 2022-11-16
SEP-24 & 6: update & add `lang` parameter to on relevant endpoints

JakeUrban:
This is a request from Lobstr. Currently, Polaris and other anchors return 400 when the specified language is not supported by the anchor. But because anchors don't provide a list of supported languages, the wallets have no choice but to attempt the request with the user's preferred language anyway. Therefore, anchors should be forgiving when an unsupported language is passed and default to english.

### DIFF
--- a/ecosystem/sep-0024.md
+++ b/ecosystem/sep-0024.md
@@ -7,7 +7,7 @@ Author: SDF
 Status: Active
 Created: 2019-09-18
 Updated: 2022-11-16
-Version 2.6.3
+Version 2.7.0
 ```
 
 ## Simple Summary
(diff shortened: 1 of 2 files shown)


ecosystem/sep-0024.md
```diff
@@ -230,7 +230,7 @@ Name | Type | Description
...
 `wallet_name` | string | (optional) In communications / pages about the deposit, anchor should display the wallet name to the user to explain where funds are going.
 `wallet_url` | string | (optional) Anchor should link to this when notifying the user that the transaction has completed.
-`lang` | string | (optional) Defaults to `en`. Language code specified using [RFC 4646] which means it can also accept locale in the format `en-US`. `error` fields in the response, as well as the interactive flow UI and any other user-facing strings returned for this transaction should be in this language.
+`lang` | string | (optional) Defaults to `en` if not specified or if the specified language is not supported. Language code specified using [RFC 4646] which means it can also accept locale in the format `en-US`. `error` fields in the response, as well as the interactive flow UI and any other user-facing strings returned for this transaction should be in this language.
```
lijamie98, line 233: I remember there was a conversation to fail and reject the transaction if `lang` is specified and not supported.

Do we want to dictate in the standard?

JakeUrban: I think its better than returning 400 in this case, otherwise wallets will have to resubmit their requests with `en` anyway.

PR #1321, opened 2022-11-16
SEP-31, 24, & 6: add optional `refund_memo` to transaction initiation requests

JakeUrban:
Receiving anchors often use custodial services to hold their on-chain assets, and these custodians often use omnibus accounts, so memos are used by the custodian to map inbound payments to their customer's balances.

To support this case, the standard allows Receiving anchors to provide a memo & memo type to use for each transaction in the `POST /transactions` request.

Sending anchors may also use custodial services to hold their on-chain assets, so in the event that a refund payment is necessary, sending anchors should be allowed to provide the receiving anchor a memo to use.

Thats why I'm adding two optional parameter to the `POST /transactions` request body, allowing the sending anchor to provide the receiving anchor with a memo to use when sending refund payments back.

### DIFF
--- a/ecosystem/sep-0024.md
+++ b/ecosystem/sep-0024.md
@@ -990,7 +990,7 @@ There is a small set of changes when upgrading from SEP-6 to SEP-24.
 
 ## Changelog
 
-* `v2.7.0`: Add `refund_memo` and `refund_memo_type` parameters to withdraw endpoint. ([]())
+* `v2.7.0`: Add `refund_memo` and `refund_memo_type` parameters to withdraw endpoint. ([#1321](https://github.com/stellar/stellar-protocol/pull/1321))
 * `v2.6.3`: Clarify `lang` defaults to `en` when specified value is not supported. ([#1320](https://github.com/stellar/stellar-protocol/pull/1320))
 * `v2.6.2`: Fixed the missing links of customer information needed. ([#1316](https://github.com/stellar/stellar-protocol/pull/1316))
 * `v2.6.0`: Add callback signature and verification requirement. ([#1263](https://github.com/stellar/stellar-protocol/pull/1263))
--- a/ecosystem/sep-0024.md
+++ b/ecosystem/sep-0024.md
@@ -6,7 +6,7 @@ Title: Hosted Deposit and Withdrawal
 Author: SDF
 Status: Active
 Created: 2019-09-18
-Updated: 2022-11-16
+Updated: 2022-11-18
 Version 2.7.0
 ```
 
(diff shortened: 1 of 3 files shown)


JakeUrban:
cc @Nopik

lijamie98 (dismissed):
LGTM after reading the messages from slack.

JakeUrban:
If we agree on this design (cc @marcelosalloum) then I will make the same changes to SEP-31 & SEP-6

ecosystem/sep-0031.md
```diff
@@ -437,6 +437,8 @@ Name | Type | Description
 `receiver_id` | `string` | (optional) The ID included in the [SEP-12 PUT /customer](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0012.md#customer-put) response for the Receiving Client. Required if the Receiving Anchor requires SEP-12 KYC on the Receiving Client.
 `fields` | object | An object containing the values requested by the Receiving Anchor in the `GET /info` endpoint.
 `lang` | string | (optional) Defaults to `en`. Language code specified using [ISO 639-1](https://en.wikipedia.org/wiki/ISO_639-1).  Any human-readable error codes or field descriptions will be returned in this language.
+`refund_memo` | (optional) The memo the Receiving Anchor must use when sending refund payments back to the Sending Anchor. If not specified, the Receiving Anchor should use the same memo the Sending Anchor used to send the original payment.
+`refund_memo_type` | (optional) The type of the `refund_memo`. Can be `id`, `text`, or `hash`. See the [memos](https://developers.stellar.org/docs/encyclopedia/memos) documentation for more information.
```
marcelosalloum, line 441: 🤔  Should anchors use a default value for the memo type? I mean, if the user sends a memo but no memo type, should it be considered a text memo?

marcelosalloum (approved):
LGTM!

Please, don't forget to update the [Changelog section](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0031.md#changelog) at the bottom of the file.

ecosystem/sep-0031.md
JakeUrban: If we do want to standardize a default type, it should probably be `id`. That way the ecosystem could migrate to using muxed account addresses for refunds eventually.

But I don't know if providing a default makes sense. Maybe I should just add a clause to the definitions saying `refund_memo` and `refund_memo_type` must be sent together.

marcelosalloum: > But I don't know if providing a default makes sense. Maybe I should just add a clause to the definitions saying `refund_memo` and `refund_memo_type` must be sent together.

Yeah, that makes sense to me.

ecosystem/sep-0024.md
```diff
@@ -6,7 +6,7 @@ Title: Hosted Deposit and Withdrawal
 Author: SDF
 Status: Active
 Created: 2019-09-18
-Updated: 2022-11-16
+Updated: 2022-11-18
 Version 2.7.0
```
marcelosalloum, line 10: I think this should be 2.8.0, right?

ecosystem/sep-0024.md
```diff
@@ -987,6 +989,8 @@ There is a small set of changes when upgrading from SEP-6 to SEP-24.
 * Solar wallet: https://solarwallet.io
 
 ## Changelog
+
+* `v2.7.0`: Add `refund_memo` and `refund_memo_type` parameters to withdraw endpoint. ([#1321](https://github.com/stellar/stellar-protocol/pull/1321))
```
marcelosalloum, line 993: I think this should be 2.8.0, right?

I believe you forget to add the 2.7.0 changelog in https://github.com/stellar/stellar-protocol/pull/1320

marcelosalloum (approved):
LGTM!

I just added a comment about the versioning of SEP-24.

ecosystem/sep-0024.md
JakeUrban: I think I messed up by actually incrementing the minor version on SEP-24 instead of the patch version. So 2.7 makes sense for this change.

PR #1331, opened 2022-12-27
SEP-24: Change the type of amount of the `/fee` endpoint to `string`

lijamie98:
The `amount_in`, `amount_out` and `amount_fee` are as type `string` in most end-points. We should make them consistent.

### DIFF
--- a/ecosystem/sep-0024.md
+++ b/ecosystem/sep-0024.md
@@ -6,8 +6,8 @@ Title: Hosted Deposit and Withdrawal
 Author: SDF
 Status: Active
 Created: 2019-09-18
-Updated: 2022-12-22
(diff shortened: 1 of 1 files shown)


JakeUrban (changes requested):
This is unfortunately not a minor version change. I would recommend keeping this as-is for now.

Most wallets & anchors do not use this endpoint anyway because it assumes the on & off-chain asset are equivalent (i.e. USD<-->USDC, BRL<-->BRLT, etc.).

We need to introduce a replacement endpoint that supports providing fees for transactions that use nonequivalent on & off-chain assets. When we do that, we can deprecate the original `GET /fee` endpoint.

PR #1333, opened 2023-01-07
SEP 6,12,24,31: replace `X-Stellar-Signature` with `Signature`

### LINKED ISSUE #1328: SEPs (6, 12, 24, 31): Update callback header from `X-Stellar-Signature` to `Signature`

Using the `X-` prefix for custom headers has been [deprecated](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers), and there is an in-progress [RFC draft](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-message-signatures) to standardize signatures to verify the origin & integrity of requests that uses the `Signature` header.

While we won't update our callback signature scheme to match the draft until the RFC is finalized, we can update the header used, since that is unlikely to change.

Update the header used for callbacks in various SEPs from `X-Stellar-Signature` to `Signature`

Keep the `X-Stellar-Signature` header.

JakeUrban:
resolves #1328 

Updates the name of the signature header for callback requests from `X-Stellar-Signature` to `Signature`.

While this is technically a breaking change, the addition of the callback signature mechanism is new enough to not be concerned making existing implementations noncompliant.

### DIFF
--- a/ecosystem/sep-0006.md
+++ b/ecosystem/sep-0006.md
@@ -7,7 +7,7 @@ Author: SDF
 Status: Active (Interactive components are deprecated in favor of SEP-24)
 Created: 2017-10-30
 Updated: 2023-01-06
-Version 4.0.0
+Version 3.17.0
 ```
 
 ## Simple Summary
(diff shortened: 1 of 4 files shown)


leighmcculloch (dismissed):
Would their be value in making this a non-breaking change, by saying that applications should publish both headers until the next breaking change that removes the older X-Stellar-Signature? If no, 👍🏻.

howardtw:
> by saying that applications should publish both headers until the next breaking change that removes the older X-Stellar-Signature

I think it's safer to do so for a smooth transition. Otherwise, the application (wallet backend) would have to check for both `X-Stellar-Signature` and `Signature` headers whenever a partner is making the change.

C0x41lch0x41:
I made a similar comment about backward compatibility [here](https://github.com/stellar/stellar-protocol/issues/1328#issuecomment-1363225415). Otherwise looks good to me.

C0x41lch0x41 (dismissed):

JakeUrban:
Ok, we'll go with the following approach:

1. Anchors send the signature using both `X-Stellar-Signature` & `Signature`
2. Wallets check for signatures using both `X-Stellar-Signature` & `Signature`. Callbacks should be discarded if their values do not match.
3. When the RFC is finalized, we'll remove `X-Stellar-Signature`

leighmcculloch (approved):

C0x41lch0x41:
I was thinking of a less restrictive/complicated approach:
1. Anchors send `X-Stellar-Signature` (old code) or `Signature` (new code)
2. Wallets check for `X-Stellar-Signature` or `Signature`. If both exists, values should be the same.
3. When RFC is finalized, wallets can drop support of `X-Stellar-Signature`

That allows Anchors to not have to update an existing code for now.

JakeUrban:
I think the approach we took allows outdated wallets & anchors still using `X-Stellar-Signature` (and not `Signature`) to remain compatible with wallets & anchors that have implemented the use of both fields.

If we specified that anchors could send `Signature` without `X-Stellar-Signature`, all existing wallets would need to update.

C0x41lch0x41:
> remain compatible with wallets & anchors that have implemented the use of both fields.

I think I am missing something here. Today we only use `X-Stellar-Signature` correct? So wallets and anchors only implemented this field. Using `Signature` is new, that's why I was suggesting an approach where either `X-Stellar-Signature` or `Signature` are valid.

JakeUrban:
This doesn't work for a new anchor (using `Signature`) & old wallet (uses `X-Stellar-Signature`). @howardtw maybe this is acceptable given that Vibrant is the first to implement this?

howardtw:
To work with multiple anchors, before all of them adopt the new `Signature` header, the wallet needs to validate both `X-Stellar-Signature` and `Signature` headers unless they both switch to the new header at the same time, which is unlikely.

JakeUrban:
Thats right, is that acceptable? MoneyGram is unlikely to send both, so Vibrant will likely need to check for both anyway. I agree with @C0x41lch0x41 that allowing anchors to send one header is the simpler approach.

howardtw:
Yes, I can have Vibrant check for both until all the supported anchors migrate to the new header.

PR #1334, opened 2023-01-09
SEP-9: add generalized account field

### LINKED ISSUE #1332: SEP-9: define a generalized account identifier format

Our existing bank-related fields are not sufficient for supporting the variety of off-chain account types that exist globally. Today, we have the following account-related fields:

Name | Type          | Description
-----|---------------|----
`bank_account_number` | string        | Number identifying bank account                                                             
`bank_number` | string        | Number identifying bank in national banking system (routing number in US)                   
`bank_branch_number` | string        |  Number identifying bank branch                                                              
`clabe_number` | string        | Bank account number for Mexico

`clabe_number` was added recently because the existing `bank_*` fields did not properly map to the scheme of Mexico's bank account system as described in #1201.

Now, we have a partner with a similar problem but related to Argentina's banking system and it's use of CBU & CVU numbers and aliases, see more [here](https://www.bcra.gob.ar/MediosPago/Politica_Pagos-i.asp).

Existing partners have attempted to decompose the account identifiers specific to their regions and use our existing fields in order to comply with the standard, but this is not ideal for them.

Ideally, SEP-9 should offer a flexible system for communicating all bank or virtual account identifiers globally. Additionally, this system should allow account identifiers to be represented using the format that is standard to the region in which the business is operating.

I suggest we use the same pattern defined in [SEP-38's Asset Identification Format](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0038.md#asset-identification-format), where a single field is used to communicate assets represented using various formats.

```
<scheme:identifier>
```

Each `scheme` defined in SEP-9 would also define a format to use for its `identifier`. Lets take the Brazilian PIX system for example. The `scheme` name could be `pix`, but a pix key can be one of a [few of values](https://help.remitly.com/s/article/sending-receiving-money-pix?language=en_US#:~:text=Your%20Pix%20key%20can%20be,Your%20mobile%20phone%20number).

- `pix:email:john@example.com`
- `pix:phone:+12345678901` --> phone numbers could be E164 formatted
- `pix:custom:john123`
- `pix:cpf:<a brazilian CPF number (tax ID)>`

I'm not familiar enough with Brazil's PIX, Argentina's CVU, or other non-US account identifiers, so we'll need to rely on our ecosystem to help define the proper formats.

JakeUrban:
resolved #1332 

Adds the `account_id` field (open to naming suggestions) and Account Identifier Format to be used for its values.

### DIFF
--- a/ecosystem/sep-0009.md
+++ b/ecosystem/sep-0009.md
@@ -127,8 +127,8 @@ The currently accepted scheme values are:
 
 Name | Description
 ------------------
-`cbu` | Clave Virtual Uniforme or Unique Virtual Key. The unique key for every user of a payment service provider (PSP) within Argentina.
-`cvu` | Clave Bancaria Uniforme or Unique Bank Key. The unique key for every bank account in Argentina used for receiving deposits, whether the owner is a natural or juridic person.
+`cbu` | Clave Bancaria Uniforme or Unique Bank Key. The unique key for every bank account in Argentina used for receiving deposits, whether the owner is a natural or juridic person.
+`cvu` | Clave Virtual Uniforme or Unique Virtual Key. The unique key for every user of a payment service provider (PSP) within Argentina.
 
 #### `cbu` Identifier Format
 
--- a/ecosystem/sep-0009.md
+++ b/ecosystem/sep-0009.md
@@ -85,7 +85,7 @@ Name | Type          | [Format](#encodings) |Description
 `sex` | string        | | `male`, `female`, or `other`                                                                
 `proof_of_income` | binary        | | Image of user's proof of income document
 `proof_of_liveness` | binary	| | video or image file of user as a liveness proof
-`offchain_account` | string | [Off-Chain Account Identifier Format](#off-chain-account-identifier-format) | An idenfier for the user's off-chain account. Could be a bank, virtual, or other account type.
+`account_id` | string | [Account Identifier Format](#off-chain-account-identifier-format) | An idenfier for the user's off-chain account. Could be a bank, virtual, or other account type.
 
 ### Organization KYC fields
 
@@ -113,7 +113,7 @@ Name | Type | [Format](#encodings) | Description
 
 Address formatting varies widely from country to country and even within each country. See [here](https://stackoverflow.com/questions/11160192/how-to-parse-freeform-street-postal-address-out-of-text-and-into-components) for details. Rather than attempting to create a field for each possible part of an address in every country, this protocol takes a middle of the road approach. Address fields that are fairly universal can be encoded with the `country_code`, `state_or_province`, `city`, and `postal_code` fields. Full addresses, however, should be encoded as a single multi-line string in the `address` field. This allows any address in the world to be represented with a limited number of fields. If address parsing is necessary, parsing will be easier since the country, city, and postal code are already separate fields.
 
-### Off-Chain Account Identifier Format
+### Account Identifier Format
 
 Information used to identify financial accounts varies widely depending on the country and type of account. Rather than adding fields for each type of account identifier, and requiring updates from the various implementations of this specification, accounts can be identified using the following string format:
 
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0009.md
```diff
@@ -112,8 +113,30 @@ Name | Type | [Format](#encodings) | Description
...
+```
+<scheme>:<identifier>
+```
+
+The currently accepted scheme values are:
+
+Name | Description
+------------------
+`cvu` |
+`cbu` |
```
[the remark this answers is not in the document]
JakeUrban: Hi @EpineiroBitso, can you expand on what you mean by "or not within Argentina"? I'm considering omitting this piece as its a bit confusing.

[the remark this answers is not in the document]
EpineiroBitso: Yes, I agree, I think is confusing how I redacted it. For the purposes of the description of the field I don't think is providing much value, I'd omit it also

ecosystem/sep-0009.md
```diff
@@ -85,6 +85,7 @@ Name | Type          | [Format](#encodings) |Description
 `sex` | string        | | `male`, `female`, or `other`                                                                
 `proof_of_income` | binary        | | Image of user's proof of income document
 `proof_of_liveness` | binary	| | video or image file of user as a liveness proof
+`account_id` | string | [Account Identifier Format](#account-identifier-format) | An idenfier for the user's off-chain account. Could be a bank, virtual, or other account type.
```
marcelosalloum, line 88: I fear we may need to keep adding new fields like this. To prevent that we could make this field an array of strings, so the anchor can request any number of fields and if something fails, they can explain what's wrong in the response body.

Also, `account` is too broad as it can be mis-understood as the user Stellar Account. Here's my suggestion:

```suggestion
`offchain_account_ids` | string[] | [Account Identifier Format](#account-identifier-format) | An array of idenfiers for the user's off-chain account. Could contain one or more identifiers of the type bank, virtual, or other(s).
```

JakeUrban: Thanks Marcelo, adding an `offchain` prefix is fine by me.

I share your concern about needing multiple `offchain_account_id` fields for different accounts. This would be the first time we add an array type to the standard though, and it raises questions about how multiple values are encoded in different content types.

JSON is straight-forward:

```js
{
    "offchain_account_ids": ["cvu:X", "cbu:Y"]
}
```

And most web frameworks handle the two variations that exist for URL parameters:

```
?a=1&a=2&a=3
?a=1,2,3
```

But I don't think the same is true for `application/x-www-form-urlencoded`, which SEP-12 requires support for. We could address this in a later PR though.

marcelosalloum: After further thought, I believe it's easier for wallets and for users if we go with a `string` type with comma-separated values. It should make it way easier for wallets to implement that when constructing the form automatically, and also easier for users to fill the form directly, without the wallet having to process the content prior to submitting it to the Anchors.

Here is how it'd look like in [SEP-12 `GET /customer` response](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0012.md#response):

```json
{
   "id": "d1ce2f48-3ff1-495d-9240-7a50d806cfed",
   "status": "NEEDS_INFO",
   "fields": {
      "offchain_account_ids": {
         "description": "The identifiers needed in order to deliver funds to a person's bank account in Argentina. Should be a list of comma-separated values in the format 'cbu:X,cvu:Y', containing the Clave Bancaria Uniforme and Clave Virtual Uniforme, respectively.",
         "type": "string"
      }
   },
   "provided_fields": {
      "last_name": {
         "description": "The customer's last name",
         "type": "string",
         "status": "ACCEPTED"
      }
   }
}
```

JakeUrban: Hey @marcelosalloum I've been thinking about this a bit, and unfortunately this is blocking a partner so we need to come to a decision ideally today.

 > It should make it way easier for wallets to implement that when constructing the form automatically, and also easier for users to fill the form directly, without the wallet having to process the content prior to submitting it to the Anchors.

I doubt wallets will ask users to enter their account IDs as a comma separated list, since its a rather awkward UX, especially if they'd need to enter the schema prefixes themselves (like `cvu:123`). They'll either ask for one value of a particular type (either CVU or CBU) or provide some kind of dynamic UI where a user can add additional values via some (+) button or something similar.

So I don't think using a comma-separated list will simplify wallet processing logic in practice, although I do see what you were going for.

---
 
I also think asking for an array of values has its own issues. The first is that if we add an `array` value to SEP-12/9, how do we specify the type for the array's values? We'd need something like this:

```js
{
    "fields": {
        "offchain_account_id": {
            "description": "Account identifiers. Supported Identifiers include CVU & CBU numbers.",
            "type": "array",
            "items": {
                 "type": "string"
            }
        }
     }
}
```

Thats not a huge deal, but something potentially more impactful is that wallets won't know how to format & validate their users' input for these types of fields unless they know which formats a specific anchor supports ahead of time. If we used something like `cvu_number`, even if the wallet didn't have a relationship with the anchor, it would know how to format / validate the value.

---

So, I'm afraid that we're making a change that adds more complexity than it removes. If you agree, I'll move forward by scratching this proposal and adding `cvu_number` and `cbu_number` SEP-9 fields that can be used in SEP-12 just like all other existing fields. I think we will eventually need to figure out how to support multiple values for the same field, but I no longer believe using generic field names is advantageous.

marcelosalloum: SGTM!

PR #1338, opened 2023-01-26
SEP-9: add `cvu_number` and `cbu_number` fields

### LINKED ISSUE #1332: SEP-9: define a generalized account identifier format

JakeUrban:
resolves #1332 

Adds `cvu_number` and `cbu_number` to SEP-9. We explored adding a generalized financial account field in #1334 and ultimately decided it added more complexity than it removed.

### DIFF
--- a/ecosystem/sep-0009.md
+++ b/ecosystem/sep-0009.md
@@ -6,8 +6,8 @@ Title: Standard KYC Fields
 Author: stellar.org
 Status: Active
 Created: 2018-07-27
-Updated: 2022-11-30
-Version 1.8.0
+Updated: 2023-01-25
+Version 1.9.0
 ```
 
 ## Simple Summary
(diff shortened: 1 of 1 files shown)


reecexlm (approved):
LGTM

PR #1343, opened 2023-02-06
SEP-24: make `account` deposit request parameter optional

### LINKED ISSUE #1342: SEP-24: make `account` for deposit request optional to match withdraw request
The `account` parameter for `POST /transactions/withdraw/interactive` is optional, and the anchor should assume the source account of the payment will be the account authenticated via SEP-10 if `account` is not specified.

The `account` parameter for `POST /transactions/deposit/interactive` should also be optional and default to the authenticated account as well.

JakeUrban:
resolves #1342 

Makes the `account` parameter for the `POST /transactions/deposit/interactive` request optional to match the `POST /transactions/withdraw/interactive` request.

SEP-24 originally did not require SEP-10, and because of that, required that the client specify the destination account of the payment via the deposit request body. When we introduced SEP-10, we should have also updated the SEP to have anchors default to using the authenticated account as the destination.

This change will make the deposit request's `account` parameter definition consistent with the withdraw request's account parameter, which is also optional.

This is technically a breaking change for anchors that assume the `account` field will always be present, although in practice this is easy for wallets to work around.

### DIFF
--- a/ecosystem/sep-0024.md
+++ b/ecosystem/sep-0024.md
@@ -1021,7 +1021,7 @@ There is a small set of changes when upgrading from SEP-6 to SEP-24.
 
 ## Changelog
 
-* `v3.0.0`: Make the `account` parameter for `POST /transactions/deposit/interactive` request optional ([]())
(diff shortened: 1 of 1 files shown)


lijamie98 (approved):
LGTM

PR #1344, opened 2023-02-10
SEP-9: add `bank_account_type`

JakeUrban:
Adds `bank_account_type` to the list of SEP-9 fields, with initial supported values of `checking` or `savings`. Some jurisdictions have banking systems that require that the account type be specified in addition to the account number.

Note that `bank_account_type` does not indicate the format of `bank_account_number`, and should not be used to validate `bank_account_number`. 

If an anchor needs to be able to validate an account identifier, it should use other SEP-9 fields added specifically for the format they're looking to validate, such as `cbu_number` or `clabe_number`. These fields often contain pieces of information other than the user's account number with their bank. 

For example, CBU numbers contain identifiers for the bank, the bank branch, the type of account, and the account number itself. Instead of trying to decompose CBU numbers and passing each piece via the `bank_account_number`, `bank_number`, `bank_branch_number`, and other SEP-9 fields, it is much easier for businesses operating in those jurisdictions to accept account identifiers in the formats that are common in those jurisdictions.

### DIFF
--- a/ecosystem/sep-0009.md
+++ b/ecosystem/sep-0009.md
@@ -117,7 +117,7 @@ Address formatting varies widely from country to country and even within each co
 
 ## Changelog
 
-* `v1.11.0`: Add `bank_account_type` for describing types of bank accounts. ([]())
+* `v1.11.0`: Add `bank_account_type` for describing types of bank accounts. ([#1344](https://github.com/stellar/stellar-protocol/pull/1344))
 * `v1.10.0`: Remove `cvu_number`, update `cbu_number` to also accept CVU numbers, and add `cbu_alias` to Natural Person KYC fields ([#1339](https://github.com/stellar/stellar-protocol/pull/1339))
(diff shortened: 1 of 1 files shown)


JakeUrban:
@theclabs this is relevant to the [comment](https://github.com/stellar/stellar-protocol/pull/1339#issuecomment-1407704445) you made on #1339.

cc @lucaszanotelli

reecexlm (approved):
LGTM

lijamie98:
@JakeUrban Do you consider having the composite bank number that contains both the `bank_account_type` and the `bank_account_number` such as `bank_account_number` is `checking:123-456-7890`?

JakeUrban:
We've entertained the idea of defining our own formats for composite fields in SEP-9 [here](https://github.com/stellar/stellar-protocol/pull/1334), but ultimately decided against it. We do have some composite fields though, like `clabe_number` and `cbu_number`, the difference is that their formats have been defined by the governments of those jurisdictions.

Missjones2829 (approved):
Merge

PR #1345, opened 2023-02-13
Update documentation

jopmiddelkamp:


Ifropc:
Hi @jopmiddelkamp thanks for raising this, but I will close this PR, as description in protocol is correct.
Couple things to note:
1. Transaction expiration is not a mandatory 24-hour period, and is decided by the Anchor.
2. `more_info_url` is a mandatory field, and must always be set, regardless of transaction status. 

The Anchor you have been working with may have not provided this field after transaction expiration date, but in correct implementation it should always be there.

jopmiddelkamp:
Ok then the Anchor is not following he protocol indeed. Thank you for your response.

PR #1347, opened 2023-02-25
Fee CAP updates

MonsieurNicolas:
Integrates various decisions from the mailing list

### DIFF
--- a/core/cap-0046-07.md
+++ b/core/cap-0046-07.md
@@ -78,10 +78,9 @@ struct ContractLedgerCostNetworkSettingsV0
     int64 minFeeWrite1KB; // minimum fee for writing 1KB
 
     int64 bucketListSizeBytes; // bucket list fees grow slowly up to that size
-    int64 bucketListBufferSizeBytes; // buffer size in bytes past bucketListSizeBytes
     int64 bucketListFeeRateLow; // fee rate in stroops when the bucket list is empty
     int64 bucketListFeeRateHigh; // fee rate in stroops when the bucket list reached bucketListSizeBytes
-    int64 bucketListGrowthFactor; // used to determine the fee rate when the entire buffer is used up
+    int64 bucketListGrowthFactor; // rate multiplier for any additional data passed the first bucketListSizeBytes
 };
 
 // historical data settings
@@ -211,13 +210,13 @@ With
 wfee_rate(s) = (bucketListFeeRateHigh - bucketListFeeRateLow)*s/bucketListSizeBytes +
 bucketListFeeRateLow +
 (if s > bucketListSizeBytes,
-     exp(bucketListGrowthFactor*(s-bucketListSizeBytes)/bucketListBufferSizeBytes),
-     0)
+    bucketListGrowthFactor*
+    (bucketListFeeRateHigh - bucketListFeeRateLow)*
+    (s-bucketListSizeBytes)/bucketListSizeBytes,
+    0)
 ```
 
 __Open:__ `wfee_rate` (and possibly `min_LedgerData_fee`) needs to be reconciled with "rent" (not finalized at the time of this writing) as "paying rent" is very similar to adding an entry to the ledger. The difference is that adding an entry to the ledger adds the entry to the "topmost" bucket, where as paying rent is adding an entry to a bucket while merging.
-
-__Open:__ use something easier to compute than "exp" (both for core, and for clients that can overshoot a bit).
 
 Validity constraints:
 * per transaction
@@ -375,6 +374,7 @@ As a consequence the final formula looks like this:
 With
 `fee_rate(s) = (feeRateM - minFeeRate)*s/M_base + minFeeRate + (if s > M_base, exp(K*(s-M_base)/B_buffer), 0)`
 
+We can simplify this even further by replacing the exponential component by a steep linear slope that causes fees to be "extremely high" at `M_buffer`.
 
 ##### Putting it together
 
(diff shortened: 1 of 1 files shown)


core/cap-0046-07.md
```diff
@@ -396,13 +425,15 @@ With
...
 
 As a consequence, we can just price transactions independently, based on the bucket list from the last closed ledger, and allocation within a transaction (for multiple bytes) is a flat rate as well.
 
-Flooding transactions in that context is straightforward as the only factor to take into account is bucket list size (determined based on the last closed ledger).
+Flooding transactions for the next ledger in that context is straightforward as the only factor to take into account is bucket list size (determined based on the last closed ledger).
+
+Note that transactions can sill be invalidated more than a ledger in the future, as a consequence validators may apply a certain level of "padding" when computing the minimum fee required before flooding those transactions.
```
dmkozh, line 436: s/can sill/can still

core/cap-0046-07.md
```diff
@@ -506,13 +483,14 @@ The model retained in this CAP is a flat rate per byte model for simplicity. It
...
+A minimum is also defined as spam countermeasure.
 
-In order to support payloads that get discarded by validators, bandwidth needs to be tracked separately.
+Note that validators may apply additional market dynamics due to implementation constraints, especially when trying to balance propagating large transactions vs smaller ones. See [CAP-0042: Multi-Part Transaction Sets](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0042.md).
 
-Transaction contains an optional `payload` (Hash + size), and the `TransactionEnvelope` gets the actual payload, that gets cleared before applying transactions.
+##### Ephemeral payload
 
-The model retained in this CAP is a flat rate per byte model for simplicity. It is expected that this fee would be orders of magnitude smaller than what is needed to persist data on chain.
+Transaction may contain a `ephemeralPayload` (Hash + size), that gets cleared before applying transactions (used in the context of proofs of availability).
```
dmkozh, line 493: s/a `ephemeral/an `ephemeral

core/cap-0046-07.md
```diff
@@ -506,13 +483,14 @@ The model retained in this CAP is a flat rate per byte model for simplicity. It
...
 
-Fees to propagate transactions as they are now can just be included in the "Historical storage" cost (that already charges for the cost of `TransactionEnvelope`).
+At the networking layer, transactions compete for bandwidth on a per ledger basis (`ledgerMaxPropagateSizeBytes`).
+A minimum is also defined as spam countermeasure.
 
-In order to support payloads that get discarded by validators, bandwidth needs to be tracked separately.
+Note that validators may apply additional market dynamics due to implementation constraints, especially when trying to balance propagating large transactions vs smaller ones. See [CAP-0042: Multi-Part Transaction Sets](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0042.md).
 
-Transaction contains an optional `payload` (Hash + size), and the `TransactionEnvelope` gets the actual payload, that gets cleared before applying transactions.
+##### Ephemeral payload
```
dmkozh, line 491: I'm not sure what this means and what are the proofs of availability. FWIW any payload that only exists for the lifespan of a function invocation doesn't need to be declared explicitly, as it would just be paid for via the regular budget metering mechanisms.

core/cap-0046-07.md
```diff
@@ -165,14 +176,13 @@ At the end of the transaction execution, refund the source account for resources
 
 Right now the following resources fall under this category:
   * historical data
-  * extended data
-  * network data
+  * extended meta data
 
 Total fee is equal to:
-total_fee =  gasFeeBid + ledgerDataAccessFeeBid + flatResourcesFee;
+total_fee =  gasFeeBid + ledgerDataAccessFeeBid + propagateFeeBid + flatResourcesFee;
```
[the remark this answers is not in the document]
MonsieurNicolas: Added the definition (bandwidth)

core/cap-0046-07.md
```diff
@@ -226,50 +244,60 @@ Validity constraints:
...
-`min_ExtendedData_fee(tx) = round_up(tx.extendedDataSize * minFeeExtendedData1KB / 1024)`
+`tx.flatResourcesFee` must be greater than
+`ledgerDataAccess_flat_fee(tx) + historical_flat_fee(txEnvelope) + extendedMetaData_flat_fee(tx)`
 
-Validity constraints:
-* `tx.extendedDataSize <= txMaxExtendedDataSizeBytes`
+## "Fee bump" and failing transactions
 
-#### Network data
+__Open:__
```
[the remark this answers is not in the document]
MonsieurNicolas: Let's not try to resolve this here. The section is basically there to capture the topic of "what about fee bump?" that we will have to go back to when we have a more fleshed out solution.  It is true that this is well beyond fees, fees just make it a lot likelier that a pre-signed transaction is going to fail at execution time and we need to decide if we care about that or not (or if we just tell people to follow a different pattern for pre-signed payloads).

core/cap-0046-07.md
MonsieurNicolas: We need to talk about this outside of this PR. I originally put it there so that we have the right extension points and concepts when we design fees. We do not need to ship this in v1. TLDR; is that contracts have access to `ephemeralPayload` but the actual binary blob that corresponds to that payload is not stored anywhere by default, only propagated. This allows to implement efficient rollup protocols. I added a link to an article about.

core/cap-0046-07.md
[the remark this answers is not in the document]
dmkozh: Well, we can leave this as is for now... Speaking of fee bumps, we probably should mention whether the classic (pre-apply) fee bumps should apply to Soroban txs (I think, why not).

core/cap-0046-07.md
dmkozh: Sounds good, it just seemed weird to include that without any context.

dmkozh (approved):

PR #1349, opened 2023-03-04
SEP-9: added bank_name

yuriescl:
Some Anchors require the Bank name for KYC compliance and sending bank transfers.

### DIFF
--- a/ecosystem/sep-0009.md
+++ b/ecosystem/sep-0009.md
@@ -63,6 +63,7 @@ Name | Type | [Format](#encodings) | Description
 `birth_country_code` | string | ISO 3166-1 alpha-3 country code | ISO Code of country of birth
 `bank_account_number` | string | | Number identifying bank account
 `bank_account_type` | string | | `checking` or `savings`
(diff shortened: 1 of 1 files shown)


yuriescl:
@JakeUrban Will need this for an Anchor implementation that requires bank name. Otherwise Polaris won't let me return `bank_name` from SEP-12 /customer

JakeUrban:
Thanks Yuri! Is the bank's name required to initiate transfers to/from these banks? Typically banks are identified via numerical identifiers like the US' routing number or the first few digits in Argentina's CBU numbers. Which banking system(s) require the bank name?

antb123:
Hi Jake - its part of the ISO 20022 standard and certain banks may require it.

In ISO its: CdtrAgt.FinInstnId.Nm
we could simply use bank_name or something from here
https://schema.org/BankOrCreditUnion

Thanks!

AbdulKiram:
HTTP//: QUEENBANK 

`)

AbdulKiram:
- QUEENBANK apps// Url / http/ "

PR #1357, opened 2023-05-29
SEP-40: Asset enum instead of Address

yuriescl:
This is a proposal for changes in [SEP-40](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0040.md) (which is in `Draft` status) from [BP Ventures](https://bpventures.us/).

Using `Address` to represent assets restricts the contract to only provide prices for assets that have a respective `Address`, this means the contract can only provide prices for assets that are issued inside the Stellar blockchain.
There's no need for such restriction.  
Instead of using `Address`, we can use a `enum`:
```
#[contracttype]
enum Asset {
   Stellar(Address),
   Other(Symbol),
}
```

This would allow us to store not just prices of Stellar assets, but also prices of fiat currencies like `USD`, as well as any other non-standard currency like Bitcoin or mutual funds tokens.
- Stellar assets could be stored as `Asset::Stellar`
- Other off-chain currencies could be stored as `Asset::Other`, e.g. `EUR`, `BTC`

```
/// Get price in base asset at specific timestamp
fn price(
    env: soroban_sdk::Env,
    asset: Asset,    // <- Asset instead of Address
    timestamp: u64
) -> Option<PriceData>;
```
For consistency, the base asset should also be of type `Asset`, this way we have `Asset` as a global type for any asset, including base.

EDIT 1 (2023-06-02): Removed suggestion to use `u128` for `price`, [as pointed out](https://github.com/stellar/stellar-protocol/pull/1357#issuecomment-1573985282) by @MonsieurNicolas 
EDIT 2 (2023-06-02): Replaces `Bytes` with `Asset` enum, [as suggested](https://github.com/stellar/stellar-protocol/pull/1357#discussion_r1214617725) by @dmkozh
EDIT 3 (2023-06-05): Simplified `Asset` struct [as suggested](https://github.com/stellar/stellar-protocol/pull/1357#issuecomment-1576716850) by @mootz12


MonsieurNicolas:
note: I would not recommend moving to `uint128` as we've standardized on `int128` in Soroban's standard token interface

ecosystem/sep-0040.md
```diff
@@ -61,12 +61,12 @@ pub trait PriceFeedTrait {
     /// Return the base asset the price is reported in
     fn base(
         env: soroban_sdk::Env,
-    ) -> Address;
+    ) -> Bytes;
```
dmkozh, line 64: I would recommend using descriptive UDTs instead of Bytes and custom encodings.

For example, if your goal is to represent on-chain assets as well as off-chain entities, I would recommend to use the following enum to represent assets:

```rust
[contracttype]
enum Asset {
   Contract(Address),
   Generic(Bytes/String)
}
```

This will provide much greater readability and reduce the risk of introducing bugs.

yuriescl: I agree

yuriescl:
> note: I would not recommend moving to `uint128` as we've standardized on `int128` in Soroban's standard token interface

That makes sense. The only reason I see to use `uint128` would be to allow higher prices, but I'm not sure if it's necessary, maybe `int128` already covers a sufficient limit.

ecosystem/sep-0040.md
yuriescl: Does Soroban support `String` type? If yes, then I think we can use `String`, otherwise `Bytes` seems to be best option to avoid storing unnecessary information.

yuriescl: > Does Soroban support String type?

Yes: https://docs.rs/soroban-sdk/latest/soroban_sdk/struct.String.html

dmkozh: > We also have to consider the performance/compute cost of using a enum instead of raw Bytes. 

It's actually likely cheaper to use UDT. Don't forget, that you need to do custom decoding of your bytes input, which is more or less what you'd end up doing with enums (but for the Soroban UDTs you're guaranteed to be using the host-based decoding, which should be fast).

mootz12:
Good catch on `Address` not being applicable to off-chain assets. We do have the option of standardizing on random `Address` values for off-chain assets, but this is much clearer.

The naming of the `enum` is a bit confusing for me. `Contract` can be ambiguous between Stellar Asset's and Soroban token contracts, and the separation of `ISO4217` and `Generic` seems too fine grained. Consider:

```rust
[contracttype]
enum Asset {
   Stellar(Address),
   Other(String) // open to a better name here
}
```

While I see a benefit for allowing the base asset to be an off-chain asset, I'm not totally convinced there is value in being able to fetch prices for assets that are not represented on the Stellar blockchain (IE - have an `Address`). Is there something you had in mind when making this change?

cc: @orbitlens

orbitlens:
Agree with @mootz12 here. 
For developers, it might be problematic to use oracles price feed inside Soroban for assets, that are not mapped to corresponding Soroban assets. In the ideal world, it would be nice to add the requirement for all anchored or "generic" assets to have corresponding Stellar assets wrapped as Soroban tokens. That way, the price feed can be utilized everywhere (for Soroban and Classic Stellar). And providers could attach any metadata to the quoted asset itself to avoid confusion. 

However, I do agree that under some circumstances it may be beneficial to have `enum Asset{Stellar(Address),Generic(String)}` as @mootz12 proposed. But wee need to enforce generic string length restriction (let's say, no more than 32 bytes), as this directly impacts storage and interface calls (we may hit max request size limitations).

yuriescl:
> While I see a benefit for allowing the base asset to be an off-chain asset, I'm not totally convinced there is value in being able to fetch prices for assets that are not represented on the Stellar blockchain (IE - have an Address). Is there something you had in mind when making this change?

The reason to have the base asset with type `Asset` would be just for consistency, that is, to have a single type to represent any asset in the contract.

> The naming of the enum is a bit confusing for me. Contract can be ambiguous between Stellar Asset's and Soroban token contracts, and the separation of ISO4217 and Generic seems too fine grained

I agree, `Stellar` and `Other` are sufficient.

> In the ideal world, it would be nice to add the requirement for all anchored or "generic" assets to have corresponding Stellar assets wrapped as Soroban tokens

I don't think we need to require this, as it would create a unnecessary restriction. By allowing fiat (off-chain) prices in the contract, we maximize flexibility and allow contract users to fetch fiat prices directly from the Oracle instead of recurring to external sources. For example, there could be contracts that really need to get prices from special currencies that don't really have issuers in Stellar (and issuing them could be very complicated depending on regulation), and if those contracts could fetch directly from an Oracle, they can put all their fiat logic inside the contract, making the contract more powerful and more open source, otherwise they'd need external (non-Soroban) scripts running outside the blockchain to get those special prices and feed them into the contract.

> But wee need to enforce generic string length restriction (let's say, no more than 32 bytes), as this directly impacts storage and interface calls (we may hit max request size limitations)

I agree. Maybe we use `Symbol` instead of `String`, since `Symbol` already has this 32 bytes restriction.

orbitlens:
@yuriescl 
> EDIT 4 (2023-06-10): Replaced timestamp with start_timestamp and end_timestamp in price() function. Renamed price() to prices(). Renamed prices() to lastprices(). Changed prices return types to `Vec<PriceData>` for simplicity.

This edit and two last commits are out of scope of the original pull request which was focused only on the `Asset` enum. It seems to me that we already reached a consensus on the proposed `Asset` struct (up to commit [992dcf9](https://github.com/stellar/stellar-protocol/pull/1357/commits/992dcf94630f5288afdf9142aac0669004a52fd5)). This probably can be merged to the SEP if we won't see any other objections.

Could you please move everything related to "Edit 4" to a separate pull-request?

yuriescl:
> Could you please move everything related to "Edit 4" to a separate pull-request?

Done

mootz12 (approved):
@orbitlens @dmkozh We should look to get this update merged into the SEP-40 draft.

dmkozh:
> @orbitlens @dmkozh We should look to get this update merged into the SEP-40 draft.

I really just left some feedback regarding the data structures and from this perspective the current design looks sane for me. I can't quite talk from the business logic requirements perspective, so leaving this up to folks who came up with this. So LGTM for me on the data structures.

dmkozh (approved):

yuriescl:
Updated the branch with changes from `master`. I imagine it's ready to merge @mootz12 @dmkozh

nguemechieu:
Hi team,
 I 'm using your service and this one of the error I have found while trying to convert some or claim some new token and I think this need to be fixed in order  to prevent scam.
1- Fist ,any token providers who website extension ending by .claim are mostly scammer.
2- If a use is claiming a particular token to that hold a certain amount in value such as 500$ in a claimable pending transaction ,on the transaction is completed that value will disappear.
Then if the user is try to remove  that token on untrust the line , he will get and error prompt 
such as "asset can't be remove when balance is 0"
3- we need some filters on assets  and those filter should be able notified the user about a particular asset states with warning such (probably scam, scam ,good ,bad for example ).This are critical for users awareness and safety.

PR #1358, opened 2023-06-08
Add SEP-38 support to SEP-24

Ifropc:
### Abstract 
Currently, there is no way for a wallet to get a quote from the anchor in a standardized way. Anchors may have in-house endpoints for the wallet, however, wallet would need to implement a new set of API for each anchor it integrates with.
What's more, current fee endpoint is not flexible enough, and doesn't offer complex functionality SEP-38 fee (part of the quote object body) has.
Having a quote and fees before proceeding to the interactive screen is very important for the user experience, as without it, user would not be able to know this information before initiating a transaction. 

### Proposal
This proposal is aimed to solve this problem. We add integration of the existing SEP-38 endpoint with the SEP-24. The wallet may now use the existing SEP-38 endpoints implemented by the anchor to query a quote.
Later, wallet may use the quote id (for the firm quote) to exchange the asset on agreed rates.

### Changelog summary
Add support for the SEP-38 API. Add optional `quote_id` parameter to `post /transactions/deposit/interactive`
  and `POST /transactions/deposit/interactive` requests. Add `quote_id` to the transaction object schema.

### Use cases
1. Wallet can query an indicative quote from the anchor
2. Anchor can support more complex fee logic via SEP-38 `/price` and `/prices` endpoints
3. Wallet can request a firm quote and initiate a transaction with this quote. 

### Backward compatibility 
This proposal is fully backward compatible


ecosystem/sep-0024.md
```diff
@@ -73,6 +74,54 @@ To support the exchange of non-equivalent assets using this protocol, the anchor
 
 Finally, anchors should populate the optional attributes listed in the `GET /transaction(s)` endpoints related to non-equivalent asset transactions, specifically `amount_in_asset`, `amount_fee_asset`, and `amount_out_asset`. This allows client applications can relay this information, in addition to the amounts specified in the associated attributes, back to their users.
 
+Alternatively, the Anchor must support `quote_id` in the `POST /transactions/deposit/interactive` and `POST /transactions/withdraw/interactive`, and the [SEP-38] Anchor RFQ API that enables the Anchor to provide quotes to the Wallet.
```
JakeUrban, line 77: Given that this is the first time this concept is introduced, I think we can provide some more detail.

```suggestion
Alternatively, the Anchor can support the [SEP-38] RFQ API and the `quote_id` attribute in `POST /transactions/deposit/interactive` and `POST /transactions/withdraw/interactive` requests. This enables wallets to check the Anchor's exchange rates prior to initiating a transaction and obtain a firm quote to use for an upcoming transaction.
```

ecosystem/sep-0024.md
```diff
@@ -651,6 +708,8 @@ The default values for the features listed above have been selected based on the
 
 ## Fee
 
+### This endpoint is deprecated. [SEP-38] quotes API should be used to communicate fees instead.
```
JakeUrban, line 711:
```suggestion
### This endpoint is deprecated. The [SEP-38] `GET /price` endpoint should be used to fetch fees instead.
```

ecosystem/sep-0024.md
```diff
@@ -73,6 +74,54 @@ To support the exchange of non-equivalent assets using this protocol, the anchor
 
 Finally, anchors should populate the optional attributes listed in the `GET /transaction(s)` endpoints related to non-equivalent asset transactions, specifically `amount_in_asset`, `amount_fee_asset`, and `amount_out_asset`. This allows client applications can relay this information, in addition to the amounts specified in the associated attributes, back to their users.
 
+Alternatively, the Anchor must support `quote_id` in the `POST /transactions/deposit/interactive` and `POST /transactions/withdraw/interactive`, and the [SEP-38] Anchor RFQ API that enables the Anchor to provide quotes to the Wallet. 
+
+### Asset Conversions Using the Market Rate
+
+To convert at the market rate, clients get an [indicative quote](https://www.investopedia.com/terms/i/indicativequote.asp) with [SEP-38] and then proceed to calling the [`GET /deposit-exchange`](#deposit-exchange) or [`GET /withdraw-exchange`](#withdraw-exchange) endpoints. Market rate conversions executed in this way do not provide a guarantee of the final price to be paid for the asset. Here is how this kind of conversion should happen in deposits and withdrawals for an asset pair like `BRL <> USDC`:
```
JakeUrban, line 81: I know this is copy-pasted, but we don't have the `-exchange` endpoints here.

ecosystem/sep-0024.md
```diff
@@ -73,6 +74,54 @@ To support the exchange of non-equivalent assets using this protocol, the anchor
...
+
+#### Deposit:
+
+Considering the user will be making a bank deposit of fiat `BRL` in exchange for `USDC` in the Stellar network:
+
+1. The Wallet checks the assets' market conversion rate with [`SEP-38 GET /price`](sep-0038.md#get-price) and shows the estimated price to the user.
+1. The Wallet initiates a [`POST /transactions/deposit/interactive`](#deposit) using the parameter `asset_code`.
```
JakeUrban, line 88: One thing that SEP-6 & SEP-31 both have is a way to specify both the on-chain & off-chain assets when initiating a transaction using an indicative quote. This would improve the SEP-24 UX too because the user wouldn't have to select the off-chain asset in the UI, it can be pre-selected.

Can we explore adding an optional parameter for specifying an off-chain asset? SEP-6 & SEP-31 used `source_asset` & `destination_asset` but that might be tricky here given that we already have `asset_code` & `asset_issuer`.

ecosystem/sep-0024.md
```diff
@@ -73,6 +74,54 @@ To support the exchange of non-equivalent assets using this protocol, the anchor
...
+Considering the user will be making a Stellar payment of `USDC` in exchange for withdrawing `BRL` in the bank account:
+
+1. The Wallet checks the assets' market conversion rate with [`SEP-38 GET /price`](sep-0038.md#get-price) and shows the estimated price to the user.
+1. The Wallet initiates a [`POST /transactions/withdraw/interactive`](#withdraw) using the parameter `asset_code`.
+1. The user proceeds with the on-chain payment, sending `USDC` to the Anchor's Stellar account.
+1. Upon receiving the on-chain withdraw funds, the Anchor converts it to `BRL` using the market rate and sends the resulting amount to the user's bank account.
+
+### Asset Conversions Using a Firm Quote
+
+With firm quotes, users are guaranteed an asset price they will pay if they submit the payment before the quote expiration. Here is how the conversion should happen in deposits and withdrawals for an asset pair like `USDC <> BRL`:
```
JakeUrban, line 103: nit: the quote is honored only if the payment was _received_ by the anchor before the quote's expiration. For off-chain payments this is important since they can take some time.

ecosystem/sep-0024.md
```diff
@@ -73,6 +74,54 @@ To support the exchange of non-equivalent assets using this protocol, the anchor
...
+1. The user proceeds with the off-chain deposit, sending `BRL` to the Anchor's bank account before the quote expires.
+1. Upon receiving the off-chain deposit funds, the Anchor converts it to `USDC` according with the agreed quote and sends the resulting amount to the user's Stellar account.
+1. If the user sends the funds after the quote expiration, the Anchor can proceed according with its own policy.
+
+#### Withdraw:
+
+Considering the user will be making a Stellar payment of `USDC` in exchange for withdrawing `BRL` in her bank account:
+
+1. The Wallet gets a firm conversion rate in [`SEP-38 POST /quote`](sep-0038.md#post-quote) and shows the price and expiration date to the user.
+1. The Wallet initiates a [`GET /withdraw-exchange`](#withdraw-exchange) using the parameter `quote_id` returned from the previous step, as well as `asset_code` and `destination_asset`.
```
JakeUrban, line 120: There is no `destination_asset` parameter in SEP-24's endpoint, and no `-exchange` endpoint.

ecosystem/sep-0024.md
```diff
@@ -207,6 +258,8 @@ SEP-24 lays out many options for how deposit and withdrawal can work. These are
 
 A deposit is when a user sends some non-stellar asset (BTC via Bitcoin network, USD via bank transfer, Cash to a teller, etc...) to an account held by an anchor. In turn, the anchor sends an equal amount of tokens on the Stellar network (minus fees) to the user's Stellar account.
 
+If the anchor supports SEP-38 quotes, it can also provide a bridge between non-equivalent tokens. For example, the anchor can receive ARS via bank transfer and in return send the equivalent value (minus fees) as USDC on the Stellar network to the user's Stellar account. That
```
JakeUrban, line 261:
```suggestion
If the anchor supports SEP-38 quotes, it can also provide a bridge between non-equivalent tokens. For example, the anchor can receive ARS via bank transfer and in return send the equivalent value (minus fees) as USDC on the Stellar network to the user's Stellar account.
```

ecosystem/sep-0024.md
```diff
@@ -227,7 +280,8 @@ Name | Type | Description
 -----|------|------------
 `asset_code` | string | The code of the stellar asset the user wants to receive for their deposit with the anchor. The value passed must match one of the codes listed in the [/info](#info) response's deposit object.
```
JakeUrban, line 281: Lets add that this must match the `buy_asset` from the quote, if `quote_id` is specified. Same for the issuer. Don't forget to also update the withdraw endpoint in the same way.

ecosystem/sep-0024.md
```diff
@@ -742,6 +801,7 @@ Name | Type | Description
 `amount_out_asset` |  string | (optional) The asset delivered or to be delivered to the user. Must be present if the deposit/withdraw was made using non-equivalent assets. The value must be in [SEP-38 Asset Identification Format](sep-0038.md#asset-identification-format). See the [Asset Exchanges](#asset-exchanges) section for more information.
 `amount_fee` | string | Amount of fee charged by anchor.
 `amount_fee_asset` | string | (optional) The asset in which fees are calculated in. Must be present if the deposit/withdraw was made using non-equivalent assets. The value must be in [SEP-38 Asset Identification Format](sep-0038.md#asset-identification-format). See the [Asset Exchanges](#asset-exchanges) section for more information.
+`quote_id` | string | (optional) The ID of the quote used to create this transaction. Should be present if a `quote_id` was included in the `POST /transactions` request. Clients should be aware though that the `quote_id` may not be present in older implementations.
```
JakeUrban, line 804:
```suggestion
`quote_id` | string | (optional) The ID of the quote used when creating this transaction. Should be present if a `quote_id` was included in the `POST /transactions/deposit/interactive` or `POST /transactions/withdraw/interactive` request. Clients should be aware that the `quote_id` may not be present in older implementations.
```

JakeUrban (commented):
This is a good start!

I don't see any changes to `GET /info` though, we wanted to make changes there right?

I think the only other change requested is to explore adding a parameter for pre-populating the off-chain asset, although its not strictly necessary since we have a UI.

ecosystem/sep-0024.md
```diff
@@ -818,7 +906,9 @@ The response should be a JSON object like:
   },
   "features": {
     "account_creation": true,
-    "claimable_balances": true
+    "claimable_balances": true,
+    "indicative_quotes_supported": true,
+    "firm_quotes_supported": true
```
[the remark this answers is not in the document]
Ifropc: > Wallets can determine if the 1st approach is possible by looking at the toml, so I don't think the flag in GET /info is necessary for that.

I agree that they can simply look at the toml file. IMO, it's a nice addition to communicate to the wallets what features anchor support

> I'm not sure if that information is useful

It's mainly to tell if `POST /quotes` endpoint is supported or not. AFAIK, all current anchors support both endpoints, so it's an extra addition to make it more flexible and understandable for the wallet.

That being said, I don't feel strongly about adding this flags. We can remove them, if majority or reviewers think it's not necessary.

ecosystem/sep-0024.md
```diff
@@ -107,6 +109,78 @@ non-equivalent asset transactions, specifically `amount_in_asset`, `amount_fee_a
...
+To convert at the market rate, clients get an
+[indicative quote](https://www.investopedia.com/terms/i/indicativequote.asp) with [SEP-38] and then proceed to calling
+the [`POST /transactions/withdraw/interactive`](#withdraw) or [`POST /transactions/deposit/interactive`](#deposit)
+endpoints. Market rate conversions executed in this way do not provide a guarantee of the final price to be paid for the
+asset. Here is how this kind of conversion should happen in deposits and withdrawals for an asset pair like
+`BRL <> USDC`:
+
+Anchors offer indicative quote support via enabling `indicative_quotes_supported` feature flag.
+
+#### Deposit:
```
lijamie98, line 128: I am not sure if this is an anti-pattern of the interactive UI flow. This would make sense in SEP-6. But for SEP-24, why is the reason not to let the anchor finalize the quote in the interactive UI?

ecosystem/sep-0024.md
```diff
@@ -214,12 +288,12 @@ banking rails so counterparties can run through the flow without a need to colla
...
-  description of the asset from the `stellar.toml` file.
-- **Use the `/info` endpoint**
-  - Fetch the asset's deposit & withdawal fee structure: if `fee_fixed`, `fee_percent`, or `fee_minimum` are provided,
-    show this to the user early in the process so they're fully informed.
-  - If the `/fee` endpoint is enabled, use it for computing fees when you need to show them to the user.
+  display the asset's quote (if possible) as well as information such as the address of the anchor and description of
+  the asset from the `stellar.toml` file.
+- **If [SEP-38] quote API is supported by the anchor**
+  - Fetch the asset's exchange rate at the market price using [`SEP-38 GET /price`](sep-0038.md#get-price) API.
+  - For the firm quote, use [`SEP-38 POST /quote`](sep-0038.md#post-quote) API and show the quote and expiration time to
```
lijamie98, line 295: I am not sure of handling the firm quote before the UI flow starts. I agree getting the indicative quote may be helpful for the wallet. However if some of the SEP-24 anchors are handling the exchange in the UI flow, wouldn't this cause divergence in how the exchanges are handled?

ecosystem/sep-0024.md
```diff
@@ -44,13 +44,15 @@ the API endpoints according to the protocol will work with any compliant wallet.
...
+- If the Anchor supports [SEP-38], `ANCHOR_QUOTE_SERVER` must be defined in their [`stellar.toml`](sep-0001.md).
 
 ## API Endpoints
 
 - [`POST /transactions/deposit/interactive`](#deposit)
 - [`POST /transactions/withdraw/interactive`](#withdraw)
 - [`GET /info`](#info)
-- [`GET /fee`](#fee): optional (only needed for complex fee structures)
+- [`GET /fee`](#fee): **deprecated** optional (only needed for complex fee structures).
+  [`SEP-38 GET /price`](sep-0038.md#get-price) endpoint should be used to communicate the fees.
```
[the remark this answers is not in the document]
JakeUrban: There is a [`GET /price`](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0038.md#get-price) endpoint

ecosystem/sep-0024.md
Ifropc: Anchors internally have a firm quote. The idea is moving firm quote from the interactive UI into the wallet UI.

ecosystem/sep-0024.md
Ifropc: They can finalize it in interactive UI, however wallet may want to fetch the firm quote and agree with anchor to use that particular quote.
FYI: in this protocol suggestion this functionality is optional. Anchor may only support indicate qoutes

ecosystem/sep-0024.md
```diff
@@ -322,22 +402,24 @@ The fields below should be placed in the request body using the `multipart/form-
...
+| `sending_asset`               | string in [Asset Identification Format] | (optional) The asset user wants to send. Note, that this is the asset user initially holds (off-chain or fiat asset).If this is not provided, it will be collected in the interactive flow. When `quote_id` is specified, this parameter is mandatory and must match the quote's `setll_asset` asset code.                                                                                                                       |
+| `quote_id`                    | string                                  | (optional) The `id` returned from a `SEP-38 POST /quote` response. If this parameter is provided and the user delivers the deposit funds to the Anchor before the quote expiration, the Anchor must respect the conversion rate agreed in that quote. If the values of `asset` and `amount` conflict with the ones used to create the [SEP-38] quote, this request should be rejected with a `400`.                              |
```
CassioMG, line 410: could we also include in the `quote_id` description that `sending_asset` must be provided when using `quote_id`? 
I'm asking it because in all the other paragraphs above we only mention `quote_id` & `asset_code` params so mentioning `sending_asset` here would make this requirement more explicit

also, what is `asset` referring to in the `quote_id` description? would it be `asset_code`, `asset_issuer` or `sending_asset`?

ecosystem/sep-0024.md
```diff
@@ -322,22 +402,24 @@ The fields below should be placed in the request body using the `multipart/form-
...
+| `sending_asset`               | string in [Asset Identification Format] | (optional) The asset user wants to send. Note, that this is the asset user initially holds (off-chain or fiat asset).If this is not provided, it will be collected in the interactive flow. When `quote_id` is specified, this parameter is mandatory and must match the quote's `sell_asset` asset code.                                                                                                                                                    |
+| `quote_id`                    | string                                  | (optional) The `id` returned from a `SEP-38 POST /quote` response. If this parameter is provided and the user delivers the deposit funds to the Anchor before the quote expiration, the Anchor must respect the conversion rate agreed in that quote. When `quote_id` is set, `asset_code`, `sending_asset` and `amount` are mandatory. In case of a conflict with the ones used to create the [SEP-38] quote, this request should be rejected with a `400`. |
```
CassioMG, line 410: @Ifropc thanks for the description update.

Is `amount` really mandatory here?
On the description of the `amount` field it says it should be omitted if a `quote_id` was used so both descriptions are conflicting now.

ecosystem/sep-0024.md
```diff
@@ -516,19 +602,21 @@ The fields below should be placed in the request body using the `multipart/form-
...
+| `destination_asset` | string in [Asset Identification Format] | (optional) The asset user wants to receive. It's an off-chain or fiat asset. If this is not provided, it will be collected in the interactive flow. When `quote_id` is specified, this parameter is mandatory and must match the quote's `buy_asset` asset code.                                                                                                                                                                                                 |
+| `quote_id`          | string                                  | (optional) The `id` returned from a `SEP-38 POST /quote` response. If this parameter is provided and the user delivers the deposit funds to the Anchor before the quote expiration, the Anchor must respect the conversion rate agreed in that quote. When `quote_id` is set, `asset_code`, `destination_asset` and `amount` are mandatory. In case of a conflict with the ones used to create the [SEP-38] quote, this request should be rejected with a `400`. |
```
CassioMG, line 610: same question about `amount` being required here

ecosystem/sep-0024.md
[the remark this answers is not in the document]
CassioMG: I don't have a strong opinion here either. I can think that this could help a bit to automate feature enabling/disabling by Anchors (e.g. an Anchor that is connected to several wallets but needs to turn quote off only to a specific wallet without having to change the TOML file), but I'm leaning towards agreeing that it's not a necessary change, I have a feeling those flags would be rarely used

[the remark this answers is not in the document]
Ifropc: That's a valid concern. I think I'll remove them in the final version of the PR.

ecosystem/sep-0024.md
Ifropc: No, I got everything mixed up. Will fix description.
amount, assets fields are not required, but if they are set, they must match the quote.

ecosystem/sep-0024.md
```diff
@@ -516,19 +602,21 @@ The fields below should be placed in the request body using the `multipart/form-
...
+| `amount`            | number                                  | (optional) Amount of asset requested to withdraw. If this is not provided it will be collected in the interactive flow. When `qoute_id` is specified, this parameter must match the quote's `quote.sell_amount` or be omitted.                                                                                                                                                                                                                                                                 |
+| `quote_id`          | string                                  | (optional) The `id` returned from a `SEP-38 POST /quote` response. If this parameter is provided and the user delivers the deposit funds to the Anchor before the quote expiration, the Anchor must respect the conversion rate agreed in that quote. When `quote_id` is set, `asset_code`, `destination_asset` and `amount` must be validated by the anchor, if present. In case of a conflict with the ones used to create the [SEP-38] quote, this request should be rejected with a `400`. |
```
CassioMG, line 611: @Ifropc I think you updated the `quote_id` description here on withdrawal by removing the "mandatory" part but did not make the same update for deposit

ecosystem/sep-0024.md
CassioMG: Looks good, thanks for fixing those!

I left another comment related to the `quote_id` description (looks like only withdrawal as updated)

CassioMG (approved):
Lgtm

ecosystem/sep-0024.md
```diff
@@ -322,22 +402,24 @@ The fields below should be placed in the request body using the `multipart/form-
...
+| `asset_issuer`                | string                                  | (optional) The issuer of the stellar asset the user wants to receive for their deposit with the anchor. If asset_issuer is not provided, the anchor should use the asset issued by themselves as described in their TOML file. If `native` is specified as the `asset_code`, `asset_issuer` must be not be set. When `quote_id` is specified, this parameter must match the quote's `buy_asset` asset code or be omitted.                                                                      |
+| `sending_asset`               | string in [Asset Identification Format] | (optional) The asset user wants to send. Note, that this is the asset user initially holds (off-chain or fiat asset).If this is not provided, it will be collected in the interactive flow. When `quote_id` is specified, this parameter must match the quote's `sell_asset` asset code or be omitted.                                                                                                                                                                                         |
```
JakeUrban, line 409: Can we use `source_asset` instead here?

ecosystem/sep-0024.md
```diff
@@ -322,22 +402,24 @@ The fields below should be placed in the request body using the `multipart/form-
...
+| `amount`                      | number                                  | (optional) Amount of asset requested to deposit. If this is not provided it will be collected in the interactive flow. When `qoute_id` is specified, this parameter must match the quote's `quote.sell_amount` or be omitted.                                                                                                                                                                                                                                                                  |
+| `quote_id`                    | string                                  | (optional) The `id` returned from a `SEP-38 POST /quote` response. If this parameter is provided and the user delivers the deposit funds to the Anchor before the quote expiration, the Anchor must respect the conversion rate agreed in that quote. When `quote_id` is set, `asset_code`, `destination_asset` and `amount` must be validated by the anchor, if present. In case of a conflict with the ones used to create the [SEP-38] quote, this request should be rejected with a `400`. |
```
JakeUrban, line 411:
```suggestion
| `quote_id`                    | string                                  | (optional) The `id` returned from a `SEP-38 POST /quote` response. If this parameter is provided and the user delivers the deposit funds to the Anchor before the quote expiration, the Anchor must respect the conversion rate agreed in that quote. When `quote_id` is set, `asset_code`, `source_asset` and `amount` must be validated by the anchor, if present. In case of a conflict with the ones used to create the [SEP-38] quote, this request should be rejected with a `400`. |
```

ecosystem/sep-0024.md
```diff
@@ -107,6 +109,78 @@ non-equivalent asset transactions, specifically `amount_in_asset`, `amount_fee_a
...
+1. Upon receiving the on-chain withdraw funds, the Anchor converts it to `BRL` using the market rate and sends the
+   resulting amount to the user's bank account.
+
+### Asset Conversions Using a Firm Quote
+
+With firm quotes, users are guaranteed an asset price they will pay if they submit the payment and anchor receives the
+payment before the quote expiration. Here is how the conversion should happen in deposits and withdrawals for an asset
+pair like `USDC <> BRL`:
+
+Anchors enable firm quotes support with `firm_quotes_supported` feature flag.
```
JakeUrban, line 156: If we remove `indicative_quotes_supported`, we can update this to `quotes_supported`.

ecosystem/sep-0024.md
```diff
@@ -322,22 +402,24 @@ The fields below should be placed in the request body using the `multipart/form-
...
+| `wallet_name`                 | string                                  | (optional) In communications / pages about the deposit, anchor should display the wallet name to the user to explain where funds are going.                                                                                                                                                                                                                                                                                                                                                    |
+| `wallet_url`                  | string                                  | (optional) Anchor should link to this when notifying the user that the transaction has completed.                                                                                                                                                                                                                                                                                                                                                                                              |
```
JakeUrban, line 416: We deprecated these parameters so we may want to merge master

ecosystem/sep-0024.md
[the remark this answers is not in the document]
JakeUrban: Coming back to this, I think I agree with cassio that these flags would not be used in practice. I think it is a valid assumption that if `ANCHOR_QUOTE_SERVER` is present, firm quotes are supported in SEP-24 transactions.

JakeUrban (approved):
Approved with some comments

ecosystem/sep-0024.md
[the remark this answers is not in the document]
Ifropc: What if anchor supports both SEP-31 and SEP-24, but doesn't support SEP-38 for SEP-24? Do we want to support such case or leave it to the anchor? (They can throw an error if context set to SEP-24)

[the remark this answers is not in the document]
JakeUrban: I feel fairly confident making the assumption that the anchor would want to enable the use of SEP-38 quotes in all the services that support it.

JakeUrban (dismissed):

JakeUrban (approved):

PR #1362, opened 2023-06-12
Replace price() timestamp with ranges

yuriescl:
Since prices can be set at any arbitrary timestamp, I don't think it's very useful to require a `timestamp` parameter in the `price()` function, because this would force the caller to either predict which timestamps have prices, or loop through timestamps until a price is found. This is not efficient.
Instead, `price()` should take a range of timestamps and return a list of prices contained in that range, ordered by oldest to newest in the `Vec`.
`price()` should be renamed to `prices()` since it returns a `Vec`.
The current `prices()` should be renamed to `lastprices()` since it returns the last N prices, and also to not conflict with the other function being renamed.

This would be the new `price()` parameter structure:
```
/// Get prices (in base asset) contained in the provided timestamp range, ordered by oldest to newest.
/// Both timestamps are inclusive, meaning if there's a price exactly at
/// `start_timestamp`, it will be included. Same for `end_timestamp`.
fn prices(
    env: soroban_sdk::Env,
    asset: Asset,
    start_timestamp: u64,
    end_timestamp: u64
) -> Vec<PriceData>;
```

Current `prices()` renamed to `lastprices()`:
```
/// Get last N price records, ordered by oldest to newest
fn lastprices(
    env: soroban_sdk::Env,
    asset: Asset,
    records: u32
) -> Vec<PriceData>;
```

Return type of both `prices()` and `lastprices()` changed to `Vec<PriceData>`. There's no need for `Option<Vec<PriceData>>`, since an empty `Vec` already means "no prices found".

### DIFF
--- a/ecosystem/sep-0040.md
+++ b/ecosystem/sep-0040.md
@@ -102,7 +102,7 @@ pub trait PriceFeedTrait {
 
 Assets have type `Bytes`, this allows representing both fiat and Stellar assets in the following formats:
 - `iso:4217:<currency code>` for fiat currencies, e.g. `iso4217:USD` for US Dollar, `iso4217:EUR` for Euro, etc
-- `<Address>` for Stellar assets, where `Address` is the asset unique address in Soroban network, obtained when an asset from Stellar Classic is deployed to Soroban via the [Stellar Asset Contract](https://soroban.stellar.org/docs/how-to-guides/stellar-asset-contract). E.g. ``d93f5c7bb0ebc4a9c8f727c5cebc4e41194d38257e1d0d910356b43bfc528813`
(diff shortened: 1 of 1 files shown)


mootz12:
@yuriescl The SEP originally intended to use the round based data feed system seen in solutions like Chainlink. However, we ultimately landed on doing uniform update periods. That being said, the current implementation would have known timestamp gaps so contracts could find exact prices.

There is some context as to why here: https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0040.md#timestamps-instead-of-rounds

I'm not strongly attached to uniform update periods over rounds. However, the benefit of having strong temporary storage guarantees and TWAP capabilities do seem like an improvement. Was there a specific use case this change was targeting?

cc: @orbitlens

yuriescl:
After internal discussion, our team now sees the advantages of having uniform update periods. Since we want to be compliant with how existing well-known Oracles work, it makes sense to keep the same standard of timestamps. Thanks for the reply and explanation.

PR #1363, opened 2023-06-14
Add native asset support to SEP-24

philipliu:
### Abstract
Today, only issued assets are supported in SEP-24 transactions. The native asset XLM cannot be specified using the current `asset_code`, `asset_issuer` pair as XLM does not have an issuer. 

### Proposal
This proposes `native` be treated as a special `asset_code`, that when specified in combination with a null `asset_issuer`, allows wallets to deposit and withdraw XLM. Although this differs from the SEP-38 asset identification format, this reduces the friction for existing anchors to onboard to support this new feature.

### Backwards compatibility
This change is fully compatible.


JakeUrban (commented):
This LGTM, I think its just missing a note in the section for `GET /info` on being able to use `native` as the asset code for XLM.

ecosystem/sep-0024.md
```diff
@@ -324,8 +324,8 @@ Request Parameters:
...
-| `asset_issuer`                | string                  | (optional) The issuer of the stellar asset the user wants to receive for their deposit with the anchor. If asset_issuer is not provided, the anchor should use the asset issued by themselves as described in their TOML file.                                                                                                                                                                                                   |
+| `asset_code`                  | string                  | The code of the stellar asset the user wants to receive for their deposit with the anchor. The value passed must match one of the codes listed in the [/info](#info) response's deposit object. `native` is a special `asset_code` that represents the native XLM token.                                                                                                                                                                                                                   |
```
Ifropc, line 327: Can you make sure the table is formatted? 
```
npx prettier  --write  ecosystem/sep-0024.md 
```

Ifropc (approved):

ecosystem/sep-0024.md
philipliu: Oops. Thanks!

Ifropc (approved):

ecosystem/sep-0024.md
```diff
@@ -324,8 +324,8 @@ Request Parameters:
...
+| `asset_code`                  | string                  | The code of the stellar asset the user wants to receive for their deposit with the anchor. The value passed must match one of the codes listed in the [/info](#info) response's deposit object. `native` is a special `asset_code` that represents the native XLM token.                                                                                                                                                         |
+| `asset_issuer`                | string                  | (optional) The issuer of the stellar asset the user wants to receive for their deposit with the anchor. If asset_issuer is not provided, the anchor should use the asset issued by themselves as described in their TOML file. If `native` is specified as the `asset_code`, `asset_issuer` must be not be set.                                                                                                                  |
```
leighmcculloch, line 328: > If `native` is specified as the `asset_code`, `asset_issuer` must be not be set.

`native` is a valid asset code on the network, so I don't think SEP-24 should prevent an issued asset with the name `native` from being used.

I think we would need to allow the case where `asset_code` is `"native"` and `asset_issuer` has a value, and in that case treat the asset as credit / issued. That would make this change truly backwards compatible.

To make sure we're talking about all options, there are two ways we've preferred to represent assets as strings:

1. Three elements:
    - `asset_type` with valid values being `native`,`credit_alphanum4`,`credit_alphanum12`
    - `asset_code`
    - `asset_issuer`

2. One string element, based on [SEP-11](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0011.md#asset-trustlineasset), where:
    - The native asset is rendered as `"native"`
    - An issued asset is rendered as `"ABC:GJKJASD..."`

I understand the latter format is unrealistic given the existing API.

The first form may be reasonable with some subset of values such as `native` and `credit` given the SEP-24 API already places code and issuer into separate fields.

cc @JakeUrban

ecosystem/sep-0024.md
```diff
@@ -1276,6 +1285,7 @@ There is a small set of changes when upgrading from SEP-6 to SEP-24.
 
 ## Changelog
 
+- `v3.1.0`: Add XLM support to SEP-24 ([#1363](https://github.com/stellar/stellar-protocol/pull/1363))
```
leighmcculloch, line 1288: I suggest removing the term XLM here and replacing it with native. We typically refer to the native asset simple as so, and not the specific pubnet network's name for its native asset.

```suggestion
- `v3.1.0`: Add native asset support to SEP-24 ([#1363](https://github.com/stellar/stellar-protocol/pull/1363))
```

leighmcculloch (commented):
Thoughts inline.

ecosystem/sep-0024.md
```diff
@@ -768,6 +768,7 @@ wallet. For example:
 ## Info
 
 Allows an anchor to communicate basic info about what their `TRANSFER_SERVER_SEP0024` supports to wallets and clients.
+If an anchor supports XLM, the `native` asset code should be included in the response.
```
leighmcculloch, line 771: I understand it is useful to include XLM here to make sure folks know what we're talking about when we say native. It might be worth tweaking this text so that the focus is on the native asset, and then we have a note educating what the native asset is called on pubnet. Suggest something like:
```suggestion
If an anchor supports the `native` asset, the `native` asset code should be included in the response.

The `native` asset on the Stellar Public Network is the [lumen (XLM)](lumen).

[lumen]: https://stellar.org/lumens
```

ecosystem/sep-0024.md
philipliu: Fixed!

ecosystem/sep-0024.md
leighmcculloch: Ah, this edge case exists because the `asset_issuer` field is optional?

philipliu: Yes, it's because `asset_issuer` is optional.

philipliu: I see two ways of supporting an issued `native` token.

1. Always require `asset_issuer` to be specified. I can't comment on if any wallets are leaving `asset_issuer` as null today, but as far as breaking changes go, I don't think it's too difficult to migrate over.
2. Use SEP-11/SEP-38 asset identification format by default, but allow clients to use `asset_code` and `asset_issuer` with the goal of eventually deprecating the `asset_code` and `asset_issuer` pair.

Of these two options, I prefer 2. It lets us move to using SEP-38/11 format everywhere.

leighmcculloch: Both options assume that it is a good thing to make the issuer required in the API. I'm wondering what the context is on how it came to be optional. @JakeUrban do you have any context?

JakeUrban: Unfortunately the `asset_issuer` field was optional from the start (at least of my time at the SDF).

The good news is that while there may be an asset with code `native` that exists on the network, the only time this becomes problematic is if an anchor wants to support both the `native` non-XLM asset _and_ XLM itself. In this corner case, the anchor would have to require the `asset_issuer` field in order to differentiate.

I think this situation is acceptable in order to avoid adding something like `asset_type` to the list parameters.

Longer-term, I think we do probably need to introduce an `asset` parameter that can be used as a unique identifier for all on-chain assets and deprecate the existing set of parameters (cc @philipliu), but I wouldn't want to make that parameter required in order to anchor XLM or other assets (such as AMM shares?), so we should find a backwards-compatible solution in the meantime.

leighmcculloch: @JakeUrban For true native (rather than credit native) should the SEP require `asset_issuer: null`?

JakeUrban: Yea I think thats appropriate, although I wouldn't be surprised if custom anchor implementations skip. Still good to include it in the specification though.

PR #1364, opened 2023-06-15
[PARTNER-71] Deprecate wallet_name

Ifropc:
### Abstract
Today, wallets can optionally provide name and url in the /interactive URLs. However, this information should be already known to the anchor, as the anchor knows the SEP-10 JWT auth token. 

### Proposal
`wallet_name` and `wallet_url` parameters become deprecated. Instead, anchors could use one of:
* `client_domain` for non-custodial wallets
* `sub` value of the auth JWT token for custodial wallets 

### Backwards compatibility
This change is fully compatible.


ecosystem/sep-0024.md
```diff
@@ -330,8 +330,8 @@ Request Parameters:
...
-| `wallet_url`                  | string                  | (optional) Anchor should link to this when notifying the user that the transaction has completed.                                                                                                                                                                                                                                                                                                                                |
+| `wallet_name`                 | string                  | (**deprecated**, optional) In communications / pages about the deposit, anchor should display the wallet name to the user to explain where funds are going. However, anchors should use client_domain (for custodial) and `sub` value of JWT (for non-custodial) to determine wallet information.                                                                                                                                |
```
leighmcculloch, line 333: How does an anchor determine that the client is custodial or not? The text seems to imply they should choose client domain or sub based on whether the client is custodial.

Is the intent that the anchor use client domain if it is present, and otherwise fall back to sub? If so, I suggest this change:

```suggestion
| `wallet_name`                 | string                  | (**deprecated**, optional) In communications / pages about the deposit, anchor should display the wallet name to the user to explain where funds are going. However, anchors should use client_domain if present, otherwise `sub` value of JWT to determine wallet information.                                                                                                                                |
```

ecosystem/sep-0024.md
```diff
@@ -289,7 +289,7 @@ banking rails so counterparties can run through the flow without a need to colla
     amount is ready for pick-up.
   - Some wallets might exchange currencies only once they're ready to send the withdrawal payment, so there might be
     slight fluctuations of value between the informed withdrawal `amount` and the actual transferred amount. It is
-    recommended for anchors to accept an amount fluctuation of up to ±10%, and adjust the amount to be transferred (and
+    recommended for anchors to accept an amount fluctuation of up to ±10%, and adjust the amount to be transferred ( and
```
leighmcculloch, line 292:
```suggestion
    recommended for anchors to accept an amount fluctuation of up to ±10%, and adjust the amount to be transferred (and
```

leighmcculloch (commented):
Seems reasonable. One small fix. One suggested change.

Thanks for updating the changelog and meta.

Defer to @JakeUrban.

ecosystem/sep-0024.md
JakeUrban: I think the better approach is actually the other way around, where the anchor maintains a set of addresses that are known to be owned by custodial services, and if the account present in the `sub` value doesn't match any of those, it should be treated as noncustodial.

I actually think this change isn't accurate. If the wallet is custodial, it won't specify a client domain. This applies to the change made to the `wallet_url` description too.

Ifropc: Do you think I should add it somewhere as a commentary for anchor? We do have it in the stellar-docs, but I think for Anchor not using the Platform, it could be a valuable information

ecosystem/sep-0024.md
Ifropc: It's prettier auto-format...

ecosystem/sep-0024.md
Ifropc: Updated description

ecosystem/sep-0024.md
leighmcculloch: This is English, not code, we don't normally put spaces after paranethesis. It's a space, can we backspace it?

Ifropc: For some reason I had a space added when I was auto-formatting with IDE but it doesn't happen anymore after running cli formatter. Weird.

ecosystem/sep-0024.md
JakeUrban: About this line:

> However, anchors should use client_domain (for custodial) and `sub` value of JWT (for non-custodial) to determine wallet information. 

I don't think this is accurate right? Custodial wallets won't be passing `client_domain` in the SEP-10 flow and it won't be in the JWT, that is for non-custodial wallets.

Ifropc: Yes I accidentally reversed it

CassioMG (approved):
Lgtm

JakeUrban (approved):

PR #1367, opened 2023-06-30
SEP-9: Defined shared financial account fields

JakeUrban:
The primary use of SEP-9 fields has been for collecting information on users of Stellar-based applications, including both KYC data and off-chain financial account data.

However, SEP-9 was also intended to be used by organizations to provide information to their users. This is particularly relevant in the context of SEP-6, where organizations need to provide off-chain financial account information to their customers in order for customers to fund their deposit transactions.

This PR clarifies the set of fields that can be used by applications such as anchors to both request that information from customers and provide that information to customers.

Unrelatedly, it also makes formatting changes for improved readability.

### DIFF
--- a/ecosystem/sep-0009.md
+++ b/ecosystem/sep-0009.md
@@ -108,25 +108,25 @@ be used for both natural persons and organizations. For these shared fields that
 
 ## Organization Fields
 
-| Name                                   | Type   | Description                                                                      |
-| -------------------------------------- | ------ | -------------------------------------------------------------------------------- |
-| `organization.name`                    | string | Full organization name as on the incorporation papers                            |
-| `organization.VAT_number`              | string | Organization VAT number                                                          |
-| `organization.registration_number`     | string | Organization registration number                                                 |
-| `organization.registration_date`       | string | Date the organization was registered                                             |
-| `organization.registered_address`      | string | Organization registered address                                                  |
-| `organization.number_of_shareholders`  | number | Organization shareholder number                                                  |
-| `organization.shareholder_name`        | string | Can be an organization or a person                                               |
-| `organization.photo_incorporation_doc` | binary | Image of incorporation documents                                                 |
-| `organization.photo_proof_address`     | binary | Image of a utility bill, bank statement with the organization's name and address |
-| `organization.address_country_code`    | string | country code for current address                                                 |
-| `organization.state_or_province`       | string | name of state/province/region/prefecture                                         |
-| `organization.city`                    | string | name of city/town                                                                |
-| `organization.postal_code`             | string | Postal or other code identifying organization's locale                           |
-| `organization.director_name`           | string | Organization registered managing director                                        |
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0009.md
```diff
@@ -6,120 +6,151 @@ Title: Standard KYC Fields
...
+| `photo_id_front`              | binary | Image of front of user's photo ID or passport                                               |
+| `photo_id_back`               | binary | Image of back of user's photo ID or passport                                                |
+| `notary_approval_of_photo_id` | binary | Image of notary's approval of photo ID or passport                                          |
+| `ip_address`                  | string | IP address of customer's computer                                                           |
+| `photo_proof_residence`       | binary | Image of a utility bill, bank statement or similar with the user's name and address         |
+| `sex`                         | string | `male`, `female`, or `other`                                                                |
+| `proof_of_income`             | binary | Image of user's proof of income document                                                    |
+| `proof_of_liveness`           | binary | video or image file of user as a liveness proof                                             |
+
+## Shared Fields
```
Ifropc, line 93: As organization fields are not widely used, maybe we should rename this header to "Financial Information Fields" or something like this?
And add in description that they could also be used for organizations too?
(I.e. we should emphasize that this fields are primarily used for financial information of a person)

ecosystem/sep-0009.md
```diff
@@ -6,120 +6,151 @@ Title: Standard KYC Fields
...
+| `organization.website`                 | string | Organization website                                                             |
+| `organization.email`                   | string | Organization contact email                                                       |
+| `organization.phone`                   | string | Organization contact phone                                                       |
+
+## Addresses
+
+Address formatting varies widely from country to country and even within each country. See
+[here](https://stackoverflow.com/questions/11160192/how-to-parse-freeform-street-postal-address-out-of-text-and-into-components)
+for details. Rather than attempting to create a field for each possible part of an address in every country, this
+protocol takes a middle of the road approach. Address fields that are fairly universal can be encoded with the
```
Ifropc, line 136: What if address is universal, but we also need to include street name + building number?
What do we place in `address? Is it only building address, or full address?

philipliu: I think building/apt/suite numbers are typically separate optional fields in address forms. Maybe we should also have a separate field so we don't have this ambiguity?

JakeUrban: This section hasn't been changed in this PR, so I suggest we keep things as-is for now.

Ifropc: Yes, but do we have any recommendations/guidance for it?

JakeUrban: I don't think we do at the moment

Ifropc: Got it. I think we should add some guidance for this field because right now it's quite confusing

Ifropc (dismissed):

Ifropc (approved):

Issue #1372, opened 2023-07-25
SEP-6: Providing deposit instructions asynchronously


philipliu:
### What problem does your feature solve?

Today, Anchors are expected to return deposit instructions (In the `how` field but may change in the [future](https://github.com/stellar/stellar-protocol/issues/1368)) in the response of the `GET /deposit` endpoint.

This approach does not work if the Anchor is not able to return the deposit instructions immediately. For example, if the Anchor needs to manually validate a user's KYC information, it should not return deposit instructions until the
validation is complete.

Additionally, the KYC fields required for a deposit may different for each deposit type as well as the arguments passed to the
deposit such as amount, the user's country, etc. The standard currently assumes that the KYC fields are the same for all deposits with the same type. This is not always the case. For example, a deposit of 100 USD may require a different set of KYC fields than a deposit of 10,000 USD.

### What would you like to see?

SEP-6 should provide a way for wallets to request deposit instructions without requiring the Anchor to return them immediately from the `GET /deposit` endpoint.

One way to do this is to immediately put the SEP-6 transaction created by the deposit request into a `pending_anchor` state with the deposit instructions omitted from the deposit response. Once the KYC information has been validated, the Anchor can transition the transaction to the `pending_user_transfer_start` state and include the deposit instructions in the `GET /transactions` response.

If the Anchor determines that more KYC information needs to be provided for the deposit, it can transition the transaction to the `pending_customer_info_needed` state and include the missing KYC fields in the `GET /transactions` response so that the wallet can prompt the user to provide the missing information using SEP-12.

### What alternatives are there?

The alternative is to require Anchors to return deposit instructions immediately from the `GET /deposit` endpoint and leverage the [403 status code](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0006.md#customer-information-needed-non-interactive) to indicate that either KYC information is missing or the deposit is not ready to be processed.

Issue #1368, opened 2023-06-30
SEP-6: standardize structured off-chain deposit instructions for users


JakeUrban:
### What problem does your feature solve?

Today, SEP-6 uses the `how` response attribute defined for the `GET /deposit` endpoint, which is a short but human-readable string used to communicate off-chain payment instructions to the user. This is an insufficient solution for reasons discussed in the [stellar-dev google group](https://groups.google.com/g/stellar-dev/c/BTIx2rrHaFs/m/78iViUchAwAJ).

### What would you like to see?

We should make a change to the SEP-6 protocol that enables wallets to communicate the anchor's deposit instructions to users _that doesn't rely on human-readable content provided by the anchor_. This likely involves defining a structured object in the response that communicates standardized off-chain financial account data. #1367 aims to clarify which SEP-9 fields could be useful in this context.

### What alternatives are there?

Members of the community as well as developers of ecosystem wallet and anchors are welcome to propose solutions. Please read the google group thread so you have full context.

PR #1379, opened 2023-08-14
SEP-6: Add support for asynchronous deposit instructions

philipliu:
### Proposal
This adds new fields to the `transaction` object to facilitate Anchors providing deposit instructions outside of `GET /deposit` response.

### Backwards Compatability
This change is backward compatible as it does not introduce new required fields if Anchors can provide deposit instructions in the `GET /deposit` response.

Resolves https://github.com/stellar/stellar-protocol/issues/1372, https://github.com/stellar/stellar-protocol/issues/1368.


ecosystem/sep-0006.md
```diff
@@ -1121,6 +1121,8 @@ Name | Type | Description
 `refunds` | object | (optional) An object describing any on or off-chain refund associated with this transaction. The schema for this object is defined in the [Refunds Object Schema](#refunds-object-schema) section below.
 `required_info_message` | string | (optional) A human-readable message indicating any errors that require updated information from the user.
 `required_info_updates` | object | (optional) A set of fields that require update from the user described in the same format as [/info](#info).  This field is only relevant when `status` is `pending_transaction_info_update`.
+`customer_info_type` | string | (optional) The [SEP-12 type](sep-0012.md#type-specification) of customer information that is required to complete the transaction. This field is only relevant when `status` is `pending_customer_info_update`. Wallets should use this field to determine which fields to request from the user.
```
JakeUrban, line 1124: I think we should provide the set of SEP-9 fields required on the transaction rather than use a SEP-12 customer type, which is just adding a level of indirection. It also requires businesses to group and label every potential set of SEP-9 fields required, which might be tedious if there are a lot of permutations.

ecosystem/sep-0006.md
```diff
@@ -1121,6 +1121,8 @@ Name | Type | Description
...
+`how` | string | (optional) A human-readable message indicating how the user should complete the transaction. This field is only required if `how` was not included in the response of a `deposit` request.
```
JakeUrban, line 1125: Rather than port `how` to the `/transaction(s)` response, I'd rather move away from having the anchor produce human-readable content that the anchor has no control over.

Instead, we should create a object that contain a set of SEP-9 account-related fields, providing information about the business' off-chain account in the case of deposit and on-chain account in the case of withdraw.

philipliu: The plan is to replace `how` with something more structured but I was thinking of addressing that in another PR. Is that okay with you?

Ifropc: I think it makes sense to do it in one PR, otherwise we will have ab in-between state we are not satisfied with

ecosystem/sep-0006.md
```diff
@@ -367,13 +368,38 @@ Name | Type | Description
...
+    "bank_number": {
+      "description": "Bank number",
+    },
+    "bank_branch_number": {
+      "description": "Bank branch number",
+    }
+  },
+  "id": "9421871e-0623-4356-b7b5-5996da122f3e",
+}
+```
```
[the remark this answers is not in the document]
philipliu: > Lets update all the examples (anyone using how already knows how to use it, everyone else can benefit from more examples using instructions.

Sure.

> Finally, we're providing information with the `instructions` object, not requesting it. So instead of having SEP-9 fields be objects with a `description`, we need the actual values from the organization:

Oops! Thanks for catching it. I think we should keep the `description` in case SEP-9 fields are not sufficient to describe the deposit instructions.

> I think we're missing SEP-9 fields for crypto addresses and memos in the case of withdraws. Can you add those to SEP-9 before merging this PR? 

Sure.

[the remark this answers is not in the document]
Ifropc: >we're missing SEP-9 fields for crypto addresses and memos in the case of withdraws
Should we also add chain identifier? If user want to receive ERC-20 (e.g stablecoin), it's unclear which chain should be used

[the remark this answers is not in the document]
JakeUrban: Lets use SEP-38's asset identification format, which has the `stellar` and `iso4217` schemes already. Other schemes can be added if an anchor wants to offer ethereum -> stellar bridging, for example.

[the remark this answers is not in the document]
philipliu: To be clear, this would be a withdrawal type listed in the `/info` response and not a SEP-9 field, right?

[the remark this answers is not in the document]
JakeUrban: Ah you're right, we already have those fields defined in SEP-6.. never mind on this request then.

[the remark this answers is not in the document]
philipliu: I think we should still add `crypto_address` and `crypto_memo` to the SEP-9 fields. Chain identifier is handled by the SEP-6 withdraw types.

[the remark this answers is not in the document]
philipliu: https://github.com/stellar/stellar-protocol/pull/1382

ecosystem/sep-0006.md
```diff
@@ -367,18 +368,50 @@ Name | Type | Description
...
+      "value": "121122676",
+      "description": "US bank routing number"
+    },
+    "organization.bank_account_number": {
+      "value": "13719713158835300",
+      "description": "US bank account number"
+    }
+  }
+}
+```
```
JakeUrban, line 402: We're not removing `how`, and its still required, so we need to have it present in our examples.

ecosystem/sep-0006.md
```diff
@@ -943,7 +1000,7 @@ All assets listed in a `deposit` and `deposit-exchange` can contain these attrib
 
 * `enabled`: `true` if SEP-6 deposit for this asset is supported
 * `authentication_required`: Optional. `true` if client must be [authenticated](#authentication) before accessing the deposit endpoint for this asset. `false` if not specified.
-* `fields` object as explained below.
+* `fields` (**Deprecated**, KYC information should be supplied to the Anchor via SEP-12) `fields` object is explained below.
```
JakeUrban, line 1003: We might want to expand on why this is deprecated (security of passing PII in URLs).

ecosystem/sep-0006.md
```diff
@@ -1123,6 +1180,9 @@ Name | Type | Description
 `refunds` | object | (optional) An object describing any on or off-chain refund associated with this transaction. The schema for this object is defined in the [Refunds Object Schema](#refunds-object-schema) section below.
 `required_info_message` | string | (optional) A human-readable message indicating any errors that require updated information from the user.
 `required_info_updates` | object | (optional) A set of fields that require update from the user described in the same format as [/info](#info).  This field is only relevant when `status` is `pending_transaction_info_update`.
+`required_customer_info_message` | string | (optional) A human-readable message indicating any errors that require updated information from the user.
```
JakeUrban, line 1183: I think this field will be used in the case where businesses haven't attempted to facilitate the transaction yet, so I think we should rephrase the description to remove mention of errors.

ecosystem/sep-0006.md
```diff
@@ -1123,6 +1180,9 @@ Name | Type | Description
...
+`required_customer_info_updates` | string | (optional) A set of SEP-9 fields that require update from the user via SEP-12.  This field is only relevant when `status` is `pending_customer_info_update`.
+`instructions` | string | (optional) JSON object containing the [SEP-9 financial account fields](sep-0009.md#financial-account-fields) that describe how to complete the off-chain deposit in the same format as the [/deposit](#deposit) response. This field is only required if `instructions` was not included in the response of a `deposit` request. Only relevant for deposit transactions.
```
JakeUrban, line 1185: I think we should change the description to say that `instructions` should be present on the object whether or not it was provided synchronously in the transaction initiation request.

Also, we should state that `instructions` should only be present once the status of the transaction is `pending_user_transfer_start` (not `incomplete` or `pending_cusotmer_info_update`, which are statuses that may proceed it).

JakeUrban (commented):
Should be the last batch of comments!

philipliu:
I appreciate the thorough review @JakeUrban! I believe I've addressed your comments now.

ecosystem/sep-0006.md
```diff
@@ -367,18 +368,52 @@ Name | Type | Description
 `fee_percent` | float | (optional) Percentage fee (if any). In units of percentage points.
 `extra_info` | object | (optional) JSON object with additional information about the deposit process.
 
+`instructions` fields:
+
+An object with SEP-9 financial account fields as keys and its values are objects with the following fields:
+
+Name | Type | Description
+-----|------|------------
+`description` | string | A human-readable description of the field. This can be used to provide any additional information about fields that are not defined in the SEP-9 standard.
```
JakeUrban, line 377: Add `value` here

ecosystem/sep-0006.md
```diff
@@ -1123,6 +1174,9 @@ Name | Type | Description
...
 `required_info_updates` | object | (optional) A set of fields that require update from the user described in the same format as [/info](#info).  This field is only relevant when `status` is `pending_transaction_info_update`.
+`required_customer_info_message` | string | (optional) A human-readable message indicating why the SEP-12 information provided by the user is not sufficient to complete the transaction.
+`required_customer_info_updates` | string | (optional) A set of SEP-9 fields that require update from the user via SEP-12.  This field is only relevant when `status` is `pending_customer_info_update`.
+`instructions` | string | (optional) JSON object containing the [SEP-9 financial account fields](sep-0009.md#financial-account-fields) that describe how to complete the off-chain deposit in the same format as the [/deposit](#deposit) response. This field should be present if the `instructions` were provided in the [/deposit](#deposit) response or if it could not have been previously provided synchronously. This field should only be present if the status is `pending_user_transfer_start`, not any status that precede it such as `incomplete`, or `pending_customer_info_update`.
```
JakeUrban, line 1179:
```suggestion
`instructions` | string | (optional) JSON object containing the [SEP-9 financial account fields](sep-0009.md#financial-account-fields) that describe how to complete the off-chain deposit in the same format as the [/deposit](#deposit) response. This field should be present if the `instructions` were provided in the [/deposit](#deposit) response or if it could not have been previously provided synchronously. This field should only be present once the status becomes `pending_user_transfer_start`, not while the transaction has any statuses that precede it such as `incomplete`, `pending_anchor`, or `pending_customer_info_update`.
```

JakeUrban (dismissed):
Ok, I think this is ready! Can we move forward with the implementation while we let this bake for a bit and allow the community to comment if they have feedback? I've let MyKobo and Beans know.

JakeUrban (approved):

ydag:
Great update! Enhances UX significantly!

PR #1386, opened 2023-08-24
[SEP-38] Switch country code ISO 

Ifropc:


ecosystem/sep-0038.md
```diff
@@ -7,9 +7,9 @@ Author: Jake Urban <@jakeurban> and Leigh McCulloch <@leighmcculloch>
 Track: Standard
 Status: Draft
 Created: 2021-04-09
-Updated: 2023-06-07
+Updated: 2023-08-24
 Discussion: https://github.com/stellar/stellar-protocol/issues/901
-Version 2.1.0
+Version 2.1.1
```
JakeUrban, line 12: I think this is a breaking change for those who expect alpha-3 codes, so we'll need to bump the major version and notify existing partners if we go with this approach.

JakeUrban:
@Ifropc where did we land with this, do we want to merge and update the anchor platform and notify clients?

Ifropc:
Yes, we want to merge this. No need to update the Platform, the schema is not validated on the platform

JakeUrban (approved):

PR #1387, opened 2023-08-31
SEP-31: deprecate per-transaction fields

JakeUrban:
This PR deprecates and makes optional the `fields` object in the `GET /info` response, the `POST /transactions` requests, and the `PATCH /transactions/:id` requests. It also deprecates the use of the `pending_transaction_info_updates` status, since this was only used to request per-transaction fields.

To our knowledge this object is not used, and was originally added to account for the possibility of non-standard per-transaction fields. However, supporting the use of non-standard fields in our tools & services meant to make implementing SEP-31 easier is difficult, and given the lack of usage in the ecosystem, it makes sense to deprecate the object and encourage the use of standard SEP-9 fields.


lijamie98 (approved):
LGTM

Ifropc (approved):

PR #1406, opened 2023-10-11
NFT-Collection interface (erc 1155 like)

silence48:
This draft proposal aims to align with the ERC-1155 standard by allowing for the creation and management of both fungible and non-fungible tokens within a single contract, and by providing functions for batch minting and transferring of NFTs. Further refinement and additional functions will be necessary to fully cater to the use-cases and requirements of NFT collections on Soroban while ensuring compatibility with the Stellar token interface and adherence to SEP0039.

It's based on my not yet functional implementation of an NFT collection creation contract which is acts like a factory for NFT contracts as defined in https://github.com/vinamogit/stellar-protocol/pull/2

### DIFF
--- a/contract/sep-nft-collection.md
+++ b/contract/sep-nft-collection.md
@@ -1,4 +1,5 @@
-Preamble
+## Preamble
+
 CIP: TBD
 Title: Token Collections (ERC-1155 like)
 Author: Stellar Development Foundation <@stellar>
--- a/contract/sep-nft-collection.md
+++ b/contract/sep-nft-collection.md
@@ -1,0 +1,114 @@
+## Summary
+
+This proposal introduces a standard interface for creating and managing NFT collections on Soroban, ensuring compatibility and interoperability within the Stellar ecosystem, and aligning with the previously defined NFT interface and the ERC-1155 standard.
+
+## Motivation
+
+A standardized interface for NFT collections on Soroban will facilitate seamless interaction within the Stellar ecosystem, enhancing user experience and developer interaction by providing a consistent API for NFT collection creation, management, and interaction, while also allowing for efficient batch transfers and management of both fungible and non-fungible tokens within a single contract.
+
+## Abstract
+
+This proposal defines a Rust trait for NFT collections, ensuring that all NFT collection contracts on Soroban adhere to a common API. This API includes functions for creating new NFTs within a collection, transferring NFTs, querying for ownership, and for accessing metadata necessary for displaying NFT information in wallets and other user interfaces.
+
+## Specification
+
+```rust
+pub trait NFTCollectionFactory {
+    // Admin interface – privileged functions.
+    fn initialize(
+        env: Env, 
+        admin: Address, 
+        collection_name: String, 
+        collection_symbol: String
+    );
+
+    fn mint_nft(
+        env: Env, 
+        to: Address, 
+        name: String, 
+        symbol: String, 
+        royalty_recipient: Address, 
+        royalty_percentage: u32, 
+        short_uri: String, 
+        detailed_uri: String, 
+        long_uri: String
+    ) -> Address; // Returns the address of the minted NFT
+
+    fn batch_mint_nft(
+        env: Env, 
+        to: Address, 
+        names: Vec<String>, 
+        symbols: Vec<String>, 
+        royalty_recipients: Vec<Address>, 
+        royalty_percentages: Vec<u32>, 
+        short_uris: Vec<String>, 
+        detailed_uris: Vec<String>, 
+        long_uris: Vec<String>
+    ) -> Vec<Address>; // Returns the addresses of the minted NFTs
+
+    // NFT Interface
+    fn transfer(env: Env, from: Address, to: Address, token_id: u32);
+
+    fn batch_transfer(env: Env, from: Address, to: Address, token_ids: Vec<u32>);
+
+    fn approve(
+        env: Env, 
+        owner: Address, 
+        approved: Address, 
+        token_id: u32
+    );
+
+    fn transfer_from(
+        env: Env, 
+        from: Address, 
+        to: Address, 
+        token_id: u32
+    );
+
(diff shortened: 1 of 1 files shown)


contract/sep-nft-collection.md
```diff
@@ -0,0 +1,114 @@
...
+A standardized interface for NFT collections on Soroban will facilitate seamless interaction within the Stellar ecosystem, enhancing user experience and developer interaction by providing a consistent API for NFT collection creation, management, and interaction, while also allowing for efficient batch transfers and management of both fungible and non-fungible tokens within a single contract.
+
+## Abstract
+
+This proposal defines a Rust trait for NFT collections, ensuring that all NFT collection contracts on Soroban adhere to a common API. This API includes functions for creating new NFTs within a collection, transferring NFTs, querying for ownership, and for accessing metadata necessary for displaying NFT information in wallets and other user interfaces.
+
+## Specification
+
+```rust
+pub trait NFTCollectionFactory {
```
willemneal, line 16: Why a factory? Doesn't that imply that it's deploying contracts?

contract/sep-nft-collection.md
```diff
@@ -0,0 +1,114 @@
...
+        collection_symbol: String
+    );
+
+    fn mint_nft(
+        env: Env, 
+        to: Address, 
+        name: String, 
+        symbol: String, 
+        royalty_recipient: Address, 
+        royalty_percentage: u32,
```
willemneal, line 31: Is this in basis points, e.g. 0-10,000?

contract/sep-nft-collection.md
```diff
@@ -0,0 +1,114 @@
...
+
+This proposal defines a Rust trait for NFT collections, ensuring that all NFT collection contracts on Soroban adhere to a common API. This API includes functions for creating new NFTs within a collection, transferring NFTs, querying for ownership, and for accessing metadata necessary for displaying NFT information in wallets and other user interfaces.
+
+## Specification
+
+```rust
+pub trait NFTCollectionFactory {
+    // Admin interface – privileged functions.
+    fn initialize(
+        env: Env,
```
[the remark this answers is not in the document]
leighmcculloch: Agreed, it's an implementation detail bleeding in.

contract/sep-nft-collection.md
silence48: It is deploying contracts. I was unable to find a way to make them compatible with token interface without having seperate addresses for minted token. 

Appreciate the feedback. Definitely not saying this is the best or correct way but rather a starting point for us to build off of.

contract/sep-nft-collection.md
silence48: > Is this in basis points, e.g. 0-10,000?

I guess it should be defined because but it is meant to be a percentage as a full number.

Basically 1 over royalty_p

  Im not sure if i should have even included royalty metadat in these interfaces as in eth it is a whole seperate eip however on eth its up to the marketplace to actually comply with royalties so if there is a way to force compliance the info would need to be accessible in all i think. 

Maybe theres a better way?

willemneal: Perhaps the metadata could include it as on option type? Then have the mint methods take metadata as an argument instead of an argument for each field.

contract/sep-nft-collection.md
```diff
@@ -0,0 +1,124 @@
...
+## Abstract
+
+This proposal defines a Rust trait for NFT collections, ensuring that all NFT collection contracts on Soroban adhere to a common API. This API includes functions for creating new NFTs within a collection, transferring NFTs, querying for ownership, and for accessing metadata necessary for displaying NFT information in wallets and other user interfaces.
+
+## Specification
+
+```rust
+pub trait NFTCollectionFactory {
+    // Admin interface – privileged functions.
+    fn initialize(
```
dmkozh, line 28: Initialization should probably not be a part of the common interface, similar to tokens.

contract/sep-nft-collection.md
dmkozh: Is the idea here that every NFT is a separate contract instance? If so, what is the purpose/value of having a contract that allows managing NFT collection as a whole? Intuitively, I'd think that the purpose of collection contract is to store and manage multiple NFTs without deploying separate contracts for each one (this is also reflected in functions like `transfer` that take `token_id` as an input). However if this just duplicates the functionality of regular (singular) tokens, then it seems redundant - the collection management would better be managed off-chain.

To be clear, doing something like `transfer` when token is a separate contract would require auth for `token.transfer`, i.e. `transfer` would act just as a router that maps `token_id` to and `Address`.  Or is the idea to have factory as admin for NFTs, so that `transfer` would do clawback+mint on the underlying token?

starpause:
Will this standard accommodate non-transferable NFTs? In my experience if non-transferable isn't addressed in the standard then people will implement that functionality in non standard ways (having the transfer function do nothing etc).

0xETHengineer (approved):

dmkozh:
>  In my experience if non-transferable isn't addressed in the standard then people will implement that functionality in non standard ways (having the transfer function do nothing etc).

What would be wrong if `transfer` returned an error when NFT is non-transferrable for whatever reason? I agree that 'do nothing' is bad behavior, but I think the error trivially follows from the properly defined standard (i.e. if 'transfer' has to succeed if and only if transfer succeeds, then any failure condition should result in an error trap).

starpause:
> What would be wrong if `transfer` returned an error when NFT is non-transferrable for whatever reason?

If that approach is considered standards compliant then non transferable nfts are already in the standard. It may be worth noting or documenting so future folks looking to implement NTNFT don't start creating their own standard for it.

dmkozh:
> If that approach is considered standards compliant then non transferable nfts are fine and functional with things as they are here.

Sure, it's probably worth mentioning here. But basically as long as we state clearly that transfer may succeed if and only if it has actually happened, then arbitrarily complex transferability conditions can be implemented while still being compliant to the standard.

PR #1410, opened 2023-10-19
PARTNER-146: Add external_account_id to SEP-9

Ifropc:


JakeUrban (commented):
What is the difference between the `id` sent in SEP-12 requests vs. the `external_account_id` parameter? They seem like the same thing to me.

CassioMG:
> What is the difference between the `id` sent in SEP-12 requests vs. the `external_account_id` parameter? They seem like the same thing to me.

Very good point. I guess it would depend on how the partner is planning on handling things on their side, but yes I believe they could potentially be the same thing as Sep-12's `id` field refers to the account id on the Anchor(partner) side.

Ifropc:
> What is the difference between the id sent in SEP-12 requests vs. the external_account_id parameter? They seem like the same thing to me.

Ideally, it should be the same (I don't see why they should be different, actually). This field is to be used in other SEPs that may not use SEP-12, to pass the customer's ID along the KYC (mainly, it's for SEP-24)
Do you think that it makes sense to just add it there instead?

JakeUrban:
I think so. Adding something like `customer_id` to the SEP-24 (and SEP-6) transaction initiation endpoint, noting that this field should match the value returned from the anchor's SEP-12 `PUT /customer` response, seems to make sense.

JakeUrban:
This matches SEP-31's design of having `sender_id` and `receiver_id` fields in its transaction initiation request.

Ifropc:
@JakeUrban do you think we need to add `customer_id` to SEP-6, considering client first calls SEP-12 in the flow?

Ifropc:
I think it actually makes sense to pass customer_id before KYC.
If it's passed after, the business should already have tied JWT to their internal account.  
However, if it's done before, business can do the assignment of Stellar account to the internal account at this step.

ecosystem/sep-0024.md
```diff
@@ -1386,9 +1388,10 @@ There is a small set of changes when upgrading from SEP-6 to SEP-24.
 
 ## Changelog
 
+- `v3.4.0`: Add `extrernal_account_id` field ([#1410](https://github.com/stellar/stellar-protocol/pull/1410))
```
CassioMG, line 1391: I think the changelog also needs an update here

ecosystem/sep-0006.md
```diff
@@ -1480,7 +1484,7 @@ If the information was malformed, or if the sender tried to update data that isn
 [SEP-38]: sep-0038.md
 
 ## Changelog
-
+- `v3.23.0`: Add `extrernal_account_id` field ([#1410](https://github.com/stellar/stellar-protocol/pull/1410))
```
CassioMG, line 1487: same here

JakeUrban:
Ah you're right. The `customer_id` attribute only needs to be passed if 

- the customer exists in the anchor's records
- the client somehow knows the ID
- the anchor doesn't know the customer's stellar account (therefore wasn't registered via SEP-12 yet, if they need to be at all)

ecosystem/sep-0006.md
```diff
@@ -331,6 +331,7 @@ Name | Type | Description
 `amount` | string | (optional) The amount of the asset the user would like to deposit with the anchor. This field may be necessary for the anchor to determine what KYC information is necessary to collect.
 `country_code` | string | (optional) The [ISO 3166-1 alpha-3](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-3) code of the user's current address. This field may be necessary for the anchor to determine what KYC information is necessary to collect.
 `claimable_balance_supported` | string | (optional) `true` if the client supports receiving deposit transactions as a claimable balance, `false` otherwise.
+| `customer_id`                 | string | (optional) ID of an account associated with this user (identified by JWT) used by the anchor. If the anchor supports [SEP-12], the `customer_id` field should match [SEP-12] Customer's id.                                                                                                                                                                                                                                                                                               |
```
JakeUrban, line 334: I don't quite understand the definition.

> ID of an account associated with this user (identified by JWT) used by the anchor.

Is the `account` supposed to be some kind of off-chain record, like a github account? Or are you referring to a Stellar account here?

When you say `(identified by JWT)`, are you referring to the user or their account mentioned above? 

Unrelatedly, lets make sure we're being consistent with how we use "id" vs. "ID` and remove the capital C in `Customer's`.

Ifropc: > Is the account supposed to be some kind of off-chain record, like a github account

Yes, off-chain account (internal anchor's account)

> When you say (identified by JWT), are you referring to the user or their account mentioned above?

Account + memo combination 

I will rephrase description so it's more clear

ecosystem/sep-0006.md
```diff
@@ -331,6 +331,7 @@ Name | Type | Description
...
 `country_code` | string | (optional) The [ISO 3166-1 alpha-3](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-3) code of the user's current address. This field may be necessary for the anchor to determine what KYC information is necessary to collect.
 `claimable_balance_supported` | string | (optional) `true` if the client supports receiving deposit transactions as a claimable balance, `false` otherwise.
+| `customer_id`                 | string | (optional) id of an off-chain account (managed internally by the anchor) associated with this user's Stellar account (identified by JWT). If the anchor supports [SEP-12], the `customer_id` field should match [SEP-12] customer's id. `customer_id` should be passed only when off-chain id is know to the client, but relationship between this id and user's Stellar account is not known to the Anchor.                                                                                                                                                                                                                                                                                               |
```
JakeUrban, line 334:
```suggestion
| `customer_id`                 | string | (optional) id of an off-chain account (managed by the anchor) associated with this user's Stellar account (identified by the JWT's `sub` field). If the anchor supports [SEP-12], the `customer_id` field should match the [SEP-12] customer's id. `customer_id` should be passed only when the off-chain id is know to the client, but the relationship between this id and the user's Stellar account is not known to the Anchor.                                                                                                                                                                                                                                                                                               |
```

JakeUrban (dismissed):
LGTM with the grammar edits made in my comment.

ecosystem/sep-0006.md
Ifropc: Keep forgetting the's...

JakeUrban (approved):

CassioMG (approved):
Lgtm

PR #1417, opened 2023-10-31
SEP-6: Make `account_id` optional in withdraw response

philipliu:
This makes the `account_id` field in the `withdraw` and `withdraw-exchange` response optional. This is to support Anchors integrating with a Custodian, in which a Stellar address other than the asset's distribution account is used as the `withdraw_anchor_account`. This is to prevent users from accidentally depositing to the wrong address in case the `withdraw_anchor_account` is changed during the SEP transaction's lifecycle.

### DIFF
--- a/ecosystem/sep-0006.md
+++ b/ecosystem/sep-0006.md
@@ -6,8 +6,8 @@ Title: Deposit and Withdrawal API
 Author: SDF
 Status: Active (Interactive components are deprecated in favor of SEP-24)
 Created: 2017-10-30
-Updated: 2023-10-20
-Version 3.23.0
+Updated: 2023-10-31
+Version 3.24.0
 ```
 
 ## Simple Summary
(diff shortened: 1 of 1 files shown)


JakeUrban:
I'm not sure we want to move forward this with this.

Ideally, `withdraw_anchor_account` should never change once its originally set in order to avoid confusion / mistakes. We only allow `withdraw_anchor_account` to be changed in the Anchor Platform because the field is set to the asset's configured distribution account when the transaction is created.

This is a bug, since `withdraw_anchor_account` should only be available once the business is ready to receive funds and the transaction is in `pending_user_transfer_start`. But because businesses already depend on that behavior, we can't change that until 3.0.

All this is to say that `account_id` should probably remain a required field for the synchronous success response. The Anchor Platform will never return this response type so we don't need to be concerned about providing the wrong `account_id`.

JakeUrban:
Actually, I misspoke when I said the Anchor Platform will never return the success response (it will).

So now I see the problem.. the Anchor Platform's implementation has a bug but we can't fix it without potentially breaking business's application logic. At the same time, allowing businesses to update `withdraw_anchor_account` means the wallet will get an incorrect account address in the synchronous response.

So I've come full circle now 😅 and think making `account_id` optional is really the only viable path forward. I think this makes sense from an API design perspective as well. We don't want to provide deposit instructions in the synchronous response, so we shouldn't do that in the withdraw response either.

JakeUrban (commented):
Can we update the description of the responses too?

> This is the correct response if the anchor is able to execute the withdrawal and needs no additional information about the user.

This isn't actually true now. This probably shouldn't be considered a "success" response since its possible for the business to request KYC info after returning 200 and the transaction's ID. 

Lets update the deposit response description too if it has the same issue.

philipliu:
@JakeUrban Thanks for catching that, I made the changes 👍

JakeUrban (approved):

PR #1422, opened 2023-11-22
Mark Soroban CAPs as Accepted

rice2000:
Final Comment Period is now complete, and there were no questions or objections raised, so this PR changes the designation of the Soroban CAPs from `FCP: Acceptance` to `Accepted`


tomerweller (approved):
LGTM

PR #1427, opened 2024-01-11
Add host functions to support BLS12-381

namankumar:
This CAP proposes adding support for BLS12-381 elliptical curve to enable advanced zero knowledge proof systems including privacy preserving constructions and zkrollups. The Soroban ecosystem has already expressed interest in these type of systems.

Next steps:
- determine appropriate size for modulus and field elements
- determine appropriate size, types, and encoding for arguments
- consider fees
- consider sample applications to aid ecosystem adoption

Issue: https://github.com/stellar/rs-soroban-env/issues/779
PR (in progress): https://github.com/stellar/rs-soroban-env/pull/1310


core/cap-tbd.md
```diff
@@ -0,0 +1,209 @@
...
+ const BLST_G1_POINT: usize = 96; (rename after adding decoding functions)
+ const BLST_G2_POINT: usize = 192;
+ const BLST_SCALAR_SIZE: usize = 255;
+```
+### Curve Operations
+
+#### Addition in G1
+
+```rust
+fn bls12_g1add( e: &Host, p0: BLST_G1_POINT, p1: BLST_G1_POINT) -> Result <BLST_G1_POINT, Error>;
```
dmkozh, line 68: `BLST_G1_POINT` is not a type, what is this signature supposed to mean?

While it's probably not important for the discussion, if we want to merge this, we should make it consistent with the host function specifications we have.

https://github.com/stellar/rs-soroban-env/blob/main/soroban-env-common/env.json can be used as a general reference. Some quick notes:
- There is no need to specify the `&Host` argument.
- There is no need to declare the return type as `Result`, just use the return value
- There is no `u8` type supported; minimal integer type is `U32Val`.
- Slices can't be passed to the host functions; use `BytesObject` instead (the length of the bytes object is just a part of the protocol specification, but can't be reflected in the function signature)

core/cap-tbd.md
```diff
@@ -0,0 +1,209 @@
...
+
+```rust
+fn bls12_pairing( e: &Host, p_g1: &[BLST_G1_POINT], p_g2: &[BLST_G2_POINT]) -> Result <u32, Error>;
+```
+
+Perform pairing between points in G1 and G2. 
+Arguments: 
+p_g1: point in G1(96 bytes)
+p_g1: point in G1(192 bytes)
+Returns padded 0x01 if pairing is successful else 0x00.
```
dmkozh, line 129: Why not return `Bool` instead?

core/cap-tbd.md
```diff
@@ -0,0 +1,209 @@
...
+### Curve Operations
+
+#### Addition in G1
+
+```rust
+fn bls12_g1add( e: &Host, p0: BLST_G1_POINT, p1: BLST_G1_POINT) -> Result <BLST_G1_POINT, Error>;
+```
+Perform addition of two points in G1. 
+Inputs should be 96 bytes each. 
+An error will be thrown if neither input is in G1, input has invalid length, or encoding rules are violated.
```
dmkozh, line 72: We don't return any errors actually - the host functions just trap in case of an error.

core/cap-tbd.md
```diff
@@ -0,0 +1,209 @@
...
+Arguments: 
+p_g1: point in G1(96 bytes)
+p_g1: point in G1(192 bytes)
+Returns padded 0x01 if pairing is successful else 0x00.
+Error will be thrown if p0_g1 and p1_g1 are not on G1 and G2 respectively, if the points are in the incorrect subgroups, have invalid length, inputs are empty, encoding rules are violated. 
+
+#### Map Base Field Element to G1
+
+```rust
+fn bls12_map_fp_to_g1( e: &Host, fp: &[u64]) -> Result <BLST_G1_POINT, Error>;
```
dmkozh, line 135: Why do we have two variants of the same 'map' function that seemingly take the same input? I think we should only need one version, that either takes a `U64Val`, or `BytesObject`, depending on which case is more useful.

core/cap-tbd.md
```diff
@@ -0,0 +1,209 @@
...
+Following types are used throughout this proposal.
+
+```
+ const BLST_G1_POINT: usize = 96; (rename after adding decoding functions)
+ const BLST_G2_POINT: usize = 192;
+ const BLST_SCALAR_SIZE: usize = 255;
+```
+### Curve Operations
+
+#### Addition in G1
```
dmkozh, line 65: It seems like there is a lot of redundancy in this interface. E.g. we could reduce the number of functions by 2 times if we take g1/g2 as an input bool parameter to functions instead of having duplicate functions for every operation.

core/cap-tbd.md
```diff
@@ -0,0 +1,209 @@
...
+
+## Simple Summary
+Add support for the BLS12-381 elliptical curve to enable zero knowledge proof systems.
+
+## Working Group
+
+This section describes the composition of the working group.
+
+## Motivation
+BLS12-381 host functions allow for cryptographic operations in the Soroban execution environment that are otherwise prohibitively expensive. It enables a variety of applications including confidentiality preserving systems, BLS signature schemes, and multiple types of scaling solutions. Pairing-friendly elliptic curves are needed for such systems and BLS12-381 is the state of the art in secure elliptical curves.
```
tomerweller, line 24: I would add a note about the motivation of using BLS12-381 as opposed to BN254

core/cap-tbd.md
jayz22: Agree with @dmkozh.
I would suggest following https://github.com/stellar/stellar-protocol/blob/master/core/cap-0046-03.md on host function definition format. The host functions are written in WAT (WebAssembly Text) format because that (not Rust, or Json) is what is exposed to the WASM contracts.

dmkozh: While WAT might be ok for bulk definitions we had in the protocol 20, I would prefer to also have a definition in terms of env-common types (used in `env.json`) - while these functions are not type-checked on the Wasm side, they are type-checked on the host-side, so it would make sense to specify what should be the signature. It is also easier to understand for the reader, which is important given that hopefully future CAPs will be focused on small sets of functions.

jayz22: `env.json` sounds good, i don't feel strongly either way

rsinha:
What are your thoughts on exposing a generic interface for pairing friendly curves? That is, instead of `bls12_g1add`, you could expose `add` for curve elements, that is implemented by the individual curve. This might make it easier to add additional curves in future. 

One reference would be the algebraic interfaces defined in [arkworks](https://github.com/arkworks-rs/algebra/).

PR #1429, opened 2024-01-25
[ANCHOR-505] Add fee details field to transaction object of SEP-6 and SEP-24

Ifropc:
## Background
Currently, both SEP-24 and SEP-6 offers an ability for anchor to charge fees. However, fees are only present to user as a single value (`transaction.fee`, without an ability to break it down. This proposal addresses this issue with ability to specify all charges being made (such as fees, taxes), on top of the existing summary value.  

## Proposed change
Deprecate `amount_fee` and `amount_fee_asset` in favor of a new field: `fee_details`. This field follows the same schema as fee details defined in SEP-38 standard.


ecosystem/sep-0024.md
```diff
@@ -1158,6 +1159,22 @@ Each object in the `transactions` array should have the following fields:
...
 | `fee`     | string | The amount charged as a fee for processing the refund, in units of `amount_in_asset`.                                                                                                                                                                                |
 
+### Fee Details Object Schema
+
+| Name        | Type   | Description                                                                                                                                                           |
+| ----------- | ------ | --------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `total`     | string | The total amount of fee applied.                                                                                                                                      |
+| `asset`     | string | The asset in which the fee is applied, represented through the [Asset Identification Format](#asset-identification-format).                                           |
```
philipliu, line 1167: This should link to `sep-0038.md#asset-identification-format`. `#asset-identification-format` does not exist in this file. The same applies to SEP-6.

philipliu (approved):
LGTM, just a minor comment.

ecosystem/sep-0024.md
```diff
@@ -1067,29 +1067,30 @@ Each object in the `transactions` array should have the following fields:
...
+| `amount_fee`              | string              | (**deprecated**) Amount of fee charged by anchor. This field is deprecated in favor of the `fee_details` field.                                                                                                                                                                                                                   |
+| `amount_fee_asset`        | string              | (**deprecated**, optional) The asset in which fees are calculated in. Must be present if the deposit/withdraw was made using non-equivalent assets. The value must be in [SEP-38 Asset Identification Format](sep-0038.md#asset-identification-format). See the [Asset Exchanges](#asset-exchanges) section for more information. |
+| `fee_details`             | object              | Description of fee charged by the anchor. The schema for this object is defined in the [Fee Details Object Schema](#fee-details-object-schema) section below.                                                                                                                                                                     |
```
JakeUrban, line 1084: Lets add a note here that `fee_details` should match the referenced quote's `fee` object.

JakeUrban (commented):
I don't love that we're duplicating information on `GET /quote/:id` and `GET /transaction?id=`, is this avoidable? Am I correct in understanding that we're adding this because having fee info on the quote doesn't work when the quote is indicative or a quote is not used?

Ifropc:
> Am I correct in understanding that we're adding this because having fee info on the quote doesn't work when the quote is indicative or a quote is not used?

Yes, you are right. I think duplicating fee object for when the quote is used is the better option

philipliu (dismissed):
LGTM

ecosystem/sep-0006.md
```diff
@@ -1462,6 +1463,22 @@ Each object in the `transactions` array should have the following fields:
...
 
+### Fee Details Object Schema
+
+| Name        | Type   | Description                                                                                                                                                           |
+| ----------- | ------ | --------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `total`     | string | The total amount of fee applied.                                                                                                                                      |
+| `asset`     | string | The asset in which the fee is applied, represented through the [Asset Identification Format](sep-0038.md#asset-identification-format).                                |
+| `breakdown` | array  | (optional) An array of objects detailing the fees that were used to calculate the conversion price. This can be used to datail the price components for the end-user. |
```
JakeUrban, line 1472: I know its redundant to have `fee_details.details` but I think its better than `breakdown`. Can we update the naming in this structure to match SEP-38's fee object as closely as possible?

JakeUrban (dismissed):
Approved pending the comment, but I don't feel strongly about it.

ecosystem/sep-0006.md
Ifropc: That's a good point, I was going to reuse exact same object, but forgot to change the field name back

JakeUrban (approved):

PR #1432, opened 2024-01-26
Update SEP-6 flow to delegate getting KYC fields to SEP-12

Ifropc:
## Background
Current SEP-6 flow for working with KYC is quite complex from the anchor standpoint: it requires to pass the same set of information (required KYC fields) via both SEP-6 and SEP-12. This proposal aims to simplify the flow for both anchor and wallet. This proposal logically separates SEP-6 and SEP-12, and all work with KYC is done via SEP-12.   

## Proposed flow
The process of providing information to the anchor looks as follows:

1. `GET /transaction` and verify transaction status is `pending_customer_info_update`
2. Get required fields via [SEP-12] `GET /customer` request. Note, that you must pass `transaction_id` as part of the
   request.
3. Collect user input for the required fields and use [SEP-12] `PUT /customer` endpoint to transfer the data
4. Repeat steps 2-3 until customer's response status becomes `ACCEPTED`
5. `GET /transaction` should transition to the next status (`pending_user_transfer_start`)

## Proposed change
This proposal includes multiple changes in order to simplify the flow.

SEP-6:
1. Deprecate response code 403 (Forbidden) from the anchor on missing customer information. Instead, anchor should return 200 and create the transaction in status `pending_customer_info_update`
2. Remove `required_customer_info_message` and `required_customer_info_updates`. This field have not being adopted yet, so it's safe to remove them. Instead, transaction in status `pending_user_transfer_start` should use existing `message` field to indicate that additional input is required, and use [SEP-12] `PUT /customer` endpoint (with transaction id)
 
SEP-12:
Add `transaction_id` field to `GET /customer` and `PUT /customer`, allowing to fetch/update fields required in the scope of specific transaction


ecosystem/sep-0006.md
```diff
@@ -769,16 +768,19 @@ The response body should be a JSON object with the following fields:
...
-`memo` | string | (optional) Value of memo to attach to transaction, for `hash` this should be base64-encoded. The
-anchor should use this memo to match the Stellar transaction with the database entry associated created to represent it.
-`id` | string | (optional) The anchor's ID for this withdrawal. The wallet will use this ID to query the
-[`/transaction`](#single-historical-transaction) endpoint to check status of the request. `eta` | int | (optional)
-Estimate of how long the withdrawal will take to credit in seconds. `min_amount` | float | (optional) Minimum amount of
-an asset that a user can withdraw. `max_amount` | float | (optional) Maximum amount of asset that a user can withdraw.
-`fee_fixed` | float | (optional) If there is a fee for withdraw. In units of the withdrawn asset. `fee_percent` | float
-| (optional) If there is a percent fee for withdraw. `extra_info` | object | (optional) Any additional data needed as an
-input for this withdraw, example: Bank Name.
+asynchonously.
```
Ifropc, line 771: Changes below fixes formatting

ecosystem/sep-0012.md
```diff
@@ -102,6 +102,7 @@ Name | Type | Description
...
 `memo_type` | string | (**deprecated**, optional) type of `memo`. One of `text`, `id` or `hash`. Deprecated because memos should always be of type `id`, although anchors should continue to support this parameter for outdated clients. If `hash`, `memo` should be base64-encoded. If a memo is present in the decoded SEP-10 JWT's `sub` value, this parameter can be ignored. See the [Shared Accounts](#shared-omnibus-or-pooled-accounts) section for more information.
 `type` | string | (optional) the type of action the customer is being KYCd for.  See the [Type Specification](#type-specification) below.
+`transaction_id` | `string` | (optional) Transacion id with which customer's info is associated. In cases when information required from the customer depends on the transaction (e.g. more infromation is required for bigger amounts), this field should be provided to indicate to the anchor transaction's context.                                                                                                                                               |
```
philipliu, line 105:
```suggestion
`transaction_id` | `string` | (optional) the transaction id with which the customer's info is associated. When information from the customer depends on the transaction (e.g., more information is required for larger amounts)
```

ecosystem/sep-0006.md
```diff
@@ -769,16 +768,19 @@ The response body should be a JSON object with the following fields:
...
-asynchonously. `memo_type` | string | (optional) Type of memo to attach to transaction, one of `text`, `id` or `hash`.
-`memo` | string | (optional) Value of memo to attach to transaction, for `hash` this should be base64-encoded. The
-anchor should use this memo to match the Stellar transaction with the database entry associated created to represent it.
-`id` | string | (optional) The anchor's ID for this withdrawal. The wallet will use this ID to query the
-[`/transaction`](#single-historical-transaction) endpoint to check status of the request. `eta` | int | (optional)
-Estimate of how long the withdrawal will take to credit in seconds. `min_amount` | float | (optional) Minimum amount of
-an asset that a user can withdraw. `max_amount` | float | (optional) Maximum amount of asset that a user can withdraw.
-`fee_fixed` | float | (optional) If there is a fee for withdraw. In units of the withdrawn asset. `fee_percent` | float
-| (optional) If there is a percent fee for withdraw. `extra_info` | object | (optional) Any additional data needed as an
-input for this withdraw, example: Bank Name.
```
philipliu, line 781: Is this related to the KYC change? Should we be deprecating this first before removing it?

JakeUrban: I think was a formatting issue that @Ifropc fixed below.

ecosystem/sep-0012.md
```diff
@@ -102,6 +102,7 @@ Name | Type | Description
 `memo` | string | (optional) the client-generated [memo](https://developers.stellar.org/docs/glossary/transactions/#memo) that uniquely identifies the customer. If a memo is present in the decoded SEP-10 JWT's `sub` value, it must match this parameter value. If a muxed account is used as the JWT's `sub` value, memos sent in requests must match the 64-bit integer subaccount ID of the muxed account. See the [Shared Accounts](#shared-omnibus-or-pooled-accounts) section for more information.
 `memo_type` | string | (**deprecated**, optional) type of `memo`. One of `text`, `id` or `hash`. Deprecated because memos should always be of type `id`, although anchors should continue to support this parameter for outdated clients. If `hash`, `memo` should be base64-encoded. If a memo is present in the decoded SEP-10 JWT's `sub` value, this parameter can be ignored. See the [Shared Accounts](#shared-omnibus-or-pooled-accounts) section for more information.
 `type` | string | (optional) the type of action the customer is being KYCd for.  See the [Type Specification](#type-specification) below.
```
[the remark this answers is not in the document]
Ifropc: > it doesn't make sense to use both in the same request.

I think it still does. If we keep type to the protocol level (sep-24, sep-6, sep-31) it makes either for the anchor to work with the request.
First, it's technically possible for transaction to have the same id between protocols
Second, in AP implementation, without specifying type, it'll need to do 3 requests, instead of 1 (because it needs to search for transaction in all 3 tables).

So, IMO we should instead advice into matching `type` with the protocol that's being used, when defining transaction, and deprecate all other usages.
WDYT?

ecosystem/sep-0006.md
```diff
@@ -6,8 +6,8 @@ Title: Deposit and Withdrawal API
 Author: SDF
 Status: Active (Interactive components are deprecated in favor of SEP-24)
 Created: 2017-10-30
-Updated: 2023-11-07
-Version 3.24.1
+Updated: 2024-01-25
+Version 3.25.0
```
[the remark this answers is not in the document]
Ifropc: 4.0 is it 👍🏼

ecosystem/sep-0006.md
Ifropc: Formatting issue!

ecosystem/sep-0012.md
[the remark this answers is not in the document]
JakeUrban: I don't think anyone will use `type` moving forward if we add `transaction_id`. For those that already using `type`, limiting its usage to indicate which SEP the customer is KYC'ing for would require them to update their implementation, and at that point they're probably better off switching to `transaction_id`.

As for the uniqueness of the ID, I think that is up to the anchor to ensure the IDs aren't the same between the SEPs they support. The anchor platform chose to use different tables for each SEP (which I don't think is the right approach), but they're random UUID4's so chance of collision is minimal.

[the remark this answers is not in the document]
Ifropc: `type` is still required for SEP-31; and we can use `type` to specify protocol when used with `transaction_id`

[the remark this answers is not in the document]
JakeUrban: Gleb and I discussed this offline. In short: we'll not deprecate `type` since the SEP-31 has not been updated to create transactions before KYC is checked / processed. 

However, I do think we should deprecated `type` once SEP-31 has been updated, since there is no need for general-purpose custom field, and we can add an alternative to `type` that is explicitly for the purpose of clarifying the SEP, like a `sep` or `transaction_type` argument.

ecosystem/sep-0012.md
```diff
@@ -102,6 +102,7 @@ Name | Type | Description
...
+`transaction_id` | `string` | (optional) The transaction id with which the customer's info is associated. When information from the customer depends on the transaction (e.g., more information is required for larger amounts)                                                                                                                                                                                                                                                                                          |
```
philipliu, line 105: nit: Should we specify what type of transaction this refers to? i.e. it refers to the "action" mentioned in the type specification section.

Ifropc: It's up to the anchor implementation. However, in Platform implementation we will require type to be specified, that matches appropriate SEP (so client application should pass both transaction_id=abc and type=sep24 for example) so we know which table to lookup, and to avoid ambiguity when different protocols have transactions with the same id

JakeUrban (approved):

JakeUrban (approved):

philipliu (dismissed):

philipliu (approved):

philipliu (approved):

PR #1433, opened 2024-01-27
[SEP-6] Add types to /info deposits, update description for withdrawal

Ifropc:
## Background
Sometimes anchor may want to receive additional information from the wallet, that's not required in the generic case. While it's good to keep protocol consistent, adding fields just to be used for one anchor seems to be excessive. For example, one of partners would like to add `location_id` to the request, which is not going to be used by the majority of anchors. Other anchors may want to add some fields specific to their use case.
Currently, it can be supported via simply passing extra field to the request. It's also supported to define type of the object via `types` of `/info` endpoint, but only for `withdrawal`. What's more, current `types` approach also extends on KYC/financial fields, which is responsibility of SEP-12

## Proposed change
1. Remove support of KYC/financial fields in `types` of `/info` endpoint. Instead, types should be defined directly in `SEP-12 /GET`.
2. Add support of `types` to deposit.
3. Rephrase description of `types`: it only supports extra arguments of withdrawal/deposit requests, that are not related to the customer, but only to the transaction


ecosystem/sep-0006.md
```diff
@@ -1606,6 +1610,16 @@ body should contain error details. For example:
 }
 ```
 
+### Extra object schema
+
+This object is used to pass any additional info to the anchor. Anchors are free to add any field to this object. Anchors
+should notify wallets about fields they are expecting to receive. However, it's recommended to submit a change to this
```
philipliu, line 1616: How will Anchors communicate this to the wallet? Should the schema be included in the `/info` response?

Ifropc: The idea was that this field should be rarely used, and if it is used, anchor should contact wallet directly or print comprehensive error message. 
However, adding schema to `/info` sounds like a good idea. We can only allow simple objects as part of `extra` to simplify definition of the schema

JakeUrban:
Its unclear if fields listed in `/info` need to be SEP-9 or can be custom. The fields `dest` and `dest_extra` are defined in the SEP itself, where as others in the example are SEP-9 fields, so its unclear if `location_id` should be defined somewhere as well.

I'm not opposed to simply adding an optional `location_id` argument. It is unclear how using an object like `extra` would work in the context of GET URL arguments.

ecosystem/sep-0006.md
```diff
@@ -1050,55 +1050,39 @@ Request parameters:
...
-        "type" : {
-          "description": "type of deposit to make",
-          "choices": ["SEPA", "SWIFT", "cash"]
+      "types": {
+        "location": {
+          "fields": {
+            "id": {
+              "description": "ID of location user must visit to pick up cash"
+            }
+          }
```
JakeUrban, line 1067: I think the example should be something like:

```suggestion
      "types": {
        "cash": {
          "fields": {
            "location_id": {
              "description": "ID of location user must visit to pick up cash"
            }
          }
```

JakeUrban: Same for the other examples

ecosystem/sep-0006.md
```diff
@@ -1110,18 +1094,6 @@ The response should be a JSON object like:
...
-              "dest": {"description": "your bank account number" },
-              "dest_extra": { "description": "your routing number" },
-              "bank_branch": { "description": "address of your bank branch" },
-              "phone_number": { "description": "your phone number in case there's an issue" },
-              "country_code": {
-                "description": "The ISO 3166-1 alpha-3 code of the user's current address",
-                "choices": ["USA", "PRI"]
-              }
-          }
-        },
```
JakeUrban, line 1124: To confirm, we're removing this because the `bank_account` type is requesting the user's KYC info like phone number and bank account number right? And we want them to do this via SEP-12?

Don't we still need the client to provide the `type=bank_account` query parameter though, so the anchor knows to request bank account fields in the `GET /customer` response? If that is the case, then we need to do something like this:

```
"bank_account": {}
```

This signals that the type should be passed but there are no extra fields that need to passed in the transaction creation request.

ecosystem/sep-0006.md
```diff
@@ -1194,10 +1154,16 @@ All assets listed in a `deposit` and `deposit-exchange` can contain these attrib
...
 - `fields` (**Deprecated**, Accepting personally identifiable information through request parameters is a security risk
-  due to web server request logging. KYC information should be supplied to the Anchor via SEP-12) `fields` object is
+  due to web server request logging. KYC information should be supplied to the Anchor via SEP-12. For extra fields `types` should be used) `fields` object is
   explained below.
+- `types`: a field with each type of deposit supported for that asset as a key. Each type can specify a `fields`
+  object as below explaining what fields are needed and what they do. Anchors can define custom fields if
+  necessary. If a `types` object is not specified, the wallet should assume that no extra fields are needed for that
+  type of deposit. In the case that the Anchor requires additional fields for a deposit, it should set the
+  transaction status to `pending_customer_info_update`. The wallet can query the `/transaction` endpoint to get the
+  fields needed to complete the transaction in `required_customer_info_updates` and then use
```
JakeUrban, line 1164: This stops mid-sentence. I also think you mean `pending_transaction_info_update` and `required_info_updates`.

JakeUrban (commented):
I'm not 100% convinced we should be adding a `types` map to deposit info objects. 

I understand that this approach makes sense because the withdraw objects use the same approach, but withdraws and deposits are different. With withdraws, you need to tell the anchor your financial account information so they can send you funds off-chain. But with deposits, the anchor actually needs to tell you their financial account information so you can send the funds off-chain. Thats why the deposit info objects don't have a `types` map today -- there is typically no extra non-KYC information that is needed, and that is why we have an `instructions` field in the deposit endpoints response.

Location selection seems like the one exception to that rule. The anchor can't tell the user which location to go to, the user needs to pick. I can't think of another type of deposit that would require extra arguments from the client.

Because of that, I think I still prefer adding `location_id` as a first-class parameter to deposit request endpoints, rather than it being a custom field defined in `GET /info`. If you can point out other potential uses of a `types` array for deposits, I'm willing to change my mind.

EDIT: actually, I know its not a requirement for MGI since you can pick up cash at any location, but should we consider adding `location_id` to withdraw request endpoints as well?

ecosystem/sep-0006.md
Ifropc: If we do it like described above, that would mean that we accept field named `bank_account` with empty body though, am I wrong? 
We still have `type` in the request, it should be used as before

Ifropc:
I don't mind just adding location_id directly, but I think it's more flexible if we allow anchors to define their own fields, so we don't need to modify protocol for any small deviation from SEP-6 protocol

ecosystem/sep-0006.md
JakeUrban: > If we do it like described above, that would mean that we accept field named bank_account with empty body though, am I wrong?

On what endpoint would anchors accept a `"bank_account": {}` key-value pair from wallet requests? I'm referring to the fact that `GET /info` responses should have all possible values of the `type` query parameter as keys. So we shouldn't remove the entire `bank_account` object, just the items within the object, since no additional parameters should be sent to the anchor other than the `type` parameter in that case.

JakeUrban:
I agree your solution provides anchors a way to request their own custom fields, and if we were starting from scratch I think this is the right approach. The reason I think we should try to avoid making this change is simply because it is breaking, and it is unclear how common the need for custom fields will be.

I also think that a `location_id` parameter makes sense to include in an on/off ramp standard API. Right now MGI is the only cash-agent network that supports a stellar on/off ramp standard, but any future cash-agent network anchor will need this. And if we don't standardize the `location_id` parameter, they'll define their own parameter in `GET /info` which could use a different name.

JakeUrban:
Ok, it sounds like we're on the same page of adding `location_id` to request parameters rather than the approach defined here, so I'll close this PR.

Ifropc:
Reopened PR to preserve discussion history

philipliu (approved):

ecosystem/sep-0006.md
```diff
@@ -828,6 +829,7 @@ Request Parameters:
...
 | `customer_id`                 | string                  | (optional) id of an off-chain account (managed by the anchor) associated with this user's Stellar account (identified by the JWT's `sub` field). If the anchor supports [SEP-12], the `customer_id` field should match the [SEP-12] customer's id. `customer_id` should be passed only when the off-chain id is know to the client, but the relationship between this id and the user's Stellar account is not known to the Anchor.                                   |
+| `location_id`                 | string                  | (optional) id of the chosen location for cash pick up cash                                                                                                                                                                                                                                                                                                                                                                                                            |
```
JakeUrban, line 832:
```suggestion
| `location_id`                 | string                  | (optional) id of the chosen location to drop off cash                                                                                                                                                                                                                                                                                                                                                                                                            |
```

ecosystem/sep-0006.md
```diff
@@ -726,6 +726,7 @@ Request parameters:
...
 | `customer_id`        | string                  | (optional) id of an off-chain account (managed by the anchor) associated with this user's Stellar account (identified by the JWT's `sub` field). If the anchor supports [SEP-12], the `customer_id` field should match the [SEP-12] customer's id. `customer_id` should be passed only when the off-chain id is know to the client, but the relationship between this id and the user's Stellar account is not known to the Anchor.                                   |
+| `location_id`        | string                  | (optional) id of the chosen location for cash pick up cash                                                                                                                                                                                                                                                                                                                                                                                                            |
```
JakeUrban, line 729:
```suggestion
| `location_id`        | string                  | (optional) id of the chosen location to pick up cash                                                                                                                                                                                                                                                                                                                                                                                                            |
```

JakeUrban (dismissed):
Approved with edits

JakeUrban:
Also, don't forget to update the updated-at and version fields.

JakeUrban (approved):

PR #1434, opened 2024-01-30
Add the start of a CAP draft for secp256r1 in the Soroban host interface

leighmcculloch:
### What
Add the start of a CAP draft for secp256r1 in the Soroban host interface

### DIFF
--- a/core/cap-0051.md
+++ b/core/cap-0051.md
@@ -18,14 +18,14 @@ interface.
 
 ## Motivation
 
-Secp256r1, also commonly known as prime256r1 or ecdsa p256, is a common elliptic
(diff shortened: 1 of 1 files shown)


core/cap-0051.md
```diff
@@ -0,0 +1,145 @@
...
+
+## Specification
+
+### XDR Changes
+
+None.
+
+### Host Function Changes
+
+```diff mddiffcheck.ignore=true mddiffcheck.base=v20.1.0
```
leighmcculloch, line 52: I'll have to come back and make it possible for this check to work, but right now it won't which is why it is set to ignore.

tomerweller (approved):

jacekn (approved):

PR #1443, opened 2024-02-15
CAP-52: Add first draft for base64 on Soroban

leighmcculloch:
### What
Add first draft for base64 on Soroban, in CAP-52.

### Why
...


core/cap-0052.md
```diff
@@ -0,0 +1,194 @@
...
+@@ -1890,6 +1890,38 @@
+           ],
+           "return": "U32Val",
+           "docs": "Return the index of a Symbol in an array of linear-memory byte-slices, or trap if not found."
++        },
++        {
++          "export": "n",
++          "name": "base64_encode",
++          "args": [
++            {
```
jayz22, line 90: It should include the `input: BytesObject`? Same for the decode function.

leighmcculloch: Ah good catch. I'll add.

leighmcculloch: Fixed thanks.

tomerweller (approved):

PR #1448, opened 2024-02-26
Improve SEP-006

christian-rogobete:
1. **fix json examples**

- remove trailing commas that break the json parser
- remove `min_amount` and `max_amount` from `withdrawal-exchange` response

2. **extend withdrawal and withdrawal-exchange request**

- add `email_address` field analogous to the deposit and deposit-exchange request

3. **fix some typos**


Ifropc:
Hey @christian-rogobete thanks for the PR, I totally missed it. 
Can't we just pass email address as part of SEP-9 field and use it for notification purposes?

christian-rogobete:
Hello @Ifropc , thank you very much for your message. 

I have noticed that the `email_address` field is currently present in the `deposit` request but not in the `withdraw` request. I have added it there for the sake of completeness.
 
At first I was also wondering why it is not used as part of SEP-09/SEP-12 and I had the following thoughts:

It may be that the anchor does not specify the email field as mandatory in SEP-12 and therefore it is not present. Furthermore, if the user wants to receive notifications only for this specific case (SEP-06), he can enter the email address in the request and receive the notifications only for the executed action (deposit or withdrawal). 

I think we should either delete the  `email_address` field from the `deposit` request, or add it to the `withdraw` request as well. 

What would perhaps be even better is to create a new SEP, through which the user can determine in fine granularity what he wants to receive notifications for. He could also specify how he wants to receive the notifications (e.g. email, phone, callback, etc.). What do you think?

Ifropc:
Hey @christian-rogobete I believe anchor can make passing the email address mandatory via SEP-12 and use it to send notifications. User may unsubscribe from email chain if they want to using the standard email functionality. 
For now, I don't see the use for this.
Thank you for suggesting the change!

christian-rogobete:
Hey @Ifropc, thank you for letting me know. Sounds good. I closed the issue because I am unsure if `min_amount` and `max_amount` can be part of a `withdraw-exchange` (and  `deposit-exchange`) asset in the `info` response.

The text suggests that they are not, but they are used in the example. I will reintegrate them into the SDKs to make sure that nothing is missing. They are also used in the Anchor Platform.

Issue #1451, opened 2024-02-29
Support for `memo` field in SEP-9 Financial Account Fields


ydag:
### Issue
The current SEP-9 Financial Account Fields lacks a `memo` field, hindering the inclusion of vital tag/memo information for regular bank transfers in deposit instructions.

### Proposal
Add a `memo` field to SEP-9 Financial Account Fields to facilitate the inclusion of necessary tag/memo details.

PR #1452, opened 2024-02-29
SEP-9 : Add `external_transfer_memo` field to Financial Account Fields

ydag:
Resolving the issue #1451

### DIFF
--- a/ecosystem/sep-0009.md
+++ b/ecosystem/sep-0009.md
@@ -109,6 +109,7 @@ applications use fields that are the most familiar, which are often specific to 
 | `bank_number`         | string | Number identifying bank in national banking system (routing number in US)      |
 | `bank_phone_number`   | string | Phone number with country code for bank                                        |
 | `bank_branch_number`  | string | Number identifying bank branch                                                 |
+| `memo`                | string | A destination tag/memo used to identify a transaction                          |
 | `clabe_number`        | string | Bank account number for Mexico                                                 |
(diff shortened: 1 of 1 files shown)


ydag:
Hi @JakeUrban,

After this change how do you feel about deprecating `crypto_memo`? I think `memo` can be used both regular bank and crypto transfers.

ecosystem/sep-0009.md
```diff
@@ -109,6 +109,7 @@ applications use fields that are the most familiar, which are often specific to
 | `bank_number`         | string | Number identifying bank in national banking system (routing number in US)      |
 | `bank_phone_number`   | string | Phone number with country code for bank                                        |
 | `bank_branch_number`  | string | Number identifying bank branch                                                 |
+| `memo`                | string | A destination tag/memo used to identify a transaction                          |
```
Ifropc, line 112: WDYT about renaming it to `external_transfer_memo`? 
So it's not confused with Stellar's transaction memo

ydag: That sounds good! And what do you think about deprecation of `crypto_memo` in that case? We can use `external_transfer_memo` for both regular bank and crypto transfers, right? @Ifropc @marcinx

ydag: Cool! It is updated.

Ifropc (dismissed):

JakeUrban (approved):

Ifropc:
@ydag could you update version and changelog of the SEP-9 file?

Ifropc (approved):

Ifropc:
Thanks for proposing this change @ydag 🎉

PR #1468, opened 2024-03-20
Add GH action to check prettier style on SEP files

Ifropc:
1. Add GH actions to check prettier style on all SEP files
2. Run prettier on all SEP files to pass the added check


JakeUrban (approved):

.github/workflows/prettier.yml
```diff
@@ -0,0 +1,20 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.head_ref }}
+
+      - name: Check code
+        uses: creyD/prettier_action@v4.3
```
leighmcculloch, line 18: Can the @v4.3 be changed to the full git commit sha? See https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions for more details on why.

Ifropc: Thanks Leigh, good point, I didn't consider that!
Fixed in https://github.com/stellar/stellar-protocol/pull/1469

PR #1470, opened 2024-03-20
[PARTNER-275] Add Authorization header for SEP-10 GET /Auth

Ifropc:
In the protocol change, an optional Authorization header was added for `GET <WEB_AUTH_ENDPOINT>` endpoint. The header should contain a signed JWT token (using **ed25519**) with an appropriate key from the request. 
For custodial applications, this is a primary Application key, provided in `account` field. 
For non-custodial, this will be the `SIGNING_KEY` from toml file hosted in the `client_domain`  

The server will validate that the signature is correct, and that URL in the JWT corresponds to the request. It can optionally filter out requests from all clients that are not allowed by the server.


ecosystem/sep-0010.md
```diff
@@ -139,7 +139,75 @@ operation of the challenge is the **Home Domain**, and that the web auth domain
...
+Authorized `WEB_AUTH_ENDPOINT` accepts `Authorization Bearer: <token>`, where token is [JSON Web Token](https://jwt.io/)
+authorizing the request.
+
+Client must specify the following claims:
+
+- `iat` (the time at which the JWT was issued
+  [RFC7519, Section 4.1.6](https://tools.ietf.org/html/rfc7519#section-4.1.6)) — current timestamp (`1530644093`)
+- `exp` (the expiration time on or after which the JWT must not be accepted for processing,
+  [RFC7519, Section 4.1.4](https://tools.ietf.org/html/rfc7519#section-4.1.4)) — a server can pick its own expiration
+  period for the token (`1530730493`)
```
JakeUrban, line 160: This JWT is generated by the client, right? So when you say "a server can pick its own expiration", are you referring to the client? We may want to disambiguate that.

ecosystem/sep-0010.md
```diff
@@ -139,7 +139,75 @@ operation of the challenge is the **Home Domain**, and that the web auth domain
...
+```
+
+This token was signed with the `GCADAIACE6CRWGOB3HJRXIOHMQEUKPNLAUIYFIE26F3OK72RSHGAMARK` key, belonging to a
+hypothetical domain `example-wallet.stellar.org`
+
+Another example of the valid signed token:
+
+```
+eyJhbGciOiJFZERTQSJ9.eyJqdGkiOiI1YWExYTNhYi0zY2Q2LTRkYWYtODQyMi0xYmI5Zjc1NTI1YjEiLCJpYXQiOjE3MTA5NzY5MDYsImV4cCI6MTAzNTA5NzY5MDYsInVybCI6Imh0dHBzOi8vYXV0aC5leGFtcGxlLmNvbS8_YWNjb3VudD1HQ1hYSDZBWUpVVlRER0lIVDQyT1pOTUYzTEhDVjRET0tDWDZISERLV0VDVVpZWERaU1daTjZIUyZtZW1vPTEyMzQ1NjcifQ.Gi7aPIwLBss54DJzKhi9lRlUViCkLbbjT0SQFrG9eGGt7TAGD1VeLfj2LFBCOoCFytXWDufplY4iJk4XG_jwDw
+```
```
JakeUrban, line 192: This token has a `jti` property that is not defined in this standard

ecosystem/sep-0010.md
```diff
@@ -139,7 +139,75 @@ operation of the challenge is the **Home Domain**, and that the web auth domain
...
+##### Verifying request header
+
+When server receives a request with an `Authorization` header, server should validate the token before processing with
+issuing a challenge transaction.  
+First, server must use the same steps of choosing the public key (see above), and verify **ed25519** signature of the
+token. Next, server must verify that:
+
+1. URL's path must match `<WEB_AUTH_ENDPOINT>`
+2. URL's query parameters must match request's parameters. Server should address possible permutations of the query
+   parameters.
```
JakeUrban, line 205: What do you mean by "address possible permutations of the query parameters"?

JakeUrban (commented):
This looks like its 90% there! I added some clarifying questions. Also don't forget to update the version and updated-at field.

ecosystem/sep-0010.md
Ifropc: Yes, I just copy-pasted it 😅  I'll just remove this part

ecosystem/sep-0010.md
Ifropc: I was thinking about if we should add it or not. I think we should allow clients add their own fields or maybe add JTI as optional? IMO it can be useful

ecosystem/sep-0010.md
Ifropc: If the server receives request
```
https://auth.example.com/?account=GCXXH6AYJUVTDGIHT42OZNMF3LHCV4DOKCX6HHDKWECUZYXDZSWZN6HS&memo=1234567
``` 
But in JWT url is
```
https://auth.example.com/?memo=1234567&account=GCXXH6AYJUVTDGIHT42OZNMF3LHCV4DOKCX6HHDKWECUZYXDZSWZN6HS
```
It should not be rejected (not sure how to word it better)

ecosystem/sep-0010.md
JakeUrban: I don't feel strongly either way, but my inclination is to only add fields that satisfy a requirement of what we're trying to solve.

ecosystem/sep-0010.md
JakeUrban: I see, maybe we can avoid this problem by moving the query parameters to actual JWT fields?

So for example, instead of having the server validate the query parameters of the URL, the client can put the account, memo, and client_domain in fields within the JWT?

```js
{
  "iat": "...",
  "exp": "...",
  "url": "...",
  "sub": "G...:12314213",
  "client_domain": "..."
}
```

JakeUrban: Or maybe if we're moving parameters outside the URL, se can use `host` instead of `url`?

Ifropc: I think we can do that. We'll have query parameters, `WEB_AUTH_ENDPOINT` and `host` as part of JWT. Then anchor will validate that query params, endpoint and host are matching

ecosystem/sep-0010.md
```diff
@@ -139,7 +139,81 @@ operation of the challenge is the **Home Domain**, and that the web auth domain
...
+
+Client must specify the following claims:
+
+- `iat` (the time at which the JWT was issued
+  [RFC7519, Section 4.1.6](https://tools.ietf.org/html/rfc7519#section-4.1.6)) — current timestamp (for example
+  `1530644093`)
+- `exp` (the expiration time on or after which the JWT must not be accepted for processing,
+  [RFC7519, Section 4.1.4](https://tools.ietf.org/html/rfc7519#section-4.1.4))
+- `host` domain of the server to which the request is being made for.
+- `path` the path of the request (`<WEB_AUTH_ENDPOINT>`)
```
JakeUrban, line 162: I'm not sure it adds any value to split the host and path, since we really only wanted to exclude query parameters. I think we should continue using the full endpoint representation so we match the toml and `iss` field in the SEP-10 token.

So I think we have a couple options. We can do `path` (including the host), `web_auth_endpoint` to match the toml naming, or `aud` which seems like the [standardized](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.3) name for what we're trying to do here. I think I prefer `aud` but I don't feel strongly.

ecosystem/sep-0010.md
```diff
@@ -139,7 +139,81 @@ operation of the challenge is the **Home Domain**, and that the web auth domain
...
+- All request's parameters (if specified in the request)
```
JakeUrban, line 163: Just to clarify, you'd rather have fields in the JWT match the query parameters of the request? We use a different representation in the actual SEP-10 token and use the standardized `sub` field name.

```js
{
  "account": "G...",
  "memo": "12345123"
}
```
vs.
```js
{
  "sub": "G...:123456"
}
```

ecosystem/sep-0010.md
```diff
@@ -139,7 +139,81 @@ operation of the challenge is the **Home Domain**, and that the web auth domain
...
+
+Client must then correctly sign the payload with appropriate Stellar private key. To choose the private key client
+application should follow this steps:
+
+- If `client_domain` is specified, the token must be signed with the `SIGNING_KEY`. (See
```
JakeUrban, line 168: Lets use the term we defined in the beginning, `*Client Domain Account*` to make sure readers don't confuse the server and client's `SIGNING_KEY`.

JakeUrban (commented):
Sorry for the repeated delay, a couple more questions / ideas

ecosystem/sep-0010.md
Ifropc: Yes, I'd rather to keep it simple and just match query parameters (unlike SEP-10 which is issued and consumed by the same entity, this token is issued by the client application, so I'd rather to make it as obvious as possible)

ecosystem/sep-0010.md
Ifropc: Agree on that

Ifropc: I decided to use `web_auth_endpoint` instead as Java library uses set of strings in `aud`, and I think we should only allow single string there.

ecosystem/sep-0010.md
```diff
@@ -139,7 +139,80 @@ operation of the challenge is the **Home Domain**, and that the web auth domain
...
+```json
+{
+  "alg": "EdDSA"
+}
+```
+
+Example of the valid signed token in base64 format:
+
+```
+eyJhbGciOiJFZERTQSJ9.eyJpYXQiOjE3MTEwNzMyMjAsImV4cCI6MTcxMTA3NDEyMCwiYWNjb3VudCI6IkdDNlVDWFZUQU1ORzVKTE9NWkJTQ05ZWFZTTk5GSEwyM1NKUFlPT0ZKRTJBVllERFMyRkZUNDVDIiwiY2xpZW50X2RvbWFpbiI6ImV4YW1wbGUtd2FsbGV0LnN0ZWxsYXIub3JnIiwiaG9zdCI6ImV4YW1wbGUuY29tIiwicGF0aCI6Ii9zZXAxMC9hdXRoIn0.SJwDQBr_xjvb-Nx23t8xhnxzVZVY_IKCzTtPk0-xHfkse7sIyJK_Aobzq3JYpxZ5F5CpNEfOi4Yu5U4mDWVnDA
```
CassioMG, line 183: it appears this token is missing the `web_auth_endpoint` claim

ecosystem/sep-0010.md
```diff
@@ -139,7 +139,80 @@ operation of the challenge is the **Home Domain**, and that the web auth domain
...
+
+This token was issued for the request URL
+`https://example.com/sep10/auth?account=GC6UCXVTAMNG5JLOMZBSCNYXVSNNFHL23SJPYOOFJE2AVYDDS2FFT45C&client_domain=example-wallet.stellar.org`
+and signed with the `GCADAIACE6CRWGOB3HJRXIOHMQEUKPNLAUIYFIE26F3OK72RSHGAMARK` key, belonging to a hypothetical domain
+`example-wallet.stellar.org`
+
+Another example of the valid signed token:
+
+```
+eyJhbGciOiJFZERTQSJ9.eyJpYXQiOjE3MTEwNzMyMjMsImV4cCI6MTcxMTA3NDEyMywiYWNjb3VudCI6IkdDWFhINkFZSlVWVERHSUhUNDJPWk5NRjNMSENWNERPS0NYNkhIREtXRUNVWllYRFpTV1pONkhTIiwibWVtbyI6IjEyMzQ1NjciLCJob3N0IjoiZXhhbXBsZS5jb20iLCJwYXRoIjoiL3NlcDEwL2F1dGgifQ.wQdzjK2iPiU2C1rpreOEgZalwAdo1nxs2cnD6N1bst4bpwHTnbBE1jU4un0cJjWBMkH_NX3J8gfX082h6VMKDw
```
CassioMG, line 194: also missing the `web_auth_endpoint` claim

ecosystem/sep-0010.md
CassioMG: it appears to be split in `host` and `path`, instead of a single `web_auth_endpoint` claim

ecosystem/sep-0010.md
```diff
@@ -139,7 +139,80 @@ operation of the challenge is the **Home Domain**, and that the web auth domain
...
+
+##### Verifying request header
+
+When server receives a request with an `Authorization` header, server should validate the token before processing with
+issuing a challenge transaction.  
+First, server must use the same steps of choosing the public key (see above), and verify **ed25519** signature of the
+token. Next, server must verify JWT claims:
+
+1. `host` must match server's domain
+2. `path` path must match `<WEB_AUTH_ENDPOINT>`
```
CassioMG, line 209: hummm are we going to use `host` + `path` or just `web_auth_endpoint` ?

ecosystem/sep-0010.md
Ifropc: I'll update it, examples are from previous integration

ecosystem/sep-0010.md
Ifropc: web_auth_endpoint

CassioMG (approved):
Lgtm

JakeUrban (dismissed):

JakeUrban (approved):

PR #1478, opened 2024-04-11
Wallet standard interface proposal

piyalbasu:


ecosystem/sep-0043.md
```diff
@@ -0,0 +1,102 @@
...
+  message: string,    // general description message returned to the client app
+  code: number,       // unique error code
+  ext?: Array<string>  // optional extended details
+}
+
+{
+  getAddress: () => Promise<{address: string;} | Error>,
+  signTransaction: (xdr: string, opts?: { network?: string, networkPassphrase?: string, address?: string }) => Promise<{signedXdr: string} | Error>,
+  signAuthEntry: (authEntryXdr: string, opts?: { network?: string, networkPassphrase?: string, address?: string }) => Promise<{signedAuthEntry: string} | Error>,
+  signMessage: (message: string, opts?: { network?: string, networkPassphrase?: string, address?: string }) => Promise<{signedMessage: string} | Error>,
```
Shaptic, line 44: :bulb: For all of these, the response should just be `Promise<{ signature: string } | Error>`. It's a better reflection of the return value, which is just an opaque signature. Also, we should describe its format: is it hex, base64, etc.? For example, [`Keypair.sign`](https://stellar.github.io/js-stellar-sdk/Keypair.html#sign) just returns raw bytes.

Unless this is implying that you would get e.g. an auth entry that is "filled out" with the signature field inside of it? This isn't explicit in the description below.

piyalbasu: @Shaptic This is a good call out for `signAuthEntry` and `signMessage`. 

`signTransaction` actually returns the signed transaction xdr from Freighter, not just the signautre, so I do think `Promise<{ xdr: string } | Error>` is still correct here. Though it may be useful to also pass just the signature

piyalbasu: Good question on the Buffer encoding piece. I know Albedo encodes as hex. I'll check with the ecosystem to see if they have a preference here

ecosystem/sep-0043.md
```diff
@@ -0,0 +1,106 @@
...
+{
+  get_address: () => Promise<{address: string;} | Error>,
+  
+  sign_transaction: (xdr: string, opts?: { network_passphrase?: string, address?: string; submit?: boolean; submit_url?: string; }) => Promise<{signed_tx_xdr: string; signer_address: string;} | Error>,
+
+  sign_auth_entry: (auth_entry: string, opts?: { network_passphrase?: string, address?: string }) => Promise<{signed_auth_entry: string; signer_address: string;} | Error>,
+
+  sign_message: (message: string, opts?: { network_passphrase?: string, address?: string }) => Promise<{signed_message: string; signer_address: string;} | Error>,
+
+  get_network: () => Promise<{network: string, network_url: string, network_passphrase: string, soroban_rpc_url: string}>
```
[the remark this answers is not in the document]
piyalbasu: Ah, that's a good point. Dapps should be able to target their own Horizon/RPC url's based on the network and what the wallet is configured to is actually kind of irrelevant. I'll change to just `network/network_passphrase`

ecosystem/sep-0043.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Abstract
+
+This standard aims to enhance developer experiences by standardizing an interface for web wallet's to use
+when interacting with other codebases.
+
+## Motivation
+
+There are numerous Stellar wallets in the ecosystem with this number expected to grow as the ecosystem does.
+Wallets expose an API to allow dapps to call methods that preform such common tasks as transmitting a wallet's adress. An issue is that wallets all use slightly different API's to preform similar functions, with different method names and call signature as well as different return values. Developers have a motivation to integrate with as many as possible, but currently, they must create abstractions to handle every wallet they wish to connect to. This proposal presents a standard API interface for all wallets to implement. If all wallets have the same interface, it removes the needs for developers to create abstractions for every individual wallet they want to integrate with. Developers can safely assume a wallet uses this agreed upon interface, creating simpler codepaths that easily connect to current wallets while also simplifying the onboarding of new wallets.
```
CassioMG, line 22:
```suggestion
Wallets expose an API to allow dapps to call methods that preform such common tasks as transmitting a wallet's address. An issue is that wallets all use slightly different API's to preform similar functions, with different method names and call signature as well as different return values. Developers have a motivation to integrate with as many as possible, but currently, they must create abstractions to handle every wallet they wish to connect to. This proposal presents a standard API interface for all wallets to implement. If all wallets have the same interface, it removes the needs for developers to create abstractions for every individual wallet they want to integrate with. Developers can safely assume a wallet uses this agreed upon interface, creating simpler codepaths that easily connect to current wallets while also simplifying the onboarding of new wallets.
```

ecosystem/sep-0043.md
```diff
@@ -0,0 +1,108 @@
...
+
+### get_network
+
+`get_network` provides details about the network that the wallet is currently configured to. This allows a dapp to confirm that they are building a transaction for the correct network.
+
+### Errors
+
+Error messaging provides clear errors to the requester. The `message` field provides an informative, human readable string that a UI could surface to the end user. The 4 types of `code` key dilineate between possible erroes: those caused by an internal wallet error, an error caused by external service (for ex: Horizon or Soroban RPC), invalid user input (for ex: a malformed XDR), and an error caused by the user rejecting the action requested. Finally, the `ext` field is an optional value that can provide more context for the error, if needed.
```
CassioMG, line 103:
```suggestion
Error messaging provides clear errors to the requester. The `message` field provides an informative, human readable string that a UI could surface to the end user. The 4 types of `code` key delineate between possible errors: those caused by an internal wallet error, an error caused by external service (for ex: Horizon or Soroban RPC), invalid user input (for ex: a malformed XDR), and an error caused by the user rejecting the action requested. Finally, the `ext` field is an optional value that can provide more context for the error, if needed.
```

ecosystem/sep-0043.md
```diff
@@ -0,0 +1,108 @@
...
+    code: -2,
+    ext: ['Operation 2: Invalid "amount" value.', "Operation 5: Asset issuer is required."] // malformed tx error example
+  },
+  invalid: { // client app request is invalid (wrong parameters, invalid transaction XDR, etc.)
+    message: 'Request is invalid.',
+    code: -3,
+    ext: ['Invalid transaction XDR'] // example error
+  },
+  rejected: { // user rejected the request, client app should not try to retry the request
+    message: 'Rejected by the user.',
```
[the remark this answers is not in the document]
CassioMG: There is also the localization related to user-facing messages, ideally they should appear on a given language based on where user is located or based on some user config. Would this be part of the scope of this SEP?

ecosystem/sep-0043.md
```diff
@@ -0,0 +1,108 @@
...
+interface Error {
+  message: string,    // general description message returned to the client app
+  code: number,       // unique error code
+  ext?: Array<string>  // optional extended details
+}
+
+{
+  get_address: () => Promise<{address: string;} | Error>,
+
+  sign_transaction: (xdr: string, opts?: { network_passphrase?: string, address?: string; submit?: boolean; submit_url?: string; }) => Promise<{signed_tx_xdr: string; signer_address: string;} | Error>,
```
CassioMG, line 43:
```suggestion
  sign_transaction: (xdr: string; opts?: { network_passphrase?: string; address?: string; submit?: boolean; submit_url?: string; }) => Promise<{signed_tx_xdr: string; signer_address: string;} | Error>,
```

ecosystem/sep-0043.md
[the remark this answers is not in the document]
piyalbasu: This is a good callout. I can make this more clear - `message` should provide users an action that can take to remedy the situation. The contents of the `message` will be defined by the wallet itself. We're really just asking them to group their errors by `unhandled|external|invalid|rejected`.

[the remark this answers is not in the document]
piyalbasu: I think localization is a good callout, as well. A wallet should be localizing the `message` and `ext` field and sending it back. I'll add that to the SEP

ecosystem/sep-0043.md
```diff
@@ -0,0 +1,145 @@
...
+confirm that they are building a transaction for the correct network.
+
+### Errors
+
+Error messaging provides clear errors to the requester. The `message` field provides an informative, human readable
+string that a UI could surface to the end user. This `message` should include actions a user can take to remedy the
+situation. The 4 types of `code` key delineate between possible errors: those caused by an internal wallet error, an
+error caused by external service (for ex: Horizon or Soroban RPC), invalid user input (for ex: a malformed XDR), and an
+error caused by the user rejecting the action requested. Finally, the `ext` field is an optional value that can provide
+more context for the error, if needed. The wallet is reponsible for localizing the `message` and `ext` fields to be
```
CassioMG, line 139:
```suggestion
more context for the error, if needed. The wallet is responsible for localizing the `message` and `ext` fields to be
```

ecosystem/sep-0043.md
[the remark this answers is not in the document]
CassioMG: Looks good, thanks for the changes!

ecosystem/sep-0043.md
```diff
@@ -0,0 +1,146 @@
...
+
+## Abstract
+
+This standard aims to enhance developer experiences by standardizing an interface for web wallet's to use when
+interacting with other codebases.
+
+## Motivation
+
+There are numerous Stellar wallets in the ecosystem with this number expected to grow as the ecosystem does. Wallets
+expose an API to allow dapps to call methods that preform such common tasks as transmitting a wallet's address. An issue
```
Shaptic, line 22: nit: typo
```suggestion
expose an API to allow dapps to call methods that perform common tasks such as transmitting a wallet's address. An issue
```

ecosystem/sep-0043.md
```diff
@@ -0,0 +1,146 @@
...
+is that wallets all use slightly different API's to preform similar functions, with different method names and call
```
Shaptic, line 23:
```suggestion
is that wallets all use slightly different API's to perform similar functions, with different method names and call
```

ecosystem/sep-0043.md
```diff
@@ -0,0 +1,146 @@
...
+signature as well as different return values. Developers have a motivation to integrate with as many as possible, but
```
Shaptic, line 24:
```suggestion
signatures as well as different return values. Developers have a motivation to integrate with as many as possible, but
```

ecosystem/sep-0043.md
```diff
@@ -0,0 +1,146 @@
...
+  signAuthEntry: (authEntry: string, opts?: { networkPassphrase?: string, address?: string }) => Promise<{signedAuthEntry: string; signerAddress: string;} | Error>,
+
+  signMessage: (message: string, opts?: { networkPassphrase?: string, address?: string }) => Promise<{signedMessage: string; signerAddress: string;} | Error>,
+
+  getNetwork: () => Promise<{network: string, networkPassphrase: string}>
+}
+```
+
+In addition, this proposal has a standard for error interfaces. Below are examples of different kinds of errors a wallet
+may surface:
```
Shaptic, line 64: We should heavily emphasize which parts of the errors are standard (i.e. the interface or categories) vs. examples that wallets should fill in themselves (i.e. the messages/codes themselves, mentioning localization etc. might be worthwhile), because it's not super clear to me here which ones are "guaranteed" for an implementor.

ecosystem/sep-0043.md
```diff
@@ -0,0 +1,146 @@
...
+interface Error {
+  message: string,    // general description message returned to the client app
+  code: number,       // unique error code
+  ext?: Array<string>  // optional extended details
+}
+
+{
+  getAddress: () => Promise<{address: string;} | Error>,
+
+  signTransaction: (xdr: string, opts?: { networkPassphrase?: string, address?: string; submit?: boolean; submitUrl?: string; }) => Promise<{signedTxXdr: string; signerAddress: string;} | Error>,
```
Shaptic, line 53: Can we wrap these so they're easier to read in the rendered markdown?

ecosystem/sep-0043.md
```diff
@@ -0,0 +1,146 @@
...
+  }
+}
+
+```
+
+### Methods
+
+### getAddress
+
+`getAddress` will provide the public key the wallet is signing for. This may be the account loaded in the wallet, or it
```
Shaptic, line 94: Always `G...`, right?

ecosystem/sep-0043.md
```diff
@@ -0,0 +1,146 @@
...
+may be a different account that the wallet is the signer for.
+
+### signTransaction
+
+`signTransaction` will take an XDR provided by the requester and apply a signature to it before returning a signed XDR
+and the signer address back to the requester. The first parameter is the XDR string while the second parameter is a set
```
Shaptic, line 100: Suggested rephrase to provide more detail, somewhat applies to everywhere you mention XDR strings

```suggestion
`signTransaction` will take a serialized transaction provided by the requester and apply a signature to it before returning a signed XDR and the signer address back to the requester. The first parameter is a base64-encoded string of an XDR `Transaction` instance (right? or is it the envelope? plus cross-reference to the `.x` file/line itself) while the second parameter is a set
```

Also, does

> signed XDR

mean a `TransactionEnvelope` with a corresponding `DecoratedSignature` appended?

ecosystem/sep-0043.md
```diff
@@ -0,0 +1,146 @@
...
+Similar to its transaction counterpart, `signAuthEntry` will take an XDR provided by the requester and apply a signature
+to it before returning a signature and the signer address back to the requester. The first parameter is the XDR string
+while the second parameter is a set of options. Options allow a user to specify what network they intend the transaction
+to be signed on. They can do so by passing the network's passphrase as `networkPassphrase`. Options also allow a wallet
+to specify which account they're seeking a signature for as `address`. If a wallet holds numerous addresses, it can use
+this param to ensure it is signing with the intended address.
+
+### signMessage
+
+Similar to its transaction and aither entry counterparts, `signMessage` will take an arbitrary message provided by the
```
Shaptic, line 119: typo:
```suggestion
Similar to its transaction and auth entry counterparts, `signMessage` will take an arbitrary message provided by the
```

also, wouldn't it be `Contrary` rather than `Similar` given that it's now arbitrary rather than well-defined or structured data?

Shaptic (commented):
Looks fantastic! Just requesting some clarifications and details. We should aim to make SEPs like this completely comprehensive and answer all of the little details with as many cross-references as necessary to cover the basis of the standard.

ecosystem/sep-0043.md
piyalbasu: Thanks - the lack of spell check in vscode is very evident in this PR 😬

ecosystem/sep-0043.md
piyalbasu: Yes that's correct. I'll specify

ecosystem/sep-0043.md
piyalbasu: I was trying to convey that all 3 are similar in that you're sending _something_ to the wallet, it signs, and then returns the signed version. But I can make that more clear

ecosystem/sep-0043.md
```diff
@@ -0,0 +1,215 @@
...
+`getAddress` will provide the public key (G...) the wallet is signing for. This may be the account loaded in the wallet,
+or it may be a different account that the wallet is the signer for.
+
+### signTransaction
+
+`signTransaction` will take an XDR provided by the requester and apply a signature to it before returning a
+base64-encoded string XDR-encoded
+[Transaction Envelope](https://github.com/stellar/js-stellar-base/blob/5efdff0b2a78ed842f7463340b4ff23eb4370635/types/next.d.ts#L11928)
+with
+[Decorated Signatures](https://github.com/stellar/js-stellar-base/blob/5efdff0b2a78ed842f7463340b4ff23eb4370635/types/next.d.ts#L5882)
```
Shaptic, line 165:
```suggestion
[Decorated Signatures](https://github.com/stellar/js-stellar-base/blob/5efdff0b2a78ed842f7463340b4ff23eb4370635/types/curr.d.ts#L5882)
```

(They're identical today, but referencing `curr` will be more accurate in the future once `next` changes.)

Shaptic (dismissed):
Besides the `next` vs. `curr` comment that applies to all links, this looks great from my perspective!

Shaptic (approved):

PR #1481, opened 2024-04-22
SEP-9 : Add `mobile_number_format` field to Natural Person Fields

CassioMG:
Depending on the integration (e.g the [Stellar Disbursement Platform](https://github.com/stellar/stellar-disbursement-platform-backend/pull/256)) parts may want to hash the mobile number to increase security during the data sharing between wallets and Anchors.

### DIFF
--- a/ecosystem/sep-0009.md
+++ b/ecosystem/sep-0009.md
@@ -54,42 +54,42 @@ top-level key.
 
 ## Natural Person Fields
 
-| Name                          | Type   | Description                                                                                 |
-| ----------------------------- | ------ | ------------------------------------------------------------------------------------------- |
-| `family_name` or `last_name`  | string | Family or last name                                                                         |
-| `given_name` or `first_name`  | string | Given or first name                                                                         |
-| `additional_name`             | string | Middle name or other additional name                                                        |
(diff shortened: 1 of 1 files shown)


reecexlm (dismissed):

Ifropc:
@CassioMG maybe adding a mobile_number format would be a more flexible approach? You can pass the hash in `mobile_number` field but set "hash" in the `mobile_number_format` field to indicate the hash format

CassioMG:
> @CassioMG maybe adding a mobile_number format would be a more flexible approach? You can pass the hash in `mobile_number` field but set "hash" in the `mobile_number_format` field to indicate the hash format

@Ifropc thanks for the suggestion. What would be the possible values for `mobile_number_format` in this case?

Ifropc:
@CassioMG 
I'd suggest it to be a usual String type with suggested values `E164` or `E.164` ("Standard" format which we used to see, e.g. +14155552671 see [wiki](https://en.wikipedia.org/wiki/E.164)), hash (you should add a small comment about this hash format) and I think that's it for now. If this value is not specified, receiver should assume it's in E164 format

Ifropc (approved):

marcelosalloum (approved):
LGTM!

PR #1484, opened 2024-05-08
[PARTNER-329] SEP-6/24: Add new `user_action_required_by` transaction field

Ifropc:
Add optional `user_action_required_by` field. This field should be used to show the time anchors gives for the user to make an action before transaction will automatically expire or be refunded. In certain statuses, such as `pending_user_transfer_start` or `incomplete`, anchor waits for the user before advancing transaction to the next status. Using new `user_action_required_by` field, anchor indicates the deadline of the action to take place. Anchor should specify the action waited on using `message` or `more_info_url`.  `user_action_required_by`  should only be set for the statuses that will lead to anchor automatically setting transaction to the next status (such as `expired`/`refunded`) once the deadline is reached, and should be not set (or set to `null`) in all other cases.

Also added missed on_hold field to SEP-6 transaction


CassioMG (approved):
Great UX improvement

ecosystem/sep-0006.md
```diff
@@ -1426,6 +1426,7 @@ Each object in the `transactions` array should have the following fields:
...
+| `action_required_by`      | UTC ISO 8601 string | (optional) The date and time by when the user action is required. This field must be omitted if no user action is required in the current status. Anchor should automatically move transaction into the next status (such as `expired`) once the deadline has been reached.                                                                                                                                                                                                                                                                                                                                                    |
```
JakeUrban, line 1429: I really like your description of `action_required_by` you wrote for the PR description, can we use that (or parts of it) here?

JakeUrban (commented):
Looks pretty good, just looking to see some more context added to the field description

ecosystem/sep-0006.md
jopmiddelkamp: Can this name be more specific like for example `user_action_required_by`?

Ifropc: That was actually one of the names we were discussing, I don't have any strong preference, so let's rename it so it's a bit more clear!

JakeUrban (approved):

PR #1486, opened 2024-05-14
[SEP-31] Add new `user_action_required_by` field

Ifropc:
New field `user_action_required_by` was added in SEP-24/SEP-6 fields previously in #1484
For consistency, this field should be added to SEP-31 as well


ecosystem/sep-0031.md
```diff
@@ -594,32 +594,33 @@ On success the response should include a `200 OK` HTTP status code and the follo
...
+| `user_action_required_by` | UTC ISO 8601 string | (optional) The date and time by when the user action is required. In certain statuses, such as `pending_customer_info_update`, anchor waits for the user action and `user_action_required_by` field should be used to show the time anchors gives for the user to make an action before transaction will automatically be moved into a different status (such as `expired` or to be `refunded`). `user_action_required_by` should only be specified for statuses where user action is required, and omitted for all other. Anchor should specify the action waited on using message or `more_info_url`. |
```
JakeUrban, line 617: SEP-31 involves two users, how should this field be used in this context? For the sender, for both?

Ifropc: I think we can do it for both users, as it's makes things more simple

PR #1489, opened 2024-06-06
🪲 Fix example phone formatting

JFWooten4:
This example phone number should be in E.164 format. Adjusted accordingly and readably.


ecosystem/sep-0001.md
```diff
@@ -251,7 +251,7 @@ ORG_LOGO="https://www.domain.com/awesomelogo.png"
 ORG_DESCRIPTION="Description of issuer"
 ORG_PHYSICAL_ADDRESS="123 Sesame Street, New York, NY 12345, United States"
 ORG_PHYSICAL_ADDRESS_ATTESTATION="https://www.domain.com/address_attestation.jpg"
-ORG_PHONE_NUMBER="1 (123)-456-7890"
+ORG_PHONE_NUMBER="+16401320789"
```
leighmcculloch, line 254: I agree, this should probably use the E.164 format. If this was to be updated for any reason we probably only need to update the format, not the contents though, so I'd keep it as:
```suggestion
ORG_PHONE_NUMBER="+11234567890"
```

On whether we can update this, or should, I don't have historical context for why this format is the way it is, or if format even matters. This field may just be an arbitrary string.

@JakeUrban, do you have any context?

JakeUrban: The field is [defined](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0001.md#organization-documentation) as a E.164-formatted field, so I think we can merge this change.

JakeUrban (dismissed):

JakeUrban:
@JFWooten4 do you mind running prettier for this file?

ecosystem/sep-0001.md
leighmcculloch: Thanks for catching that. ❤️

JakeUrban (approved):

PR #1492, opened 2024-06-14
SEP-10: Add support for contracts (storage bound)

leighmcculloch:
### What
Add support for contracts to SEP-10 using storage to store a fixed key.

### Why
So that contract accounts can authenticate with SEP-24, SEP-30, etc.

### State
This PR is a early draft and idea for how SEP-10 could be modified.

cc @JakeUrban @tomerweller @philipliu @reecexlm @mtwtfss 

Related:
- https://github.com/stellar/stellar-protocol/pull/1494

### DIFF
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -7,7 +7,7 @@ Author: Sergey Nebolsin <@nebolsin>, Tom Quisel <tom.quisel@gmail.com>, Leigh Mc
 Status: Active
 Created: 2018-07-31
 Updated: 2024-03-20
-Version: 3.4.0
+Version: 3.5.0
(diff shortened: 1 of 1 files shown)


JakeUrban (commented):
👏 this was much simpler than I expected.

To confirm, things that would require updates are:

- SDKs
  - to use the G-zero address and add the additional manage data op when a C-address is passed as the client account
  - fetch the `sep-10-signers` key-val pair from contract storage
- SEP-10 servers
  - need to run an RPC to fetch the contract's signers

leighmcculloch:
Closing in favor of proposing:
- https://github.com/stellar/stellar-protocol/pull/1494.

PR #1494, opened 2024-06-14
SEP-10: Add support for contracts (Soroban Auth bound)

leighmcculloch:
### What
Add support for contracts to SEP-10 using Soroban Auth.

### Why
So that contract accounts can authenticate with SEP-24, SEP-30, etc.

### State
This PR is a early draft and idea for how SEP-10 could be modified.

cc @JakeUrban @tomerweller @philipliu @reecexlm @mtwtfss @dmkozh 

Related:
- https://github.com/stellar/stellar-protocol/pull/1492

### DIFF
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -7,7 +7,7 @@ Author: Sergey Nebolsin <@nebolsin>, Tom Quisel <tom.quisel@gmail.com>, Leigh Mc
 Status: Active
 Created: 2018-07-31
 Updated: 2024-03-20
-Version: 3.5.0
+Version: 4.0.0
 ```
 
 ## Simple Summary
--- a/ecosystem/sep-0010.md
+++ b/ecosystem/sep-0010.md
@@ -7,7 +7,7 @@ Author: Sergey Nebolsin <@nebolsin>, Tom Quisel <tom.quisel@gmail.com>, Leigh Mc
 Status: Active
 Created: 2018-07-31
 Updated: 2024-03-20
-Version: 3.4.0
+Version: 3.5.0
 ```
 
 ## Simple Summary
@@ -33,7 +33,7 @@ It involves the following components:
   document. The server's domain may be the **Home Domain**, a sub-domain of the **Home Domain**, or a different domain.
   - The `SIGNING_KEY` from the **Home Domain** is the **Server Account**.
 - A **Client Account**: the account being authenticated.
-  - A Stellar account (`G...`) or a muxed account (`M...`).
+  - A Stellar account (`G...`) or a muxed account (`M...`) or a contract account (`C...`).
   - If a Stellar account (`G...`), may be accompanied by a memo to scope the authentication to a user or sub-account of
     the account.
 - A **Client**: the software used by the holder of the **Client Account** being authenticated by the **Server**.
@@ -59,9 +59,15 @@ The authentication flow is as follows:
    the transaction isn't malicious.
 1. The **Client** verifies that the transaction is signed by the **Server Account** obtained through discovery flow.
 1. The **Client** verifies that the transaction's first operation is a Manage Data operation that has its:
-   1. Source account set to the **Client Account**
+   1. Source account set to the **Client Account** or the G-zero address
    1. Key set to `<home domain> auth` where the home domain is the **Home Domain**.
    1. Value set to a nonce value.
+1. If the source account is set to the G-zero address, the **Client** verifies that the transaction has an Invoke Host Function operation with:
+   1. Host Function set to `HOST_FUNCTION_TYPE_INVOKE_CONTRACT` with:
+      1. Contract address set to the **Client Account**.
+      1. Function name set to `webauth_verify`.
+      1. Args set to an empty list.
+   1. Auth containing a single Soroban Authorization Entry that was provided by the **Client** in the challenge request.
 1. The **Client** verifies that if the transaction has a Manage Data operation with key `web_auth_domain` that it has:
    1. Source account set to the **Server Account**.
    1. Value set to the **Server**'s domain that the client requested the challenge from.
@@ -84,6 +90,8 @@ The authentication flow is as follows:
 1. If the **Client Account** does not exist (optional):
 1. The **Server** verifies the signature count is two
 1. The **Server** verifies that one signature is correct for the master key of the **Client Account**
+1. If the **Client Account** is the G-zero address:
+1. The **Server** verifies the account by submitting the Invoke Host Function operation to a Stellar RPC for simulation with auth verification enabled and checking that auth succeeds in simulation.
 1. The **Server** verified that the other signature is from the **Server Account**
 1. If the transaction has a Manage Data operation with key `client_domain`, the **Server** verifies that the source
    account of the operation signed the transaction and includes an additional `client_domain` claim in the JWT included
@@ -216,6 +224,7 @@ should respond with an error.
 | Name            | Type          | Description                                                                                                                                                                                                                                                                                                                                                                                                                |
 | --------------- | ------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | `account`       | `G...` string | The **Client Account**, which can be a stellar account (`G...`) or muxed account (`M...`) that the **Client** wishes to authenticate with the **Server**.                                                                                                                                                                                                                                                                  |
+| `auth`       | XDR string | (optional) The Soroban Authorization Entry encoded as XDR base64. Required when the `account` is a `C...` address.                                                                                                                                                                                                                                                                  |
 | `memo`          | string        | (optional) The memo to attach to the challenge transaction. Only permitted if a Stellar account (`G...`) is used. The memo must be of type `id`. Other memo types are not supported. See the [Memo](#memos) section for details.                                                                                                                                                                                           |
 | `home_domain`   | string        | (optional) a **Home Domain**. Servers that generate tokens for multiple **Home Domain**s can use this parameter to identify which home domain the **Client** hopes to authenticate with. If not provided by the **Client**, the **Server** should assume a default for backwards compatibility with older **Clients**.                                                                                                     |
 | `client_domain` | string        | (optional) a **Client Domain**. Supplied by **Clients** that intend to verify their domain in addition to the **Client Account**. See [Verifying the Client Domain](#verifying-the-client-domain). **Servers** should ignore this parameter if the **Server** does not support **Client Domain** verification, or the **Server** does not support verification for the specific **Client Domain** included in the request. |
@@ -475,6 +484,27 @@ with less than any threshold may allow a third-party to authenticate.
 
 A **Server** that needs to support validating accounts that do not exist can require a signature of the master key of
 the account address for accounts that do not exist.
+
+### Verifying Contract Accounts
+
+A **Server** can verify a contract account if the contract has a `webauth` function that has a `require_auth` called on its own address.
+
+When requesting the challenge the **Client** must include the `auth` parameter containing a prebuilt and presigned Soroban Authorization Entry to authorize a call to the contracts `webauth` function.
+
+The Soroban Authorization Entry should contain:
+1. Credentials set to a Soroban Address Credentials with:
+   1. Address set to the **Client Account**.
+   1. Nonce set to a random value.
+   1. Signature Expiration Ledger set to a value in the near future to define the length of time of validatity of the signature.
+   1. Signature set to the values the account requires to authenticate, as defined by the contract.
+1. Root invocation with:
+   1. Host Function set to `HOST_FUNCTION_TYPE_INVOKE_CONTRACT` with:
+      1. Contract address set to the **Client Account**.
+      1. Function name set to `webauth_verify`.
+      1. Args set to an empty list.
+   1. Sub-invocations set as required.
+
+The server will consider the contract authenticated and authorized if the simulation of invoking its `webauth` function succeeds.
 
 ### Verifying the Client Domain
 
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0010.md
```diff
@@ -59,9 +59,15 @@ The authentication flow is as follows:
...
 1. The **Client** verifies that the transaction is signed by the **Server Account** obtained through discovery flow.
 1. The **Client** verifies that the transaction's first operation is a Manage Data operation that has its:
-   1. Source account set to the **Client Account**
+   1. Source account set to the **Client Account** or the G-zero address
    1. Key set to `<home domain> auth` where the home domain is the **Home Domain**.
    1. Value set to a nonce value.
+1. If the source account is set to the G-zero address, the **Client** verifies that the transaction has an Invoke Host Function operation with:
+   1. Host Function set to `HOST_FUNCTION_TYPE_INVOKE_CONTRACT` with:
+      1. Contract address set to the **Client Account**.
+      1. Function name set to `webauth_verify`.
```
leighmcculloch, line 68:
```suggestion
      1. Function name set to `web_auth_verify`.
```

ecosystem/sep-0010.md
```diff
@@ -476,6 +485,27 @@ with less than any threshold may allow a third-party to authenticate.
...
+The Soroban Authorization Entry should contain:
+1. Credentials set to a Soroban Address Credentials with:
+   1. Address set to the **Client Account**.
+   1. Nonce set to a random value.
+   1. Signature Expiration Ledger set to a value in the near future to define the length of time of validatity of the signature.
+   1. Signature set to the values the account requires to authenticate, as defined by the contract.
+1. Root invocation with:
+   1. Host Function set to `HOST_FUNCTION_TYPE_INVOKE_CONTRACT` with:
+      1. Contract address set to the **Client Account**.
+      1. Function name set to `webauth_verify`.
```
leighmcculloch, line 503:
```suggestion
      1. Function name set to `web_auth_verify`.
```

ecosystem/sep-0010.md
```diff
@@ -476,6 +485,27 @@ with less than any threshold may allow a third-party to authenticate.
 A **Server** that needs to support validating accounts that do not exist can require a signature of the master key of
 the account address for accounts that do not exist.
 
+### Verifying Contract Accounts
+
+A **Server** can verify a contract account if the contract has a `webauth` function that has a `require_auth` called on its own address.
+
+When requesting the challenge the **Client** must include the `auth` parameter containing a prebuilt and presigned Soroban Authorization Entry to authorize a call to the contracts `webauth` function.
```
leighmcculloch, line 492:
```suggestion
When requesting the challenge the **Client** must include the `auth` parameter containing a prebuilt and presigned Soroban Authorization Entry to authorize a call to the contracts `web_auth_verify` function.
```

ecosystem/sep-0010.md
```diff
@@ -476,6 +485,27 @@ with less than any threshold may allow a third-party to authenticate.
 A **Server** that needs to support validating accounts that do not exist can require a signature of the master key of
 the account address for accounts that do not exist.
 
+### Verifying Contract Accounts
+
+A **Server** can verify a contract account if the contract has a `webauth` function that has a `require_auth` called on its own address.
```
leighmcculloch, line 490:
```suggestion
A **Server** can verify a contract account if the contract has a `web_auth_verify` function that has a `require_auth` called on its own address.
```

ecosystem/sep-0010.md
```diff
@@ -59,9 +59,15 @@ The authentication flow is as follows:
...
-   1. Source account set to the **Client Account**
+   1. Source account set to the **Client Account** or the G-zero address
    1. Key set to `<home domain> auth` where the home domain is the **Home Domain**.
    1. Value set to a nonce value.
+1. If the source account is set to the G-zero address, the **Client** verifies that the transaction has an Invoke Host Function operation with:
+   1. Host Function set to `HOST_FUNCTION_TYPE_INVOKE_CONTRACT` with:
+      1. Contract address set to the **Client Account**.
+      1. Function name set to `webauth_verify`.
+      1. Args set to an empty list.
+   1. Auth containing a single Soroban Authorization Entry that was provided by the **Client** in the challenge request.
```
leighmcculloch, line 70:
```suggestion
   1. Auth containing one or more Soroban Authorization Entries that were provided by the **Client** in the challenge request.
```

ecosystem/sep-0010.md
```diff
@@ -216,6 +224,7 @@ should respond with an error.
...
 | `account`       | `G...` string | The **Client Account**, which can be a stellar account (`G...`) or muxed account (`M...`) that the **Client** wishes to authenticate with the **Server**.                                                                                                                                                                                                                                                                  |
+| `auth`       | XDR string | (optional) The Soroban Authorization Entry encoded as XDR base64. Required when the `account` is a `C...` address.                                                                                                                                                                                                                                                                  |
```
leighmcculloch, line 227:
```suggestion
| `auth`       | XDR string | (optional) An XDR var-array of Soroban Authorization Entries encoded as XDR base64. Required when the `account` is a `C...` address.                                                                                                                                                                                                                                                                  |
```

ecosystem/sep-0010.md
```diff
@@ -476,6 +485,27 @@ with less than any threshold may allow a third-party to authenticate.
 A **Server** that needs to support validating accounts that do not exist can require a signature of the master key of
 the account address for accounts that do not exist.
 
+### Verifying Contract Accounts
+
+A **Server** can verify a contract account if the contract has a `webauth` function that has a `require_auth` called on its own address.
+
+When requesting the challenge the **Client** must include the `auth` parameter containing a prebuilt and presigned Soroban Authorization Entry to authorize a call to the contracts `webauth` function.
+
+The Soroban Authorization Entry should contain:
```
leighmcculloch, line 494:
```suggestion
The Soroban Authorization Entries should contain at least one entry that has:
```

ecosystem/sep-0010.md
```diff
@@ -476,6 +485,27 @@ with less than any threshold may allow a third-party to authenticate.
...
+   1. Nonce set to a random value.
+   1. Signature Expiration Ledger set to a value in the near future to define the length of time of validatity of the signature.
+   1. Signature set to the values the account requires to authenticate, as defined by the contract.
+1. Root invocation with:
+   1. Host Function set to `HOST_FUNCTION_TYPE_INVOKE_CONTRACT` with:
+      1. Contract address set to the **Client Account**.
+      1. Function name set to `webauth_verify`.
+      1. Args set to an empty list.
+   1. Sub-invocations set as required.
+
```
leighmcculloch, line 506:
```suggestion

The auth entries may contain other necessary auths for sub-invocations if they are present.

```

ecosystem/sep-0010.md
```diff
@@ -476,6 +485,27 @@ with less than any threshold may allow a third-party to authenticate.
...
+The server will consider the contract authenticated and authorized if the simulation of invoking its `webauth` function succeeds.
```
leighmcculloch, line 507:
```suggestion
The server will consider the contract authenticated and authorized if the simulation of invoking its `web_auth_verify` function succeeds.

It is the clients responsibility to ensure the `web_auth_verify` succeeds only if a valid and appropriate auth and signature are provided alongside it's invocation.

It is the **Servers** and Soroban Env's responsibility to ensure that if the `web_auth_verify` function is invoked with auth entries that are unused, or missing auth entries that the invocation requires using `require_auth`, that the simulation and auth fails.
```

ecosystem/sep-0010.md
```diff
@@ -59,9 +59,15 @@ The authentication flow is as follows:
    the transaction isn't malicious.
 1. The **Client** verifies that the transaction is signed by the **Server Account** obtained through discovery flow.
 1. The **Client** verifies that the transaction's first operation is a Manage Data operation that has its:
-   1. Source account set to the **Client Account**
+   1. Source account set to the **Client Account** or the G-zero address
    1. Key set to `<home domain> auth` where the home domain is the **Home Domain**.
    1. Value set to a nonce value.
+1. If the source account is set to the G-zero address, the **Client** verifies that the transaction has an Invoke Host Function operation with:
+   1. Host Function set to `HOST_FUNCTION_TYPE_INVOKE_CONTRACT` with:
```
dmkozh, line 66: Operation and transaction are irrelevant for the custom account signer. I think it would make more sense to specify what is the invocation that server has to build and use the auth entry with (i.e. only a signed SorobanAuthEntry is actually necessary to be provided)

ecosystem/sep-0010.md
```diff
@@ -476,6 +485,33 @@ with less than any threshold may allow a third-party to authenticate.
...
+
+The Soroban Authorization Entries should contain at least one entry that has:
+1. Credentials set to a Soroban Address Credentials with:
+   1. Address set to the **Client Account**.
+   1. Nonce set to a random value.
+   1. Signature Expiration Ledger set to a value in the near future to define the length of time of validatity of the signature.
+   1. Signature set to the values the account requires to authenticate, as defined by the contract.
+1. Root invocation with:
+   1. Host Function set to `HOST_FUNCTION_TYPE_INVOKE_CONTRACT` with:
+      1. Contract address set to the **Client Account**.
```
dmkozh, line 502: I'm a bit confused; is the idea here for the account contract to implement `web_auth_verify` function? It would be sufficient to call a well-known contract that has a function like
```rust
fn web_auth_verify(addr: Address) {
  addr.require_auth();
}
```
The upside is that most custom accounts would work fine without adding a special codepath. The downside is the additional risk of needing to deal with state archival, but OTOH the same is true for the account contract and its data deps as well (which is probably worth specifying here as well - simulation can imitate entry restoration, so that might be ok, but we need to spell that explicitly).

ecosystem/sep-0010.md
```diff
@@ -59,9 +59,15 @@ The authentication flow is as follows:
    the transaction isn't malicious.
 1. The **Client** verifies that the transaction is signed by the **Server Account** obtained through discovery flow.
 1. The **Client** verifies that the transaction's first operation is a Manage Data operation that has its:
-   1. Source account set to the **Client Account**
+   1. Source account set to the **Client Account** or the G-zero address
    1. Key set to `<home domain> auth` where the home domain is the **Home Domain**.
    1. Value set to a nonce value.
+1. If the source account is set to the G-zero address, the **Client** verifies that the transaction has an Invoke Host Function operation with:
```
leighmcculloch, line 65: This part of the process is a bit broken, because the signed auth is provided up front without the server being the one that is allowed to choose the unique and random parts of what gets signed. SEP-10 is designed so that the server defines the random component, and can later verify that it generated the transaction by simply generating it's own signature, and then hand to the client for it to sign. But the client doesn't need to sign at this point because they need to sign the internal auth. So yeah we need to rethink this.

ecosystem/sep-0010.md
leighmcculloch: The intent was to add a codepath so custom accounts have to choose to participate in SEP-10 and can choose to specify a specific signer. But that might be overkill. You're right, it could be some other arbitrary contract call that is being signed for, instead of requiring signature of that specific fn, and then most custom accounts could work without additional code. I'll make that modification.

dmkozh: > The intent was to add a codepath so custom accounts have to choose to participate in SEP-10 and can choose to specify a specific signer. 

Yeah, I don't think that's a reasonable place to opt out as it's possible to come up with SEP-N that allows auth without opt-in. I think opt-in should happen at the wallet level.

ecosystem/sep-0010.md
leighmcculloch: Given this more thought and I agree. The reason I was avoiding changing the entity to an auth entry only is because there's a bunch of data that the transaction and operatons hold, and we need to retain all that information in the contract auth version.

But most of the information can be encoded in the invocation that's being defined. The invocation arguments can be a map of key-values with each of the things that the transaction currently contains, like the `web_auth_domain` which is important for preventing a forwarding vulnerability.

So the flow would be something like this:
```mermaid
sequenceDiagram
Client ->> Server: Get Challenge
Server ->> Server: Build SorobanAuthorizedInvocation<br>(with web_auth_domain,<br> home_domain,<br> random value,<br> authenticating account,<br> etc...)
Server ->> Server: Sign SorobanAuthorizedInvocation
Server ->> Client: Return SorobanAuthorizedInvocation + (Server Sig)
Client ->> Client: Verify SorobanAuthorizedInvocation
Client ->> Client: Sign SorobanAuthorizedInvocation
Client ->> Client: Build SorobanAddressCredentials
Client ->> Server: Get Token<br/>(with SorobanAuthorizedInvocation,<br>Server Sig,<br>SorobanAddressCredentials)
Server ->> Server: Verify Server Sig
Server ->> Server: Verify Client Sig<br>with Simulation
Server ->> Client: Return Token
```

It's convenient today that everything lives inside a Transaction. But I guess that isn't strictly required. And an advantage of not putting it in a transaction is if we do we have to duplicate all the information into the SorobanAuthorizationInvocation to make sure that the client is signing over it all.

ecosystem/sep-0010.md
```diff
@@ -216,6 +224,7 @@ should respond with an error.
...
 | --------------- | ------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | `account`       | `G...` string | The **Client Account**, which can be a stellar account (`G...`) or muxed account (`M...`) that the **Client** wishes to authenticate with the **Server**.                                                                                                                                                                                                                                                                  |
```
accordeiro, line 226: Should we allow for `C...` accounts as the account param too?

ecosystem/sep-0010.md
accordeiro: Can we make it so that if the initial request contain a `C...` address as `account`, the initial server response is a payload with the random params that the client will use to assemble the transaction?

It does cause a bigger divergence between `C` and `G/M` SEP-10 which I don't necessarily love.

leighmcculloch:
Closing to approach the change as a new SEP.

PR #1495, opened 2024-06-25
cap-57 initial draft: State Archival Persistent Entry Eviction

SirTyson:
See https://github.com/stellar/stellar-protocol/discussions/1480 for discussion.


MonsieurNicolas (approved):

anupsdf:
nit: In the `verifyProofOfExistence`, the indentation is slightly off where `expectedHash == r.hash` condition is showing as inside the for loop of `proof.entries`.

anupsdf:
> If an archived entry is restored and the entry currently exists in the Hot Archive, the INITENTRY previously stored in the Hot Archive is overwritten by a RESTORED entry.

How about rephrasing to something like,
If an archived entry is restored and the entry exists in the Hot Archive, then a RESTORED entry is inserted into the Hot Archive bucketlist. This overrides the previously stored INITENTRY in the Hot Archive.

anupsdf:
nit: 4.3% **change** a single miss

PR #1502, opened 2024-07-12
[ANCHOR-719] Update SEP31 to replace SEP12 with async kyc flow

JiahuiWho:
### What's Changed:

**SEP-6, 8, 30** 
Nits and prettify
    
**SEP-12** 
Nits
 
 **SEP-31**
1. KYC completion is no longer required before creating a SEP-31 transaction. Sending anchor can optionally pre-register customers and use the obtained `sender_id` or `receiver_id` to create transaction, or leave these fields empty. 

2. After the transaction is created, the Sending anchor will call `GET /customer?transaction_id=<sep31-txn-id>&type=` to determine the required fields for each type of client, and then make a `PUT` call to the same endpoint to update the information

### DIFF
--- a/ecosystem/sep-0006.md
+++ b/ecosystem/sep-0006.md
@@ -86,16 +86,16 @@ sequenceDiagram
     Anchor-->>-Wallet: transaction id
 
     note right of Wallet: If User has already <br> been accepted, <br> KYC is unnecessary.
-    Wallet->>+Anchor: GET [SEP-12]/customer?transaction_id=<id>
+    Wallet->>+Anchor: GET [SEP-12]/customer?transaction_id=<id>&type=
     Anchor-->>-Wallet: fields required
     Wallet->>+User: requests fields
     User-->>-Wallet: provides field values
-    Wallet->>+Anchor: PUT [SEP-12]/customer?transaction_id=<id>
+    Wallet->>+Anchor: PUT [SEP-12]/customer?transaction_id=<id>&type=
     Anchor-->>Anchor: validates KYC values
     Anchor-->>-Wallet: id, ACCEPTED
 
     Wallet->>+Anchor: GET [SEP-6]/transaction?id=
-    Anchor-->>-Wallet: transaction object containing amount_in, receving account, memo
+    Anchor-->>-Wallet: transaction object containing amount_in, receiving account, memo
     note right of Wallet: Use the info amount_in returned <br> from GET [SEP-6]/transaction?id={id} <br> to make the stellar payment.
 
     Wallet->>+Stellar: submit Stellar payment
--- a/ecosystem/sep-0006.md
+++ b/ecosystem/sep-0006.md
@@ -86,11 +86,11 @@ sequenceDiagram
     Anchor-->>-Wallet: transaction id
 
     note right of Wallet: If User has already <br> been accepted, <br> KYC is unnecessary.
-    Wallet->>+Anchor: GET [SEP-12]/customer?transaction_id=<id>&type=
+    Wallet->>+Anchor: GET [SEP-12]/customer?transaction_id=<id>
     Anchor-->>-Wallet: fields required
     Wallet->>+User: requests fields
     User-->>-Wallet: provides field values
-    Wallet->>+Anchor: PUT [SEP-12]/customer?transaction_id=<id>type=
+    Wallet->>+Anchor: PUT [SEP-12]/customer?transaction_id=<id>
     Anchor-->>Anchor: validates KYC values
     Anchor-->>-Wallet: id, ACCEPTED
 
@@ -1792,7 +1792,7 @@ object containing an error message.
 - `v4.1.0` Add `user_action_required_by` field to transaction responses and add new `on_hold` status
   ([#1484](https://github.com/stellar/stellar-protocol/pull/1484/))
 - `v4.0.0`: Update flow to delegate getting KYC fields to SEP-12
-  ([#1432](https://github.com/stellar/stellar-protocol/pull/1431))
+  ([#1432](https://github.com/stellar/stellar-protocol/pull/1432))
 - `v3.26.0`: Add `location_id` to deposit/withdrawal requests
   ([#1433](https://github.com/stellar/stellar-protocol/pull/1433))
 - `v3.25.0` Add `fee_details` field to the transaction object
@@ -1820,7 +1820,8 @@ object containing an error message.
   ([#1333](https://github.com/stellar/stellar-protocol/pull/1333))
 * `v3.16.0`: Add `refund_memo` and `refund_memo_type` to requests initiating transactions.
   ([#1321](https://github.com/stellar/stellar-protocol/pull/1321))
-* `v3.15.0`: Add `lang` to `/transactions` & `/transaction` parameters, update format to [RFC 4646]. ([#1320](https://github.com/stellar/stellar-protocol/pull/1320))
+* `v3.15.0`: Add `lang` to `/transactions` & `/transaction` parameters, update format to [RFC 4646].
+  ([#1320](https://github.com/stellar/stellar-protocol/pull/1320))
 * `v3.14.0`: Add `quote_id` to the transaction object schema.
   ([#1268](https://github.com/stellar/stellar-protocol/pull/1268))
 * `v3.13.0`: Add callback signature requirement. ([#1262](https://github.com/stellar/stellar-protocol/pull/1262))
--- a/ecosystem/sep-0012.md
+++ b/ecosystem/sep-0012.md
@@ -6,8 +6,8 @@ Title: KYC API
 Author: Interstellar
 Status: Active
 Created: 2018-09-11
-Updated: 2024-02-09
-Version 1.13.0
+Updated: 2024-07-23
+Version 1.14.0
 ```
 
 ## Abstract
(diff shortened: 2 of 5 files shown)


ecosystem/sep-0012.md
```diff
@@ -140,21 +140,17 @@ returned in the response instead.
 
 Different types of customers may have different KYC requirements depending on the action a given customer wants to take.
 The `type` parameter is used to specify the action a customer wants to make so the server can identify what information
-must be collected. SEP-12 does not define what `type` values are accepted and instead leaves that to the other
-protocols, such as [SEP-6](sep-0006.md) and [SEP-31](sep-0031.md), that use SEP-12 for the actions associated with those
-protocols.
+must be collected.
```
philipliu, line 143: Let's leave the part about SEP-31 defining the types in. Otherwise, we have to define SEP-31 types here (line 152), creating coupling that we don't necessarily want.

ecosystem/sep-0012.md
```diff
@@ -140,21 +140,17 @@ returned in the response instead.
...
-receiver would require banking information in order to receive a direct deposit. The `type` parameter could also be used
-to indicate that a customer is an organization instead of a person.
+receiver would require banking information in order to receive a direct deposit.
 
 Note that it is possible for the same customer to have different `status` values for different `type` parameters. For
-example, a customer could have an `ACCEPTED` status for the `small-transaction-amount` `type` parameter but have a
-`NEEDS_INFO` status for the `large-transaction-amount` type. Therefore it is recommended to always pass the `type`
-parameter when making requests to `GET /customer` or `PUT /customer`, even though the field is optional to accomodate
-for implementations that do not require `type`.
+example, a customer could have an `ACCEPTED` status for the `seo31-sender` `type` parameter but have a `NEEDS_INFO`
```
philipliu, line 149: typo: `sep31-sender`.

ecosystem/sep-0031.md
```diff
@@ -103,7 +92,16 @@ sequenceDiagram
 
     Sending Anchor->>+Receiving Anchor: POST [SEP-31]/transactions
     Receiving Anchor-->>Receiving Anchor: checks customer statuses, <br> links quote, <br> creates transaction record
-    Receiving Anchor-->>-Sending Anchor: transaction id, receiving account & memo
+    Receiving Anchor-->>-Sending Anchor: transaction id
+
+    note right of Sending Anchor: If Sending Client or Receiving Client <br >has already been <br> accepted, this section <br> is unnecessary.
```
philipliu, line 97: Should we be looping once for the Sending Client and once for the Receiving Client? If we don't then it looks like this says we can move forward with the transfer when the Sending Client has provided KYC even though we don't know anything about the Receiving Client.

ecosystem/sep-0031.md
```diff
@@ -329,32 +317,7 @@ The response should be a JSON object like:
...
-            "sep31-receiver": {
-              "description": "U.S. citizens receiving USD"
-            },
-            "sep31-foreign-receiver": {
-              "description": "non-U.S. citizens receiving USD"
-            }
-          }
-        }
-      }
+      "max_amount": 1000
```
philipliu, line 320: I think it's possible to use the new KYC flow and keep the `sep12` block right? The new flow works if the SEP-12 request is `type=sep31-large-sender` and `transaction_id=...`. The special types `sep31-sender` and `sep31-receiver` can be hardcoded values we use in the Anchor Platform but we don't need to restrict them at the protocol level. Otherwise, this would be a breaking change for SEP-31 clients.

ecosystem/sep-0031.md
```diff
@@ -857,9 +802,18 @@ In certain cases the Receiving Anchor might need to request updated information
 the bank tells the Receiving Anchor that the provided Receiving Client's name is incorrect or missing a middle initial.
 Since this information was sent via SEP-12, the transaction should go into the `pending_customer_info_update` state
 until the Sending Anchor makes another `SEP-12 PUT /customer` request to update. The Sending Anchor can check which
-fields need to be updated by making a `SEP-12 GET /customer` request including the `id` or `account` & `memo`
-parameters. The Receiving Anchor should respond with a `NEEDS_INFO` status and `last_name` included in the fields
-described.
+fields need to be updated by making a `SEP-12 GET /customer` request including either the customer id `id`, or a
+combination of SEP-31 transaction id `transaction_id` and customer type `type`, which can only be `sep31-sender` or
```
philipliu, line 806: I think we can recommend this way of collecting KYC info, but it doesn't have to be the _only_ way.

philipliu (commented):
This change makes a lot of sense and brings SEP-31 closer to the new KYC standard set by SEP-6, although the current approach would break existing clients. I left a recommendation for how we can avoid this.

ecosystem/sep-0031.md
JiahuiWho: It's possible, but I don't think it's necessary because the sending anchor no longer needs to classify customers.

In the new flow, sending anchors can pre-register the sender and receiver with some basic information. Then, it’s up to the receiving anchor to determine what additional information is needed (such as whether the customer is foreign or a large sender) based on the transaction details and return the required fields to the sending anchor. This way, the receiving anchor can manage its own classified KYC customer decision-making process.

What are your thoughts?

Additionally, since SEP-31 is not currently being adopted by any clients, it's a good opportunity to make these improvements.

ecosystem/sep-0006.md
```diff
@@ -86,11 +86,11 @@ sequenceDiagram
     Anchor-->>-Wallet: transaction id
 
     note right of Wallet: If User has already <br> been accepted, <br> KYC is unnecessary.
-    Wallet->>+Anchor: GET [SEP-12]/customer?transaction_id=<id>&type=
```
Ifropc, line 89: Type should still be passed. Theoretically anchor can use numerical ids for transactions (by type), therefore it'd need to distinguish between sep-6 transaction id N and SEP-31.
In addition, it makes implementation on AP more straightforward (otherwise we need to query both tables to find transaction with this ID)

ecosystem/sep-0012.md
Ifropc: I agree, we should leave this part here

ecosystem/sep-0012.md
```diff
@@ -140,21 +140,17 @@ returned in the response instead.
...
+status for the `sep31-receiver` type.
 
-If the implementor requires the same set of fields for all customers, there is no need for the `type` parameter.
+Currently, the use of the `type` parameter is limited to SEP-31 only, supporting only two types: `sep31-sender` and
```
Ifropc, line 152: This is only true for the anchor platform. We can't change the previous descriptions as there could be different implementations (e.g. Polaris) that define their own types.

ecosystem/sep-0031.md
```diff
@@ -216,21 +214,11 @@ earlier than the quote's expiration.
...
-   Receiving Clients. The response includes the [SEP-9](sep-0009.md) KYC attributes required for registering a client of
-   the associated `type`.
-1. The Sending Anchor collects all required information from the Sending Client. This includes the custom fields listed
-   in the Receiving Anchor's `GET /info` response as well as the KYC fields described in the `GET /customer` responses
-   for both the Sending and Receiving Clients. How the Sending Anchor collects this information from the Sending Client
-   is out of the scope of this document, but the Receving Client should not be required to take any action in order for
-   the Sending Client to make the payment to the Receiving Anchor.
-1. The Sending Anchor makes [SEP-12 PUT /customer](sep-0012.md#customer-put) requests containing the collected
-   information for each client.
+1. Optionally the Sending Anchor can pre-register clients using [SEP-12 PUT /customer](sep-0012.md#customer-put) and
```
Ifropc, line 220: I think it'd be nice to have a bit more detailed description of how it works (basically explaining flow chart above)

ecosystem/sep-0031.md
Ifropc: I don't mind getting rid of custom-define types entirely, but we would need to deprecate it, rather than straight removing

Ifropc (changes requested):

ecosystem/sep-0006.md
JiahuiWho: Sure I can add it back. Just curious, what should the customer type be for SEP-6? Currently, there is no configuration to define a SEP-6 customer type

ecosystem/sep-0012.md
JiahuiWho: Will revert it back

ecosystem/sep-0006.md
Ifropc: Should just pass types as sep-6 and sep-31. We can make recommendation in the standard to follow this rule, but that's generally how AP should be implemented to use the correct table

ecosystem/sep-0031.md
JiahuiWho: I'm also worry about some gaps in this flow with customized customer type.

If the sending anchor create a SEP-31 transaction without pre-register the customers, and is doing kyc now, how would they know what customer `type` they should use in the `GET /cutomer?transaction_id=<>&type=<> ` call?

Ifropc: It's up for the anchor to decide, generally. Client is only supposed to either use `sep-6` or `sep-31` when interacting with business that uses AP

philipliu (commented):
The changes look good. We should also update the SEP-31 version and changelog.

Ifropc (commented):
LGTM. 
Could you update versions, run formatting (`./bin/prettier.sh`) and write a comment that for integrations using Anchor Platform `type` should be `sep6` and `sep31`? This would be a guideline for wallets who sometimes look at this spec.

JiahuiWho:
> LGTM. Could you update versions, run formatting (`./bin/prettier.sh`) and write a comment that for integrations using Anchor Platform `type` should be `sep6` and `sep31`? This would be a guideline for wallets who sometimes look at this spec.

How to distinguish `sender` from `receiver` then? Or is this something up for anchor to decide?

Ifropc:
The anchor that receives KYC should know if it's receiving or sending anchor based on transaction id I think

philipliu (approved):

PR #1515, opened 2024-08-01
[SEP-31] Delegate Kyc flow to SEP-12 using `transaction_id`

JiahuiWho:
As title

PR #1518, opened 2024-08-06
Update prettier job

Ifropc:
We saw some failures with original job in https://github.com/stellar/stellar-protocol/pull/1515#event-13782930566
Which shouldn't fail (and doesn't fail locally), I think it's something with `prettier-action` so replaced it with more straightforward approach of just calling yarn script

### DIFF
--- a/.github/workflows/prettier.yml
+++ b/.github/workflows/prettier.yml
@@ -1,4 +1,3 @@
-# https://github.com/marketplace/actions/prettier-action
 name: 'Prettier for SEP files'
 
 on:
@@ -7,12 +6,10 @@ on:
 jobs:
   prettier:
     runs-on: ubuntu-latest
-
     steps:
-      - name: Checkout
-        uses: actions/checkout@v3
-
-      - name: Check code
-        uses: creyD/prettier_action@31355f8eef017f8aeba2e0bc09d8502b13dbbad1
+      - uses: actions/checkout@v3
+      - name: sep-check
+        uses: DerYeger/yarn-setup-action@18fb87f9a712d8cc8ad1e7407a06470b397b4a7d
         with:
-          prettier_options: --check ecosystem/*.md
+          node-version: 20
+      - run: yarn sep-check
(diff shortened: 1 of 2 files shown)


.github/workflows/prettier.yml
```diff
@@ -7,12 +6,10 @@ on:
...
-
     steps:
-      - name: Checkout
-        uses: actions/checkout@v3
-
-      - name: Check code
-        uses: creyD/prettier_action@31355f8eef017f8aeba2e0bc09d8502b13dbbad1
+      - uses: actions/checkout@v3
+      - name: sep-check
+        uses: DerYeger/yarn-setup-action@18fb87f9a712d8cc8ad1e7407a06470b397b4a7d
```
leighmcculloch, line 12: This action appears to add little functionality. I think there's some value in keeping the dependencies light, especially ones that don't add much value. We could use the GitHub maintained `actions/setup-node` that also automatically installs yarn, and then we just need to have two `-run` steps, for `yarn install --frozen-lockfile` and `yarn sep-check`.

leighmcculloch (approved):
One suggestion, but looks good.

.github/workflows/prettier.yml
Ifropc: SG, I will add a follow-up PR. I'm not familiar with node ecosystem tbh so I just added the first one I found

leighmcculloch:
This PR auto merged with the comment outstanding. I'm fixing the GitHub setup so that the auto merge doesn't kick in if there's an outstanding comment:
- https://github.com/stellar/terraform/pull/2750

.github/workflows/prettier.yml
Ifropc: https://github.com/stellar/stellar-protocol/pull/1519

copiden:
> We

SEP: 0024
Title: Hosted Deposit and Withdrawal
Author: SDF
Status: Active
Created: 2019-09-18
Updated: 2024-05-07
Version 3.7.0

PR #1520, opened 2024-08-07
SEP-24: Make `amount_out` optional in transaction responses

philipliu:
SEP-24 supports indicative quotes. `amount_out` should not be set before users transfer funds.


JakeUrban (dismissed):
I think some additional context in the field description may be useful.

ecosystem/sep-0024.md
```diff
@@ -1077,7 +1077,7 @@ Each object in the `transactions` array should have the following fields:
...
+| `amount_out`              | string              | (optional) Amount sent by anchor to user at end of transaction as a string with up to 7 decimals. Excludes amount converted to XLM to fund account and any external fees.                                                                                                                                                                                                                                                                                                                                                                                                                                              |
```
Ifropc, line 1080: We should definitely clarify that it's only optional if quote_id is set, and the quote was indicative

ecosystem/sep-0024.md
```diff
@@ -1077,7 +1077,7 @@ Each object in the `transactions` array should have the following fields:
...
+| `amount_out`              | string              | (optional) Amount sent by anchor to user at end of transaction as a string with up to 7 decimals. Excludes amount converted to XLM to fund account and any external fees. This field is not set only when there is an indicative quote associated with this transaction.                                                                                                                                                                                                                                                                                                                                               |
```
JakeUrban, line 1080: The current wording implies the field shouldn't ever be set for transactions using indicative quotes. My suggestion is a bit wordy but I think its important to highlight that the field should always be set eventually.

```suggestion
| `amount_out`              | string              | (optional) Amount sent by anchor to user at end of transaction as a string with up to 7 decimals. Excludes amount converted to XLM to fund account and any external fees. This field should be set as soon as the anchor can calculate it using `amount_in`, the exchange rate between `amount_in_asset` and `amount_out_asset`, and `fee.total`.                                                                                                                                                                                                                                                                                                                                               |
```

ecosystem/sep-0024.md
```diff
@@ -1391,8 +1391,6 @@ There is a small set of changes when upgrading from SEP-6 to SEP-24.
 1. Removed `external_extra` transaction property since this should all live in a human readable ` more_info_url`.
 1. Changed the response of the deposit and withdraw endpoints from 403 to 200 since this is the expected flow.
 1. `/transactions` and `/transaction` are now required endpoints.
-1. `Transaction` properties `more_info_url, amount_in, amount_out, amount_fee, and stellar_transaction_id` are now
-   non-optional.
```
JakeUrban, line 1395: This `Changes from SEP-6` section isn't comprehensive anymore and its unlikely to be useful given most businesses implement one or the other rather than migrate from SEP-6 to SEP-24.

This is out of scope of the current PR but at some point we should clean up the spec so its more focused.

ecosystem/sep-0024.md
philipliu: Thanks!

ecosystem/sep-0024.md
```diff
@@ -6,8 +6,8 @@ Title: Hosted Deposit and Withdrawal
 Author: SDF
 Status: Active
 Created: 2019-09-18
-Updated: 2024-05-07
-Version 3.7.0
+Updated: 2024-08-07
+Version 3.7.1
```
Ifropc, line 10: Should it be 3.8.0 ?

Ifropc (approved):

PR #1529, opened 2024-08-20
Add CAP for wasmi register machine.

graydon:
This is a short CAP about the Wasmi register machine. It's mostly not perceptible to users but will alter costs some.


tomerweller (commented):
Seems safe to tag it CAP-0060, no?

tomerweller (approved):

PR #1541, opened 2024-09-04
SEP-44, CAP-61: Move memo into fifth event topic

leighmcculloch:
### What
Move memo into fifth event topic field in SEP-44 and CAP-61.

### Why
The fourth event topic already being used in the built-in token implementation is problematic as it causes existing applications already ingesting transfer events to need to change which topic field they expect the sep11 asset name to be in.

Moving the memo to the fifth topic field is a small change that increases the chance SEP-44 is a forward compatible change.

### DIFF
--- a/core/cap-0061.md
+++ b/core/cap-0061.md
@@ -63,7 +63,7 @@ The Stellar Asset Contract interface is extended with one function:
 ///
 /// # Events
 ///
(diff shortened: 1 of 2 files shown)


tomerweller (approved):
makes sense

PR #1560, opened 2024-10-07
[SEP-31] Add `methods` to info response and update `fields` desc

JiahuiWho:
1. This is part of the ticket https://stellarorg.atlassian.net/browse/ANCHOR-809, where `methods` were introduced to assets config to represents the deposit methods supported by anchor. To align the config and info response across SEPs, adding `methods` to SEP-31 info response

2. Also adjust the desc of `fields` to reflect the change.

3. Minor fix of broken `pending-transaction-info-update` reference

### DIFF
--- a/ecosystem/sep-0031.md
+++ b/ecosystem/sep-0031.md
@@ -6,8 +6,8 @@ Title: Cross-Border Payments API
 Author: SDF
 Status: Active
 Created: 2020-04-07
-Updated: 2024-07-23
-Version 3.0.0
+Updated: 2024-10-07
+Version 3.1.0
 ```
 
 ## Simple Summary
@@ -249,7 +249,7 @@ earlier than the quote's expiration.
 1. If `error`, the Receiving Anchor should be contacted to resolve the situation.
 1. If `pending_transaction_info_update`, the `transaction.fields` values collected from the Sending Client were invalid
    and must be corrected by the Sending Client. See the
-   [Pending Transaction Info Update](#pending-transaction-info-update) section for more information.
+   [Pending Transaction Info Update](#pending-transaction-info-update-deprecated) section for more information.
 1. If `pending_customer_info_update`, the SEP-9 KYC values collected were invalid and must be corrected by the Sending
    Client. See the [Pending Customer Info Update](#pending-customer-info-update) section for more information.
 1. After providing the Receiving Anchor with updated values, the status should ultimately change to `completed`.
@@ -326,6 +326,13 @@ The response should be a JSON object like:
       "fee_percent": 1,
       "min_amount": 0.1,
       "max_amount": 1000,
+      "fields": {
+        "type": {
+          "description": "methods supported by the anchor for asset deposits",
+          "choices": ["SEPA", "SWIFT"],
+          "optional": false
+        }
+      },
       "sep12": {
         "sender": {
           "types": {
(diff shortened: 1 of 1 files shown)


JakeUrban (commented):
SEP-6 uses the `fields` object provided in `GET /info` responses to inform clients what the supported values are for the `type` request parameter in `GET /deposit-exchange` and `GET /withdraw-exchange`.

There is no `type` parameter in SEP-31's `POST /transactions` endpoint, so the `fields` object currently doesn't serve a purpose. I think in order for this to be valuable, we need to add a `type` parameter to this endpoint.

JiahuiWho:
Can we use the fields object just for display purposes? Even without a `type` parameter in SEP-31, it's meaningful to inform clients about the deposit types supported by the anchor. Otherwise the `methods` is not being used anywhere in the anchor platform either.

I remember that `fee_fixed` and `fee_percent` used to be included in the info response, and they are not being used at all (https://github.com/stellar/java-stellar-anchor-sdk/pull/1483). I wonder if those two are for informational use as well?

JakeUrban:
Clients can already use SEP-38 to see what an anchor's supported buy and sell methods are, so providing that information in the SEPs `/info` endpoints is already a bit redundant. Thats also the reason why we don't support `fee_fixed` and `fee_percent` in the anchor platform -- SEP-38 provides better support for communicating fee details.

philipliu:
> Can we use the fields object just for display purposes? Even without a type parameter in SEP-31, it's meaningful to inform clients about the deposit types supported by the anchor. Otherwise the methods is not being used anywhere in the anchor platform either.

I think we need the client to provide the `type`, otherwise the anchor would not know what KYC fields to request.

ecosystem/sep-0031.md
```diff
@@ -376,7 +383,7 @@ the corresponding off-chain asset, after fees have been applied.
...
 | `sender_sep12_type`   | string  | (**deprecated**, optional) The value of the `type` parameter the Sending Anchor should use for a `SEP-12 GET /customer` request. This field can be omitted if no KYC is necessary. Use a value from `sep12.sender.types` instead if any are present.    |
 | `receiver_sep12_type` | string  | (**deprecated**, optional) The value of the `type` parameter the Sending Anchor should use for a `SEP-12 GET /customer` request. This field can be omitted if no KYC is necessary. Use a values from `sep12.receiver.types` instead if any are present. |
-| `fields`              | object  | (**deprecated**, optional) An object containing the per-transaction parameters required in `POST /transactions` requests. Pass [SEP-9] fields via [SEP-12 PUT /customer](sep-0012.md#customer-put) instead.                                             |
+| `fields`              | object  | (**deprecated**, optional) An object that contains multiple key-value pairs, where each key represents a field related to the transaction, and the value is an `transaction` object describing the field's details.                                     |
```
lijamie98, line 386: Since we are adding `type` to `fields`, I think we should remove `deprecated`.

JakeUrban: Neither of these descriptions quite make sense. The original version of the `fields` object schema looked like this:

```js
{
  "fields": {
    "transaction": {}
  }
}
```

We deprecated `fields` because its only purpose was for `transaction` to specify request parameters required to be sent in `POST /transactions` requests.

Now that we're adding `type`, I think we have a couple options:

- Deprecate `fields.transaction` specifically, and un-deprecated `fields`. This way we can add the `type` key to the `fields` object:

```js
{
  "fields": {
    "type": {
      "description": "",
      "choices": [],
      "optional"
    }
  }
}
```

- Un-deprecate both `fields` and `field.transaction`, and add `type` as a subfield of `field.transactions`:

```js
{
  "fields": {
    "transaction": {
      "type": {
        "description": "",
        "choices": [],
        "optional"
      }
    }
  }
}
```

PR #1561, opened 2024-10-08
SEP-45: Stellar Web Authentication for Contract Accounts

philipliu:
This adds the initial draft for SEP-0045, which is a proposal for a new API to allow for web authentication for contract accounts. It is based on SEP-10 which is the web authentication API for classic accounts but with modifications to support Soroban's authentication model.

This is based on a [previous PR](https://github.com/stellar/stellar-protocol/pulls?q=is%3Apr+is%3Aclosed+soroban+auth+) authored by @leighmcculloch.

### DIFF
--- a/ecosystem/sep-0045.md
+++ b/ecosystem/sep-0045.md
@@ -36,8 +36,7 @@ and responses.
 
 It involves the following components:
 
-- A **Home Domain**: a domain hosting a [SEP-1 stellar.toml](sep-0001.md) containing a `WEB_AUTH_ENDPOINT_SEP0045` (URL)
-  and `SIGNING_KEY` (`G...`).
+- A **Home Domain**: a domain hosting a [SEP-1 stellar.toml](sep-0001.md) containing a `WEB_AUTH_FOR_CONTRACTS_ENDPOINT` (URL), `WEB_AUTH_CONTRACT_ID`, and `SIGNING_KEY` (`G...`).
 - A **Server**: a server providing the `WEB_AUTH_FOR_CONTRACTS_ENDPOINT` that implements the GET and POST operations discussed
   - The `SIGNING_KEY` from the **Home Domain** is the **Server Account**.
   in this document. The server's domain may be the **Home Domain**, a sub-domain of the **Home Domain**, or a different
@@ -55,7 +54,7 @@ The discovery flow is as follows:
 
 1. The **Client** retrieves the `stellar.toml` from the **Home Domain** in accordance with
    [SEP-1 stellar.toml](sep-0001.md).
-1. The **Client** looks up the `WEB_AUTH_ENDPOINT_SEP0045`, `WEB_AUTH_CONTRACT_ID` and `SIGNING_KEY` (i.e. **Server Account**) from the `stellar.toml`.
+1. The **Client** looks up the `WEB_AUTH_FOR_CONTRACTS_ENDPOINT`, `WEB_AUTH_CONTRACT_ID` and `SIGNING_KEY` (i.e. **Server Account**) from the `stellar.toml`.
 
 The authentication flow is as follows:
 
@@ -88,11 +87,11 @@ The flow achieves several things:
 ## Authentication Endpoint
 
 The organization with a **Home Domain** indicates that it supports authentication via this protocol by specifying
-`WEB_AUTH_ENDPOINT_SEP0045` in their [`stellar.toml`](sep-0001.md) file. This is how a wallet knows where to find the
+`WEB_AUTH_FOR_CONTRACTS_ENDPOINT` in their [`stellar.toml`](sep-0001.md) file. This is how a wallet knows where to find the
 **Server**. A **Server** is required to implement the following behavior for the web authentication endpoint:
 
-- [`GET <WEB_AUTH_ENDPOINT_SEP0045>`](#challenge): request a challenge (step 1)
-- [`POST <WEB_AUTH_ENDPOINT_SEP0045>`](#token): exchange a signed challenge for session JWT (step 2)
+- [`GET <WEB_AUTH_FOR_CONTRACTS_ENDPOINT>`](#challenge): request a challenge (step 1)
+- [`POST <WEB_AUTH_FOR_CONTRACTS_ENDPOINT>`](#token): exchange a signed challenge for session JWT (step 2)
 
 ## Cross-Origin Headers
 
@@ -119,7 +118,7 @@ the **Server** domain.
 #### Request
 
 ```
-GET <WEB_AUTH_ENDPOINT_SEP0045>
+GET <WEB_AUTH_FOR_CONTRACTS_ENDPOINT>
 ```
 
 ##### Request Parameters:
@@ -192,7 +191,7 @@ This endpoint accepts a signed challenge transaction, validates it and responds 
 [JSON Web Token](https://jwt.io/) authenticating the account.
 
 The **Client** submits a challenge (that was previously returned by the [`challenge`](#challenge) endpoint) as a HTTP
-POST request to `WEB_AUTH_ENDPOINT_SEP0045` using one of the following formats (both should be equally supported by the
+POST request to `WEB_AUTH_FOR_CONTRACTS_ENDPOINT` using one of the following formats (both should be equally supported by the
 server):
 
 - Content-Type: `application/x-www-form-urlencoded`, body:
--- a/ecosystem/sep-0045.md
+++ b/ecosystem/sep-0045.md
@@ -40,6 +40,7 @@ It involves the following components:
 
 - A **Home Domain**: a domain hosting a [SEP-1 stellar.toml](sep-0001.md) containing a `WEB_AUTH_FOR_CONTRACTS_ENDPOINT`
   (URL), `WEB_AUTH_CONTRACT_ID`, and `SIGNING_KEY` (`G...`).
+  - The `SIGNING_KEY` from this domain is referred to as the **Home Domain Address** in this document.
 - A **Server**: a server providing the `WEB_AUTH_FOR_CONTRACTS_ENDPOINT` that implements the GET and POST operations
   discussed in this document. The server's domain may be the **Home Domain**, a sub-domain of the **Home Domain**, or a
   different domain.
@@ -48,7 +49,7 @@ It involves the following components:
 - A **Client**: the software used by the holder of the **Client Account** being authenticated by the **Server**.
 - A **Client Domain** (optional): a domain hosting a [SEP-1 stellar.toml](sep-0001.md) containing a `SIGNING_KEY` used
   for [Verifying the Client Domain](#verifying-the-client-domain)
-  - The `SIGNING_KEY` from this domain is the **Client Domain Address**
+  - The `SIGNING_KEY` from this domain is referred to as the **Client Domain Address** in this document.
 - A **Web Auth Contract**: a contract that implements the `web_auth_verify` function as described in this document. The
   contract must be deployed at the `WEB_AUTH_CONTRACT_ID` address specified in the **Server**'s `stellar.toml`.
 
@@ -63,7 +64,7 @@ The authentication flow is as follows:
 
 1. The **Client** requests a unique [`challenge`](#challenge) from the **Server** which includes a list of authorization
    entries XDR-encoded strings
-1. If the request contains a `client_domain` parameter, the **Server** may fetch the **Client Domain Account** and
+1. If the request contains a `client_domain` parameter, the **Server** may fetch the **Client Domain Address** and
    generate the challenge with an additional authorization entry as described in the [Response](#response) section.
 1. The **Server** responds with the challenge
 1. The **Client** verifies that each authorization entry does not contain any sub-invocations
@@ -72,7 +73,7 @@ The authentication flow is as follows:
 1. The **Client** verifies that the `function_name` in each authorization entry is `web_auth_verify`
 1. The **Client** verifies that the `args` map in each authorization entry match the expected values and is the same
    across all authorization entries:
-   1. The `address` value matches the **Client Account** address
+   1. The `account` value matches the **Client Account** address
    1. The `home_domain` value matches the **Home Domain**
    1. The `home_domain_address` value matches the **Home Domain Address**
    1. The `web_auth_domain` value matches the **Server**'s domain
@@ -86,7 +87,7 @@ The authentication flow is as follows:
    - Note: **Client** signatures may not be required if the contract's `__check_auth` implementation does not require
      them
 1. The **Client** obtains a signature from the **Client Domain Address** for the authorization entry where
-   `credentials.address.address` is the **Client Domain Account** if the **Client** included a client domain in the
+   `credentials.address.address` is the **Client Domain Address** if the **Client** included a client domain in the
    request
 1. The **Client** simulates the transaction with the signed authorization entries and verifies the following to ensure
    the transaction does not have any unintended side effects:
@@ -102,7 +103,7 @@ The authentication flow is as follows:
 1. The **Server** verifies that the `function_name` in each authorization entry is `web_auth_verify`
 1. The **Server** verifies that the `args` map in each authorization entry match the expected values and are the same
    across all authorization entries:
-   1. The `address` value matches the **Client Account** address
+   1. The `account` value matches the **Client Account** address
    1. The `home_domain` value matches the **Home Domain**
    1. The `home_domain_address` value matches the **Home Domain Address**
    1. The `web_auth_domain` value matches the **Server**'s domain
--- a/ecosystem/sep-0045.md
+++ b/ecosystem/sep-0045.md
@@ -15,15 +15,20 @@ Discussion: TBA
 ## Simple Summary
 
 This SEP defines the standard way for clients such as wallets or exchanges to create authenticated web sessions on
-behalf of a user who holds a Contract account. A wallet may want to authenticate with any web service which requires a
+behalf of a user who holds a contract account. A wallet may want to authenticate with any web service which requires a
 contract account ownership verification, for example, to upload KYC information to an anchor in an authenticated way as
 described in [SEP-12](sep-0012.md).
 
-This SEP also supports authenticating users of shared contract accounts. Clients can use [memos](#memos) to distinguish
-users or sub-accounts of shared accounts.
-
 This SEP is based on [SEP-10](sep-0010.md), but only supports contract accounts and aims to provide a JWT that is fully
 compatible with existing protocols that use SEP-10.
+
+## Definitions
+
+- **Address**: An identifier representing a possible account on the Stellar network. Addresses begin with specific
+  prefixes like G for public keys (accounts controlled by secret keys), M for muxed accounts and C for contract
+  accounts.
+- **Account**: A stateful entity on the Stellar network that can hold balances. This can refer to a Stellar account
+  (`G...` addresses or `M...` addresses) or Soroban contract that abstracts an account (`C...` addresses).
 
 ## Abstract
 
@@ -34,17 +39,15 @@ It involves the following components:
 
 - A **Home Domain**: a domain hosting a [SEP-1 stellar.toml](sep-0001.md) containing a `WEB_AUTH_ENDPOINT_SEP0045` (URL)
   and `SIGNING_KEY` (`G...`).
+  - The `SIGNING_KEY` from the **Home Domain** is the **Home Domain Address**.
 - A **Server**: a server providing the `WEB_AUTH_ENDPOINT_SEP0045` that implements the GET and POST operations discussed
   in this document. The server's domain may be the **Home Domain**, a sub-domain of the **Home Domain**, or a different
   domain.
-  - The `SIGNING_KEY` from the **Home Domain** is the **Server Account**.
-- A **Client Account**: the account being authenticated.
-  - A Contract address (`C...`) that may be accompanied by a memo to scope the authentication to a user or sub-account
-    of the account.
+- A **Client Account**: A Soroban contract that has been deployed to the network which is being authenticated.
 - A **Client**: the software used by the holder of the **Client Account** being authenticated by the **Server**.
 - A **Client Domain** (optional): a domain hosting a [SEP-1 stellar.toml](sep-0001.md) containing a `SIGNING_KEY` used
   for [Verifying the Client Domain](#verifying-the-client-domain)
-  - The `SIGNING_KEY` from this domain is the **Client Domain Account**
+  - The `SIGNING_KEY` from this domain is the **Client Domain Address**
 - A **Web Auth Contract**: a contract that implements the `web_auth_verify` function as described in this document. The
   contract must be deployed at the `WEB_AUTH_CONTRACT_ID` address specified in the **Server**'s `stellar.toml`.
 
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0045.md
```diff
@@ -0,0 +1,414 @@
...
+It involves the following components:
+
+- A **Home Domain**: a domain hosting a [SEP-1 stellar.toml](sep-0001.md) containing a `WEB_AUTH_ENDPOINT_SEP0045` (URL)
+  and `SIGNING_KEY` (`G...`).
+- A **Server**: a server providing the `WEB_AUTH_ENDPOINT_SEP0045` that implements the GET and POST operations discussed
+  in this document. The server's domain may be the **Home Domain**, a sub-domain of the **Home Domain**, or a different
+  domain.
+  - The `SIGNING_KEY` from the **Home Domain** is the **Server Account**.
+- A **Client Account**: the account being authenticated.
+  - A Contract address (`G...`) that may be accompanied by a memo to scope the authentication to a user or sub-account
```
willemneal, line 40:
```suggestion
  - A Contract address (`C...`) that may be accompanied by a memo to scope the authentication to a user or sub-account
```

philipliu: Thanks!

ecosystem/sep-0045.md
```diff
@@ -0,0 +1,443 @@
...
+
+## Simple Summary
+
+This SEP defines the standard way for clients such as wallets or exchanges to create authenticated web sessions on
+behalf of a user who holds a Contract account. A wallet may want to authenticate with any web service which requires a
+contract account ownership verification, for example, to upload KYC information to an anchor in an authenticated way as
+described in [SEP-12](sep-0012.md).
+
+This SEP also supports authenticating users of shared contract accounts. Clients can use [memos](#memos) to distinguish
+users or sub-accounts of shared accounts.
```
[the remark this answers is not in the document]
marcelosalloum: Nevermind, I see this is only used to generate distinct JWT tokens as the final result.

philipliu:
> Should SEP-45 eventually fully replace SEP-10? if so, I believe we should replace occurrences of C-address account(s) with just Account and define what it means at the top of the document. A definition that includes C-address, G-address, and mux accounts would make it flexible from the starting point, WDYT?

I think we should initially support C-addresses only. As the ecosystem begins to adopt the protocol, we may uncover issues. By starting with C-addresses only, it will be easier to support G-addresses later, rather than removing support if there are issues.

That being said, I think a definition section is helpful. I will add it.

ecosystem/sep-0045.md
```diff
@@ -0,0 +1,443 @@
...
+  expected values and are in the same order as described in the
+  [Web Authentication Contract](#web-authentication-contract) section.
+
+### Simulate the Transaction
+
+Reconstruct the transaction using the authorization entries and simulate the transaction. By simulating the transaction,
+the Client can verify:
+
+- The ledger footprint's `read_write` set contains a single `contract_data` entry where the contract ID is the Client
+  account and the key is `ledger_key_nonce`.
```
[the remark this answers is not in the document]
philipliu: Yes, I believe so.

JakeUrban (commented):
Aside from my other comments, I didn't see a good place to raise this one in particular so I'm making it here.

The `SorobanAuthorizationEntry` XDR struct contains a `SorobanCredentials` struct within it, but we're not using that field in the authorization entries we pass between the client and server. Why is that?

Couldn't we do something like this instead?

- The response to the `GET` request is a list of authorization entires where:
  - the first one is the entry for the client account, which doesn't have a signature present in its `SorobanCredentials` struct
  - the second one is the entry for the server account, which has a signature present in its `SorobanCredentials` struct
  - the third one is optional, depending on whether the client domain account is being verified, and if present, does not have a signature present in its `SorobanCredentials` struct

In the `POST` request, the auth entires can include the signatures from the client in their credentials struct.

This way, we're only passing auth entry XDRs back and forth, rather than entires + signatures + credentials.

ecosystem/sep-0045.md
```diff
@@ -0,0 +1,443 @@
...
+  the following with no additional sub invocations:
+  - `contract_address` is the `WEB_AUTH_CONTRACT_ID` from the **Server**'s `stellar.toml`
+  - `function_name` is `web_auth_verify`
+  - `args`:
+    - `account` matches the **Client Account** from the request
+    - `memo` matches the `memo` from the request
+    - `home_domain` matches the `home_domain` from the request
+    - `web_auth_domain` matches the **Server**'s domain
+    - `client_domain` matches the **Client**'s domain
+    - `client_domain_address` is the **Client Domain**'s address
```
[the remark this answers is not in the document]
philipliu: > I don't see parameters in the request that provide either of these values. In SEP-10, the client provides the client's domain and the server looks up the client domain account.

Thanks for catching this, `client_domain` was supposed to be included in the GET request.

ecosystem/sep-0045.md
```diff
@@ -0,0 +1,443 @@
...
+   Account**) from the `stellar.toml`.
+
+The authentication flow is as follows:
+
+1. The **Client** requests a unique [`challenge`](#challenge) from the **Server** which includes a list of Soroban
+   authorization entries and the server's signatures represented as XDR strings.
+1. The **Server** responds with the challenge.
+1. The **Client** verifies that each authorization entry is signed by the **Server Account**.
+1. The **Client** verifies that each authorization entry does not contain any sub-invocations.
+1. The **Client** signs the first authorization entry using the secret key(s) of the signer(s) for the **Client
```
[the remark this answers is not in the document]
philipliu: Good point, we only need to verify `__check_auth` succeeds, which does not always require client signatures. The current API assumes that each authorization entry requires a signature, so that must be fixed before I rephrase this.

ecosystem/sep-0045.md
```diff
@@ -0,0 +1,443 @@
...
+1. The **Server** constructs a transaction with a single Invoke Host Function operation using the credentials and the
+   challenge authorization entries returned by the client and simulates the transaction.
+1. If the simulation succeeds, the **Server** responds with a [JWT](https://jwt.io) that represents the authenticated
+   session
+
+The flow achieves several things:
+
+- Both **Client** and **Server** can be implemented using well-established Stellar libraries
+- The **Client** can verify that the **Server** holds the secret key to the **Server Account**
+- The **Server** can verify that the **Client** holds the secret key(s) to signer(s) of the **Client Account**
```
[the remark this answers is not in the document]
philipliu: Yes 👍

philipliu:
> the second one is the entry for the server account, which has a signature present in its SorobanCredentials struct

There wouldn't be an authorization entry for the server because the `require_auth` calls are only on the client address and an optional client domain address, right?

> In the POST request, the auth entires can include the signatures from the client in their credentials struct.
>
> This way, we're only passing auth entry XDRs back and forth, rather than entires + signatures + credentials.

Are you suggesting that we include the server signatures in each of the `SorboanCredentials` arrays?

JakeUrban:
> There wouldn't be an authorization entry for the server because the require_auth calls are only on the client address and an optional client domain address, right?

That depends on the implementation of the server's contract. We could require implementations to require authorization from the client _and server_ (and optionally the client domain's address).

> Are you suggesting that we include the server signatures in each of the SorboanCredentials arrays?

I'm suggesting we use the credentials struct with the auth entries structs so the API looks like this:

`GET <WEB_AUTH_ENDPOINT_SEP0045>` Response
```js
{
  "authorization_entries": [
"AAAAAQAAAAHw6CVqzY+dCq3myVJBo1kb3nEGE7oO6obmJeUNvYQ0uiUmvl6zrExtAAAAAAAAAAEAAAAAAAAAAahbWC88MmIccL/85Jd6Myxjq5sHF+2dnq1+2H8osKvEAAAAD3dlYl9hdXRoX3ZlcmlmeQAAAAAHAAAAEgAAAAHw6CVqzY+dCq3myVJBo1kb3nEGE7oO6obmJeUNvYQ0ugAAAA4AAAADMTIzAAAAAA4AAAAObG9jYWxob3N0OjgwODAAAAAAAA4AAAAObG9jYWxob3N0OjgwODAAAAAAAAEAAAABAAAADgAAAAozMDc4NjE1NjE1AAAAAAAA",
"AAAAAQAAAAHw6CVqzY+dCq3myVJBo1kb3nEGE7oO6obmJeUNvYQ0uiUmvl6zrExtAAAAAAAAAAEAAAAAAAAAAahbWC88MmIccL/85Jd6Myxjq5sHF+2dnq1+2H8osKvEAAAAD3dlYl9hdXRoX3ZlcmlmeQAAAAAHAAAAEgAAAAHw6CVqzY+dCq3myVJBo1kb3nEGE7oO6obmJeUNvYQ0ugAAAA4AAAADMTIzAAAAAA4AAAAObG9jYWxob3N0OjgwODAAAAAAAA4AAAAObG9jYWxob3N0OjgwODAAAAAAAAEAAAABAAAADgAAAAozMDc4NjE1NjE1AAAAAAAA"
  ]
}
```

Where the first entry contains a `SorobanCredentials` struct with a missing signature field from the client, and the second entry contains a credentials struct with a signature from the server present. That way, the request to the `POST <WEB_AUTH_ENDPOINT_SEP0045>` would have an identical format, but the first entry would be updated to include the client's signature.

philipliu:
Implemented your suggestion, @JakeUrban 👍

ecosystem/sep-0045.md
```diff
@@ -0,0 +1,431 @@
...
+
+#### Response
+
+##### Success
+
+On success the endpoint must return `200 OK` HTTP status code and a JSON object with the following fields:
+
+- `authorization_entries`: a list of XDR-encoded `SorobanAuthorizationEntry`. It contains an entry for each of the
+  **Client Account**'s signers in addition to a signed entry from the **Home Domain Address** and optionally the
+  **Client Domain Address**. The entries are not guaranteed to be in any specific order.
```
JakeUrban, line 148: > an entry for each of the  **Client Account**'s signers

I'm confused by this. If the client account is a contract (which it must be in SEP-45), then how does the server know whether the account has multiple signers, one signer, or no signers at all?

ecosystem/sep-0045.md
```diff
@@ -0,0 +1,431 @@
...
+
+- `contract_address` is the `WEB_AUTH_CONTRACT_ID` from the **Server**'s `stellar.toml`
+- `function_name` is `web_auth_verify`
+- `args`:
+  - `address` matches the `address` from the request
+  - `home_domain` matches the `home_domain` from the request
+  - `home_domain_address` matches the **Home Domain Address**
+  - `web_auth_domain` matches the **Server**'s domain
+  - `client_domain` matches the **Client**'s domain
+  - `client_domain_address` matches the **Client Domain Address**
```
JakeUrban, line 160: This could also be a `C...` address if we add a new field to the stellar.toml file.

JakeUrban, line 160: ~Should there be a step in the 'authorization' flow that says the client must verify that the function arguments in each invocation match the arguments provided to the endpoint?~

Nevermind, I see we encourage the client to do this in the recommendation section. We should probably find a way to reference this section in the authentication flow section since it include important guidance.

ecosystem/sep-0045.md
philipliu: You're right, it does not know. At most, there would be one auth entry, independent of the account's number of signers. I'll fix this.

ecosystem/sep-0045.md
```diff
@@ -0,0 +1,431 @@
...
+   [SEP-1 stellar.toml](sep-0001.md).
+1. The **Client** looks up the `WEB_AUTH_ENDPOINT_SEP0045`, `WEB_AUTH_CONTRACT_ID` and `SIGNING_KEY` (i.e. **Home Domain
+   Address**) from the `stellar.toml`.
+
+The authentication flow is as follows:
+
+1. The **Client** requests a unique [`challenge`](#challenge) from the **Server** which includes a list of Soroban
+   authorization entries XDR-encoded strings
+1. The **Server** responds with the challenge
+1. The **Client** verifies that each authorization entry does not contain any sub-invocations
```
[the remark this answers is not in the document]
philipliu: If I understand correctly, a contract could be considered benign if there are no unexpected state changes as a result of submitting the transaction.

ecosystem/sep-0045.md
```diff
@@ -0,0 +1,431 @@
...
+## Abstract
+
+This protocol is a variation of mutual challenge-response, which uses Soroban authorization entries to encode challenges
+and responses.
+
+It involves the following components:
+
+- A **Home Domain**: a domain hosting a [SEP-1 stellar.toml](sep-0001.md) containing a `WEB_AUTH_ENDPOINT_SEP0045` (URL)
+  and `SIGNING_KEY` (`G...`).
+  - The `SIGNING_KEY` from the **Home Domain** is the **Home Domain Address**.
```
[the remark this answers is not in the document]
philipliu: I like the simplicity of a single `SIGNING_KEY`. Besides the advantage of verifying the server signature without simulation, client domain verification would "just work"  as they would need to update the `stellar.toml` with a new `SIGNING_KEY_SEP0045`. In most cases, I'd imagine `SIGNING_KEY` === `SIGNING_KEY_SEP0045`.

ecosystem/sep-0045.md
```diff
@@ -0,0 +1,431 @@
...
+1. The **Client** retrieves the `stellar.toml` from the **Home Domain** in accordance with
+   [SEP-1 stellar.toml](sep-0001.md).
+1. The **Client** looks up the `WEB_AUTH_ENDPOINT_SEP0045`, `WEB_AUTH_CONTRACT_ID` and `SIGNING_KEY` (i.e. **Home Domain
+   Address**) from the `stellar.toml`.
+
+The authentication flow is as follows:
+
+1. The **Client** requests a unique [`challenge`](#challenge) from the **Server** which includes a list of Soroban
+   authorization entries XDR-encoded strings
+1. The **Server** responds with the challenge
```
leighmcculloch, line 65: I notice that SEP-10 and SEP-45 differ at step 2, where SEP-10 includes client account information and SEP-45 doesn't. Why does that difference happen?

ecosystem/sep-0045.md
```diff
@@ -0,0 +1,431 @@
...
+  contract must be deployed at the `WEB_AUTH_CONTRACT_ID` address specified in the **Server**'s `stellar.toml`.
+
+The discovery flow is as follows:
+
+1. The **Client** retrieves the `stellar.toml` from the **Home Domain** in accordance with
+   [SEP-1 stellar.toml](sep-0001.md).
+1. The **Client** looks up the `WEB_AUTH_ENDPOINT_SEP0045`, `WEB_AUTH_CONTRACT_ID` and `SIGNING_KEY` (i.e. **Home Domain
+   Address**) from the `stellar.toml`.
+
+The authentication flow is as follows:
```
leighmcculloch, line 61: The auth flow in SEP-10 contains a really important step:

> The Client verifies that the transaction has an invalid sequence number 0. This is extremely important to ensure the transaction isn't malicious.

A similar step doesn't appear in the auth flow for SEP-45. Does the same problem get tackled another way?

ecosystem/sep-0045.md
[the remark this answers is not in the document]
leighmcculloch: I think @JakeUrban's original question is worth considering. I'm reading the auth flow section, and if you implement auth just as is described in these steps then I don't think enough is being verified. This section in SEP-10 goes into much greater detail with each element that the client and server verify, and that makes implementing the SEP straightforward, as the code follows the steps. The auth flow here is a little too minimal I think. We need to state every single thing that the implementation needs to check.

ecosystem/sep-0045.md
```diff
@@ -0,0 +1,431 @@
...
+| --------------- | ------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `address`       | `C...` string | The **Client Account** address (`C...`) that the **Client** wishes to authenticate with the **Server**.                                                                                                                                                                                                                                                                                                                    |
```
leighmcculloch, line 130: I can understand why renaming this field from account (SEP-10) to address (SEP-45), but it also makes the SEP-45 and SEP-10 APIs use different terms for the same thing. I don't think we benefit much from renaming concepts in SEP-45, it could use the same vocab as SEP-10 unless necessary.

ecosystem/sep-0045.md
leighmcculloch, line 160: The list of args here are stored in an array, and so are not self describing, and are difficult to insert or remove new values over time if needed to. One of the advantages of SEP-10 was that because data was stored in data operations, new fields could be added as needed.

I think there's value in considering the args be a single value, a map, and that map contains mappings of symbols to values.

Similar to SEP-10, it should state that the list is not exhaustive so that implementations do not assume it is, and so that it can be expanded in the future.

ecosystem/sep-0045.md
```diff
@@ -0,0 +1,431 @@
...
+```
+
+#### Response
+
+##### Success
+
+On success the endpoint must return `200 OK` HTTP status code and a JSON object with the following fields:
+
+- `authorization_entries`: a list of XDR-encoded `SorobanAuthorizationEntry`. It contains an entry for each of the
+  **Client Account**'s signers in addition to a signed entry from the **Home Domain Address** and optionally the
```
leighmcculloch, line 147: > It contains an entry for each of the **Client Account**'s signers

I would expect only a single entry for the client account, and that single entry can have multiple signatures attached to it. An entry authorizes an action by an account, not by a single signer.

Is there another reason it needs multiple signers?

ecosystem/sep-0045.md
```diff
@@ -0,0 +1,431 @@
...
+  **Client** or **Server** have been configured incorrectly.
+
+Example:
+
+```json
+{
+  "authorization_entries": [
+    "AAAAAQAAAAHw6CVqzY+dCq3myVJBo1kb3nEGE7oO6obmJeUNvYQ0unUQ/ZB5+1lbAAAAAAAAAAEAAAAAAAAAAahbWC88MmIccL/85Jd6Myxjq5sHF+2dnq1+2H8osKvEAAAAD3dlYl9hdXRoX3ZlcmlmeQAAAAAHAAAAEgAAAAHw6CVqzY+dCq3myVJBo1kb3nEGE7oO6obmJeUNvYQ0ugAAAA4AAAAObG9jYWxob3N0OjgwODAAAAAAABIAAAAAAAAAANKw4wpgrtrVoAuJ7zcXgZAOgF0etbzpRfZJhXKBED5VAAAADgAAAA5sb2NhbGhvc3Q6ODA4MAAAAAAAAQAAAAEAAAAOAAAACTg5MjE5NDQ1MQAAAAAAAAA=",
+    "AAAAAQAAAAAAAAAA0rDjCmCu2tWgC4nvNxeBkA6AXR61vOlF9kmFcoEQPlUEH887Ol7oawAT5vYAAAAQAAAAAQAAAAEAAAARAAAAAQAAAAIAAAAPAAAACnB1YmxpY19rZXkAAAAAAA0AAAAg0rDjCmCu2tWgC4nvNxeBkA6AXR61vOlF9kmFcoEQPlUAAAAPAAAACXNpZ25hdHVyZQAAAAAAAA0AAABA42dUagEASp9894JchR8zl8qIqSVc5KhuWgxOu1YWDH2nLBlwwQXAMX4h5i1zA5Z86LU84aSPoVQtdzCeJZG2BgAAAAAAAAABqFtYLzwyYhxwv/zkl3ozLGOrmwcX7Z2erX7Yfyiwq8QAAAAPd2ViX2F1dGhfdmVyaWZ5AAAAAAcAAAASAAAAAfDoJWrNj50KrebJUkGjWRvecQYTug7qhuYl5Q29hDS6AAAADgAAAA5sb2NhbGhvc3Q6ODA4MAAAAAAAEgAAAAAAAAAA0rDjCmCu2tWgC4nvNxeBkA6AXR61vOlF9kmFcoEQPlUAAAAOAAAADmxvY2FsaG9zdDo4MDgwAAAAAAABAAAAAQAAAA4AAAAJODkyMTk0NDUxAAAAAAAAAA=="
```
leighmcculloch, line 171: These auth entries could be wrapped in a variable length array so that they're a single unit to decode. I've noticed in some other APIs where we designed them to be an array of XDR, that it's more annoying to interact with than a single XDR value. This is because you have to iterate over the values and decode each one, rather than decode a single blob.

ecosystem/sep-0045.md
```diff
@@ -0,0 +1,431 @@
...
+- Content-Type: `application/x-www-form-urlencoded`, body:
+  `authorization_entries=<XDR (URL-encoded)>,authorization_entries=<XDR (URL-encoded)>`)
+- Content-Type: `application/json`, body: `{"authorization_entries": ["<XDR>", "<XDR>"]}`
+
+To validate the challenge transaction the following steps are performed by the **Server**. If any of the listed steps
+fail, then the authentication request must be rejected — that is, treated by the **Server** as an invalid input.
+
+1. Construct a transaction with a single Invoke Host Function operation using the authorization entries returned by the
+   client;
+1. Simulate the transaction and verify that it succeeds.
```
leighmcculloch, line 208: The list seems pretty small compared to SEP-10's list of things to verify.

This should state how to fill out the InvokeHostOp and transaction. Along with the steps the server needs to take to verify the incoming signed values. At the very least it'll be important to check the server value is signed, as well as consistency between the input parameters across all the authorization entries.

leighmcculloch (commented):
👏🏻 Few comments in line.

ecosystem/sep-0045.md
philipliu: > Is there another reason it needs multiple signers?

Nope, this was a mistake. Jake had also pointed this out.

ecosystem/sep-0045.md
philipliu: I couldn't think of a way to construct the challenge such that the signed transaction succeeds in simulation but errors when submitted to the network. Instead, I recommend that the client check that the challenge is not malicious by simulating it before returning it to the server.

philipliu: An alternative approach would require `web_auth_verify` to return a specific error code when all signatures are valid. The server would check for this error code when it verifies the client's response.

ecosystem/sep-0045.md
philipliu: When I wrote this SEP, I did not include client domain verification and deleted this step 🤦 . It should be added back in, thanks.

ecosystem/sep-0045.md
philipliu: This was the approach I went with initially but I discovered during implementation that the args map would not deserialize when it exceeded a certain size. Maybe there's a workaround or I implemented it incorrectly, I'll continue looking into this.

ecosystem/sep-0045.md
philipliu: An alternative approach would be to implement `web_auth_verify` to fail with a specific error when all signatures are valid. The server would have to manually construct the authorization entries which can be challenging due to the nonces.

ecosystem/sep-0045.md
leighmcculloch: Highly recommend doing a diff of the two SEPs and checking for anything else that doesn't align, or could align such that diffing the two highlights the things that are unique. There may be some art rather than science to doing that, and it's not the be-all-end-all, but just an idea if possible.

ecosystem/sep-0045.md
leighmcculloch: This is very interesting. Do you happen to have an example of this in code I could experiment with?

ecosystem/sep-0045.md
leighmcculloch: If helpful we can add a new type to the XDR such as:
```
type SorobanAuthorizationEntries SorobanAuthorizationEntry<>;
```

ecosystem/sep-0045.md
philipliu: Yep. https://gist.github.com/philipliu/5524addbfe41abb48f5e292c8f3d2670

The only difference between the transactions is that the failing one has a nonce argument.

leighmcculloch: When I simulate the transaction with the follow command, I get the following resources required:

```
$ stellar contract invoke \
    --id CCUFWWBPHQZGEHDQX76OJF32GMWGHK43A4L63HM6VV7NQ7ZIWCV4JLOM \
    --build-only \
    -- \
    web_auth_verify \
    --args '{"address":"CDYOQJLKZWHZ2CVN43EVEQNDLEN544IGCO5A52UG4YS6KDN5QQ2LUWKY","home_domain":"localhost:8080","home_domain_address":"GDJLBYYKMCXNVVNABOE66NYXQGIA5AC5D223Z2KF6ZEYK4UBCA7FKLTG","web_auth_domain":"localhost:8080"}' \
    | stellar tx simulate \
    | stellar xdr decode --type TransactionEnvelope \
    | jq -r '.tx.tx.ext.v1.resources | "| Instructions | Read Bytes | Write Bytes |\n|---|---|---|\n| \(.instructions) | \(.read_bytes) | \(.write_bytes) |"'
```
| Instructions | Read Bytes | Write Bytes |
|---|---|---|
| 1,791,370 | 3,564 | 148 |

Those limits don't look like they're anywhere near the limits posted here: https://developers.stellar.org/docs/networks/resource-limits-fees

What am I doing differently?

leighmcculloch: We worked out offline that the issue is the map wasn't sorted in the failing tx.

ecosystem/sep-0045.md
philipliu: I like this idea. Since this will require an XDR change and likely a discussion, what do you think about keeping the auth entries in an array and updating this SEP with the new type later?

leighmcculloch: The XDR change is not protocol related, so it can be made at anytime, and I think is relatively uncontroversial, so I think you could rely on it here. Also, in most SDKs the XDR change is not required, it's just convenient for the CLI that can only decode things with a name. Most SDKs should be able to decode an array of a type as is today.

ecosystem/sep-0045.md
```diff
@@ -0,0 +1,489 @@
...
+1. The **Server** responds with the challenge
+1. The **Client** verifies that each authorization entry does not contain any sub-invocations
+1. The **Client** verifies that the `contract_address` in each authorization entry matches the `WEB_AUTH_CONTRACT_ID`
+   from the **Server**'s `stellar.toml`
+1. The **Client** verifies that the `function_name` in each authorization entry is `web_auth_verify`
+1. The **Client** verifies that the `args` map in each authorization entry match the expected values and is the same
+   across all authorization entries:
+   1. The `address` value matches the **Client Account** address
+   1. The `home_domain` value matches the **Home Domain**
+   1. The `home_domain_address` value matches the **Home Domain Address**
```
JakeUrban, line 77: I don't think we've defined the home domain address. Is that the home domain's `SIGNING_KEY`?

ecosystem/sep-0045.md
```diff
@@ -0,0 +1,489 @@
...
+   from the **Server**'s `stellar.toml`
+1. The **Server** verifies that the `function_name` in each authorization entry is `web_auth_verify`
+1. The **Server** verifies that the `args` map in each authorization entry match the expected values and are the same
+   across all authorization entries:
+   1. The `address` value matches the **Client Account** address
+   1. The `home_domain` value matches the **Home Domain**
+   1. The `home_domain_address` value matches the **Home Domain Address**
+   1. The `web_auth_domain` value matches the **Server**'s domain
+   1. The `client_domain_address` value matches the **Client Domain Address** if the **Client** included a
+      `client_domain` in the request, otherwise it is not present
```
JakeUrban, line 110: Is this necessary? In SEP-10, the server only has to verify that its own signature is present on the challenge. Is it because the server's signature is not present on 2 of the 3 auth entires?

JakeUrban (commented):
Looks almost ready to go. Just a couple notes / questions

ecosystem/sep-0045.md
philipliu: Yes, it's the `SIGNING_KEY`. "home domain address" is confusing, I'll replace all instances of it with `SIGNING_KEY`.

ecosystem/sep-0045.md
philipliu: >  Is it because the server's signature is not present on 2 of the 3 auth entires?

Yeah, in SEP-10 the server signs over the transaction envelope while in SEP-45, the server only signs one of the authorization entries. 

If some of the auth entries returned by client are invalid (for any of those reasons checked by the steps above) it would be caught by the simulation. Checking these prior to simulation allows the server to return specific error messages without needing to parse the simulation response.

ecosystem/sep-0045.md
JakeUrban: I'm actually not against "home domain address", given we have "client domain address", I just think its needs to be defined with the rest of them.

philipliu: There was an earlier comment from Leigh about this term being confusing https://github.com/stellar/stellar-protocol/pull/1561#discussion_r1870662377, that's why I thought it would be better to remove it, but we would need to rename "client domain address" if we do so.

SEP-10 calls the client domain's `SIGNING_KEY`, "client domain account" so I think it makes sense to use similar terminology here. I'll keep "home domain address" for now and add the definition.

JakeUrban (approved):

PR #1564, opened 2024-10-27
CAP-0046-06: Fix events emitted

leighmcculloch:
### What

Change the events emitted by Stellar Asset Contracts to match the implementation.

### Why

If you look at the code of what events are emitted, they don't currently match the specification, with data values incorrect, as well as parameters.

https://github.com/stellar/rs-soroban-env/blob/v21.2.2/soroban-env-host/src/builtin_contracts/stellar_asset_contract/event.rs#L56-L71

### Reviewing

When reviewing, please thoroughly check the changes with the code. This is the second time we're correcting the events in the CAP with the implementation, having already attempted to get it aligned at least one other time since implementation, and these docs should be reliable, so it's worth two people thoroughly confirming.

cc @stellar/contract-committers

### DIFF
--- a/core/cap-0046-06.md
+++ b/core/cap-0046-06.md
@@ -330,13 +330,14 @@ fn mint(env: Env, to: Address, amount: i128);
 ///
 /// # Arguments
 ///
+/// * `admin` - The address initiating the clawback.
 /// * `from` - The address holding the balance from which the clawback will
 ///   take tokens.
 /// * `amount` - The amount of tokens to be clawed back.
 ///
 /// # Events
 ///
-/// Emits an event with topics `["clawback", admin: Address, to: Address, sep0011_asset: String],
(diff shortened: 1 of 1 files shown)


core/cap-0046-06.md
```diff
@@ -171,15 +171,15 @@ fn allowance(e: &Host, from: Address, spender: Address) -> Result<i128, HostErro
 /// * `spender` - The address being authorized to spend the tokens held by
 ///   `from`.
 /// * `amount` - The tokens to be made availabe to `spender`.
-/// * `expiration_ledger` - The ledger number where this allowance expires. Cannot
+/// * `live_until_ledger` - The ledger number where this allowance expires. Cannot
```
leighmcculloch, line 174: This isn't an event, but noticed the parameter is incorrectly named vs the contract.

dmkozh (approved):

PR #1567, opened 2024-10-30
[SEP-6][SEP-31] Deprecating `fields` and adding `funding_method`

JiahuiWho:
**What's Changed**
- SEP-6
  - GET `info`: Add `funding_methods` to each asset 
  - GET `deposit` and `deposit-exchange`: deprecate `type`, add required param `funding_method` to replace
  - GET `withdraw`  and `deposit-exchange`:  deprecate `type`, add required param `funding_method` to replace
- SEP-31
  - GET `info`: Clean up deprecated `fields` and `sep12` in the code example, adding `funding_methods` to each asset 
  - POST `transaction`: add required param `funding_method`  
- SEP-38
  - GET `info`: Add note that `buy_delivery_methods` and `sell_delivery_methods` should be in line with the `funding_methods` in the protocol where the quote will be applied

**Context**
SEP-31 and SEP-6 have both delegate KYC to SEP-12, there’s no need for these SEPs to include KYC-specific details that are not directly related to transaction creation in their info responses.

The only exception is the `funding_method` which may be needed to decide what KYC info required.

Also change the name from `offchain_delivery_method` to `funding_method`, as withdrawals may not necessarily involve cash.

More details:
https://docs.google.com/document/d/15FJ4p1Iw6Gkv-6vuRor49ENKlImGwbgeGY74tHjeQyk/edit?usp=sharing


lijamie98 (commented):
LGTM.

ecosystem/sep-0031.md
```diff
@@ -367,50 +343,15 @@ the corresponding off-chain asset, after fees have been applied.
...
-| `fee_percent`         | number  | (optional) A percentage fee in percentage points. Leave blank if there is no fee or fee calculation cannot be modeled using a fixed and percentage fee.                                                                                                 |
-| `sender_sep12_type`   | string  | (**deprecated**, optional) The value of the `type` parameter the Sending Anchor should use for a `SEP-12 GET /customer` request. This field can be omitted if no KYC is necessary. Use a value from `sep12.sender.types` instead if any are present.    |
-| `receiver_sep12_type` | string  | (**deprecated**, optional) The value of the `type` parameter the Sending Anchor should use for a `SEP-12 GET /customer` request. This field can be omitted if no KYC is necessary. Use a values from `sep12.receiver.types` instead if any are present. |
-| `fields`              | object  | (**deprecated**, optional) An object containing the per-transaction parameters required in `POST /transactions` requests. Pass [SEP-9] fields via [SEP-12 PUT /customer](sep-0012.md#customer-put) instead.                                             |
```
Ifropc, line 379: We need to keep deprecated fields in the document

Ifropc (commented):
SEP-6 doc is missing version update

ecosystem/sep-0031.md
JiahuiWho: Hmm I thought these fields has been marked as deprecated for a long time. Is there's a plan/timeline to finally get rid of it?

Ifropc: cc @JakeUrban I don't think we have a field removal strategy ATM

JakeUrban: Agreed with Gleb, we need to keep the deprecated fields in the doc for now.

ecosystem/sep-0006.md
```diff
@@ -1078,6 +1082,7 @@ The response should be a JSON object like:
       "authentication_required": true,
       "min_amount": 0.1,
       "max_amount": 1000,
+      "funding_method": ["SEPA", "SWIFT", "cash"],
       "fields": {
```
JakeUrban, line 1086: Lets remove these `fields` objects from examples given that the object is deprecated

ecosystem/sep-0006.md
```diff
@@ -410,7 +410,8 @@ Request Parameters:
...
+| `funding_method`              | string                  | (optional) Method supported by the anchor for transferring or settling assets. If the anchor supports multiple deposit methods (e.g. `SEPA` or `SWIFT`), the wallet should specify `funding_method`. This field may be necessary for the anchor to determine which KYC fields to collect.                                                                                                                                                                             |
+| `type`                        | string                  | (**deprecated** in favor of `funding_method` ,optional) Type of deposit. If the anchor supports multiple deposit methods (e.g. `SEPA` or `SWIFT`), the wallet should specify `type`. This field may be necessary for the anchor to determine which KYC fields to collect.                                                                                                                                                                                             |
```
JakeUrban, line 414: Something we could do to make things more readable is to move all deprecated fields to the bottom on the list?

Eventually we do want to remove these deprecated fields but I want to be intentional about when that happens and why its the right time.

ecosystem/sep-0031.md
```diff
@@ -1040,7 +982,8 @@ unexpected downtime, it is recommended to poll all in-progress transactions to f
 
 ## Changelog
 
-- `v3.0.0`: Update SEP31 with async kyc flow ([#1502](https://github.com/stellar/stellar-protocol/pull/1502))
+- `v3.1.0`: Add methods to info response ()
```
JakeUrban, line 985: Lets not forget to add the PR number and URL here

JakeUrban (commented):
Overall I think this change looks good but I think we need a better explanation for why we're introducing `funding_method` and deprecating `type`, because their parameter descriptions in the deposit and withdrawal are almost identical.

My understanding is that we're deprecating `GET /info`'s `types` object, because we don't want KYC sent in transaction initiation requests. However, it is still necessary to indicate what your funding/delivery method is, so the anchor can request necessary info via SEP-12, so we're introducing a new field that serves the same purpose as the original `type` parameter but removing the connection to the related object in `GET info`.

Another thing I think is missing is the connection to SEP-38. We should make it clear that the `funding_methods` defined in SEP-6 should match the delivery methods defined in SEP-38, if SEP-38 is supported.

I'd also argue that we should reuse the `delivery_methods` name in order to stay consistent with SEP-38.

JiahuiWho:
Regarding `funding_method`, 
- For the identical description, I intentionally kept it the same because it serves the same purpose across deposit, withdrawal, and receive requests, with consistent behavior. A uniform explanation improves clarity and avoids redundancy. I can add a note or example tailored to each request to highlight the different context.
- For the naming, I still prefer `funding_method` over `delivery_method` because the latter lacks intuitiveness without prefixes like `offchain`, `buy`, `sell`, or specific SEP-38 context. 

> Another thing I think is missing is the connection to SEP-38. We should make it clear that the funding_methods defined in SEP-6 should match the delivery methods defined in SEP-38, if SEP-38 is supported.

I would argue that it's actually the other way around: the `buy/sell_delivery_method` defined in SEP-38 should match the `funding_method` specified in SEP-6. This serves as the definitive guide for both wallets and anchors to determine which SEP should be the authoritative source in the event of any conflicts.

JakeUrban:
I think its more important to be consistent rather than accurate with our naming. Its confusing that SEP-38's "delivery methods" are the same thing as SEP-6 "funding methods".

It also doesn't matter which SEP is the authoritative source because they should always match. If they don't, its a bug, and wallets should report it to the anchor rather than assume one is correct and the other isn't. Can we have SEP-6/24/31 reference SEP-38 and have SEP-38 reference SEP-6/24/31?

lijamie98:
`delivery_method` feels a little confusing to me because it is like the `non-stellar-asset-funding-method` which is too long.

I think it would be good if we also rename SEP-38 to `funding_method`. The AP can support both names for backward compatibility for SEP-6, 31, 38.

JiahuiWho:
How about we also make change to SEP-38 so they are consistent?

JiahuiWho:
> If we were to use different words ("funding" in SEP-6 and "delivery" in SEP-38), it might be confusing at first but we can make it clear that they're the same thing using the descriptions. So this would be my 2nd pick.

Ok I can add that. But isn't that more a guidance for anchor not for wallet? I think stellar dev doc is where we should add it

ecosystem/sep-0006.md
```diff
@@ -820,9 +820,10 @@ Request Parameters:
...
 | `account`                     | `G...` or `M...` string | The stellar or muxed account ID of the user that wants to deposit. This is where the asset token will be sent. Note that the account specified in this request could differ from the account authenticated via SEP-10.                                                                                                                                                                                                                                                |
+| `quote_id`                    | string                  | (optional) The `id` returned from a `SEP-38 POST /quote` response. If this parameter is provided and the user delivers the deposit funds to the Anchor before the quote expiration, the Anchor should respect the conversion rate agreed in that quote. If the values of `destination_asset`, `source_asset` and `amount` conflict with the ones used to create the [SEP-38] quote, this request should be rejected with a `400`.                                     |
```
JakeUrban, line 826: Can we add `funding_method` to the list of values that can't conflict with SEP-38 here?

JiahuiWho: Is there such a list?

JakeUrban: > If the values of `destination_asset`, `source_asset` and `amount` conflict with the ones used to create the [SEP-38] quote, this request should be rejected with a `400`.

PR #1576, opened 2024-12-02
CAP 62

SirTyson:
This CAP introduces "partial state archival".

Instead of deleting evicted state initially in protocol 23, I'm proposing only implementing the "Hot Archive" initially. This will allow prioritization of live Soroban state (as in caching all live state in memory). See more context [here](https://docs.google.com/document/d/1hTCh-Ya-ulavZL_Gtb7gBM75chocMcG300RXsjPVPp0/edit?usp=sharing). Full State Archival will still be necessary long term, but can be implemented in a later protocol release.

This CAP discusses this Hot Archive implementation and related changes to the storage fee system.

### DIFF
--- a/core/README.md
+++ b/core/README.md
@@ -94,6 +94,8 @@
 | [CAP-0058](cap-0058.md) | Constructors for Soroban Contracts | Dmytro Kozhevin | Implemented |
 | [CAP-0059](cap-0059.md) | Host functions for BLS12-381 | Jay Geng | Implemented |
 | [CAP-0060](cap-0060.md) | Update to Wasmi register machine| Graydon Hoare | Accepted |
+| [CAP-0061](cap-0061.md) | Smart Contract Standardized Asset (Stellar Asset Contract) Extension: Memo | Tomer Weller | Draft |
+| [CAP-0062](cap-0062.md) | Soroban Live State Prioritization | Garand Tyson | Draft |
 
 ### Rejected Proposals
 | Number | Title | Author | Status |
--- a/core/cap-0062.md
+++ b/core/cap-0062.md
@@ -1,0 +1,283 @@
+```
+CAP: 0062
+Title: Soroban Live State Prioritization
+Working Group:
+    Owner: Garand Tyson <@SirTyson>
+    Authors: Garand Tyson <@SirTyson>
+    Consulted: Dmytro Kozhevin <@dmkozh>, Nicolas Barry <@MonsieurNicolas>
+Status: Draft
+Created: 2024-12-02
+Discussion: https://github.com/stellar/stellar-protocol/discussions/1575
+Protocol version: 23
+```
+
+## Simple Summary
+
+This proposal allows the network to evict `PERSISTENT` entries from the live state BucketList to a separate BucketList
+containing archival state. Note that evicted entries are not deleted from validators.
+
+## Working Group
+
+As specified in the Preamble.
+
+## Motivation
+
+This is a first step towards full State Archival, which will lower the storage requirements of validators
+and decrease the growth of History Archives. Additionally, this opens the door for database optimizations of
+live Soroban state, increasing limits and throughput, as well as automatic entry restoration.
+
+### Goals Alignment
+
+This change is aligned with the goal of lowering the cost and increasing the scale of the network.
+
+## Abstract
+
+Currently both live and archived Soroban state are maintained in a single BucketList DB. This makes data prioritization difficult
+since all state exists in a singular data structure. This proposal separates archived state and live state into two separate DBs.
+While these DBs are still both persisted on disk by validators, optimizing live state access is significantly easier. In particular,
+live state can be entirely cached in memory, removing disk reads from the Soroban transaction execution path, greatly increasing
+read limits and throughput.
+
+Each validator will maintain two BucketLists, the Live BucketList and the Hot Archive BucketList. `PERSISTENT` Soroban
+data entries and code entries that have expired will be "evicted" from the Live BucketList and moved to the Hot Archive
+BucketList. Both the Live BucketList and Hot Archive BucketList are persisted on disk and written to History Archives.
+
+Note that this proposal is a subset of [CAP-0057](cap-0057.md). The Hot Archive here is the same as CAP-0057, and this CAP
+is a strict subset of CAP-0057. The purpose is not to replace CAP-0057, but to offer a smaller first step implementation wise.
+
+## Specification
+
+### XDR changes
+
+```
+enum BucketListType
+{
+    LIVE = 0,
+    HOT_ARCHIVE = 1,
+};
+
+enum HotArchiveBucketEntryType
+{
+    HOT_ARCHIVE_METAENTRY = -1, // Bucket metadata, should come first.
+    HOT_ARCHIVE_ARCHIVED = 0,   // Entry is Archived
+    HOT_ARCHIVE_LIVE = 1,       // Entry was previously HOT_ARCHIVE_ARCHIVED but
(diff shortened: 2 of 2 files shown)


core/README.md
```diff
@@ -94,6 +94,8 @@
 | [CAP-0058](cap-0058.md) | Constructors for Soroban Contracts | Dmytro Kozhevin | Implemented |
 | [CAP-0059](cap-0059.md) | Host functions for BLS12-381 | Jay Geng | Implemented |
 | [CAP-0060](cap-0060.md) | Update to Wasmi register machine| Graydon Hoare | Accepted |
+| [CAP-0057](cap-0057.md) | Smart Contract Standardized Asset (Stellar Asset Contract) Extension: Memo | Tomer Weller | Draft |
+| [CAP-0057](cap-0057.md) | Soroban Live State Prioritization | Garand Tyson | Draft |
```
Shaptic, line 98: Weird upstream merge and wrong link in your CAP?

core/cap-0062.md
```diff
@@ -0,0 +1,281 @@
...
+BucketList. Both the Live BucketList and Hot Archive BucketList are persisted on disk and written to History Archives.
+
+Note that this proposal is a subset of [CAP-0057](cap-0057.md). The Hot Archive here is the same as CAP-0057, and this CAP
+is a strict subset of CAP-0057. The purpose is not to replace CAP-0057, but to offer a smaller first step implementation wise.
+
+## Specification
+
+### XDR changes
+
+```
```
Shaptic, line 52: This is usually rendered as a `diff` to make it easier to read :thinking:

core/cap-0062.md
```diff
@@ -0,0 +1,281 @@
...
+#### Live State BucketList
+
+The Live State BucketList most closely resembles the current BucketList. It will contain all “live” state of the ledger, including
+Stellar classic entries, live Soroban entries, network config settings, etc.
+The Live State BucketList is published to the History Archive on every checkpoint ledger via the "history" category. The Live BucketList
+functions identically to the current BucketList.
+
+#### Hot Archive BucketList
+
+The Hot Archive is a BucketList that stores evicted `PERSISTENT` entries
```
Shaptic, line 144: nit: missing period 
```suggestion
The Hot Archive is a BucketList that stores evicted `PERSISTENT` entries.
```

core/cap-0062.md
```diff
@@ -0,0 +1,281 @@
...
+#### Meta
+
+Whenever a `PERSISTENT` entry is evicted (i.e. removed from the Live State BucketList and added to the Hot Archive),
+the entry key and its associated TTL key will be emitted via `evictedLedgerKeys`
+(this field has been renamed and was previously named `evictedTemporaryLedgerKeys`).
+
+Whenever an entry is restored via `RestoreFootprintOp`, the `LedgerEntry` being restored and its
+associated TTL will be emitted as a `LedgerEntryChange` of type `LEDGER_ENTRY_RESTORE`. Note that
+entries being restored may not have been evicted such that entries currently in the Live State BucketList
+can see `LEDGER_ENTRY_RESTORE` events.
```
Shaptic, line 203: For this

> Note that entries being restored may not have been evicted such that entries currently in the Live State BucketList can see `LEDGER_ENTRY_RESTORE` events.

I'd like to make sure I understand it correctly: Are you saying that `LEDGER_ENTRY_RESTORE` does not necessarily mean restoration from eviction because live state may have an expired-yet-not-evicted entry that can be restored before it pops off the end of the BL?

If yes, I'd like to suggest a rewording:

```suggestion
Whenever an entry is restored via `RestoreFootprintOp`, the `LedgerEntry` being restored and its
associated TTL will be emitted as a `LedgerEntryChange` of type `LEDGER_ENTRY_RESTORE`. Note that
entries can be restored from both the Live State and Hot Archive BucketList, because archived entries 
continue to exist in Live State until they are evicted as part of natural BucketList growth. The emitted
meta does not distinguish between these.
```

core/cap-0062.md
```diff
@@ -0,0 +1,281 @@
...
+first as a step towards CAP-0057. At current usage levels, "full state archival" as specified by CAP-0057 (i.e. evicted entries are deleted
+from validators) is not a necessary or useful optimization. The space savings of validators and the History Archive would be
+minimal, and the additional complexity, both from a computational and UX standpoint, do not justify the minimal storage savings.
+Long term, should usage increase, the tradeoffs introduced by CAP-0057 will make more sense, but that is not the case currently
+or in the near term future.
+
+The reasoning behind introducing CAP-0057 before it was strictly necessary was to prevent breakage of applications. If applications
+and users were not prepared for State Archival, it would be very difficult to enable CAP-0057. However, because
+[CAP-0046-12](./cap-0046-12.md) is implemented and State Archival semantics are already required, upgrading to CAP-0057 should not
+be breaking, especially since much of the complexity, like proof generation, is handled automatically by RPC endpoints.
```
Shaptic, line 225:
```suggestion
be breaking, especially since much of the complexity, like proof generation, will be handled automatically by new Stellar Core HTTP endpoints.
```

Is that correct? While RPC will be the facilitator via `simulateTransaction`, the proofs are actually generated by `/getrestorationproof` per https://github.com/stellar/go/issues/5426, right?

core/cap-0062.md
```diff
@@ -0,0 +1,281 @@
...
+seem to address this issue. First, there is currently no way to remove state from the Hot Archive. Only
+restore operations remove Hot Archive state, but this state is just added back to the Live BucketList. Thus
+fees would continually increase without any benefit to network sustainability, as no action or system is
+currently in place to reduce Hot Archive size.
+
+While not improving network sustainability, Hot Archive fees could prevent state based DOS attacks, where
+state is rapidly added to the network. However, unlike the live BucketList, operations cannot directly
+write state to the Hot Archive (with the exception of restores, which always decrease Hot Archive size).
+The rate of eviction (i.e. the rate at which state is added to the Hot Archive) can already be controlled
+via network config settings. The rate at which state is added to the Live BucketList is also already
```
Shaptic, line 261: Can you elaborate on this? I was under the impression that eviction is controlled by state TTL rather than network configurations. Or is this referencing the eviction of already-archived entries in the live BL?

dmkozh (approved):

core/cap-0062.md
SirTyson: Yeah, I just have it in this format for now since I haven't actually made the XDR changes yet (since we're still finalizing them). Once we have consensus I'll update the actual XDR and provide a nice diff.

core/cap-0062.md
SirTyson: Your understanding is correct, I'll reword it.

core/cap-0062.md
SirTyson: Yeah, I meant from a contract invoker's perspective it's handled by RPC. I'll clarify.

core/cap-0062.md
SirTyson: Ya this is the eviction of already archived entries. The Live BucketList at any given time may hold an arbitrary number of archived entries (since the TTL alone dictates this). However the rate at which we evict these archived entries (remove them from the Live BucketList and add them to the Hot Archive BucketList) is restricted via network config settings (specifically there are settings for both max entries to evict and max bytes to evict).

SirTyson: Note that this is an implementation detail, the restore Op looks the same to the invoker whether or not the archived entry has been evicted or not.

Shaptic: Right yep, understood that and hence wanted to clarify that that's exactly what was being discussed here

dmkozh (approved):

PR #1587, opened 2024-12-10
Initial draft cap 63

SirTyson:
This is the initial draft CAP for in-memory Soroban state and automatic entry restoration via `InvokeHostFunctionOp`. It depends on partial state archival from CAP-0062, but not full state archival of CAP-0057.

### DIFF
--- a/core/README.md
+++ b/core/README.md
@@ -97,7 +97,9 @@
 | [CAP-0061](cap-0061.md) | Smart Contract Standardized Asset (Stellar Asset Contract) Extension: Memo | Tomer Weller | Draft |
 | [CAP-0062](cap-0062.md) | Soroban Live State Prioritization | Garand Tyson | Draft |
 | [CAP-0063](cap-0063.md) | Parallelism-friendly Transaction Scheduling | Dmytro Kozhevin | Draft |
+| [CAP-0064](cap-0064.md) | Memo Authorization for Soroban | Dmytro Kozhevin | Draft |
 | [CAP-0065](cap-0065.md) | Reusable Module Cache | Graydon Hoare | Draft |
+| [CAP-0066](cap-0066.md) | Soroban In-memory Read Resource | Garand Tyson | Draft |
 
 ### Rejected Proposals
 | Number | Title | Author | Status |
--- a/core/cap-0066.md
+++ b/core/cap-0066.md
@@ -1,0 +1,489 @@
+```
+CAP: 0066
+Title: Soroban In-memory Read Resource
+Working Group:
+    Owner: Garand Tyson <@SirTyson>
+    Authors: Garand Tyson <@SirTyson>
+    Consulted: Dmytro Kozhevin <@dmkozh>, Nicolas Barry <@MonsieurNicolas>
+Status: Draft
+Created: 2024-12-09
+Discussion: https://github.com/stellar/stellar-protocol/discussions/1585
+Protocol version: 23
+```
+
+## Simple Summary
+
+This proposal introduces a new resource type for Soroban reads, distinguishing between in-memory and
+disk reads. This also proposes automatic restoration for archived entries via `InvokeHostFunctionOp`.
+
+## Working Group
+
+As specified in the Preamble.
+
+## Motivation
+
+By distinguishing disk and in-memory reads, this proposal allows for significant increase in Soroban
+read limits. This distinction also allows for safe automatic entry restoration, significantly improving UX.
+
+### Goals Alignment
+
+This change is aligned with the goal of lowering the cost and increasing the scale of the network.
+
+## Abstract
+
+[CAP-0062](cap-0062.md) introduces partial State Archival, where evicted and live Soroban state is stored in separate databases.
+This separation allows live Soroban state to be efficiently cached entirely in memory, removing disc reads entirely for all
+live Soroban state. Todays eviction scan can then delete entries from both the live BucketList disk and in-memory cache efficiently.
+However, classic state and evicted entries are still subject to disk reads.
+
+This CAP introduces a new resource type for reads, distinguishing between disk reads and in-memory reads. By making this
+distinction at the protocol level, read limits for Soroban data can greatly increase. This also opens the door for other
+optimizations, such as a complete module cache for all live contracts.
+
+Additionally, this CAP introduces automatic restoration via `InvokeHostFunctionOp`, where any archived key present in the
+footprint is automatically restored. Initially in protocol 20, the state archival design was not solidified enough to
+enable automatic restoration, so an explicit restore operation was required. While this was not technically required in
+protocol 20, it was important for contract interfaces to reason properly about restoration such that when full state
+archival was introduce, it would not break preexisting deployments. Given that the full State Archival design has been
+mostly finalized in [CAP-0057](cap-0057.md), it is now appropriate to introduce automatic restoration.
+
+## Specification
+
+### XDR changes
+
+```
+--- a/Stellar-contract-config-setting.x
++++ b/Stellar-contract-config-setting.x
+@@ -26,38 +26,48 @@ struct ConfigSettingContractComputeV0
+ // Ledger access settings for contracts.
+ struct ConfigSettingContractLedgerCostV0
+ {
+-    // Maximum number of ledger entry read operations per ledger
+-    uint32 ledgerMaxReadLedgerEntries;
+-    // Maximum number of bytes that can be read per ledger
+-    uint32 ledgerMaxReadBytes;
++    // Maximum number of disk entry read operations per ledger
++    uint32 ledgerMaxDiskReadEntries;
++    // Maximum number of bytes of disk reads that can be performed per ledger
++    uint32 ledgerMaxDiskReadBytes;
+     // Maximum number of ledger entry write operations per ledger
+     uint32 ledgerMaxWriteLedgerEntries;
+     // Maximum number of bytes that can be written per ledger
+     uint32 ledgerMaxWriteBytes;
+ 
+-    // Maximum number of ledger entry read operations per transaction
+-    uint32 txMaxReadLedgerEntries;
+-    // Maximum number of bytes that can be read per transaction
+-    uint32 txMaxReadBytes;
++    // Maximum number of disk entry read operations per transaction
++    uint32 txMaxDiskReadEntries;
++    // Maximum number of bytes of disk reads that can be performed per transaction
++    uint32 txMaxDiskReadBytes;
+     // Maximum number of ledger entry write operations per transaction
+     uint32 txMaxWriteLedgerEntries;
+     // Maximum number of bytes that can be written per transaction
+     uint32 txMaxWriteBytes;
+ 
+-    int64 feeReadLedgerEntry;  // Fee per ledger entry read
+-    int64 feeWriteLedgerEntry; // Fee per ledger entry write
++    int64 feeDiskReadLedgerEntry;  // Fee per disk ledger entry read
++    int64 feeDiskRead1KB;          // Fee for reading 1KB disk
++    int64 feeWriteLedgerEntry;     // Fee per ledger entry write
+ 
+-    int64 feeRead1KB;  // Fee for reading 1KB
+ 
+     // The following parameters determine the write fee per 1KB.
+-    // Write fee grows linearly until bucket list reaches this size
+-    int64 bucketListTargetSizeBytes;
+-    // Fee per 1KB write when the bucket list is empty
+-    int64 writeFee1KBBucketListLow;
+-    // Fee per 1KB write when the bucket list has reached `bucketListTargetSizeBytes` 
+-    int64 writeFee1KBBucketListHigh;
+-    // Write fee multiplier for any additional data past the first `bucketListTargetSizeBytes`
+-    uint32 bucketListWriteFeeGrowthFactor;
++    // Write fee grows linearly until soroban state reaches this size
++    int64 sorobanStateTargetSizeBytes;
++    // Fee per 1KB write when the soroban state is empty
++    int64 writeFee1KBSorobanStateLow;
++    // Fee per 1KB write when the soroban state has reached `sorobanStateTargetSizeBytes`
++    int64 writeFee1KBSorobanStateHigh;
++    // Write fee multiplier for any additional data past the first `sorobanStateTargetSizeBytes`
++    uint32 sorobanStateWriteFeeGrowthFactor;
++};
++
++// Ledger access settings for contracts.
++struct ConfigSettingContractLedgerCostExtV0
++{
++    // Maximum number of in-memory entry read operations per ledger
++    uint32 ledgerMaxInMemoryReadEntries;
++
++    // Maximum number of in-memory ledger entry read operations per transaction
++    uint32 txMaxInMemoryReadEntries;
+ };
+ 
+ // Historical data (pushed to core archives) settings for contracts.
+@@ -302,7 +312,8 @@ enum ConfigSettingID
+     CONFIG_SETTING_STATE_ARCHIVAL = 10,
+     CONFIG_SETTING_CONTRACT_EXECUTION_LANES = 11,
+     CONFIG_SETTING_BUCKETLIST_SIZE_WINDOW = 12,
+-    CONFIG_SETTING_EVICTION_ITERATOR = 13
++    CONFIG_SETTING_EVICTION_ITERATOR = 13,
++    CONFIG_SETTING_CONTRACT_LEDGER_COST_EXT_V0 = 14
+ };
+ 
+ union ConfigSettingEntry switch (ConfigSettingID configSettingID)
+@@ -335,5 +346,7 @@ case CONFIG_SETTING_BUCKETLIST_SIZE_WINDOW:
+     uint64 bucketListSizeWindow<>;
+ case CONFIG_SETTING_EVICTION_ITERATOR:
+     EvictionIterator evictionIterator;
++case CONFIG_SETTING_CONTRACT_LEDGER_COST_EXT_V0:
++    ConfigSettingContractLedgerCostExtV0 contractLedgerCostExt;
+ };
+ }
+diff --git a/Stellar-ledger.x b/Stellar-ledger.x
+--- a/Stellar-ledger.x
++++ b/Stellar-ledger.x
+@@ -528,12 +528,11 @@ struct LedgerCloseMetaV1
+     // systems calculating storage fees correctly.
+     uint64 totalByteSizeOfBucketList;
+ 
+-    // Temp keys that are being evicted at this ledger.
+-    LedgerKey evictedTemporaryLedgerKeys<>;
++    // TTL and data/code keys that have been evicted at this ledger.
++    LedgerKey evictedKeys<>;
+ 
+-    // Archived restorable ledger entries that are being
+-    // evicted at this ledger.
+-    LedgerEntry evictedPersistentLedgerEntries<>;
++    // Maintained for backwards compatibility, should never be populated.
++    LedgerEntry unused<>;
+ };
+ 
+ union LedgerCloseMeta switch (int v)
+diff --git a/Stellar-transaction.x b/Stellar-transaction.x
+index 7d32481..f966640 100644
+--- a/Stellar-transaction.x
++++ b/Stellar-transaction.x
+@@ -882,16 +882,30 @@ struct SorobanResources
+     // The maximum number of instructions this transaction can use
+     uint32 instructions; 
+ 
+-    // The maximum number of bytes this transaction can read from ledger
+-    uint32 readBytes;
++    // The maximum number of bytes this transaction can read from disk backed entries
++    uint32 diskReadBytes;
+     // The maximum number of bytes this transaction can write to ledger
+     uint32 writeBytes;
+ };
+ 
++struct SorobanResourcesExtV0
++{
++    // Vector of uint8_t indices representing what Soroban
++    // entries in the footprint are evicted, based on the
++    // order of keys provided in the readWrite footprint.
++    opaque<> evictedSorobanEntries;
++};
++
+ // The transaction extension for Soroban.
+ struct SorobanTransactionData
+ {
+-    ExtensionPoint ext;
++    union switch (int v)
++    {
++    case 0:
++        void;
++    case 1:
++        SorobanResourcesExtV0 resourceExt;
++    } ext;
+     SorobanResources resources;
+     // Amount of the transaction `fee` allocated to the Soroban resource fees.
+     // The fraction of `resourceFee` corresponding to `resources` specified 
+```
+
+### Semantics
+
+#### In-memory vs Disk Read Resources
+
+Read resources are now separated between memory backed and disk backed state as follows:
+
+- in-memory entries:
+  - Live Soroban entries
+  - Archived entries that have not yet been evicted (i.e. still exist in the live BucketList)
+- disk entries:
+  - Archived Soroban entries that have been evicted
+  - Classic entries
+
+In-memory entries are subject to the corresponding in-memory limits, but charge no entry or byte based
+read fees. On disk entries are subject to the corresponding limits and also charge entry and byte based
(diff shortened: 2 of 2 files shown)


core/cap-0063.md
```diff
@@ -0,0 +1,496 @@
...
+@@ -26,28 +26,28 @@ struct ConfigSettingContractComputeV0
+ // Ledger access settings for contracts.
+ struct ConfigSettingContractLedgerCostV0
+ {
+-    // Maximum number of ledger entry read operations per ledger
+-    uint32 ledgerMaxReadLedgerEntries;
+-    // Maximum number of bytes that can be read per ledger
+-    uint32 ledgerMaxReadBytes;
++    // Maximum number of live Soroban entry read operations per ledger
++    uint32 ledgerMaxReadLiveSorobanEntries;
```
dmkozh, line 65: Maybe we should name this in a bit more explicit and high-level fashion, like `diskReadLedgerEntries`, `inMemoryLedgerEntries`. This will also allow us to change what's actually on disk and in-memory without changing the names and semantics of these fields.

core/cap-0063.md
```diff
@@ -0,0 +1,496 @@
...
+-    int64 feeRead1KB;  // Fee for reading 1KB
++    int64 feeReadLiveSoroban1KB;    // Fee for reading 1KB of live Soroban state
+ 
+     // The following parameters determine the write fee per 1KB.
+     // Write fee grows linearly until bucket list reaches this size
+@@ -60,6 +60,26 @@ struct ConfigSettingContractLedgerCostV0
+     uint32 bucketListWriteFeeGrowthFactor;
+ };
+ 
++struct ConfigSettingContractLedgerCostExtV1
```
dmkozh, line 100: I don't have too strong opinion, but it might make sense to use V1 struct with all the fields instead of ext here. Or at least call it V0Ext to clarify the connection between these two.

core/cap-0063.md
```diff
@@ -0,0 +1,496 @@
...
++    // The maximum number of bytes of evicted Soroban state
++    // and classic state that can be read
++    uint32 readEvictedAndClassicBytes;
++
++    // Bitmap representing what Soroban entries in the
++    // footprint are evicted, based on the order of keys
++    // provided in the LedgerFootprint vectors. readOnly
++    // vector is first, followed by readWrite.
++    // If no keys are evicted, BitMap can be empty.
++    BitMap evictedSorobanEntries;
```
dmkozh, line 160: We don't need hacky indexing here because only RW entries can be restored.

Also, I've mentioned this already offline, I think we should have a vec of indices of entries that need to be restored.  This will reduce the size of 99% of transactions as almost no transactions need to restore any entries. I think a slight increase in size for the txs that do restore entries is worth it. Also, it will be easier to interpret a sequence of indicies vs an opaque blob.

core/cap-0063.md
```diff
@@ -0,0 +1,496 @@
...
+     SorobanResources resources;
+     // Amount of the transaction `fee` allocated to the Soroban resource fees.
+     // The fraction of `resourceFee` corresponding to `resources` specified 
+diff --git a/Stellar-types.x b/Stellar-types.x
+index 73c4e79..a4b7c7c 100644
+--- a/Stellar-types.x
++++ b/Stellar-types.x
+@@ -134,4 +134,7 @@ struct SerializedBinaryFuseFilter
+     // Array of uint8_t, uint16_t, or uint32_t depending on filter type
+     opaque fingerprints<>;
```
dmkozh, line 184: This change shouldn't belong to this CAP, should it?

core/cap-0063.md
```diff
@@ -0,0 +1,496 @@
...
+ // The transaction extension for Soroban.
+ struct SorobanTransactionData
+ {
+-    ExtensionPoint ext;
++    union switch (int v)
++    {
++    case 0:
++        void;
++    case 1:
++        SorobanResourcesExtV1 resourceExt;
```
dmkozh, line 172: What about the proofs, do we allocate an ext for these?

dmkozh (commented):
I haven't looked through the full CAP yet (and the semantics probably can be discussed separately), but I have some issues with the XDR that probably are easier to resolve now.

core/cap-0063.md
```diff
@@ -0,0 +1,496 @@
...
+ struct ConfigSettingContractLedgerCostV0
+ {
+-    // Maximum number of ledger entry read operations per ledger
+-    uint32 ledgerMaxReadLedgerEntries;
+-    // Maximum number of bytes that can be read per ledger
+-    uint32 ledgerMaxReadBytes;
++    // Maximum number of live Soroban entry read operations per ledger
++    uint32 ledgerMaxReadLiveSorobanEntries;
++    // Maximum number of live Soroban bytes that can be read per ledger
++    uint32 ledgerMaxLiveSorobanBytes;
```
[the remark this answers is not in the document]
SirTyson: I guess I'm just worried about potential "hidden" costs that could bite us later on if we have no limit. These include things like copies, hashing keys if we do something like a hashset for cache implementation, etc. Given how tightly we've set limits on other resources, it seems like a potential footgun if we allow very large footprints that load many in memory keys. Further down the line, if we do end up needing to limit reads for some unforseen implementation reason, we might break contracts if we apply something retroactively.

That being said, not metering or charging for reads is definitely appealing from a UX perspective, and you're more familiar with resource structure than I am, so I'll defer to you wrt risks of having implicit limits.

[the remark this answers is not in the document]
SirTyson: Also those implicit limits only help us once we're inside the VM. Currently core loads all data prior to entering the VM, metering things like instructions and memory would not apply when loading the footprint. We could potentially Cha get this and vector in and out of the VM for in-memory loads, but that seems like a significant implementation change. In the current load everything and pass it to the VM approach a malicious tx could allocate minimum instructions and memory to reduce the fee but fill the footprint to the max size of 64kb WASM keys to maximize the amount of "free" work.

core/cap-0063.md
```diff
@@ -0,0 +1,496 @@
...
+It will also report whether an entry is archived, evicted, or live, and return the entry's current TTL value.
+
+A POST request with the following body:
+
+```http
+ledgerSeq=NUM&key=Base64&key=Base64...
+```
+
+- `ledgerSeq`: An optional parameter, specifying the ledger snapshot to base the query on.
+  If the specified ledger in not available, a 404 error will be returned. If this parameter
```
tamirms, line 343: it seems that a 404 error can be ambiguous because it can be caused by the ledger not being available or by the key not being present.

core/cap-0063.md
tamirms: I agree with @dmkozh I think a separate list of just the evicted entries is easier to understand and will be more space efficient in the average case

core/cap-0063.md
```diff
@@ -0,0 +1,496 @@
...
+break dApps, ad preflight would be required for proof generation and these "one shot" transactions would no longer be valid.
+By maintaining smaller fees for live Soroban state, we encourage developers to preflight Soroban transactions for lower fees
+in order to ensure ecosystem projects do not break with the addition of archival proofs.
+
+TODO: Exact fee and limit calculations
+
+#### `evictedSorobanEntries` BitMap
+
+For the purposes of block creation, the `LedgerKey` alone is sufficient to distinguish between classic and Soroban data.
+However, a disk read would be required prior to tx application in order to determine if a Soroban entry was live or evicted.
```
tamirms, line 263: is the disk read strictly necessary? Couldn't you do an in-memory lookup and the absence of data would indicate that is either evicted or non-existent?

core/cap-0063.md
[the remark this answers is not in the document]
dmkozh: Most of the potential costs are proportional to the count of keys, not the entry size. Given that entries are immutable, I don't see why should we be concerned with the size specifically.

> metering things like instructions and memory would not apply when loading the footprint

'Loading' the footprint is just a bunch of map lookups, isn't it? The metering will kick in as soon as we start decoding the XDR for these (btw it would make sense to store entries as encoded XDR buffers, given that we don't need to interpret them outside of Soroban env). I don't quite see why should this be problematic. And again, I agree that we may still have the limit for the total number of entries accessed per tx and per ledger - the difference is that they're automatically inferred from the footprint, unlike read bytes that need to be accounted for separately.

core/cap-0063.md
SirTyson: The issue is, if something is absent from the cache, it could be archived, or it could just be a new entry. You could imagine a tx that reads the maximum amount of classic state, doesn't restore anything, and write a new soroban entry. This tx should succeed, but when you check the cache for the new entry key, it won't exist. If you conservatively assume non-existent entries are always archived, then you filter out legitimate TXs. If you assume every cache miss is just a non-existent entry, then you will populate the block with TXs guaranteed to fail because they access too much Soroban state. This would limit our ability to construct optimal blocks.

core/cap-0063.md
SirTyson: I didn't include it in this CAP just because they aren't necessary until full state archival.

core/cap-0063.md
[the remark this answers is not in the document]
SirTyson: Ah sorry, I originally misread your comment as  "don't limit in-memory reads at all." Ya I agree limiting entry count but not size is fine.

core/cap-0063.md
tamirms: I thought writing a new entry would still require a disk read because you need to make sure that you are not attempting to create a new entry which already exists. And in order to check if the entry already exists you would need to do a disk lookup to see if the ledger key is in the hot archive bucketlist. I remember in another design doc you mentioned that core would maintain a binary fuse filter to make the duplicate ledger entry check more efficient.

core/cap-0063.md
```diff
@@ -0,0 +1,496 @@
...
+read limits. This distinction also allows for safe automatic entry restoration, significantly improving UX.
+
+### Goals Alignment
+
+This change is aligned with the goal of lowering the cost and increasing the scale of the network.
+
+## Abstract
+
+[CAP-0062](cap-0062.md) introduces partial State Archival, where evicted and live Soroban state is stored in separate databases.
+This separation allows live Soroban state to be efficiently cached entirely in memory, removing disc reads entirely for all
```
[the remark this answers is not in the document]
SirTyson: I agree, the separate BucketList is not strictly necessary for in-memory caching and fee changes, but it does have some tangible benefits. First, implementation becomes simpler on core side. 2nd, it allows us to fragment archival state, which has lower churn, vs live state, which has higher churn. Specifically, live state can be created, modified and deleted, all which require writes. Archival state can only be initially added to the archive via eviction or removed from the archive via restoration, but never updated. This means by separating archival and live state at the DB level, the lower churn archival BucketList can have different spill times, index sizes, etc. compared to the Live BucketList with higher churn. Essentially we have an expensive, small DB for live state, and a less expensive, larger DB for archived state.

That being said, it's still a fairly small optimization relatively speaking. The primary motivation is that we still plan to implement [full state archival](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0057.md) relatively soon, at that requires the Hot Archive. If we need to create the hot archive anyway, and the in-memory cache is easier with the hot archive, it makes more sense to do it know given that there doesn't appear to be strong drawbacks.

core/cap-0063.md
SirTyson: True, but checking if an entry doesn't exist in the database is a very different and much less expensive query then checking if something does exist in the database.

The new entry write process looks as follows:

```python
for each soroban key in readWrite:
    # Could either be archived entry or new entry
    if key not in cache:
        # Check in memory bloom filter for Hot Archive for key
        if hotArchiveBloomFilter.contains(key):
            entry = hotArchive.loadFromDisk(key)
            # Entry does not actually exist, filter false positive
            if not entry:
                continue
            # Entry does exist in archive, fail TX
            else:
                failTx
```

If a new entry actually does exist in the Hot Archive, we do a max of one disk read that the resources didn't pay for, then fail the TX. Note this happens at apply time where we've already charged fees, not a tx queue submission time. There may also be a bloom miss where a non-existent key requires a disk read, but this is rare enough to not be factor. TLDR: In the apply path, we make at most up to 1 non metered disk read and charge fees.

The key point with new entry creation is we assume no entries are archived so we can efficiently leverage the bloom filter and exit early at the first disk read. However, in your proposal we can't exit early at the first entry that is actually archived. Rather, we need to mark every key as non-archived vs archived. So for each archived entry, we need to do a disk read. The bloom filter is guaranteed to return positive for each archived entry, but you still need to do a disk read to see if it's a false positive or not. Additionally, these reads need to happen during tx flooding before we actually charge fees, which is a huge DOS angle.

Note that the bloom filters I'm talking about here are not the bloom filters in the state archival cap. These are local bloom filters used by validator DBs and are not deterministic between validators.

core/cap-0063.md
SirTyson: Did a little digging and we're currently averaging 1.367 keys per restore op. I thought it would be way higher so I agree, vector of indices is better. Should we go uint8 or uint16? Do we ever forsee footprints going beyond 255?

dmkozh: I'd go for u16 just in case. The overhead is marginal, but I'd say accessing 255+ keys is not out of the realm of possibility Though restoring 255+ entries seems unlikely... Still, I feel like it might create unnecessary confusion in the future.

core/cap-0063.md
```diff
@@ -0,0 +1,436 @@
...
+TODO:  In-memory read entries must be measured in order to find a reasonable starting point.
+
+#### `evictedSorobanEntries` Vector
+
+For the purposes of block creation, the `LedgerKey` alone is sufficient to distinguish between classic and Soroban data.
+However, a disk read would be required prior to tx application in order to determine if a Soroban entry was live or evicted.
+This is necessary for determining what resources limits a Soroban key counts towards, but
+exposes a significant DOS angle. To prevent this attack, the footprint must statically declare which entries are live vs. evicted.
+
+This is accomplished via the `evictedSorobanEntries` vector. If any evicted Soroban entry is present in a TX's footprint,
```
tamirms, line 198: what happens if you mark a ledger entry as evicted in the invoke host function operation but during application it turns out the ledger entry has already been restored? will the operation still succeed?

I assume that if a ledger entry is not present in  `evictedSorobanEntries` and it turns out that the ledger entry is evicted, the transaction will fail regardless of whether there are enough fees to pay for the unexpected write. Is that correct?

core/cap-0063.md
leighmcculloch: Are you planning to have 16-bit integer values in the XDR? XDR doesn't support a 16-bit integer type, the smallest integer is a 32-bit, because all values in XDR are padded to 4 bytes, so a 1 or 2 byte value would be 4 bytes anyway.

core/cap-0063.md
SirTyson: Yeah, if you don't mark an entry as evicted you fail. If you do mark it as evicted and it's not, you're spending more in fees and resources, so we shouldn't fail. I'll update this section, but if an entry is marked evicted, we should charge it as if it's evicted no matter what, but not fail if it's actually live.

SirTyson: See added "Incorrect Entries in `evictedSorobanEntries` Vector" section.

core/cap-0063.md
```diff
@@ -0,0 +1,447 @@
...
+
+### Semantics
+
+#### In-memory vs Disk Read Resources
+
+Read resources are now separated between memory backed and disk backed state as follows:
+
+- in-memory entries:
+  - Live Soroban entries
+  - Archived entries that have not yet been evicted (i.e. still exist in the live BucketList)
```
anupsdf, line 169: Should we say `expired` entries here instead of `archived`?

core/cap-0063.md
```diff
@@ -0,0 +1,447 @@
...
+
+```
+--- a/Stellar-contract-config-setting.x
++++ b/Stellar-contract-config-setting.x
+@@ -60,6 +60,46 @@ struct ConfigSettingContractLedgerCostV0
+     uint32 bucketListWriteFeeGrowthFactor;
+ };
+ 
++// Ledger access settings for contracts.
++struct ConfigSettingContractLedgerCostV1
```
anupsdf, line 61: Is it possible to split limits and fees in different structs? That way the one with fees can have cost in the name (ConfigSettingContractLedgerCost) while the other one could be named ConfigSettingContractLedgerLimitV1.

(Probably a question for @dmkozh)

SirTyson: Originally I did that, but Dima suggested one struct instead (I unresolved the comment above for context). I don't care particularly strongly either way.

dmkozh: > Is it possible to split limits and fees in different structs?

We didn't do that initially, and I don't think it makes much sense to change this now. Conceptually the 'cost' here concerns both resource limits ('cost' for the validators) and fees ('cost' for the user).

> Originally I did that, but Dima suggested one struct instead 

FWIW I suggested to use one struct for the setting version migration, this wasn't about separating the fees and limits. But this is the first time we do a setting update and there is no well-defined process yet. Giving this a second thought, I think going the extension way is probably less disruptive for anyone who consumes these settings, as having two versions introduces the risk of using the deprecated value. The name should be `ConfigSettingContractLedgerCostV0Ext` then. Version bumps should probably be reserved for non-incremental and truly breaking changes. Sorry for the confusion, I should have thought more about this initially.

core/cap-0063.md
```diff
@@ -0,0 +1,447 @@
...
+This will include the complete LedgerEntry of the restored entry. It does not matter if the entry has been
+evicted or not, the meta produced will be the same.
+
+If a restored entry is modified during host function invocation, another `LedgerEntryChangeType` will be
+emitted corresponding to the entry change in addition to the `LEDGER_ENTRY_RESTORE` restore event.
+This is similar to the current meta schema, where the starting value of a LedgerEntry is emitted as
+`LEDGER_ENTRY_STATE` followed by the change. This is similar, except that `LEDGER_ENTRY_STATE` is replaced
+by `LEDGER_ENTRY_RESTORE`.
+
+#### Deprecation of RestoreFootprintOp(?)
```
anupsdf, line 246: Worth asking the ecosytem on deprecating this Op. Only thing I can think of is that if admin doesn't want the users to incur the restoration fees and keeps expired/archived entries restored on behalf of the users.

core/cap-0063.md
```diff
@@ -0,0 +1,447 @@
...
+
+#### getledgerentry Endpoint
+
+In order to facilitate state queries for RPC preflight, captive-core will expose the following HTTP endpoint:
+
+`getledgerentry`
+
+Used to query both live and archived LedgerEntries. While `getledgerentryraw` does simple key-value lookup
+on the live ledger, `getledgerentry` will query a given key in both the live BucketList and Hot Archive BucketList.
+It will also report whether an entry is archived, evicted, or live, and return the entry's current TTL value.
```
anupsdf, line 264: Should we refer to the states as live, expired, archived? This way `expired` indicates those still in live BL but not evicted and `archived` refers to those in hot archive BL.

core/cap-0063.md
```diff
@@ -0,0 +1,447 @@
...
+
+- `entries`: A list of entries for each queried LedgerKey. Every `key` queried is guaranteed to
+have a corresponding entry returned.
+- `e`: Either the `LedgerEntry` or `LedgerKey` for a given key encoded as a Base64 string. If a key
+is live or archived, `e` contains the corresponding `LedgerEntry`. If a key does not exist
+(including expired temporary entries) `e` contains the corresponding `LedgerKey`.
+- `state`: One of the following values:
+  - `live`: Entry is live.
+  - `new`: Entry does not exist. Either the entry has never existed or is an expired temp entry.
+  - `archived`: Entry is archived, but not yet evicted, counts towards in-memory resources.
```
anupsdf, line 300: Same thought here if we should call expired for entries in live BL and archived for entries that are evicted from live and are in hot archive.

anupsdf: This way archived entries count towards disk resources while expired ones are in-memory.

core/cap-0063.md
```diff
@@ -0,0 +1,447 @@
...
+it seems highly unlikely that an attacker would be able to DOS the network via in-memory entry copies.
+
+Given that in-memory state lookups are cheap, it also does not seem necessary to have an explicit fee for
+in-memory reads. In-memory reads will already be implicitly charged fees due to instruction and memory costs
+associated with interacting with the entry. An attacker could potentially exploit these implicit fees
+by creating TXs with large readOnly footprints that don't actually interact with the loaded state. The
+TX would consume little resources and would be cheap, but would load the maximum amount of in-memory state
+and avoid these implicit fees. Due to the low cost of these loads, it seems unlikely this would negatively
+affect the network. Even the TX size costs associated with the large footprint may make the attack economically
+nonviable relative to the amount of stress on the network. However, if this were to become and issue,
```
anupsdf, line 336: typo: an issue

core/cap-0063.md
```diff
@@ -0,0 +1,447 @@
...
+     // The maximum number of bytes this transaction can write to ledger
+     uint32 writeBytes;
+ };
+ 
++struct SorobanResourcesExtV0
++{
++    // Vector of uint16_t indices representing what Soroban
++    // entries in the footprint are evicted, based on the
++    // order of keys provided in the readWrite footprint.
++    opaque<> evictedSorobanEntries;
```
anupsdf, line 142: Should we call this `archivedEntries` dropping `Soroban` in the name since its already inside _Soroban_ResourcesExtV0 struct?

core/cap-0063.md
```diff
@@ -0,0 +1,447 @@
...
+
+Note that the RPC itself is not required to implement this behavior, but will call captive-core endpoints.
+
+### Host Functions that are too expensive for automatic restore
+
+Suppose an expensive contract is written, such that it uses the maximum in-memory resources and the maximum
+disk resources when no entries are evicted. Should any Soroban entry be evicted, automatic
+restoration would not be possible, as the restoration would exceed resource limits. In this case, it would be
+necessary to manually issue a `RestoreFootprintOp` (or equivalent in the case of deprecation) prior to the
+host function invocation.
```
anupsdf, line 364: One reason to not deprecated `RestoreFootprintOp`.

core/cap-0063.md
```diff
@@ -0,0 +1,447 @@
...
+### Memory Based DOS Attack
+
+By distinguishing between disk backed and memory backed reads at the protocol level, it is now required that
+validators maintain all live Soroban state in memory. A significant increase in live Soroban state could
+lead to an OOM based attack on the network.
+
+However, this is not a significant concern. The `sorobanLiveStateTargetSizeBytes` (see [CAP-0062](cap-0062.md))
+provides a soft cap on the amount of live Soroban state that can exist at any given time. This prevents any
+sort of runaway memory issue. Additionally, because of the eviction process, there is a natural back pressure
+applied to in-memory state, the rate at which can be controlled via network config settings. Finally, the rate
```
anupsdf, line 437: Maybe we should highlight this on top that background eviction will remove in-memory Soroban entries.

MonsieurNicolas:
CAP needs to be renumbered. Please avoid discussions in PRs and prefer discussions to make it easier to retroactively understand points that were debated

core/cap-0063.md
SirTyson: This is not removing the functionality, but just deprecating the operation. If we automatically restore archived entries in the footprint, we can do this via the `ExtendFootprintOp`, where the extension would just have a ttl value of 0 to mimic today's `RestoreFootprintOp`.

core/cap-0063.md
SirTyson: Given the significant feedback we heard when we initially called this project "State Expiration," I'd like to stay away from that term and keep the phrasing as "live", "archived", and "evicted". Generally speaking, the "evicted" state will be abstracted away from users anyway such that "live" and "archived" are really all that's relevant to developers and users anyway.

SirTyson:
Addressed some of the comments, but let's merge this and move this over to discussion per Nico.

dmkozh (approved):

dmkozh (approved):

PR #1593, opened 2024-12-12
Update SEP statuses

chowbao:
* Update SEP readme with the current statuses of SEPs committed in the repo
* New status names
  * `Active` --> `Final`
  * `Rejected` --> `Abandoned`

### DIFF
--- a/ecosystem/README.md
+++ b/ecosystem/README.md
@@ -31,56 +31,62 @@
 
 ## List of Proposals
 
-| Number                  | Title                                                                  | Author                                                        | Track    | Status                                                            |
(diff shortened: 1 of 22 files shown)


chowbao:
TODO: will update the old SEPs in draft as I read through the threads and see if they should be revived or abandoned

leighmcculloch:
If rejected is being replaced with abandoned (which I support), the README docs describing the statuses should be updated. Fwiw I think it's worth doing that change alone in a single PR, so that the what and why are hyperfocused on one decision. One decision, one PR, is easier for folks to track why something is happening, and debate it in isolation of other decisions.

PR #1600, opened 2025-01-03
[ANCHOR-812] Deprecate `on_chagne_callback` of SEP-6 and SEP-24  in favor of `STATUS_CALLBACK_URL` of SEP-1

lijamie98:
Currently, in AP we configure wallet callback in the config and don’t support passing it as a request parameter. However, in the SEP that’s still the case. We should discuss the strategy, and potentially deprecate/remove callback argument from request parameters in favor of communicating URL between Anchor and Wallet (potentially utilizing SEP-1) 

In the withdraw and deposit requests, the wallet may optionally provide the URL for the anchor to call back when the transaction status is changed. However, this may pose a potential threat when requesting the anchor's server to call an arbitrary server externally. This issue was raised by one of the anchors requesting AP  not to allow the `on_change_callback` from the withdraw/deposit requests but instead add the wallet callback URL in the configuration.

In this PR, the [SEP-6/on_change_callback](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0006.md#request) and [SEP-24/on_change_callback](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0024.md#adding-parameters-to-the-url) are deprecated. The SEP-1 is amended to specify the wallet's callback URLs in [the General Information of the TOML](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0001.md#general-information) in wallet's stellar.toml file.

Slack Discussion: https://stellarfoundation.slack.com/archives/C036T11AY84/p1733509062701459


ecosystem/sep-0006.md
```diff
@@ -183,11 +183,25 @@ Access-Control-Allow-Origin: *
...
 This protocol involves the optional use of callbacks that the Anchor can issue to update the wallet on `status` of a
-transaction. In order to validate the integrity and provenance of the callback, the Anchor MUST include a signature in
-the HTTP Header `Signature` or `X-Stellar-Signature` (**Deprecated**).
+transaction.
+
+### On Change Callback URL
+
+The wallet server can optionally provide the Anchor with a URL to send a callback to when the status of a transaction
+has changed via the `STATUS_CALLBACK_URL` field in the wallet's [`stellar.toml`](sep-0001.md). The Anchor should POST a
+JSON message to when the status property of the transaction created as a result of this request changes. The JSON
```
philipliu, line 195: > The Anchor should POST a JSON message to when the status property of the transaction created as a result of this request changes.

Not sure I understand this sentence. What does "this request" refer to?

ecosystem/sep-0006.md
```diff
@@ -183,11 +183,25 @@ Access-Control-Allow-Origin: *
...
+message should be identical to the response format for the /transaction endpoint. The callback needs to be signed by the
+anchor and the signature needs to be verified by the wallet according to the callback signature specification.
+
+The payload of the callback should be the same as the response of the `GET /transaction` endpoint.
```
philipliu, line 199: This is repeated in the paragraph above.

ecosystem/sep-0006.md
```diff
@@ -403,24 +418,24 @@ GET TRANSFER_SERVER/deposit
...
+| `lang`                        | string                  | (optional) Defaults to `en` if not specified or if the specified language is not supported. Language code specified using [RFC 4646]. `error` fields and other human readable messages in the response should be in this language.                                                                                                                                                                                                                       |
+| `on_change_callback`          | string                  | (**deprecated** in favor of `STATUS_CALLBACK_URL` of the [SEP-1](sep0001.md)).                                                                                                                                                                                                                                                                                                                                                                           |
```
philipliu, line 430: For other deprecated fields, we leave the description. Do we want to do that here?

ecosystem/sep-0001.md
```diff
@@ -74,6 +74,7 @@ These are global fields in the `stellar.toml` file.
 | URI_REQUEST_SIGNING_KEY | Stellar public key     | The signing key is used for [SEP-7](sep-0007.md) delegated signing                                                                                        |
 | DIRECT_PAYMENT_SERVER   | uses `https://`        | The server used for receiving [SEP-31](sep-0031.md) direct fiat-to-fiat payments. Requires [SEP-12](sep-0012.md) and hence a `KYC_SERVER` TOML attribute. |
 | ANCHOR_QUOTE_SERVER     | uses `https://`        | The server used for receiving [SEP-38](sep-0038.md) requests.                                                                                             |
+| STATUS_CALLBACK_URL     | uses `https://`        | The URL used for receiving [SEP-6](sep-0006.md) and [SEP-24](sep-0024.md) transaction status change callbacks.                                            |
```
philipliu, line 77: We should specify that including this field in the `stellar.toml` does not guarantee that Anchors will send callback events to the wallet server. Anchors are not required to support callbacks and some of them only send callbacks to URLs that are whitelisted.

ecosystem/sep-0006.md
lijamie98: Rephrased to avoid the confusion.

ecosystem/sep-0001.md
lijamie98: This makes sense

lijamie98: Changes the `### On Change Callback URL` section of the related SEPs.

ecosystem/sep-0024.md
```diff
@@ -203,12 +203,10 @@ transaction.
...
 
 The wallet server can optionally provide the Anchor with a URL to send a callback to when the status of a transaction
-has changed via the `STATUS_CALLBACK_URL` field in the wallet's [`stellar.toml`](sep-0001.md). The Anchor should POST a
-JSON message to when the status property of the transaction created as a result of this request changes. The JSON
-message should be identical to the response format for the /transaction endpoint. The callback needs to be signed by the
-anchor and the signature needs to be verified by the wallet according to the callback signature specification.
-
-The payload of the callback should be the same as the response of the `GET /transaction` endpoint.
+has changed via the `STATUS_CALLBACK_URL` field in the wallet's [`stellar.toml`](sep-0001.md). Depending on the policy,
+the Anchor may POST a JSON message to the callback URL when the status property of the transaction changes. The JSON
```
philipliu, line 207: > Depending on the policy, the Anchor may POST a JSON message to the callback URL when the status property of the transaction changes.

It's not clear what the policy refers to, so I think we can remove it. We can rephrase it to something like "The Anchor should POST a JSON message to the callback URL when the transaction's status changes`.

PR #1618, opened 2025-01-15
Fix broken link to SAC's event definitions

Shaptic:


leighmcculloch (commented):
Is the PR missing the fix? The change looks like a no-op functionally, and nothing is changing about SAC event definitions.

Shaptic:
@leighmcculloch notice the casing in the link: `CAP` doesn't work, `cap` does

leighmcculloch (approved):

PR #1630, opened 2025-01-22
SEP-45: Fix server auth entry verification

philipliu:
The server does not know about the origin request when verifying the challenge returned by the client. It cannot verify that the `account` matched the original request. This also adds a check for the presence of the `client_domain`.


leighmcculloch (commented):
The server should know that the values matched an earlier request because the server signed the original challenge it generated and all the components related to it. If that  correct, then it should still be possible for the server to have confidence?

Could you share an example / pseudo-example of what the generated auth entries would be at the first step and what the server, what the client would add, and what the server would validate?

joetosly414 (approved):

joetosly414:
What am I suppose to do I have my old pc in California who I have my good workingon it and I have the laptop here in Wisconsin

PR #1639, opened 2025-01-30
SEP-45: Replay prevention + argument name fixes

philipliu:
- Rename `Home Domain Address` to `Server Account`
- Rename `Client Domain Address` to `Client Domain Account`
- Rename `home_domain_address` argument to `web_auth_domain_account`
- Rename `client_domain_address` argument to `client_domain_account`
- Add `home_domain` argument to `web_auth_verify` function
- Mark nonce verification as required and add a replay prevention section
- Clarify client domain signing sentence

### DIFF
--- a/ecosystem/sep-0045.md
+++ b/ecosystem/sep-0045.md
@@ -76,10 +76,10 @@ The authentication flow is as follows:
    1. The `account` value matches the **Client Account** address
    1. The `home_domain` value matches the **Home Domain**
    1. The `web_auth_domain` value matches the **Server**'s domain
-   1. The `web_auth_domain_signer` value matches the **Server Account**
+   1. The `web_auth_domain_account` value matches the **Server Account**
    1. If the **Client** included a `client_domain` in the request:
       1. The `client_domain` value matches the **Client**'s domain
-      1. The `client_domain_signer` value matches the **Client Domain Account**
+      1. The `client_domain_account` value matches the **Client Domain Account**
 1. The **Client** verifies that there is an authorization entry where `credentials.address.address` is the **Server
    Account** and contains a valid signature from the **Server Account**
 1. The **Client** signs the authorization entry where `credentials.address.address` is the **Client Account** using the
@@ -106,8 +106,8 @@ The authentication flow is as follows:
    1. The `account` value matches the **Client Account** address
    1. The `home_domain` value matches the **Home Domain**
    1. The `web_auth_domain` value matches the **Server**'s domain
-   1. The `web_auth_domain_signer` value matches the **Server Account**
-   1. The `client_domain_signer` value matches the **Client Domain Account** if the **Client** included a
+   1. The `web_auth_domain_account` value matches the **Server Account**
+   1. The `client_domain_account` value matches the **Client Domain Account** if the **Client** included a
       `client_domain` in the request, otherwise it is not present
 1. The **Server** verifies that the `nonce` argument is the same across all authorization entries and is unique
 1. The **Server** verifies that there is an authorization entry where `credentials.address.address` is the **Server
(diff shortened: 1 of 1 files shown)


marcelosalloum:
I found an occurrence of **Home Domain Address** here:
https://github.com/philipliu/stellar-protocol/blob/6abbb17c6ec833eb282b11a2a656039b08e113ab/ecosystem/sep-0045.md?plain=1#L60-L61

ecosystem/sep-0045.md
```diff
@@ -75,27 +75,27 @@ The authentication flow is as follows:
...
       1. The `client_domain` value matches the **Client**'s domain
-      1. The `client_domain_address` value matches the **Client Domain Address**
-1. The **Client** verifies that there is an authorization entry where `credentials.address.address` is the **Home Domain
-   Address** and contains a valid signature from the **Home Domain Address**
+      1. The `client_domain_signer` value matches the **Client Domain Account**
+1. The **Client** verifies that there is an authorization entry where `credentials.address.address` is the **Server
+   Account** and contains a valid signature from the **Server Account**
 1. The **Client** signs the authorization entry where `credentials.address.address` is the **Client Account** using the
    secret key(s) of the signer(s) for the **Client Account**
    - Note: **Client** signatures may not be required if the contract's `__check_auth` implementation does not require
```
marcelosalloum, line 87: Can we make it clear that this `contract` refers to the Client's contract account? When I read it I assumed this was referring to the SEP-45 contract.

philipliu: Done!

ecosystem/sep-0045.md
```diff
@@ -397,35 +394,44 @@ impl WebAuthContract {
...
 
-### Signature Expiration Ledger
+### Setting Signature Expiration Ledger
 
-When signing the authorization entries, the **Client** should set the signature expiration ledger to the current ledger
-plus a buffer. This ensures the signature remains valid for an appropriate duration while preventing replay attacks.
-Setting this to the current ledger plus 1 is a reasonable default.
+The **Client** should set the signature expiration ledger to a near-future ledger, minimizing the window where a
+signature can be reused if nonces verification is improperly implemented. Setting this to the current ledger plus 1 is a
+reasonable default.
```
leighmcculloch, line 434: I think we should tweak the replay prevention.

Nonces are a feature available. A server should always provide a nonce so that the experience of using the APIs is consistent for clients and so that every signature is forced to be unique even for the same invocation otherwise. However storing all nonces and verifying them for uniqueness may not be practical for all implementations, so servers may choose not to use that approach.

However, the server should be able to use its signature with its signature expiration ledger in the same way the client does to limit the time/ledger window that it's own signature is valid for. Or, if that is not possible for some reason the server should be able to reject client signatures that have a signature expiration ledger that is too old.

I guess my point is the client has one way to prevent reuse, sig exp ledg, and the server has two methods: nonce, sig exp ledg. And the server and client should both at least use sig exp ledg.

ecosystem/sep-0045.md
```diff
@@ -254,16 +253,16 @@ fail, then the authentication request must be rejected — that is, treated by t
...
    1. The `home_domain` value matches the **Home Domain**;
-   1. The `home_domain_address` value matches the **Home Domain Address**;
    1. The `web_auth_domain` value matches the **Server**'s domain;
+   1. The `web_auth_domain_signer` value matches the **Server Account**;
    1. The `client_domain` is present if `client_domain_address` is present;
-   1. The `client_domain_address` value matches the **Client Domain Address** if `client_domain` is present;
-1. (Optional) Verify that the `nonce` argument is the same across all authorization entries and is valid;
-1. Verify that there is an authorization entry where `credentials.address.address` is the **Home Domain Address** and
-   contains a valid signature from the **Home Domain Address**;
+   1. The `client_domain_signer` value matches the **Client Domain Account** if `client_domain` is present;
```
leighmcculloch, line 259: I think using the term signer here can be ambiguous. A signer is usually the key signing. For example in classic accounts the account and account signers can be different keys. However, it says here the signer is the account, which may not be the case.

I think we should name the field to match the name, i.e. `client_domain_account`.

Lmk if I'm missing something where this doesn't make sense.

ecosystem/sep-0045.md
```diff
@@ -254,16 +253,16 @@ fail, then the authentication request must be rejected — that is, treated by t
 1. Verify that the `args` in each authorization entry match the expected values and is the same across all authorization
    entries:
    1. The `home_domain` value matches the **Home Domain**;
-   1. The `home_domain_address` value matches the **Home Domain Address**;
    1. The `web_auth_domain` value matches the **Server**'s domain;
+   1. The `web_auth_domain_signer` value matches the **Server Account**;
```
leighmcculloch, line 257: See comment below about signer vs account ambiguity.

leighmcculloch (commented):
👏🏻 Couple comments inline.

ecosystem/sep-0045.md
philipliu: Done!

ecosystem/sep-0045.md
philipliu: Makes sense. Done!

leighmcculloch (approved):

PR #1646, opened 2025-02-07
SEP-23: Add Liquidity Pool and Claimable Balance Strkeys

leighmcculloch:
### What
Add two new strkey types for liquidity pool and claimable balance, including examples of valid and invalid addresses.

### Why
In support of the upcoming changes to SEP-23 for CAP-67.

### Related

- https://github.com/stellar/rs-stellar-strkey/pull/74

### Todo

- [x] Add invalid test cases.
- [x] Validate with rs-stellar-strkey implementation, tests, and fuzzing. 
- [ ] Get agreement and sign off prior to merge from reviewers involved in CAP-67.

### DIFF
--- a/ecosystem/sep-0023.md
+++ b/ecosystem/sep-0023.md
@@ -7,8 +7,8 @@ Author: David Mazières, Tomer Weller <@tomerweller>, Leigh McCulloch <@leighmcc
 Track: Standard
 Status: Active
 Created: 2019-09-16
-Updated: 2022-12-02
-Version: 1.2.0
+Updated: 2025-02-07
+Version: 1.3.0
 Discussion: https://groups.google.com/g/stellar-dev
 ```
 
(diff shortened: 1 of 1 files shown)


dmkozh (approved):

ecosystem/sep-0023.md
```diff
@@ -186,6 +191,37 @@ software to accept the invalid test cases, which could in turn cause security pr
...
+       0x1c, 0x83, 0x8a, 0xad, 0x97, 0x34, 0xa4, 0xa2,
+       0xfb, 0x0d, 0x7a, 0x03, 0xfc, 0x7f, 0xe8, 0x9a,
+   }
+   ```
+
+1. Valid claimable balance address
+
+   - Strkey `BAAD6DBUX6J22DMZOHIEZTEQ64CVCHEDRKWZONFEUL5Q26QD7R76RGR4TU`
+   - type: `HASH_TYPE_SHA256`
+   - claimable balance type: `v0`
```
karthikiyer56, line 213: Do we need to include the CB type field as a 1 byte value when generating strkey for claimable balances?
Currently, yes, it is always set to 0. do we anticipate having another type of CB in the future, i.e `v1` ?

ecosystem/sep-0023.md
```diff
@@ -65,7 +67,10 @@ The following steps transform a binary key into a strkey:
 1.  Start with the appropriate version byte computed by the OR of a key type base value and algorithm selector from the
     tables above.
 
-2.  Append the binary bytes of the key (e.g., 32-bytes for ED25519).
+2.  Append the binary bytes of the key (e.g., 32-bytes for ED25519, 32-bytes for SHA256).
+
+    If we are encoding a claimable balance, the binary bytes of the key has a length of 33-bytes: 1-byte value
```
karthikiyer56, line 72: Do we need to include the CB type field as a 1 byte value when generating strkey for claimable balances?
Can we not simply omit adding the CB type?
Currently, yes, it is always set to 0. do we anticipate having another type of CB in the future, i.e `v1` ?


Also, the claimable balance type in XDR is 4 bytes. 
If we do indeed choose to include the type in the strkey representation, shudnt be be including 4 bytes, instead of 1, just for consistency purposes?

leighmcculloch: In the XDR the type field is 4-bytes because XDR doesn't represent any field in less than 4-bytes. It is highly unlikely we'd ever store more than 1-byte of data in the field, so I don't think it is necessary for us to allocate 4-bytes to the type field.

leighmcculloch: > Do we need to include the CB type field as a 1 byte value when generating strkey for claimable balances? Can we not simply omit adding the CB type? Currently, yes, it is always set to 0. do we anticipate having another type of CB in the future, i.e `v1` ?

The 1-byte 0x00 is easy enough to add to avoid us needing to create a new strkey in the future for claimable balances v1 if they ever happen.

@dmkozh @sisuresh Do you folks have an opinion on @karthikiyer56's question?

dmkozh: I think 1 byte is a reasonable tradeoff, even if we don't have v1 in plans.

karthikiyer56 (approved):

ecosystem/sep-0023.md
leighmcculloch, line 72: @stanford-scs The claimable balance strkey will not be padded to a multiple of 4. For some other strkeys like the signed payload we did padding, but I don't remember why or if that is truly important. I'm inclined to not pad for smaller strkeys in this case.

PR #1654, opened 2025-02-13
Remove contract spec and meta XDR files from CAP-46

leighmcculloch:
### What
Remove two XDR files (contract-spec.x and contract-meta.x) from CAP-46.

### Why
The contract spec and contract meta XDR are not part of the Stellar protocol. No part of the Soroban Environment or Stellar Core depends on any detail of either. The protocol does depend on the definition and validatity of the contract env meta XDR, but it is distinct and separate from the contract spec and contract meta XDR.

The contract spec and contract meta are ecosystem standards that are, or will, be captured in SEPs, not a CAP.

I don't think we intentionally included them in CAP-46 more than they were just part of the diff at the time and became included by accident. Neither are discussed anywhere in the doc or have any semantic value other than being listed as part of the diff.

So as not to conflate that the spec and meta are part of the protocol, I suggest patching CAP-46 as is done in this change to simply remove their mention.

Related:
- https://github.com/stellar/stellar-protocol/pull/1655


core/cap-0046.md
```diff
@@ -107,11 +107,9 @@ the following sub-CAPs:
...
 ### XDR changes
 
-There are four entirely new XDR files:
+There are three entirely new XDR files:
 
   - [Stellar-contract.x](../contents/cap-0046/Stellar-contract.x)
-  - [Stellar-contract-spec.x](../contents/cap-0046/Stellar-contract-spec.x)
-  - [Stellar-contract-meta.x](../contents/cap-0046/Stellar-contract-env-meta.x)
   - [Stellar-contract-env-meta.x](../contents/cap-0046/Stellar-contract-env-meta.x)
   - [Stellar-contract-config-setting.x](../contents/cap-0046/Stellar-contract-config-setting.x)
```
leighmcculloch, line 114: >four

Classic off-by-one error. 🤣 

cc @graydon

PR #1655, opened 2025-02-13
SEP-46: Contract Meta

leighmcculloch:
### What
Retroactively add a SEP for the contract meta as it is already used in the soroban-sdk and stellar-cli.

### Why
The soroban-sdk and stellar-cli already implement a way that any contract can be attached with meta. The ecosystem is already using this, and other proposals are forming (thanks @OrbitLens!) that are using the meta. It would be helpful to document the behavior of the contract meta so that it does not remain as tribal / hidden knowledge, and so that as other contract SDKs are built there is a clear outline of how they can be compatible and seamless as to providing meta.


ElliotFriend:
Is the intent here to outline or suggest _which_ `contractmeta` entries are common or useful? Or just to describe the _way_ `contractmeta` works? I've typically thought of `contractmeta` being a title/description kind of thing, so I'm curious about what the "regular" way to include that information might be (or even _if_ it's done).

As you mentioned in #1659, `sep` doesn't conflict with anything in use (at time of writing). I'm wondering what meta entries _are_ in use for deployed pubnet contracts.

leighmcculloch:
> describe the _way_ `contractmeta` works

Just describe how it works, as a point of interoperability.

tomerweller (approved):
makes perfect sense

PR #1657, opened 2025-02-13
Removed evicted state from cap 66

SirTyson:
Updates CAP 66 as follows:

- Removes "evicted" state from protocol, all archived entries are considered disk reads
- Clarifies meta changes for entries that are restored and modified in the same transaction
- adds detail on changes required for `restorePreamble` for TXs that restore too many entries for auto restore to work
- Discuses DOS attacks and mitigations wrt in memory read limits

### DIFF
--- a/core/cap-0066.md
+++ b/core/cap-0066.md
@@ -114,9 +114,6 @@ mostly finalized in [CAP-0057](cap-0057.md), it is now appropriate to introduce 
 +// Ledger access settings for contracts.
 +struct ConfigSettingContractLedgerCostExtV0
 +{
-+    // Maximum number of in-memory entry read operations per ledger
-+    uint32 ledgerMaxInMemoryReadEntries;
-+
 +    // Maximum number of in-memory ledger entry read operations per transaction
 +    uint32 txMaxInMemoryReadEntries;
  };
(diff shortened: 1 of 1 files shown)


core/cap-0066.md
```diff
@@ -228,18 +224,23 @@ up to `txMaxReadEntries - 2` classic entries today (all classic entries + wasm +
 carefully measured assuming disk access only, so the addition of in-memory state should not negatively affect the network
 in any way should we maintain these values via disk read limits.
 
-TODO:  In-memory read entries must be measured in order to find a reasonable starting point.
+In-memory read limits are not required wrt execution time, as they are very inexpensive. However, the size of TX footprints
+does impact the cost and complexity of assembling transactions sets. For this reason, no ledger wide read limit is necessary,
+and no byte based read limit is necessary. However, a tx entry limit is still necessary to ensure efficient transaction set
+construction.
+
+TODO: Determine initial tx in-memory read limit.
```
dmkozh, line 232: This value should be just inherited from the current value of `txMaxReadLedgerEntries` at the moment of network upgrade. It is the minimum viable value (or else we're risking to break someone).

dmkozh (approved):

core/cap-0066.md
SirTyson: Sounds reasonable. I'll merge for now since we'll need to update this anyway with the new default values for write fees anyway with changing BucketList average to Soroban state size average

PR #1663, opened 2025-02-24
sorobanLiveStateTargetSizeBytes now measures in-memory live soroban state

SirTyson:
Changes `sorobanLiveStateTargetSizeBytes` from tracking on-disk Soroban state to tracking in-memory live Soroban state.

### DIFF
--- a/core/cap-0062.md
+++ b/core/cap-0062.md
@@ -203,11 +203,46 @@ entries can be restored from both the Live State and Hot Archive BucketList, bec
 continue to exist in Live State until they are evicted as part of natural BucketList growth. The emitted
 meta does not distinguish between these.
 
+#### `sorobanLiveStateTargetSizeBytes`
+
+`bucketListTargetSizeBytes` has been renamed to `sorobanLiveStateTargetSizeBytes`. Wrt to fees and limits,
+this value will be used in the same way. However, the value of `sorobanLiveStateTargetSizeBytes` is calculated
+differently.
+
+The value of `sorobanLiveStateTargetSizeBytes` is the total size of live Soroban state in the ledger, in bytes.
+Previously, `bucketListTargetSizeBytes` was based on the total size of the BucketList. In addition to no longer
+counting classic state, `sorobanLiveStateTargetSizeBytes` is also now based on the "ledger state" instead of on-disk
+size of entries. The "ledger state" corresponds to live `CONTRACT_DATA` entries, `TTL` entries for both `CONTRACT_DATA`
+and `CONTRACT_CODE`, and instantiated contract code. Archived and deleted state is not included, even if a `BucketEntry`
+for the archived or deleted state exists in the Live BucketList.
+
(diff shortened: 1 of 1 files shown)


core/cap-0062.md
```diff
@@ -203,12 +203,47 @@ entries can be restored from both the Live State and Hot Archive BucketList, bec
...
+`sorobanLiveStateTargetSizeBytes` is the sum of the size of each live `CONTRACT_DATA` entry, `CONTRACT_WASM` entry, and
+corresponding `TTL` entry, where the size of each entry is calculated as follows:
+ - `CONTRACT_DATA`: XDR serialized size of `CONTRACT_DATA` LedgerEntry.
+ - `TTL`: XDR serialized size of `TTL` LedgerEntry.
+ - `CONTRACT_WASM`: The memory cost of the given code based on the module's `ContractCodeCostInputs`, as calculated by the cost model.
+    Note that this is usually much larger than the `CONTRACT_WASM` LedgerEntry size.
+
+#### Changes to `CONTRACT_WASM` Fees and Limits
+
+`contractMaxSizeBytes` will now be based on the size of instantiated contract code, not the `CONTRACT_WASM` LedgerEntry size.
```
dmkozh, line 228: I don't think there is a good way for the safe initial value of this setting. I'm not really sure if this change is necessary, given that issue.

core/cap-0062.md
```diff
@@ -203,12 +203,47 @@ entries can be restored from both the Live State and Hot Archive BucketList, bec
...
+for the archived or deleted state exists in the Live BucketList.
+
+`sorobanLiveStateTargetSizeBytes` is the sum of the size of each live `CONTRACT_DATA` entry, `CONTRACT_WASM` entry, and
+corresponding `TTL` entry, where the size of each entry is calculated as follows:
+ - `CONTRACT_DATA`: XDR serialized size of `CONTRACT_DATA` LedgerEntry.
+ - `TTL`: XDR serialized size of `TTL` LedgerEntry.
+ - `CONTRACT_WASM`: The memory cost of the given code based on the module's `ContractCodeCostInputs`, as calculated by the cost model.
+    Note that this is usually much larger than the `CONTRACT_WASM` LedgerEntry size.
+
+#### Changes to `CONTRACT_WASM` Fees and Limits
```
dmkozh, line 226: I think this update should also include the changes to write fee and rent computations for the contract code entries, as I assume these are affected as well.

core/cap-0062.md
SirTyson: Well, could we not just see what the largest instantiated contract currently is on main net and set it to that + some buffer? It's possible that a more expensive contract could be uploaded in between cutting the release and the vote, but it should be a better starting point than what we have today, which is probably too low for many contracts

dmkozh: I think we need to discuss this separately, as there is a few different things to consider. There is no need to block this update though.

dmkozh (approved):

PR #1682, opened 2025-03-14
Fix the status on CAPs 62-69 and the protocol version

anupsdf:
### What
Flip `Protocol Version` of draft CAPs to TBD and move the statuses to Awaiting Decision. 
Once CAP is accepted and implemented we would set the Protocol Version that they would be available in.


MonsieurNicolas (approved):

REDMEE1:
[
IMG_0682
](url)

PR #1699, opened 2025-04-10
SEP-51: XDR-JSON Specification

leighmcculloch:


ecosystem/sep-XDRJSON.md
```diff
@@ -0,0 +1,1058 @@
...
+JSON:
+
+```json
+9223372036854775807
+```
+
+_Note: JavaScript runtime implementations of JSON do not support numbers that
+require more than 53-bits. JavaScript applications should use a JSON decoder
+supporting 64-bit numbers in JSON numbrs. JSON decoders in other languages do
+not typically have this constraint and support numbers up to 64-bits._
```
Shaptic, line 141: Unfortunately, I don't think this is an appropriate constraint to place on the most popular SDK language :/ negating the ability to use the built-in [`JSON.parse`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse) and requiring a separate parsing library leads to all of the classic JS world problems: environment compatibility, increased bundle sizes, etc. (Though they [do exist](https://stackoverflow.com/a/18755261).)

It also counters conventions in Horizon and RPC which encode all >= 64-bit values as strings. (Separately, I can't find the discussion anymore but I do think 32-bit integers should stay `number`.)

leighmcculloch: Yup, that's why we have this open issue to change it, and also change the presentation of the 128 and 256 bit integer types too:
- https://github.com/stellar/xdrgen/issues/181#issuecomment-2819617408

This PR captures the existing XDR-JSON format which has this wart unfortunately.

tomerweller (approved):

PR #1720, opened 2025-04-29
CAP 70 - Config params for SCP timing

SirTyson:


core/cap-0070.md
```diff
@@ -0,0 +1,141 @@
...
+@@ -362,5 +371,7 @@ case CONFIG_SETTING_CONTRACT_PARALLEL_COMPUTE_V0:
+     ConfigSettingContractParallelComputeV0 contractParallelCompute;
+ case CONFIG_SETTING_CONTRACT_LEDGER_COST_EXT_V0:
+     ConfigSettingContractLedgerCostExtV0 contractLedgerCostExt;
++case CONFIG_SETTING_SCP_TIMING:
++    ConfigSettingSCPTiming contractSCPTiming;
+ };
+```
+
+## Semantics
```
dmkozh, line 86: Can you mention here that the initial values match the current (hardcoded) values? I saw this described below, but it also would be appropriate here.

core/cap-0070.md
```diff
@@ -0,0 +1,141 @@
...
+
+- **ledgerTargetCloseTimeMilliseconds**
```
dmkozh, line 88: I think this CAP should also define the min/max valid ranges for these.

MonsieurNicolas (approved):

core/cap-0070.md
```diff
@@ -0,0 +1,141 @@
...
+- Per-round nomination timeout increment (milliseconds)
+- Per-round ballot timeout increment (milliseconds)
+
+These settings are introduced as ledger config parameters, allowing the Stellar network to adjust consensus timings incrementally.
+
+## Specification
+
+### XDR Changes
+
+```diff
```
leighmcculloch, line 49: What is this diff based on? I tried to apply it to the curr and next branches of stellar-xdr with no success.

SirTyson: This is based on next, `8903b65`. I removed some of the formatting from the actual diff, I'll add it back. I think that's probably what's giving you issues.

dmkozh: FWIW you have to use patch for this, as CI doesn't seem to pass otherwise.

core/cap-0070.md
SirTyson: I can put in like rough min/max based on my intuition, but actual min/maxes will require some significant experimentation.

core/cap-0070.md
SirTyson: I think CI has in issue. I tried to give it the stellar-xdr next branch, but it failed because it couldn't find the ref. It looks like the tool only looks at stellar-core and expects a stellar-core commit. I don't think this makes much sense, since core often lags behind XDR (and does so currently). I had to add `mddiffcheck.ignore=true` to pass the test.

dmkozh (approved):

core/cap-0070.md
SirTyson: Ahh I think you have to pass the repo manually, default base is stellar-core. Let me give it another try

SirTyson: Hmm so the tool supports specifying a repo, but the GI actions has stellar-core repo hardcoded. I've modified the action with a slightly hacky script that runs the check against both core (to preserve preexisting hashes) and stellar-xdr and fails only if both fail

leighmcculloch (approved):

.github/workflows/mddiffcheck.yml
```diff
@@ -16,4 +16,12 @@ jobs:
...
-    - run: mddiffcheck -repo https://github.com/stellar/stellar-core core/*.md
+    - run: |
+          failed=0
+          for f in core/*.md; do
+            mddiffcheck -repo https://github.com/stellar/stellar-core "$f" || \
+            mddiffcheck -repo https://github.com/stellar/stellar-xdr "$f" || {
+              failed=1
+            }
+          done
+          exit $failed
```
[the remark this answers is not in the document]
leighmcculloch: Follow up updated mdiffcheck to handle multiple repos better:
- https://github.com/stellar/mddiffcheck/pull/10
- https://github.com/stellar/stellar-protocol/pull/1722

PR #1727, opened 2025-05-06
SEP-51: Encode 128/256bit integer types as strings

leighmcculloch:
### What
Encode 128/256bit integer types as strings.

### Why
See https://github.com/stellar/stellar-protocol/discussions/1702#discussioncomment-12964633


leighmcculloch:
Combined into #1729.

PR #1735, opened 2025-05-15
🐛 Capitalization style of network modifiers

JFWooten4:
In conversation with Bri Wylde, I've affirmed that it's best practice to keep network/protocol lowercase when modifying `Stellar`. The existing `README`s defer to this style in [CAP#L168](https://github.com/stellar/stellar-protocol/blob/f25395dab5dfb1282599ee6b0b89a6f81d9f5b1d/core/README.md?plain=1#L168) and the root README. However, some reminiscent double-capitalizations remain, which this PR fixes.

### DIFF
--- a/core/README.md
+++ b/core/README.md
@@ -120,50 +120,50 @@
 
 # Contribution Process
 
-The Stellar Protocol, like most software in the world, continues to evolve over time to meet the
+The Stellar protocol, like most software in the world, continues to evolve over time to meet the
 needs of our network's participants and to drive technology forward into new territory. Given the
 importance of the reliability and safety of the network, we ask that all of those who have ideas
 towards pushing Stellar's protocol development forward adhere to the following:
 
-- Consider your idea and how it serves the fundamental goals of the Stellar Network and aligns with
-  values of the Stellar Protocol (which are listed below). If you cannot show how your proposal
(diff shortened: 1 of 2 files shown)


leighmcculloch (dismissed):
Thanks. 

The exception would be Stellar Consensus Protocol (SCP) which is capitalised.

JFWooten4:
For clarity, after much review of past discussions, I'm:

- Keeping capitalization of `Core` (standalone item with direct reference),
- Leaving `protocol` to context (Bri’s sentiments make sense that "the protocol is what is built into the code" and have support in existing CAP language[^1]), and
- Lowercasing network types like `testnet` or `mainnet` (futurenet was also always lowercase in Discord[^2]).

[^1]: This one is really weird, and I'm not happy with the implementation—but I'm also not about to challenge the status quo on it right now. The SDF's announcement posts delineate between "versions of the Stellar protocol" (lowercase) while _always_ referring to specific versions as Protocol X (uppercase), which I think is the right approach. But every single CAP that references a protocol version in their body uses lowercase identifiers.

[^2]: At least at the start and by core team members. There's some contortion in the Stellar replies because of back-and-forth with less experimental devs just learning. But it's pretty apparent in the six SEPs which use lowercase testnet.

PR #1740, opened 2025-05-29
SEP: Modify the SEP process to shift SDFs role

leighmcculloch:
~This proposal is still very early, and this PR is a sandbox of sorts where I'm experimenting through words with changes to the SEP process. It hasn't been pre-reviewed by anyone yet, so doesn't represent a viewpoint of anyone other than me with my experimental thinking hat on. I am not attached to this proposal in its current form.~

This proposal is an exploration of what on paper looks like a more radical change to the SEP process might look like where SDFs role is documented to move mostly into the role of preventing abuse and performing the administrative tasks of assigning SEP numbers, merging PRs, and reminding folks to follow the process. Even though this looks radical on paper, I think it aligns reasonably well with what's been actually happening, and so is less radical in practice.

The informational and standards tracks that exist today distinguish between SEPs that get "approved" / "blessed" by SDF, and those that don't. This distinction is imo not helpful, so this PR removes it.

The changes to statuses that exist today are largely described as being driven by a SEP Team, but in reality they are often driven by authors, whoever they may be either inside or outside of SDF. The same is true of adoption, marketing, etc, that are also often driven by authors and not any centralized group of people.

SEPs are a place to share ideas, and for the ecosystem to rally around ideas. Starting as GitHub Discussions, then becoming documented when the author feels like they are ready to do so. The statuses should serve the author in communicating what the state of their proposal is. So many mentions of "buddies" and SDF making decisions about when a status occurs has been removed.

Feedback appreciated.

cc @rice2000 @accordeiro @JakeUrban @johncanneto @janewang

### DIFF
--- a/ecosystem/README.md
+++ b/ecosystem/README.md
@@ -40,57 +40,52 @@ All SEPs have individuals fulfilling the following roles:
 
 ## List of Proposals
 
-| Number                  | Title                                                                  | Author                                                        | Status |
-| ----------------------- | ---------------------------------------------------------------------- | ------------------------------------------------------------- | ------ |
-| [SEP-0001](sep-0001.md) | Stellar Info File                                                      | SDF                                                           | Draft  |
-| [SEP-0002](sep-0002.md) | Federation Protocol                                                    | SDF                                                           | Final  |
-| [SEP-0004](sep-0004.md) | Tx Status Endpoint                                                     | SDF                                                           | Final  |
-| [SEP-0005](sep-0005.md) | Key Derivation Methods for Stellar Accounts                            | SDF                                                           | Final  |
-| [SEP-0006](sep-0006.md) | Deposit and Withdrawal API                                             | SDF                                                           | Draft  |
-| [SEP-0007](sep-0007.md) | URI Scheme to facilitate delegated signing                             | Interstellar                                                  | Final  |
-| [SEP-0008](sep-0008.md) | Regulated Assets                                                       | Interstellar                                                  | Final  |
-| [SEP-0009](sep-0009.md) | Standard KYC Fields                                                    | SDF                                                           | Draft  |
-| [SEP-0010](sep-0010.md) | Stellar Authentication                                                 | Sergey Nebolsin, Tom Quisel                                   | Draft  |
-| [SEP-0011](sep-0011.md) | Txrep: Human-Readable Low-Level Representation of Stellar Transactions | David Mazières                                                | Draft  |
-| [SEP-0012](sep-0012.md) | KYC API                                                                | Interstellar                                                  | Draft  |
-| [SEP-0014](sep-0014.md) | Dynamic Asset Metadata                                                 | OrbitLens, Paul Tiplady                                       | Draft  |
-| [SEP-0018](sep-0018.md) | Data Entry Namespaces                                                  | Mister.Ticot                                                  | Draft  |
-| [SEP-0020](sep-0020.md) | Self-verification of validator nodes                                   | Johan Stén                                                    | Draft  |
-| [SEP-0023](sep-0023.md) | Muxed Account Strkeys                                                  | David Mazières, Tomer Weller, Leigh McCulloch, Alfonso Acosta | Draft  |
-| [SEP-0024](sep-0024.md) | Hosted Deposit and Withdrawal                                          | SDF                                                           | Draft  |
-| [SEP-0028](sep-0028.md) | XDR Base64 Encoding                                                    | SDF                                                           | Final  |
-| [SEP-0029](sep-0029.md) | Account Memo Requirements                                              | OrbitLens, Tomer Weller, Leigh McCulloch, David Mazières      | Draft  |
-| [SEP-0031](sep-0031.md) | Cross-Border Payments API                                              | SDF                                                           | Draft  |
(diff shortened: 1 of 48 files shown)


aolieman:
This doesn't look radical to me at all. A reasonable step in the evolution of the SEP process. Good to designate SDF as a "maintainer of last resort."

orbitlens:
Thank you for taking an active stance on this and incorporating feedback into the updated SEP process. These changes look very well thought.

rdurscki:
Great initiative Leigh, SDF stepping back means it’s stepping up on its mission.

ecosystem/README.md
```diff
@@ -1,45 +1,51 @@
...
+  making significant changes. After at least one week has passed, during which
+  any new concerns should be addressed, the SEP's status should move to
+  `Active` or `Final`, or back to `Draft`.
+- **Active** - A SEP ready to be adopted, by may still receive changes. The
+  author intends the SEP in its current form to be actively adopted. Changes
+  can be made without changing the SEP number, although the author should
+  endeavor to make changes backwards compatible so that participants who have
+  already adopted the SEP can continue to participate. Where changes cannot be
+  backwards compatible, the major version should be updated to clearly
+  distinguish new incompatible versions.
```
JakeUrban: I would like to argue for the removal of the `Active` status.

There seems to be two differences between `Draft` and `Active`:

- `Draft` proposals should not be adopted, while `Active` proposals can be adopted
- `Draft` proposals can change dramatically, while `Active` proposals can change subtly.

Ecosystem participants decide for themselves if they're going to adopt a SEP, and while the opinion of the author is a useful signal because it indicates how dramatically or frequently the author believes a SEP may change, participants can decide to adopt a SEP knowing it is subject to change and update their implementations as it evolves if they choose.

It is also difficult for the author to maintain their implicit promise to keep the magnitude of changes aligned with the status of the SEP. The author may believe the SEP should still be in `Draft` because they expect to receive significant feedback or aren't confident enough in their original proposal, only for the original proposal to be finalized without changes. In the same way, the author may believe the SEP should be in `Active` because they believe they've fully understood the problem space only to later realize that wasn't true and want to make breaking changes.

If we do this, I think there will be numerous benefits:

- Ecosystem participants will recognize and expect the SEPs they adopt to change until they're in `Final`. We won't have cases where participants thought the proposal was stable only later to learn it isn't.
- More SEPs will be authored. If an author doesn't actively pursue developing consensus among participants and the SEP sits in `Draft` or is prematurely moved to `Final`, participants will create competing SEPs. `Active` SEPs give the impression that they are the defacto standard and that participants need to either convince the author to make a change or give up.
- SEPs will be more narrowly scoped, because the larger the scope of a SEP, the less likely consensus on the proposal in its entirety is achieved, which would result in competing SEPs being authored.

leighmcculloch: I think there is some benefit in the Draft vs Active distinction in the case that an author opens a draft early, then asks for feedback, and that feedback results in radical changes.

However, in practice this doesn't happen anymore. At some point the SEP process was changed to introduce that pre-draft phase, where authors are encouraged to hold off opening a draft document until they've discussed the idea on the GitHub Discussion forum. A step encouraged in the Ethereum and Rust ecosystems.

---

> may believe the SEP should be in Active because they believe they've fully understood the problem space only to later realize that wasn't true and want to make breaking changes.

I think we've definitely done this, and so relate to it.

---

Looking over the draft and active SEPs that exist today, there is no meaningful difference between them. So I'm :+1: to removing active. I've updated the PR to remove active.

This means there are now only 3 statuses, and 2 statuses that are not temporary.

- Draft
- FCP
- Final

18f9c66aefc4e09bb77d3f07a6e3b4674d572c49

leighmcculloch:
I've updated the README to remove the Active status, after feedback and a compelling argument from @JakeUrban about why that status is not useful (https://github.com/stellar/stellar-protocol/pull/1740#discussion_r2119446671).

tomerweller (dismissed):
Awesome work. Thanks @leighmcculloch

ecosystem/README.md
```diff
@@ -1,107 +1,99 @@
...
-  has a higher bar towards acceptance, and it requires approval by 2 SDF
-  members of the SEP Team.
+SEPs are ideas, standards, and specifications in the form of proposals that the
+author is intending to be adopted by participants in the Stellar ecosystem.
+
+## Roles
+
+All SEPs have individuals fulfilling the following roles:
+
+- **Author** - The author is the indiviudal(s) who created the proposal. The
```
accordeiro:
```suggestion
- **Author** - The author is the individual(s) who created the proposal. The
```

accordeiro (approved):
I love this. My only concern is still around SEP statuses: when does something get promoted to final, and how can we determine that a SEP has reached final status? I don't think this should block this change though, it is by all measures a big improvement.

leighmcculloch:
@tomerweller and I were discussing the proposal today and he made a strong argument for why we need the Active status: we need a way for an author to signal that their proposal is ready to be adopted.

This PR already received a lot of positive response without removing Active, so I've added it back and I think this proposal is good to merge as it is with Active still an important status.

Let's consider changes to the Active status after this PR merges, as we can discuss and agree/disagree on it independent to the other changes in this proposal.

@tomerweller @accordeiro @JakeUrban ☝🏻

accordeiro (approved):

JFWooten4:
I agree with Tomer, but have been too caught up lately to [explain why](https://github.com/JFWooten4/free-markets/pull/51).

PR #1747, opened 2025-06-16
Add metadata key registry to SEP-46

ifropc:
This PR adds metadata key registry to SEP-46.
We should not make any attempts to strictly standardize it (i.e. it's okay for SEP to not list some fields using by some libraries or tools, but they are free to add the keys they are using to this list). With SEP-9 we discovered that it's sometimes hard to predict the usage of the SEP by the ecosystem, and very often people are using some fields that are not defined by the SEP. Given that, I strongly suggest we do not make any keys mandatory, at least not for the 'minimal' contract (or use case).
That being said, another SEPs (such as contract version SEP-49) should define mandatory (for this SEP) metadata keys, and I think it's a good idea to keep this "common" keys in once place.

This PR adds table of 'commonly used keys', that currently consists of:
1. Keys that support Cargo.toml key inheritance introduced in [this scaffold-stellar PR](https://github.com/AhaLabs/scaffold-stellar/pull/108) 
2. Keys set by stellar-cli (`rsver` and `rssdkver`)
3. Keys introduced by SEP-47 and SEP-49 (`sep` and `binver` respectfully)

### DIFF
--- a/ecosystem/sep-0046.md
+++ b/ecosystem/sep-0046.md
@@ -154,14 +154,16 @@ through the next type. As such no length prefix or counter are required.
 
 ## Commonly used keys
 
-| Key name     | Description                                        | Required by SEP         |
-| ------------ | -------------------------------------------------- | ----------------------- |
-| `name`       | The name of the contract                           | N/A                     |
-| `authors`    | The author(s) of the contract, often with an email | N/A                     |
-| `homepage`   | The relevant domain to relate to this contract     | N/A                     |
-| `repository` | The source repository URL for the contract         | N/A                     |
-| `binver`     | The version of the WASM bytecode of the contract   | [SEP-49](./sep-0049.md) |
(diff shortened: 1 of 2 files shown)


leighmcculloch:
> With SEP-9 we discovered that it's sometimes hard to predict the usage of the SEP by the ecosystem, and very often people are using some fields that are not defined by the SEP. Given that, I strongly suggest we do not make any keys mandatory, at least not for the 'minimal' contract (or use case).

The reasons discussed in the PR description for not requiring all keys to be defined in SEP-46 is part of the reason why SEP-46 already sets its scope to exclude a registry.

Another reason for why to rely on SEPs defining their own meta keys instead of creating a list in SEP-46 is that keys that get defined will serve a specific purpose and be defined within the context of that purpose. This PR adds keys that don't currently get used in the ecosystem, and aren't defined in any SEP, so their purpose is unclear as to how folks would use them. Also, some of the keys being proposed already exist as defined by other SEPs or are already in use, e.g. `repository` vs `source_repo`, `homepage` vs `home_domain`.

PR #1748, opened 2025-06-17
SEP: Replace Active status with Living status

leighmcculloch:
### What
  Replace all instances of 'Active' status with 'Living' status across SEP documentation.

  ### Why
  The term Active is not descriptive, or communicate the intent well. Living more accurately describes SEPs that authors intend to maintain and evolve over time. Living is also the term used by the Ethereum ecosystem for the same type of document, although it is used more rarely there than the Stellar ecosystem will.

cc @tomerweller  @rice2000 @accordeiro @JakeUrban @johncanneto @janewang


JakeUrban (dismissed):
It makes sense to change the terminology if we're keeping the status, but I don't think we should keep it. Should we start a github discussion?

PR #1749, opened 2025-06-18
CAP-66: Specify that Wasm entries are also cached in memory.

dmkozh:
Previously it was assumed that Wasm entries are only stored in the module cache, i.e. the actual XDR doesn't leave a memory footprint. However, this would have been hard to implement in practice, which is why code entries will also be stored in the Soroban entry cache. Since the minimum size increase for Wasm is 3x now, updated the rent discount from 2x to 3x and adjusted the fees accordingly, such that the overall rent cost decrease for Wasm stays at 20x.


SirTyson (approved):
LGTM

PR #1758, opened 2025-06-30
SEP-53: Updated examples to display signatures in both base64 and hex formats for clarity.

overcat:
In SEP-43, signatures are represented using hex strings, so it would be better to update the examples in SEP-53 as well.


leighmcculloch (approved):

leighmcculloch:
@overcat Can you update your branch?

Fyi, PRs opened from organisations, rather than individual accounts, don't allow maintainers to update / modify the PR, so I can't update it as part of the merge process.

overcat:
> @overcat Can you update your branch?
> 
> Fyi, PRs opened from organisations, rather than individual accounts, don't allow maintainers to update / modify the PR, so I can't update it as part of the merge process.

Thanks for the heads-up! I've updated the branch. I wasn't aware of that limitation for PRs opened from organizations, so that's good to know for future reference.

PR #1764, opened 2025-07-19
Update SEP 0054 to use fixed config file path

tamirms:
In response to feedback from @leighmcculloch in https://github.com/orgs/stellar/discussions/1678#discussioncomment-13812481 and https://github.com/orgs/stellar/discussions/1678#discussioncomment-13812511 , I have decided to update the sep with the following changes:

- only `/<ledgers-path>` is configurable and the config file will located at `/<ledgers-path>/.config.json`. The advantage of this approach is that the config file is always co-located with the partitions. Also, because the config filename starts with a period, it will always precede the partitions lexicographically, whereas a filename of "config.json" could be interspersed lexicographically with the partition folders. Having a fixed config file path is easier for the operator because they only need to configure `/<ledgers-path>` as opposed to both `/<ledgers-path>` and `/<config-path>`
- added a `version` field to the config file which is aligned with the SEP. This will make it easier to introduce modifications to the config schema in the future.

Note that I opted against including a "server" field because I thought it would be confusing. The config file is only written once while ledgers can be continuously added to the data store. I think if we included a galexie version in the "server" field it would not be clear whether this is the version that initialized the config file or if it's the version which added the most recent ledgers to the data store.

### DIFF
--- a/ecosystem/sep-0054.md
+++ b/ecosystem/sep-0054.md
@@ -7,7 +7,7 @@ Author: Tamir Sen <@tamirms>
 Status: Draft
 Created: 2025-03-11
 Updated: 2025-03-11
-Version: 0.2.0
+Version: 0.1.0
 Discussion: https://github.com/orgs/stellar/discussions/1678
 ```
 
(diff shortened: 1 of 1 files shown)


urvisavla:
I really like the idea of prefixing the config file with a period and keeping it in the same directory as the ledger files. I also support the idea of versioning, but I think the version should also be included in the metadata of each file as well.

Regarding the configurable ledger path, is there a real advantage/need to make it configurable? As it stands, you only need to provide the bucket name, which also includes the path, making it a single config parameter. One of the strengths of Galexie is its simplicity with very few configuration options and would prefer to keep it that way unless adding new options is really necessary.

leighmcculloch:
Replied to @urvisavla's comment about versioning over at https://github.com/orgs/stellar/discussions/1678#discussioncomment-13820232 to keep the versioning conversation in one place.

tamirms:
> As it stands, you only need to provide the bucket name, which also includes the path, making it a single config parameter.

I don't think the proposal is different than the current implementation in this aspect. The bucket path parameter which is used to create the data store is a concatenation of the bucket name and the ledgers path.

urvisavla (approved):

PR #1766, opened 2025-07-22
SEP-48: Add events

leighmcculloch:
### What

Add support for contract event specifications to SEP-48. Extend the contract interface specification to include events alongside functions and user-defined types.

### Why  

Events are a critical part of smart contract interfaces that off-chain systems, indexers, and developer tools need to understand. This enables better tooling support and self-describing events.

Close stellar/stellar-protocol#1724

### Alternatives

It would be relatively easy to propose the changes to SEP-48 in a new SEP, although I have not done that here because of the precedence in this repo to update existing SEPs that are still in draft, and SEP-48 is still very new and still in draft.

### DIFF
--- a/ecosystem/sep-0048.md
+++ b/ecosystem/sep-0048.md
@@ -1684,4 +1684,4 @@ do not exist. Or a contract may omit spec entries for functions that do exist.
 
 [`soroban-sdk`]: https://github.com/stellar/rs-soroban-sdk
 [`stellar-cli`]: https://github.com/stellar/stellar-cli
-[`SEP-41`]: ./sep-0041.md
+[SEP-41]: ./sep-0041.md


ecosystem/sep-0048.md
```diff
@@ -817,6 +862,212 @@ For example, the `SCVal` in XDR, expressed in XDR-JSON:
...
+The schema will be encoded to the following `SCSpecEntry` XDR when stored in
+the custom section. The `SCSpecEntry` expressed as XDR-JSON:
+
+```json
+{
+  "event_v0": {
+    "doc": "Transfer event published when tokens are transferred.",
+    "lib": "",
+    "name": "Transfer",
+    "prefixTopics": ["transfer"],
```
Shaptic, line 976: Will it camelCase or snake_case on longer structure names?

Shaptic: Whoops, shoulda kept reading heh

ecosystem/sep-0048.md
```diff
@@ -817,6 +862,212 @@ For example, the `SCVal` in XDR, expressed in XDR-JSON:
...
+          }
+        ]
+      }
+    }
+  }
+}
+```
+
+Note: The Soroban Rust SDK, as demonstrated above, defaults to setting a single
+`prefixTopics` that is the snake_case transformation of the struct name. This
```
leighmcculloch, line 1072: @Shaptic snake_case, same as XDR-JSON, Horizon, etc, except RPC because RPC decided to be unique and go camelCase 😞 .

Shaptic: lol I have no idea why we did that

ecosystem/sep-0048.md
```diff
@@ -817,6 +862,212 @@ For example, the `SCVal` in XDR, expressed in XDR-JSON:
...
+is not a feature of this proposal, but a feature of the SDK. The name of the
```
Shaptic, line 1073: Can we make this a required spec? I know it doesn’t make a difference now given the Rust SDK is the only SDK, but a consistent experience downstream means a consistent parse+convert to the language idioms (like `name.split(“_”).map(name => name.capitalize()).join()` or whatever)

Shaptic (approved):
Super clear! I’ll probably have more Q&A as I progress on the parser but this is precise and easy to follow so far.

ecosystem/sep-0048.md
leighmcculloch: It's not obvious to me what the statement would be to include in the SEP, because the SEP defines the spec for an event including two distinct fields:
- `name`
-  `prefixTopics`

In the Rust SDK when `#[contractevent] struct Transfer` is defined, the spec XDR that gets generated has those fields set to:
- `name` = `Transfer`
- `prefixTopics` = `["transfer"]`

The spec itself doesn't infer a prefix topic or make any assumptions, because for the spec it is always specified.

As part of the note the SEP could make a recommendation that any implementation defining events reduce redundant information in the default case and encourage the pattern where the type's name prefix the first topic, and is transformed using snake case. Is that what you're thinking? Or does that sound good to you?

leighmcculloch: Downstream systems matching back to the event should imo match the topics and field types, not the 'name'. The 'name' field is there to make code generation more consistent.

leighmcculloch: Added a note in 26f5053.

leighmcculloch: I've expanded further in 8bc76382ceb3ca5ad2e3510f04165a906f713bad. Let me know what you think.

Shaptic: Oh I see what you mean about not being describable as part of the SEP. I understand! This part,

> make a recommendation that any implementation defining events reduce redundant information in the default case and encourage the pattern where the type's name prefix the first topic, and is transformed using snake case

is perfect! We can't predict how other languages will define events that conform to the XDR in this spec.

PR #1775, opened 2025-08-27
SEP-41: Add muxed address support to token interface

leighmcculloch:
### What
  Add support for muxed addresses to SEP-41 Soroban token interface in the transfer and mint actions.

  ### Why
  Updating SEP-41 for consistency with the proposals in CAP-67. SEP-41 is updated to completely encompass a unified view into the movement of value in the same reasons that mint and clawback events were recently added to SEP-41. The MuxedAddress type is a special Soroban type in that it is compatible with the Address type and can contain it and be converted to it automatically when no muxed address details are present. This makes the changes to the transfer function largely backwards compatible with existing token implementations. The changes to the events expand the possible events consumers should expect to see, but existing token contracts and implementations can continue to emit the existing events if desired.

### DIFF
--- a/ecosystem/sep-0041.md
+++ b/ecosystem/sep-0041.md
@@ -3,11 +3,11 @@
 ```
 SEP: 0041
 Title: Soroban Token Interface
-Authors: Jonathan Jove <@jonjove>, Siddharth Suresh <@sisuresh>, Simon Chow <@chowbao>
+Authors: Jonathan Jove <@jonjove>, Siddharth Suresh <@sisuresh>, Simon Chow <@chowbao>, Leigh McCulloch <@leighmcculloch>
 Status: Draft
 Created: 2023-09-22
-Updated: 2025-03-13
-Version 0.3.0
+Updated: 2025-08-28
+Version 0.4.0
 Discussion: https://discord.com/channels/897514728459468821/1159937045322547250, https://github.com/stellar/stellar-protocol/discussions/1584
 ```
 
@@ -97,15 +97,17 @@ pub trait TokenInterface {
     ///
     /// - `from` - The address holding the balance of tokens which will be
     /// withdrawn from.
-    /// - `to` - The address which will receive the transferred tokens.
+    /// - `to` - The address which will receive the transferred tokens. A MuxedAddress or Address.
     /// - `amount` - The amount of tokens to be transferred.
     ///
     /// # Events
     ///
     /// Emits an event with:
     /// - topics - `["transfer", from: Address, to: Address]`
-    /// - data - `amount: i128`
-    fn transfer(env: Env, from: Address, to: Address, amount: i128);
+    /// - data - `amount: i128` or `{ amount: i128, to_muxed_id: Option<u64 | String | BytesN<32>> }`
+    /// If the transfer involves a muxed address the address and muxed details
+    /// are separated in the event.
+    fn transfer(env: Env, from: Address, to: MuxedAddress, amount: i128);
 
     /// Transfer `amount` from `from` to `to`, consuming the allowance of
     /// `spender`. Authorized by spender (`spender.require_auth()`).
@@ -200,6 +202,9 @@ The event has topics:
 The event has data:
 
 - `i128` the amount transferred.
+- or `map` containing:
+  - `amount: i128` the amount minted.
+  - `to_muxed_id: Option<u64 | String | Bytes<32>>` void is no muxed ID, or the u64 muxed ID, or for Stellar assets also string and bytes.
 
 #### Burn Event
 
@@ -226,6 +231,9 @@ The event has topics:
 The event has data:
 
 - `i128` the amount minted.
+- or `map` containing:
+  - `amount: i128` the amount minted.
+  - `to_muxed_id: Option<u64 | string | bytes>` void is no muxed ID, or the u64 muxed ID, or for Stellar assets also string and bytes.
 
 #### Clawback Event
 
@@ -254,6 +262,7 @@ functions as they see fit. The contract is free to define how they want to
 - `v0.1.0` - Initial draft based on [CAP-46-6].
 - `v0.2.0` - Remove `spendable_balance`.
 - `v0.3.0` - Add `mint` and `clawback` event
+- `v0.4.0` - Add muxed support to transfer, and mint.
 
 ## Implementations
 


leighmcculloch:
At the very least I think we should make it more explicit that the data fields of any events with data fields can be extendable with additional fields. I think this was our goal, but it is not explicitly stated. I don't think this change should make that statement for all events, just so that this change is narrowly scoped. But I think we could start a discussion making that proposal generically across all the SEP-41 events, where any event's data can be the single value (if currently defined as so), vec (if currently defined as so), or a map of named elements.

I've pushed a commit stating that the data entries are not rigid but can be extended by implementations.

dmkozh:
> Because the SAC writes the string and bytes memos into the to_muxed_id field, I still think they need to be defined here, otherwise the SAC would be incompatible with the definition here

SAC doesn't write strings/bytes to `to_muxed_id`. Only classic events do so. Classic events generally can't be fully specified in Soroban-only terms, so this is just 'yet another' difference. The tradeoff here is between keeping things more streamlined via using just i64 ids and explicit types in Soroban, or bringing all the classic legacy to Soroban for the sake of being able to do a bit more things that are possible in classic events (which also don't have an obvious utility).

leighmcculloch:
> SAC doesn't write strings/bytes to `to_muxed_id`.

From an events consumer perspective there is no distinction between SAC and Classic, the event is from the SAC contract ID, so I think what code path generated it is an implementation detail, what matters is consumers ingesting events. Part of the goal of CAP-67 was to finally address the one-way-to-consume-memo usage, so I think it's critical in service of that to not introduce more fragmentation for event consumers.

dmkozh (approved):

dmkozh:
> so I think what code path generated it is an implementation detail, what matters is consumers ingesting events

I don't really disagree here. What I'm saying that that there is a disconnect between the Soroban interface that uses i64-id muxed addresses and only C/G addresses, and the 'consumer' interface, that may need to care about the classic event intricacies as well (additional address types, additional memo types, fee events, maybe something else). I've had the same comment for the SDK, but basically I don't see a fundamental reason for these to be tightly coupled. I feel like the coupling might lead to confusion, e.g. it's not obvious from reading SEP-41 what an `Address` is (and to make things worse, it actually means two different things in the interface and event sections). Fee events are also not defined in SEP-41 and it would be weird to define them here.

> Part of the goal of CAP-67 was to finally address the one-way-to-consume-memo usage, so I think it's critical in service of that to not introduce more fragmentation for event consumers.

Sure, but does that mean that we also have to expose all these details to the event producers? Event consumption can be unified, but event production from the Soroban side can't really be fully consistent with what consumers have to expect. Basically SEP-41 was historically oriented towards the contract developers, so it's important for them to know what is the minimum set of features that a token should support. But it is unfortunately not sufficient for properly ingesting the unified events.

ecosystem/sep-0041.md
```diff
@@ -200,6 +202,12 @@ The event has topics:
 The event has data:
 
 - `i128` the amount transferred.
+- or `map` containing entries:
+  - `amount: i128` the amount minted.
+  - `to_muxed_id: Option<u64 | String | BytesN<32>>` void or absent if no muxed
```
leighmcculloch: Have specified that the field can be absent to make this compatible with:
- https://github.com/stellar/stellar-protocol/discussions/1750

Shaptic: Is this implying that `to_muxed_id` is allowed to be present with no value? I feel like a value should be required if there’s a key, and the key itself is the optional part?

Also it’s worth clarifying the data type of the keys here

leighmcculloch: Clarified in 08a70eb6fd9ae8cdfa3ca86110512efcdf2ce8eb. Due to implementation limitations the value can be `void`, although after https://github.com/stellar/stellar-protocol/discussions/1750 the whole entry will be absent.

leighmcculloch: Void is an ScVal type, and it is the value used when `Option::None`.

ecosystem/sep-0041.md
```diff
@@ -97,15 +97,17 @@ pub trait TokenInterface {
     ///
     /// - `from` - The address holding the balance of tokens which will be
```
anupsdf, line 98: Just a suggestion here. How about adding `source` here to say, 
`The source address holding the balance of tokens which will be withdrawn from`

ecosystem/sep-0041.md
```diff
@@ -97,15 +97,17 @@ pub trait TokenInterface {
...
     /// withdrawn from.
-    /// - `to` - The address which will receive the transferred tokens.
+    /// - `to` - The address which will receive the transferred tokens. A MuxedAddress or Address.
```
anupsdf, line 100: Just a suggestion here.  If we added source for `from` above then we could say `destination` here, 
`The destination Address or MuxedAddress that will receive the transferred tokens.`

ecosystem/sep-0041.md
```diff
@@ -226,6 +234,12 @@ The event has topics:
 The event has data:
 
 - `i128` the amount minted.
+- or `map` containing entries with `Symbol` keys, and values defined as
+  - `amount: i128` the amount minted.
+  - `to_muxed_id: Option<u64 | String | BytesN<32>>` void, or entry absent, if
```
anupsdf, line 239: I am curious as to why are we referring to the field as `to_muxed_id` in both CAP-0067 and in here. Asking because the original [CAP-27](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0027.md) and [SEP-23](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0023.md) call the field `memo ID`. 
Is it because `memo id` is associated with a text/string data type.

dmkozh: While I don't recall this being discussed in much detail, I'd say it's better to stay away from 'memo' term to avoid confusion.

dmkozh (approved):
I think this is good enough for now. I think the flexibility concerns can be addressed later if the need arises.

ecosystem/sep-0041.md
```diff
@@ -200,6 +202,12 @@ The event has topics:
 The event has data:
 
 - `i128` the amount transferred.
+- or `map` containing entries with `Symbol` keys, and values defined as
+  - `amount: i128` the amount minted.
+  - `to_muxed_id: Option<u64 | String | BytesN<32>>` void, or entry absent, if
+    no muxed ID, or the u64 muxed ID, or for historical legacy reasons, the
+    string or bytes memo for Stellar assets.
```
JakeUrban, line 209: > the string or bytes memo for Stellar assets.

I don't understand how the string or bytes memo relates to stellar assets. A string or bytes type will only be emitted when the text or hash memo types are used on the transaction memo right? Should this say "the string or bytes memo from the transaction" instead?

ecosystem/sep-0041.md
```diff
@@ -200,6 +202,12 @@ The event has topics:
...
+  - Other entries allowed as defined by the implementation.
```
JakeUrban, line 210: Is it necessary to call out that other keys may be present in the map in this SEP? As a general rule I think its implied that other key-val pairs can be included in maps referenced in SEPs, but they aren't a part of the specification and therefore don't need to be explicitly acknowledged.

ecosystem/sep-0041.md
```diff
@@ -3,11 +3,11 @@
 ```
 SEP: 0041
 Title: Soroban Token Interface
-Authors: Jonathan Jove <@jonjove>, Siddharth Suresh <@sisuresh>, Simon Chow <@chowbao>
+Authors: Jonathan Jove <@jonjove>, Siddharth Suresh <@sisuresh>, Simon Chow <@chowbao>, Leigh McCulloch <@leighmcculloch>
 Status: Draft
```
JakeUrban, line 7: At what point should we finalize this SEP?

ecosystem/sep-0041.md
dmkozh: I don't have a too strong opinion, but if we look at this from the perspective of someone who wants to consume the events, they shouldn't need to worry about where `to_muxed_id` is coming from. Pointing out that this exists for legacy reasons to support specifically Stellar assets hopefully clears up things for the token implementers. At least that's my interpretation. If we want to talk about memos, we could, but we need to be precise there then (i.e. point out that the memo comes from the 'classic' transactions).

ecosystem/sep-0041.md
dmkozh: I think it is important. We don't have any other SEPs that define events (please correct me if I'm wrong). Also this change sets a precedent of returning a map in the event data payload, which is new as well. I think if we don't specify that, it might automatically make any event with additional fields non-compliant, at least for a fraction of readers.

ecosystem/sep-0041.md
leighmcculloch: +1 A question that needs an answer, but lets discuss that independent of this change.

PR #1783, opened 2025-09-10
[SEP-0024] Add Soroban contract account and SEP-0045 support

lijamie98:
This PR adds SEP-45 and contract account support for SEP-24.


ecosystem/sep-0024.md
```diff
@@ -48,8 +48,9 @@ protocol will work with any compliant wallet.
 - An anchor must define the location of their `TRANSFER_SERVER_SEP0024` in
   their [`stellar.toml`](sep-0001.md). This is how a wallet knows where to find
   the anchor's server.
-- Anchors and clients must support [SEP-10](sep-0010.md) web authentication to
-  enable authenticated deposits, withdrawals, or transaction history lookups.
+- Anchors and clients must support [SEP-10](sep-0010.md) or
```
JiahuiWho: and/or ?

ecosystem/sep-0024.md
```diff
@@ -87,28 +89,28 @@ be one-time use, or at least short-lived.
 
 ### Shared, Omnibus, or Pooled Accounts
 
-Client applications can use a single Stellar account to hold multiple users'
-funds. To distinguish users that use the same Stellar account, the decoded
-SEP-10 JWT's `sub` value may contain a memo value after the Stellar account
-(`G...:2810101841641761712`) OR the `sub` value will be a
+Client applications can use a single classic Stellar account to hold multiple
+users' funds. To distinguish users that use the same Stellar account, the
+decoded SEP-10 JWT's `sub` value may contain a memo value after the Stellar
```
JiahuiWho, line 97:
```suggestion
decoded SEP-10 JWT's `sub` value may contain a memo value after the classic Stellar
```

So that user know what classic stellar account is referring to in the following context

ecosystem/sep-0024.md
```diff
@@ -519,8 +521,8 @@ Request Parameters:
...
-| `memo_type`                   | string                                  | (optional) Type of memo that anchor should attach to the Stellar transaction, one of `text`, `id` or `hash`.                                                                                                                                                                                                                                                                                                                                                                              |
+| `account`                     | `G...` or `M...` or `C...` string       | (optional) The classic account, contract account or muxed account the client wants to use as the destination of the payment sent by the anchor. Defaults to the account authenticated via SEP-10 or SEP-45 if not specified.                                                                                                                                                                                                                                                              |
```
JiahuiWho: l
```suggestion
| `account`                     | `G...` or `M...` or `C...` string       | (optional) The classic account, contract account or muxed account the client wants to use as the destination of the payment sent by the anchor. Defaults to the account used in authentication if not specified.                                                                                                                                                                                                                                                              |
```

ecosystem/sep-0024.md
```diff
@@ -768,7 +770,7 @@ Request parameters:
...
-| `account`           | `G...` or `M...` string                 | (optional) The Stellar or muxed account the client will use as the source of the withdrawal payment to the anchor. Defaults to the account authenticated via SEP-10 if not specified.                                                                                                                                                                                                                                                                                                          |
+| `account`           | `G...` or `M...` or `C...` string       | (optional) The class account, contract account, or muxed account the client will use as the source of the withdrawal payment to the anchor. Defaults to the account authenticated via SEP-10 or SEP-45 if not specified.                                                                                                                                                                                                                                                                       |
```
JiahuiWho:
```suggestion
| `account`           | `G...` or `M...` or `C...` string       | (optional) The classic account, contract account, or muxed account the client will use as the source of the withdrawal payment to the anchor. Defaults to the account used in authentication if not specified.                                                                                                                                                                                                                                                                       |
```

ecosystem/sep-0024.md
```diff
@@ -1551,10 +1553,10 @@ details. For example:
...
 transaction at an anchor.
 
-Anchors must ensure that the SEP-10 JWT included in the request contains the
-Stellar account and optional memo value used when making the original deposit
-or withdraw request that resulted in the transaction requested using this
-endpoint.
+Anchors must ensure that the SEP-10 or SEP-45 JWT included in the request
+contains the Stellar account and optional memo value used when making the
+original deposit or withdraw request that resulted in the transaction requested
+using this endpoint.
```
JiahuiWho:
```suggestion
Anchors must verify that the SEP-10 or SEP-45 JWT in the request contains the Stellar account and, if applicable, the memo that were used in the original deposit or withdrawal request for the transaction being queried.
```
This was a really loooooong sentence without any break...

ecosystem/sep-0024.md
lijamie98: It should be `or` because they MUST do one of them.

ecosystem/sep-0024.md
lijamie98: The account of SEP-10 is always classic. There are quite a few places like this. If the context is obvious, I think we may omit the `classic`.

ecosystem/sep-0024.md
JiahuiWho: but they can support both right?

ecosystem/sep-0024.md
JiahuiWho: Ok just curious is "classic" a well-known wording to external developers? I thought "classic" is a relative concept because we are talking about contract account. 

External developers might not have the same context we do. It could be clearer to use a more explicit, descriptive name for the contract account so it’s obvious what’s being referred to.

ecosystem/sep-0024.md
lijamie98: Yes, anchors and wallets can support both. But when we say they must support SEP-10 and SEP-45, that means they cannot just support one of the SEP-10 or SEP-45.

ecosystem/sep-0024.md
lijamie98: I think classic is a relative term. It is more accurate to say "non-contract" account or G-account. In the context of SEP-10, it is clear. 

But if we want to make clear of the account everywhere, it can be very verbose. I think within the SEP readers' context, the account should mean classic if not specified.

ecosystem/sep-0024.md
lijamie98: modified to make it more clear and avoid confusion.

ecosystem/sep-0024.md
lijamie98: The definition of the [classic account ](https://github.com/stellar/stellar-protocol/pull/1783/files#diff-8eff781a81bdbf7d9a0f55d126cde3920a205a82b92ce13156269d725a50383bR56)is also added to avoid confusion.

PR #1785, opened 2025-09-12
Add CAP-71 - Authentication delegation for custom accounts

dmkozh:
This CAP improves the developer experience for delegating the authentication logic from one account contract to another. A new use case for the authentication delegation is introduced in the upcoming CAP-72, that adds the authentication delegation capabilities to G-accounts.

### DIFF
--- a/core/cap-0071.md
+++ b/core/cap-0071.md
@@ -27,7 +27,7 @@ This feature has been experimented with in some custom account implementations, 
 
  However, since the current delegation support happens to be just a by-product of the authorization framework design, there are significant user and developer experience flaws with the approach:
 
-- It is hard to properly simulate delegated authorization. When simulation records the necessary authorization payloads, it doesn't every enter `__check_auth`, as it doesn't have sufficient information to call it. Thus users need to build the authorization payload of the inner call themselves and then they also need to run simulation again.
+- It is hard to properly simulate delegated authorization. When simulation records the necessary authorization payloads, it doesn't ever enter `__check_auth`, as it doesn't have sufficient information to call it. Thus users need to build the authorization payload of the inner call themselves and then they also need to run simulation again.
     - This could be somewhat simplified on the simulation side by introducing a mode in which authorization data may be partially initialized, which would simplify the inner payload building. However, this approach still comes with complexity for developers (such as the need to do more simulation calls), as well as the added simulation complexity.
 - Every account for which `__check_auth` is called requires an authorization entry in the transaction, with its own nonce. That increases transaction cost and further increases complexity.
 - It is hard to forward the authorization context to the delegated signers. To do that, one needs to pass the whole context into the `require_auth_for_args` call. But that means that the whole context needs to be present in the authorization payload of the delegated signer, which bloats the transaction size unnecessarily, and increases the complexity for the developers if they want to write logic dependent on the context.
@@ -170,12 +170,13 @@ More formally, the algorithm for handling `delegate_account_auth` is defined as 
 - Define 'top level' `__check_auth` call for non-source address `A` as a call that occurs when `require_auth[_for_args]` is matched to `A` for the first time, as per [CAP-46-11](./cap-0046-11.md).
   - Note, that implementation of `__check_auth` for G-accounts is built into host and is not observed as a contract call; however that doesn't affect the algorithm.
 - Define top level call arguments as `A.__check_auth(P, S(A), C)` where `P` is the 32-byte signature payload, `S(A)` is the signature from the `A`s auth entry, and `C` is the authorization context derived from the auth entry.
-- For every `__check_auth` call, define `current_delegate_list` as a list of delegated signer addresses with the corresponding signatures and a 'used' flag.
-- For top level call, define `current_delegate_list` as the value of `delegates` inside `SorobanAddressCredentialsWithDelegates` in `A`s auth entry, or an empty list if `SorobanAddressCredentials` is used. Mark every delegate as unused. 
+- For every `__check_auth` call, define `current_delegate_list` as a list of delegated signer addresses with the corresponding signatures and a `used` flag.
+- For top level call, define `current_delegate_list` as the value of `delegates` inside `SorobanAddressCredentialsWithDelegates` in `A`s auth entry, or an empty list if `SorobanAddressCredentials` is used. Mark every delegate as unused by setting their `used` flag to `false`.
 - Any time when `delegate_account_auth` is called for an address `B`:
-  1. Search for the first unused occurrence of `B` in `current_delegate_list` and store the found index `i`. If there is no such occurrence, fail authentication process.
-  2. Call `B.__check_auth(P, current_delegate_list[i].signature, C)`.
-  3. For the call, set `current_delegate_list` to the value of `current_delegate_list[i].nestedDelegates` and mark every delegate as unused.
+  1. Search for the first unused (`used == false`) occurrence of `B` in `current_delegate_list` and store the found index `i`. If there is no such occurrence, fail authentication process.
+  2. Set `current_delegate_list[i].used` to `true`.
+  3. Call `B.__check_auth(P, current_delegate_list[i].signature, C)`.
+     - For the call, set `current_delegate_list` to the value of `current_delegate_list[i].nestedDelegates` and mark every delegate as unused.
 
 
 #### Signature payload for `SorobanAddressCredentialsWithDelegates`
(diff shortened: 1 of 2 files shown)


sisuresh (dismissed):

core/cap-0071.md
```diff
@@ -0,0 +1,242 @@
...
+
+This feature has been experimented with in some custom account implementations, and is generally deemed useful, as it allows for creating more flexible and 'modular' custom accounts. A significant motivating use case for the authentication delegation is also introduced in [CAP-72](./cap-0072.md): it adds the authentication logic customization capability to the Stellar accounts via adding contracts as additional account signers. This is the same case of authentication delegation, with the only difference being that the authentication procedure for the Stellar accounts is built into the protocol.
+
+ However, since the current delegation support happens to be just a by-product of the authorization framework design, there are significant user and developer experience flaws with the approach:
+
+- It is hard to properly simulate delegated authorization. When simulation records the necessary authorization payloads, it doesn't every enter `__check_auth`, as it doesn't have sufficient information to call it. Thus users need to build the authorization payload of the inner call themselves and then they also need to run simulation again.
```
anupsdf:
```suggestion
- It is hard to properly simulate delegated authorization. When simulation records the necessary authorization payloads, it doesn't enter `__check_auth`, as it doesn't have sufficient information to call it. Thus users need to build the authorization payload of the inner call themselves and then they also need to run simulation again.
```

core/cap-0071.md
```diff
@@ -0,0 +1,242 @@
...
+
+More formally, the algorithm for handling `delegate_account_auth` is defined as follows:
+
+- Define 'top level' `__check_auth` call for non-source address `A` as a call that occurs when `require_auth[_for_args]` is matched to `A` for the first time, as per [CAP-46-11](./cap-0046-11.md).
+  - Note, that implementation of `__check_auth` for G-accounts is built into host and is not observed as a contract call; however that doesn't affect the algorithm.
+- Define top level call arguments as `A.__check_auth(P, S(A), C)` where `P` is the 32-byte signature payload, `S(A)` is the signature from the `A`s auth entry, and `C` is the authorization context derived from the auth entry.
+- For every `__check_auth` call, define `current_delegate_list` as a list of delegated signer addresses with the corresponding signatures and a 'used' flag.
+- For top level call, define `current_delegate_list` as the value of `delegates` inside `SorobanAddressCredentialsWithDelegates` in `A`s auth entry, or an empty list if `SorobanAddressCredentials` is used. Mark every delegate as unused.
```
anupsdf:
```suggestion
- For top level call, define `current_delegate_list` as the value of `delegates` inside `SorobanAddressCredentialsWithDelegates` in `A`s auth entry, or an empty list if `SorobanAddressCredentials` is used. Mark every delegate as unused by setting the 'used' flag to false. 
```

core/cap-0071.md
```diff
@@ -0,0 +1,242 @@
...
+- Define top level call arguments as `A.__check_auth(P, S(A), C)` where `P` is the 32-byte signature payload, `S(A)` is the signature from the `A`s auth entry, and `C` is the authorization context derived from the auth entry.
+- For every `__check_auth` call, define `current_delegate_list` as a list of delegated signer addresses with the corresponding signatures and a 'used' flag.
+- For top level call, define `current_delegate_list` as the value of `delegates` inside `SorobanAddressCredentialsWithDelegates` in `A`s auth entry, or an empty list if `SorobanAddressCredentials` is used. Mark every delegate as unused. 
+- Any time when `delegate_account_auth` is called for an address `B`:
+  1. Search for the first unused occurrence of `B` in `current_delegate_list` and store the found index `i`. If there is no such occurrence, fail authentication process.
+  2. Call `B.__check_auth(P, current_delegate_list[i].signature, C)`.
+  3. For the call, set `current_delegate_list` to the value of `current_delegate_list[i].nestedDelegates` and mark every delegate as unused.
```
anupsdf: Suggestion: Make this part of above point 2 or use 2.a since its related to the call.

core/cap-0071.md
```diff
@@ -0,0 +1,242 @@
...
+
+## Design Rationale
+
+### New signature payload type
+
+While it would be easier to just re-use the existing payload, that would open up a potential vulnerability. 
+
+Specifically, if an account uses only delegated signers, than signatures for a delegated signer could be replayed by using the delegated signer's address in place of the original account's address in the invocation. For example, if account `A` needs to sign a token transfer `token.transfer(A, to, 100)` via delegated signer `D` **and** token implementation only calls `A.require_auth_for_args((to, 100))`, then the signature of `D` would be reusable in `token.transfer(D, to, 100)` call. That would happen because `ENVELOPE_TYPE_SOROBAN_AUTHORIZATION` is not _explicitly_ attached to the account address, it's implicitly linked to it via cryptographic signature. Thus the issue would only happen for the contracts that don't explicitly require authorization for the address as an invocation argument.
```
anupsdf:
```suggestion
Specifically, if an account uses only delegated signers, then signatures for a delegated signer could be replayed by using the delegated signer's address in place of the original account's address in the invocation. For example, if account `A` needs to sign a token transfer `token.transfer(A, to, 100)` via delegated signer `D` **and** token implementation only calls `A.require_auth_for_args((to, 100))`, then the signature of `D` would be reusable in `token.transfer(D, to, 100)` call. That would happen because `ENVELOPE_TYPE_SOROBAN_AUTHORIZATION` is not _explicitly_ attached to the account address, it's implicitly linked to it via cryptographic signature. Thus the issue would only happen for the contracts that don't explicitly require authorization for the address as an invocation argument.
```

core/cap-0071.md
```diff
@@ -0,0 +1,242 @@
...
+
+Note, that in the 'legacy' delegation flow that currently exists in the protocol the same kind of linking is achieved due to the need for `D` to explicitly authorize `A.__check_auth` call (i.e. linking happens in `D`'s authorized invocation specification within the signed payload).
+
+### Delegated signer recursive nesting
+
+The ability to nest the delegated signers makes protocol a bit more complex and requires 4 bytes of additional space in credentials even if it's not being used. This is introduced mostly for the sake of completeness and consistency, so that any accounts could be used as delegated signers (including ones that have delegated signers of their own). This may also potentially help some complex protocols that may come in the future. The relative protocol complexity increase is pretty marginal and thus should be tolerable.
+
+## Protocol Upgrade Transition
+
+The proposed host function will use the standard mechanism for protocol-gating the host functions and will become available in protocol <TBD>. The new type of credentials will also only be available starting from protocol <TBD>, otherwise including it into transaction will invalidate the transaction.
```
anupsdf: TBD is not showing up in the preview mode. Maybe use _TBD_  (underscore on both sides) instead?

core/cap-0071.md
```diff
@@ -0,0 +1,242 @@
...
+
+## Implementation
+
+TBD
+
+## Appendix: Simulation flow
+
+While simulation is not a part of the protocol, it is an important part of the overall Stellar developer experience, so it's useful to mention it in order to put this CAP into context.
+
+Even with this CAP 2 simulation runs are necessary in order to come up with the correct transaction footprint and resources. Note, that two simulation runs is a general issue that any custom account likely needs to deal with. There is a possibility to avoid the second simulation run via supplying all the signers and 'dummy' signatures for them, but that change is unrelated to this CAP.
```
anupsdf:
```suggestion
Even with this CAP two simulation runs are necessary in order to come up with the correct transaction footprint and resources. Note, that two simulation runs is a general issue that any custom account likely needs to deal with. There is a possibility to avoid the second simulation run via supplying all the signers and 'dummy' signatures for them, but that change is unrelated to this CAP.
```

anupsdf (commented):
Looks good. Added some minor suggestions.

core/cap-0071.md
dmkozh: I meant 'ever', fixed

core/cap-0071.md
dmkozh: Updated

core/cap-0071.md
dmkozh: That doesn't quite work with markdown. I've made it a sub-bullet point instead.

core/cap-0071.md
dmkozh: Fixed

core/cap-0071.md
dmkozh: Fixed

core/cap-0071.md
dmkozh: Fixed

PR #1788, opened 2025-09-17
CAP-71 - Add delegated signers getter.

dmkozh:
This is not strictly required, but not doing this would introduce some unnecessary annoyance for the contract implementation and transaction building, including CAP-72 as well.

### DIFF
--- a/core/cap-0071.md
+++ b/core/cap-0071.md
@@ -48,7 +48,7 @@ Each of the delegated addresses has its own signature and its own authentication
 
 `delegate_account_auth` calls can also be nested recursively, and every authentication call will still inherit the 'top-level' arguments.
 
-Another host function called `get_delegated_signers` is also introduced for standardizing and simplifying the delegated signers access by providing the direct access to the delegated signers populated in the authorization payload.
+Another host function called `get_delegated_signers_for_current_auth_check` is also introduced for standardizing and simplifying the delegated signers access by providing the direct access to the delegated signers populated in the authorization payload.
 
 With these changes only a single authorization entry is necessary to accommodate an arbitrary number of potentially nested delegated signers, which reduces the transaction size greatly and simplifies the simulation logic.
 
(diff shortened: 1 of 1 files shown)


core/cap-0071.md
```diff
@@ -133,16 +135,23 @@ The diff is based on commit `f5153ff476ce7e68ee0bda278188cded135ff301` of `rs-so
...
 +++ b/soroban-env-common/env.json
-@@ -2467,6 +2467,18 @@
+@@ -2467,6 +2467,25 @@
                      "return": "Val",
                      "docs": "Returns the executable corresponding to the provided address. When the address does not exist on-chain, returns `Void` value. When it does exist, returns a value of `AddressExecutable` contract type. It is an enum with `Wasm` value and the corresponding Wasm hash for the Wasm contracts, `StellarAsset` value for Stellar Asset contract instances, and `Account` value for the 'classic' (G-) accounts.",
                      "min_supported_protocol": 23
 +                },
 +                {
 +                    "export": "7",
++                    "name": "get_delegated_signers",
```
anupsdf: Should we call it get_delegated_signers_in_payload so that its clear that its not fetching the delegated signers from the contact data entry.

core/cap-0071.md
```diff
@@ -199,6 +213,12 @@ Note, that in the 'legacy' delegation flow that currently exists in the protocol
...
 
+### Delegated signer getter
+
+``get_delegated_signers` function is technically optional for this CAP, as smart accounts could get away with defining a special signature type to account for the delegated signers.
+
+ However, this creates some unnecessary duplication in the transaction data (as delegated signers would need to be specified twice), and introduces unnecessary fragmentation into how delegated signers work, as different contracts may use slightly different UDTs to represent them in signatures. This CAP already provides a standardized approach for storing the delegated signers in the transaction, so there is no good reason not to benefit from that on the contract implementation side as well, while also avoiding duplication.
```
anupsdf, line 220: You mean't CAP-0072 provides an approach for storing delegated signers, right?

dmkozh: No, `SorobanAddressCredentialsWithDelegates` XDR is defined in this CAP.

core/cap-0071.md
dmkozh: I don't like using 'payload' term in the host function name as it breaks the abstraction a bit. I've tried to rename it to emphasize the fact that the signers are associated with the current auth check (and not with the whole account).

anupsdf: sounds good!

anupsdf (approved):

PR #1555, opened 2024-09-24
Change bullet * to - for readability due to adjacent **

JFWooten4:
In re [D799](https://github.com/stellar/stellar-docs/issues/799), stars for bullet points make my insides repulse, especially when the label thereater is bold with stars.[^existing]

[^existing]: _See also_ parallel code at [C118&ndash;123](https://github.com/stellar/stellar-protocol/blob/eb401f932258c827a5b4a2e14aea939affcd2b02/core/README.md?plain=1#L118-L123) and [all SEP info](https://github.com/stellar/stellar-protocol/blob/eb401f932258c827a5b4a2e14aea939affcd2b02/ecosystem/README.md?plain=1#L5-L219) employing hyphens.

PR #1803, opened 2025-10-20
📝 How can we document the stability upgrade in clear English?

JFWooten4:
The existing CAP76 doc is extremely technically competent and quite a work of rigor in the limited compilation timeframe. That said, there are some obvious syntax upgrades which help!

I think these pages are more [than](https://github.com/stellar/stellar-protocol/pull/1555#pullrequestreview-2328226662) just "short-lived documents that propose a change" that "are not significantly referenced after implementation and vote." I know from experience these last years working on the docs updates that they can be a source of core information publication. For instance, nowhere else on the internet would we have the direct inflation info on fees burned.

I just started going through things, and I really like the `semantics` section explaining crucial state-archival concepts and functions. I'll polish it up in my free time as quickly as possible these next few busy days for our team. Feel free to use edit access to my repo too.


briwylde08:
@JFWooten4 I didn't actually write that one but I'm sure whoever did appreciates the comments! As for where the English-world names are, do you mean something like this? https://stellar.org/protocol-upgrades

PR #1812, opened 2025-11-06
SEP-56: Update event representation

ozgunozerk:
Consolidated event descriptions into the `#[contractevent]` format inline code.

@leighmcculloch, let me know if you also want to get rid of `[` and `]` in the trait comments, but we have them in other SEPs, and throughout our library as well. Nevertheless, we are open for this change if you think it still creates confusion even with the `#[contractevent]` notation down below 👍

### DIFF
--- a/ecosystem/sep-0056.md
+++ b/ecosystem/sep-0056.md
@@ -7,8 +7,8 @@ Author: OpenZeppelin, Boyan Barakov <@brozorec>, Özgün Özerk <@ozgunozerk>, S
 Track: Standard
 Status: Draft
 Created: 2025-11-04
-Updated: 2025-11-04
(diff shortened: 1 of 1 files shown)


leighmcculloch:
> @leighmcculloch, let me know if you also want to get rid of `[` and `]` in the trait comments, but we have them in other SEPs, and throughout our library as well. Nevertheless, we are open for this change if you think it still creates confusion even with the `#[contractevent]` notation down below 👍

Given that this has been a long running point of confusion, not just with these SEPs, but even going back to the very first SEP-41 draft, it might be a good idea we remove the textual representations of the events and only use the `#[contractevent]` representation moving forward.

PR #1813, opened 2025-11-06
SEP-56: Remove textual event representation

ozgunozerk:
also removes the `#[derive]` from the macros

### DIFF
--- a/ecosystem/sep-0056.md
+++ b/ecosystem/sep-0056.md
@@ -146,7 +146,7 @@ pub trait TokenizedVault: TokenInterface {
     ///
     /// # Events
     ///
-    /// Emits `[Deposit]` event.
+    /// Emits [`Deposit`] event.
     fn deposit(e: &Env, assets: i128, receiver: Address, from: Address, operator: Address) -> i128;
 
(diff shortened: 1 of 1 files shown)


ecosystem/sep-0056.md
```diff
@@ -146,8 +146,7 @@ pub trait TokenizedVault: TokenInterface {
     ///
     /// # Events
     ///
-    /// * topics - `["deposit", operator: Address, from: Address, receiver: Address]`
-    /// * data - `[assets: i128, shares: i128]`
+    /// Emits `deposit` event.
```
leighmcculloch: Given these are typically defined in the same module, we could describe it by the type name. Wdyt?
```suggestion
    /// Emits [`Deposit`] event.
```

ecosystem/sep-0056.md
```diff
@@ -146,8 +146,7 @@ pub trait TokenizedVault: TokenInterface {
     ///
     /// # Events
     ///
-    /// * topics - `["deposit", operator: Address, from: Address, receiver: Address]`
-    /// * data - `[assets: i128, shares: i128]`
```
leighmcculloch, line 150: 💯

PR #1827, opened 2025-12-16
Handle archived SEP-45 contract instances

philipliu:
This allows restoration of the SEP-45 contract instance in the client-side verification of the ledger footprint. The example authorization entries are also updated to use the new `SorobanAuthorizationEntries` XDR.


ecosystem/sep-0045.md
```diff
@@ -95,6 +95,8 @@ The authentication flow is as follows:
      - The `contract` is the **Server Account**, and the `key` is `ledger_key_nonce`.
      - (Optional) if an authorization entry for the **Client Domain Account** was present in the challenge, the
        `contract` is the **Client Domain Account**, and the `key` is `ledger_key_nonce`.
+     - (Optional) The `contract` is the `WEB_AUTH_CONTRACT_ID`, and the `key` is `ledger_key_contract_instance`
+       if the web authentication contract instance has been archived.
```
leighmcculloch, line 99: ❓ Why is the contract instance only in the footprint if it is being restored? I would have thought it would always be in the footprint.

ecosystem/sep-0045.md
```diff
@@ -186,7 +188,7 @@ GET https://auth.example.com/?account=CCIBUCGPOHWMMMFPFTDWBSVHQRT4DIBJ7AD6BZJYDI
 
 On success the endpoint must return `200 OK` HTTP status code and a JSON object with the following fields:
 
-- `authorization_entries`: an XDR-encoded array of `SorobanAuthorizationEntry`s. It contains an entry for the **Client
+- `authorization_entries`: XDR-encoded `SorobanAuthorizationEntries`. It contains an entry for the **Client
```
leighmcculloch, line 191: 👍🏻

ecosystem/sep-0045.md
philipliu: The contract instance is always included in `read_only`, but it only appears in `read_write` if the simulation restores it. This part of the doc explains how to inspect the read_write entries.

leighmcculloch (approved):

PR #1828, opened 2025-12-17
Add CAP-78 - Host function for performing limited TTL extensions

dmkozh:


core/cap-0078.md
```diff
@@ -0,0 +1,145 @@
...
+    Consulted: 
+Status: Draft
+Created: 2025-12-16
+Discussion: https://github.com/orgs/stellar/discussions/1825
+Protocol version: 26
+```
+
+## Simple Summary
+
+Add a way to limit the maximum TTL extension for persistent contract data entries.
```
anupsdf: Should contract code and contract instance entries also have this fine-grained control?

dmkozh: Good catch, I've been too focused on the extension semantics and forgot that we need a different function for instance/code. FWIW code is probably the biggest offender in terms of the rent fee spikes, so we definitely want to support it.

anupsdf:
We also have one more host function to extending `current` contract (extend_current_contract_instance_and_code_ttl). Should we provide this min/max extension to that one as well?

core/cap-0078.md
```diff
@@ -0,0 +1,204 @@
...
+- `liveUntilLedgerSeq` is the last ledger sequence number for which the entry is still considered to be alive, after that it is considered expired.
+- TTL is defined as `TTL = liveUntilLedgerSeq - currentLedgerSeq`, where `currentLedgerSeq` is the sequence number of the ledger where transaction is executed.
+- TTL extension is defined as `TTL_ext = TTL_new - TTL_curr = liveUntilLedgerSeq_new - liveUntilLedgerSeq_curr`, where `_curr/_new` are the values before and after executing the function respectively.
+
+With these definitions, the extension algorithm is defined as follows:
+
+- If  `t` is `StorageType::Instance`, the function traps
+- If `min_extension > extend_to`, or `max_extension < min_extension`, the function traps
+- If `extend_to <= TTL_curr` the function returns without performing any changes
+- Otherwise, compute the initial TTL extension `TTL_ext_init = TTL_curr - extend_to`
```
anupsdf: This should be reverse, isn't it? `TTL_ext_init = extend_to - TTL_curr`

dmkozh: Fixed

PR #1845, opened 2026-01-15
Update CAP-78

dmkozh:
There was some inconsistency around `min_extension` that I've clarified.

Also clarify the validation logic and justification for temporary entries.

### DIFF
--- a/core/cap-0079.md
+++ b/core/cap-0079.md
@@ -83,16 +83,16 @@ index 945ada2c..b8d76d4d 100644
 
 ### Semantics
 
-#### `strkey_to_address_v2` host function
+#### `address_to_strkey_v2` host function
 
-`strkey_to_address_v2` accepts a single `Val` argument that can be either `AddressObject`, or `MuxedAddressObject`. `AddressObject` will be converted to an account (`G...`) or contract (`C...`) strkey corresponding to the address. `MuxedAddressObject` will be converted to a muxed account strkey (`M...`).
+`address_to_strkey_v2` accepts a single `Val` argument that can be either `AddressObject`, or `MuxedAddressObject`. `AddressObject` will be converted to an account (`G...`) or contract (`C...`) strkey corresponding to the address. `MuxedAddressObject` will be converted to a muxed account strkey (`M...`).
 
 The function traps if a `Val` of invalid type is provided.
 
 
-#### `address_to_strkey_v2` host function
+#### `strkey_to_address_v2` host function
 
-`address_to_strkey_v2` accepts a single `Val` containing strkey in `BytesObject` or `StringObject`.
+`strkey_to_address_v2` accepts a single `Val` containing strkey in `BytesObject` or `StringObject`.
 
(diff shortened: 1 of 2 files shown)


core/cap-0078.md
```diff
@@ -165,7 +165,7 @@ If a built-in contract instance is being extended (currently, only Stellar Asset
...
 
 Maximum extension argument addresses the extension strategy mentioned in the 'Motivation' section. Developers may set extension strategies like 'extend TTL to 30 days with min extension of 1 day and max extension of 1 day', which would result in any user extending the entry TTL by just 1 day as long as its current TTL is anywhere between 0 and 29 days.
```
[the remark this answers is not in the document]
dmkozh: Looks like false positive

core/cap-0078.md
```diff
@@ -140,18 +140,18 @@ index d72886e3..3f2d0d48 100644
...
+- If `max_extension < min_extension`, the function traps
 - If `extend_to <= TTL_curr` the function returns without performing any changes
-- Otherwise, compute the initial TTL extension `TTL_ext_init = extend_to - TTL_curr`
-- If `TTL_ext_init < min_extension`, the function returns without performing any changes
-- Otherwise, compute the final extension `TTL_ext_final = min(TTL_ext_init, max_extension)`
-- If `TTL_ext_final` exceeds `maxEntryTTL` State Archival network setting value:
-     - If storage type `t` is `Persistent`, then `TTL_ext_final` is assigned to be `maxEntryTTL`
-     - If storage type `t` is `Temporary`, then function traps
+- Compute the initial TTL extension `TTL_ext_init = extend_to - TTL_curr`
+- Compute maximum extension allowed by the network `max_network_extension = maxEntryTTL - TTL_curr`
```
anupsdf, line 151: max_network_extension should just be maxEntryTTL coming from network configs, isn't it? Without subtracting TTL_curr.

dmkozh: No, TTL is not TTL extension. Let's say maximum TTL allowed is 500, we're on ledger 1000, and live-until ledger of the entry is 1400 (so its TTL is already 400). Maximum live-until ledger for the entry is 1500 because maximum TTL allowed is 500. But the maximum _extension_ is just 500-400=100 ledgers, extending by any value higher than 100 ledgers would exceed the maximum TTL.

FWIW clamping logic could be written in a different way, by e.g. clamping the final max live-until ledger to be the network max. But since we have another extension limit now (`max_extension`), it seems more natural to define clamping only in terms of the extensions.

anupsdf: I see. Maybe we should this (and a few more?) examples to make it easier to follow.

dmkozh: I'm not sure if examples are necessary in the CAP (there are many possible scenarios to cover; the spec is more comprehensive and then the tests can be observed for the actual scenarios). But some examples are likely useful for discussion, so I posted them in the discussion thread. Not sure if max network TTL clamping is very interesting though for discussion, its semantics are the same as for the existing TTL functions.

PR #1846, opened 2026-01-15
Update CAP template to specify an XDR example section

sisuresh:


cap-template.md
```diff
@@ -86,6 +86,13 @@ To apply diffs, use the `git apply --reject --whitespace=fix` command.
 For large changes, it may be beneficial to link to actual XDR files copied
 in the relevant "contents" folder.
 
+#### XDR Example
+If this update to XDR or any logical change to meta will have any effect on Horizon, RPC, etc., include some base64-encoded `LedgerCloseMeta` examples. The examples should reflect the updated XDR structures for downstream testing purposes. The feasibility of generating this XDR will depend on the XDR change itself, but the CAP author should do their best to give downstream consumers enough to test with so they don't need to wait for the Stellar Core implementation before starting testing.
+
+The `stellar-xdr` CLI can convert between base64 XDR and JSON, so that can be used to modify an existing testnet or mainnet `LedgerCloseMeta` to use as an example. Note that you may need to build or install a version of `stellar-xdr` that contains the changes in this CAP.
```
anupsdf, line 92: stellar/stellar-cli has the command without the hyphen, `stellar xdr`
https://github.com/stellar/stellar-cli/blob/v23.4.1/FULL_HELP_DOCS.md?plain=1#L3944

sisuresh: Ah I actually use the CLI from rs-stellar-xdr - https://github.com/stellar/rs-stellar-xdr?tab=readme-ov-file#examples, which is `stellar-xdr`. 

I'm not sure it makes sense to specify install instructions and mention the two CLIs here.  What do you think?

anupsdf: There was some difference between the two that I had noticed recently and @leighmcculloch had suggested to use `stellar xdr` which is an improvement over `stellar-xdr`. But I agree that we don't need to specify the install instructions here.

sisuresh: Oh I wasn't aware that one was better than the other.

leighmcculloch: The tools are identical.

The `stellar-xdr` tool exists for every commit of the https://github.com/stellar/rs-stellar-xdr repo.

The `stellar-cli`, which is what the `stellar xdr` command is, is updated and released less frequently.

Most people building on Stellar will find the `stellar xdr` command sufficient.

People building Stellar probably want to use `stellar-xdr` because the moment the XDR is updated in https://github.com/stellar/rs-stellar-xdr you'll have a `stellar-xdr` to use with it.

anupsdf: Thanks @leighmcculloch !

PR #1855, opened 2026-01-22
Add CAP-0080

sisuresh:


core/cap-0080.md
```diff
@@ -0,0 +1,218 @@
...
+
+BN254 MSM (Multi-Scalar Multiplication) and modular arithmetic are used in a variety of ZK applications. Adding host support for these will greatly improve the performance of these use cases.
+
+## Working Group
+
+As described in the preamble section.
+
+## Motivation
+
+Stellar has host support for the BN254 functions available on the EVM (G1 Add, G1 Mul, and Pairing), but there are use cases that require many G1 add and multiplies, as well as lots of modular arithmetic. These use cases end up being expensive due to all the work done on the guest. By adding a BN254 MSM host function, as well as BN254 host functions for modular arithmetic (add, sub, mul, pow, and inv), we'll enable more interesting BN254 use cases on Stellar.
```
jayz22: guest work isn't the issue for many add-and-multiply -- they are still calling the host functions. the main issue that the points and scalars has to be converted to internal form (for efficient operation) that conversion is relatively expensive. For MSM the conversion only happens once, and all intermediate operations all happen in internal form, until the final is returned. That saves much of the cost.

jayz22: I think the motivation for MSM and fr-arithmetic are different, and we should keep them distinct. 

For Fr-arithmetic it is enablement - contracts rely on them and they can't affort to implement in the guest side.
For MSM it is more of optimization, as mentioned above.

core/cap-0080.md
```diff
@@ -0,0 +1,218 @@
...
+**Description**: performs inversion of a BN254 scalar element (Fr).
+
+**Cost**: conversion of fr from U256 (`Bn254FrFromU256`), scalar inversion `Bn254FrInv`, and conversion back to U256 (`Bn254FrToU256`).
+
+**Error condition**: if the provided input `fr` is zero.
+
+#### New metering `CostType` introduced
+
+- `Bn254G1Msm` - Cost of performing BN254 G1 multi-scalar multiplication (MSM). Type: linear w.r.t the length of the input vectors.
+
```
jayz22, line 195: Missing "Bn254FrInv"

core/cap-0080.md
sisuresh: Good point. Updated.

core/cap-0080.md
sisuresh: `Bn254FrInv` is already in our xdr.

PR #1884, opened 2026-02-19
Add SLP-5 for raising the tx footprint limit.

dmkozh:


limits/slp-0005.md
```diff
@@ -0,0 +1,51 @@
...
+- One-click LP maintenance (Soroswap/Phoenix): reduce multi-transaction maintenance flows (fee collection, range adjustment, re-adding liquidity).
+- Routed trade reliability at scale (Soroswap/Phoenix): make heavier routes and mature pool state less brittle under current caps.
+- Volatility-adjusted collateral management (Blend): allow volatility estimation and lending actions to execute atomically on-chain, reducing reliance on off-chain keepers.
+
+
+## Evaluation
+
+### Benchmarking
+
+We're omitting benchmarking for this SLP as currently the overall footprint size per ledger is limited by the transaction sizes. With 266 KB of transactions per ledger and ~40B keys there can be at most ~6500 keys per ledger (less in practice due to transactions having other mandatory fields besides footprints). With the current limit of 200 footprint entries per transaction just 32 transactions are necessary to maximize the amount of keys per ledger, so raising the limit can't have any material impact on the amount of work done per ledger.
```
anupsdf, line 43: > With the current limit of 200 footprint entries per transaction just 32 transactions are necessary to maximize the amount of keys per ledger, so raising the limit can't have any material impact on the amount of work done per ledger.

So, with 400 footprint entries per tx, only 16 transactions (down from 32) would be necessary to maximize amount of keys per ledger. Why do we think this is not a material impact on the amount of work done per ledger?

dmkozh: As long as it's possible to include 32 transactions into ledger (which is indeed possible) there is no material impact, as it's possible to maximize the number of entries in all the footprints in both scenarios. Basically the maximum number of entries in all the footprints is `min(tx_count * max_footprint, ledger_tx_size_limit / min_key_size)`. For now we're limited by `ledger_tx_size_limit / min_key_size`, so it doesn't really matter what we have in the other part.

anupsdf: I see. Thanks for the clarification.

PR #1887, opened 2026-02-24
Add CAP-0082 - checked 256-bit integer arithmetic

jayz22:
Discussion: https://github.com/orgs/stellar/discussions/1834

### DIFF
--- a/core/cap-0082.md
+++ b/core/cap-0082.md
@@ -57,7 +57,7 @@ The following eight host functions are added to the `int` module (export `"i"`),
         { "name": "rhs", "type": "U256Val" }
     ],
     "return": "Val",
-    "docs": "Performs checked addition. Computes `lhs + rhs`, returning `Void` if overflow occurred.",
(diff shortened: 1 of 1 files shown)


core/cap-0082.md
```diff
@@ -0,0 +1,264 @@
...
+```
+{
+    "export": "H",
+    "name": "u256_checked_add",
+    "args": [
+        { "name": "lhs", "type": "U256Val" },
+        { "name": "rhs", "type": "U256Val" }
+    ],
+    "return": "Val",
+    "docs": "Performs checked addition. Computes `lhs + rhs`, returning `Void` if overflow occurred.",
```
leighmcculloch:
```suggestion
    "docs": "Performs checked addition. Computes `lhs + rhs`, returning `Void` if overflow occurred otherwise returns U256Val.",
```

jayz22: done.

PR #1895, opened 2026-03-10
Add SKILL.md and copilot-instructions.md for core CAPs

anupsdf:
Add SKILL.md and copilot-instructions.md for core CAPs.


anupsdf:
> > I guess we could try this out, but I'm generally really skeptical of pre-emptively adding skills or copilot instructions without solving a specific issue.
> 
> +1 plus, we should be experimenting with skills locally with ai tooling, because iterating is important, and then pushing to repos when we find something that works.
> 
> Is this a skill we have used or is this written but never used?

Oh. I have actually tested the SKILL before pushing. I used it for generating the changes in https://github.com/stellar/stellar-protocol/pull/1894 and then separated the SKILL.md and copilot-instructions.md into a separate GH branch so I could have a separate PR. So, its not untested. Sorry if I didn't clarify that earlier. cc. @dmkozh , @leighmcculloch

PR #1909, opened 2026-04-21
[CAP-71] Add new credentials type to use with the new signature payload.

dmkozh:
While technically this could be a separate CAP, it makes sense as a part of CAP-71, as the XDR is dependent on CAP-71 XDR.

### DIFF
--- a/core/cap-0071-02.md
+++ b/core/cap-0071-02.md
@@ -61,7 +61,7 @@ Thus, in order to reduce disruption and simplify the migration, we introduce a n
 
 ### No deprecation of `SOROBAN_CREDENTIALS_ADDRESS`
 
-The existing `SOROBAN_CREDENTIALS_ADDRESS` credential type is still valid most of the time, and as described in the previous section, the hard switch to the new preimage type would be too disruptive. Thus, we keep the existing credential type and preimage type valid, so that clients can choose to switch to the new preimage type at their own pace, and there is no rush to update clients immediately after protocol 27 upgrade.
+The existing `SOROBAN_CREDENTIALS_ADDRESS` credential type is safe for the overwhelming majority of use cases, and as described in the previous section, the hard switch to the new preimage type would be too disruptive. Thus, we keep the existing credential type and preimage type valid, so that clients can choose to switch to the new preimage type at their own pace, and there is no rush to update clients immediately after protocol 27 upgrade.
 
 However, in the future protocol (28 or later) we may want to consider deprecating the old credential type and preimage type, given that the clients will have had enough time to migrate to the new credentials and preimage type.
 
--- a/core/cap-0071-02.md
+++ b/core/cap-0071-02.md
@@ -1,0 +1,90 @@
+```
+CAP: 0071-02
+Title: Address-bound Soroban address credentials
+Working Group:
+    Owner: Dmytro Kozhevin <@dmkozh>
+    Authors: Dmytro Kozhevin <@dmkozh>
+    Consulted: 
+Status: Draft
+Created: 2026-04-27
+Discussion: https://github.com/orgs/stellar/discussions/1899
+Protocol version: 27
+```
+
+## Simple Summary
+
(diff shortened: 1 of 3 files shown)


leighmcculloch (commented):
Can this be a separate CAP? 

We've done cap dependent on cap before. That's not a big deal.

And especially because this is a security fix it's important we discuss it thoroughly and it isn't a background detail that gets lost in the noise for readers.

dmkozh:
> Can this be a separate CAP?

Maybe, but I find that quite awkward logistically, especially with the XDR feature gating. I think this highlights an issue with the feature gating approach and the idea of independently excluded CAPs in general. 

> And especially because this is a security fix it's important we discuss it thoroughly and it isn't a background detail that gets lost in the noise for readers.

We can and will still put this out for discussion, but the relevant part of the discussion is really how to handle this downstream. I'm not sure if there is much to discuss about the protocol/CAP side of the things, so it's not clear to me if it actually matters which CAP this belongs to.

core/cap-0071.md
```diff
@@ -98,20 +102,23 @@ index 9a14d6e..d59d1d5 100644
...
++    SOROBAN_CREDENTIALS_ADDRESS_WITH_DELEGATES = 3
  };
  
  union SorobanCredentials switch (SorobanCredentialsType type)
-@@ -581,6 +594,8 @@ case SOROBAN_CREDENTIALS_SOURCE_ACCOUNT:
+@@ -581,6 +595,10 @@ case SOROBAN_CREDENTIALS_SOURCE_ACCOUNT:
      void;
  case SOROBAN_CREDENTIALS_ADDRESS:
      SorobanAddressCredentials address;
++case SOROBAN_CREDENTIALS_ADDRESS_V2:
```
anupsdf: SorobanAddressCredentials already has the SCAddress in it, so why do we need a SOROBAN_CREDENTIALS_ADDRESS_V2?

anupsdf:
The XDR feature gating should be for independent feature changes so that the features that are fully tested can be enabled at their own cadence. Since `ENVELOPE_TYPE_SOROBAN_AUTHORIZATION_WITH_ADDRESS` is meant to be used for both delegated and non-delegated SorobanCredentials, they should go together behind the same feature #ifdef. 

I agree that the security part needs more open discussion. We could do a sub-CAP (like CAP-71-01) so that we have dedicated focus on the security part. Similar to the CAP-46 series we did for Soroban launch. But if its a hassle to create a sub-CAP (and both CAPs have to launch together), we could just keep it in a single CAP and just broadcast the security part more.

dmkozh:
> We could do a sub-CAP (like CAP-71-01) so that we have dedicated focus on the security part. 

I like that idea and I gave it a try, PTAL.

core/cap-0071-02.md
```diff
@@ -0,0 +1,90 @@
...
+
+Thus, in order to reduce disruption and simplify the migration, we introduce a new credential type, so that clients can explicitly choose to use the new preimage type at any point in time after protocol 27 upgrade.
+
+### No deprecation of `SOROBAN_CREDENTIALS_ADDRESS`
+
+The existing `SOROBAN_CREDENTIALS_ADDRESS` credential type is still valid most of the time, and as described in the previous section, the hard switch to the new preimage type would be too disruptive. Thus, we keep the existing credential type and preimage type valid, so that clients can choose to switch to the new preimage type at their own pace, and there is no rush to update clients immediately after protocol 27 upgrade.
```
anupsdf: > credential type is still valid most of the time

`credential type is still valid most of the time` indicates that there might be edge cases when it won't be valid. How about rephrasing to something like `credential type will remain valid for some time`?

anupsdf (commented):
LGTM. Left a small suggestion.

core/cap-0071-02.md
dmkozh: That's not the intended meaning. I've updated the wording to be more clear.

anupsdf (approved):

PR #1911, opened 2026-04-22
fix variables

JFWooten4:


leighmcculloch (approved):

leighmcculloch:
Thanks for the contribution, and the other recent many contributions as well. CAP-38 is final and so isn't receiving any edits so I excluded that change from the pull request and retained just the typo fix to SEP-51.

PR #1914, opened 2026-04-22
SLP process proposal to encourage participation from ecosystem

anupsdf:
### what
Added SLP process proposal where anyone in the ecosystem can propose changes to the network limits.

### why
To encourage participation from ecosystem. To clarify the SLP process. To empower Tier-1 organizations to govern the network.


limits/README.md
```diff
@@ -202,3 +202,152 @@ The evaluation may have several outcomes:
...
+The process to approve SLPs is a modified version of the [CAP-Process]( ../core/README.md#contribution-process ).
+From a technical standpoint, network settings are easier to understand than CAP changes but still require a well balanced panel of reviewers and participants.
+ 
+The Stellar Protocol, like most software in the world, continues to evolve over time to meet the
+needs of our network's participants and to drive technology forward into new territory. Given the
+importance of the reliability and safety of the network, we ask that all of those who have ideas
+towards pushing Stellar's protocol development forward adhere to the following:
+ 
+- Consider your idea and how it serves the fundamental [Stellar Network Goals](https://github.com/stellar/stellar-protocol/blob/master/core/README.md#stellar-network-goals) and aligns with
+  [Stellar Protocol Values](https://github.com/stellar/stellar-protocol/blob/master/core/README.md#stellar-protocol-development-values). If you cannot show how your proposal
```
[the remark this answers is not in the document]
MonsieurNicolas: suggestion is right but suggested change seems wrong (should just be `../core/` not `/blob/core`)

limits/README.md
```diff
@@ -202,3 +202,152 @@ The evaluation may have several outcomes:
 
 **Step 4 \- Nominate the proposal**
```
MonsieurNicolas: most of the existing README just get moved into an SLP-template.md file? That way, we can keep the README pretty clean and focused on process, and the template can get into the details of what is expected from the author of the SLP (performance evaluation etc)

limits/README.md
```diff
@@ -202,3 +202,152 @@ The evaluation may have several outcomes:
...
+
+## SLP Process
+These are the steps from [idea to deployment](https://www.youtube.com/watch?v=Otbml6WIQPo) on how
+to create a Stellar Limit Proposal (SLP).
+
+### Pre-SLP (Initial Discussion)
+
+Introduce your idea in one of
+* [stellar-dev mailing list](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!forum/stellar-dev)
+* [Stellar Protocol Discussion]( https://github.com/stellar/stellar-protocol/discussions)
```
MonsieurNicolas: need to discuss with tier-1 on venue -- we may want to force discussion to converge towards the mailing list or github -- Discord is great for early discussion, but lacks traceability and may not be a great fit for enterprise.

limits/README.md
```diff
@@ -202,3 +202,152 @@ The evaluation may have several outcomes:
...
+ 
+In the event of a live discussion being scheduled, as the owner of the SLP, you will be invited to share your SLP and participate in discussion during the meeting.
+You may invite any other members of your working group.
+ 
+At some point, the SLP committee:
+* votes FOR the proposal, in which case its status is modified to “Accepted”. This signals validators to organize a vote to deploy changes to the public network.
+* votes AGAINST the proposal, in which case its status is modified to “Rejected”. The SLP will not be discussed further by the committee.
+
+### Awaiting Decision -> Awaiting Evaluation
+Once the proposal is "Accepted", the Core team will evaluate the proposed numbers against
```
MonsieurNicolas: We should push the evaluation earlier in the process, and make it that the submitter has done impact analysis *before* this goes to the SLP committee as the committee members need to understand the impact before voting.

MonsieurNicolas (changes requested):
we should not merge this until we've good rough alignment with tier-1.
let's get more feedback from people --> it's also an opportunity to simplify things

also, left some comments (in addition to Copilot comments that are also relevant)

limits/README.md
```diff
@@ -202,3 +202,152 @@ The evaluation may have several outcomes:
...
+
+From there, the following process will happen.
+
+#### SLP gets merged
+If you properly followed the steps above, your PR will get merged.
+
+#### Iterating on the SLP
+ You should continue the discussion of the draft SLP at the location referenced from the “Discussion” contained in the SLP
+with an attempt at reaching consensus.
+
```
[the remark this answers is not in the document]
anupsdf: This is a new process that requires "Discussion" for SLPs proposed moving forward. The previous ones might have discussion threads but we didn't have a field back then to update.

limits/README.md
```diff
@@ -202,3 +202,152 @@ The evaluation may have several outcomes:
 
 **Step 4 \- Nominate the proposal**  
 In case if the proposal has gone through evaluation, it can be nominated for the vote by the validators. The nomination can be supported by the evaluation results summary.  
+
+# Contribution Process
```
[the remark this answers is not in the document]
anupsdf: Fixed.

limits/README.md
anupsdf: Good point. I have created a `slp-template.md` file and moved most contents from the README over there.

limits/README.md
anupsdf: Good point. Removed discord reference. Also, the `soroban-itself` channel has been archived for a while.

limits/README.md
anupsdf: Done.

limits/README.md
```diff
@@ -202,3 +208,152 @@ The evaluation may have several outcomes:
...
+* review an SLP.
+* provide feedback on SLP, especially when voting against a proposal.
+ 
+Votes for/against should be recorded in a publicly accessible location.
+ 
+This will be achieved by sending messages to the developer mailing list alongside feedback.
+ 
+### SLP Committee Members
+ 
+As the goal of the committee is to maximize the chances of success for proposals, members will be individuals representing each “tier 1” organization as calculated at the time of the decision.
```
rice2000: I think we should specify "tier 1" validator organizations.  We can also link to Obsrvr to make it clear which orgs that includes: https://radar.withobsrvr.com/

Reason: "tier 1" as a concept gets applied to a lot of different classes of orgs, not just validators.  For instance, people sometimes talk about "tier 1" issuers, or "tier 1" partners.  We want to make sure it's clear which orgs participate in the SLP committee.

limits/README.md
```diff
@@ -202,3 +208,152 @@ The evaluation may have several outcomes:
...
+  [Stellar Protocol Values](../core/README.md#stellar-protocol-development-values). If you cannot show how your proposal
+  aligns with those goals and values, it's unlikely to ever be implemented.
+- Gather feedback from discussion on the dev mailing list and other forums, and utilize it to begin
+  a draft SLP proposal.
+- Follow the proposal process listed below.
+
+## SLP Committee
+Before limits get voted on by validators, an SLP needs to be signed-off by the SLP committee.
+Committee members have the following roles and responsibilities:
+* vote for/against an SLP, taking into account alignment to Stellar Network Goals, Protocol Development Values and critical parties such as validator operators and domain experts.
```
rice2000: I think the order should be revised so it's step-wise:
- Review and SLP
- Provide feedback on an SLP, especially when voting against a proposal
- Vote for/agains and SLP...

rice2000 (commented):
Two quick notes

limits/README.md
anupsdf: Fixed. Please review,
```
As the goal of the committee is to maximize the chances of success for proposals, members will be individuals representing each “tier 1” validator organization as calculated at the time of the decision. You can see the current members of the “tier 1” validator organizations in the left pane under the [Organizations](https://radar.withobsrvr.com/) section.
```

limits/README.md
anupsdf: Good point. Ordering should be fixed now.

limits/README.md
```diff
@@ -14,191 +20,152 @@ Smart contract execution on Stellar is guarded by various resource limits that e
...
+## SLP Committee
+Before limits get voted on by validators, an SLP needs to be signed-off by the SLP committee.
+Committee members have the following roles and responsibilities:
+* review the SLP.
+* provide feedback on the SLP, especially when voting against a proposal.
+* vote for/against a SLP, taking into account alignment to [Stellar Network Goals](../core/README.md#stellar-network-goals), [Stellar Protocol Values](../core/README.md#stellar-protocol-development-values) and critical parties such as validator operators and domain experts.
+ 
+Votes for/against should be recorded in a publicly accessible location.
+ 
+This will be achieved by sending messages to the developer mailing list alongside feedback.
```
rice2000: I thought we decided to host everything in Github, including votes, so it's all in one place.  I think that might make things easier to follow.

limits/README.md
```diff
@@ -14,191 +20,152 @@ Smart contract execution on Stellar is guarded by various resource limits that e
...
+
+
+## SLP Process
+These are the steps from [idea to deployment](https://www.youtube.com/watch?v=Otbml6WIQPo) on how
+to create a Stellar Limit Proposal (SLP).
+
+### Pre-SLP (Initial Discussion)
+
+Introduce your idea in one of
+* [stellar-dev mailing list](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!forum/stellar-dev)
```
rice2000: Should we limit this to Github rather than providing the mailing list option?  Same reason as above: to keep discussions in a single place so they're easier to follow.

limits/README.md
```diff
@@ -14,191 +20,152 @@ Smart contract execution on Stellar is guarded by various resource limits that e
...
+
+
+### Awaiting Evaluation -> Awaiting Decision
+ 
+When your SLP has undergone technical evaluation by the Core team,
+you'll need to present it to the SLP Committee for review.
+ 
+For that, when you're ready, you should
+1. submit a PR changing the status in the draft to `Awaiting Decision`.
+2. send an email to the developer mailing list with a subject line starting with “SLP TBD: Ready for review”
```
rice2000: Same q: are we managing through Github rather than the mailing list?  Or is it a combo of both>

rice2000 (commented):
I tagged 3 spots for the same reason: we may want to consolidate discussion and SLP committee approval to Github rather than offering the mailing list as an option.  That way, all the relevant info is in one place, which might make it easier to follow and participate.

limits/README.md
anupsdf: Good point. I have removed dev mailing list references.

limits/README.md
anupsdf: Fixed.

limits/README.md
anupsdf: Fixed.

magofox:
Interesting proposal.  I like the idea of more transparency and quicker action, particularly around the ledger key freeze component of SLP.  

1. I am curious about how many of these SLP evaluation and votes are likely to occur in a month and the expected turn around time.  Clearly a ledger key freeze proposal would need to be executed quickly, but how about a more generic request to increase transaction limits. 
2. I'm also curious about the qualifications someone would ideally have that is representing the Tier 1 voting org.  Ledger freeze actions seem more like a jury role, where any community member can make a call based on the facts presented to them.  But a transaction limit feels very technical where hardware engineering knowhow would likely be needed.  Or is this more like an elected board member of a public utility, where the people making the decision are just normal citizens being presented cases to decide by the general manager and staff at the utility (in this case, SDF). 
3. While I agree decentralized governance is important, I cannot help but call out the elephant in the room.  Stellar already asks a lot from its Tier 1 Validators with no inherent incentive to run a validator.  We've seen a lot of Tier 1 validators come and go already, many times due to financial circumstances.  The list of T1 validators is quite small with a pretty sizable barrier to entry for new entrance - technical knowhow, available capital and human resources, and, most importantly, trust.   One way to view this proposal is placing another responsibility (requirement?) on the entities that are already the backbone of the network.  This proposal, combined with some added incentive from the network to the validators, would feel like a more balanced and rounded proposal that would incentivize more robust and sustainable participation in running and governing the network.

earrietadev:
I like the idea of trying to encourage participation, but I think that most of the projects building on the network are not part of the validators, and that's also because of what @magofox says about the lack of incentives.

So if people building aren't part of the validators, then it will be difficult to know what might be their blockers and what we could increase for that. On my side, I've made multiple contracts that are now in mainnet, and I have found that most of the time it's easier to just try to optimize or change the whole logic than to think about increasing the limits of the network, so it's going to be difficult to get feedback from builders about what to increase unless we face the issue ourselves.

That being said, I'm ok with it.

MonsieurNicolas:
answering questions from @magofox above. Thank you for helping shape this!

>     1. I am curious about how many of these SLP evaluation and votes are likely to occur in a month and the expected turn around time.  Clearly a ledger key freeze proposal would need to be executed quickly, but how about a more generic request to increase transaction limits.

emergency situations aside, based on historical votes, I would expect at most 1 vote/month. From a turnaround time point of view, the faster the better. I'd say the goal would be for the group to decide over a couple weeks, which probably means surfacing issues (if any) upstream from the informal vote. For CAPs, we try to get a few people from the group engaged during the early presentation for that reason (with a few days turnaround time), so that by the time a proposal gets to the "voting phase", there is enough data to make an informed decision.


> 
>     2. I'm also curious about the qualifications someone would ideally have that is representing the Tier 1 voting org.  Ledger freeze actions seem more like a jury role, where any community member can make a call based on the facts presented to them.  But a transaction limit feels very technical where hardware engineering knowhow would likely be needed.  Or is this more like an elected board member of a public utility, where the people making the decision are just normal citizens being presented cases to decide by the general manager and staff at the utility (in this case, SDF).

Do note that this vote already happens today, but at a later stage. The difference here is that we make it possible to *discuss* before the official on-chain vote, and have an informal vote before changes get finalized.

The author of the SLP should be the one providing as much information as possible to justify it.
The tier-1 group should have enough information to debate.

From an impact/assessment point of view, this is where having a more diverse group at that layer will help: while SDF can help with evaluating impact of technical proposals, others will likely require thinking about business impact over their org and balance that with a best guess at overall ecosystem impact.

> 
>     3. While I agree decentralized governance is important, I cannot help but call out the elephant in the room.  Stellar already asks a lot from its Tier 1 Validators with no inherent incentive to run a validator.  We've seen a lot of Tier 1 validators come and go already, many times due to financial circumstances.  The list of T1 validators is quite small with a pretty sizable barrier to entry for new entrance - technical knowhow, available capital and human resources, and, most importantly, trust.   One way to view this proposal is placing another responsibility (requirement?) on the entities that are already the backbone of the network.  This proposal, combined with some added incentive from the network to the validators, would feel like a more balanced and rounded proposal that would incentivize more robust and sustainable participation in running and governing the network.

Yeah it's an area where it's worth experimenting. Actively participating at that layer allows entities to actively protect their interests so it looks a lot more like a perk than a burden from that point of view.
The topic of "how much should it cost" to participate at that layer is a good one, is orthogonal to this proposal and should continue in a different discussion.

MattBlockdaemon:
I largely agree with what has already been raised so far, though I do have a separate concern. having a single nominated point of contact for SLPs could cause bottlenecks in the process, particularly for emergency cases.
There are not many Tier 1 organisations, and depending on the timezones they are primarily located in, this could introduce frequent gaps where not enough of the committee are available to make a fast decision.

I think higher Tier1 involvement is sensible if addressed properly with the other concerns above addressed appropriately. I would suggest that each Org is able to appoint multiple decision makers to weigh in though, to help prevent spots with no coverage. I'd also be curious to know what the protocol is for if an urgent decision is required, but those involved are unable to agree on a path forward.

magofox:
Thanks for the response, @MonsieurNicolas .  I agree, the incentive to participate can be discussed separately.  

I also agree with @MattBlockdaemon statement, that it would be smart to be able to put more than one person per validator set to help speed up the response.  Plus, just like validation, it would also be smart to allow an issue to be resolved, either pushed forward or killed, with a majority and not require 100% consensus.

rice2000:
Thanks everyone for engaging here!  I think you've raised some good topics to discuss, including participation incentives and redundant contacts to improve response capability.  We should definitely continue to discuss and take action on those things.

But also, we do need to move forward with the next SLP, which freezes the accounts involved in the YieldBlox exploit:
https://github.com/stellar/stellar-protocol/blob/master/limits/slp-0006.md

You already blocked these accounts at the overlay layer; moving that block to a network configuration is a cleaner and more sustainable approach.  I will message you with next steps shortly.

PR #1917, opened 2026-04-27
Clarify title and summary for freezing ledger entries

tomerweller:
Updated the title and summary to clarify that the feature relates to freezing ledger entries instead of keys.


tomerweller:
@copilot change the CAP index with the new title

dmkozh (approved):

PR #1918, opened 2026-04-28
SEP-41: Clarify allowance behavior

aristidesstaffieri:
## Summary

Documents how allowances behave in SEP-41 — most importantly, that approving a spender a second time replaces the previous allowance rather than adding to it.

The question that prompted this: _if I approve a spender for 40 XLM, then approve the same spender 50 XLM, is their total allowance 90 or 50?_ The answer is 50, but that is answerable today only from a security caution on developers.stellar.org — not from SEP-41, and not from CAP-46-6, which specifies the interface rather than these semantics. Discussion: stellar/stellar-protocol#1919.

The behavior is described in the doc comments of the functions it applies to, in the register [ERC-20](https://github.com/ethereum/ERCs/blob/master/ERCS/erc-20.md) uses: what a caller can do, rather than what values change internally. Each function keeps a brief one-line summary for rustdoc, with the added detail in a following paragraph.

Wording only — no new normative requirements and no interface change — so this is a patch bump to `Version 0.5.1`.

## What's in this PR

One file, `ecosystem/sep-0041.md`:

- `allowance` — returns the amount `spender` is _still_ allowed to transfer from `from`.
- `approve` — permits repeated transfers up to `amount`, and a second call overwrites the first rather than adding to it.
- `approve` notes — an allowance is a spending limit, not a reservation of balance; and zeroing-then-checking before replacing one is a client responsibility, explicitly not a contract requirement.
- `transfer_from` / `burn_from` — reduce the allowance without changing when it expires, and fail if the allowance or the balance is short.
- Preamble and changelog — `Version 0.5.1`, `Updated: 2026-08-03`, with a matching changelog entry.

## Deliberately not specified

Review feedback was that the first version of this PR risked over-specifying by reaching into implementation details. These are left to implementations:

- **Storage layout** — allowances are not described as keyed per `(from, spender)` pair, and values are not described as recorded entries.
- **Auth requirements** — no `require_auth` requirement is added to `approve`.
- **Failure atomicity** — that a failed call moves nothing follows from the host, and is not restated as a token-level rule.
- **Allowance events** — no event is emitted when an allowance is consumed. Mandating one would be an interface change rather than a clarification, so it is raised separately on #1919.
- **Expiry, in `allowance`'s return description** — the `live_until_ledger` argument doc already states that an expired entry is treated as a 0 amount allowance. Restating it on the return value adds nothing, and enumerating every case that returns 0 (never set, revoked, reset) would be noise.

## Test plan

- [x] `yarn sep-check` clean
- [x] `ecosystem/sep-0041.md` is the only file touched, rebased on current `master`
- [x] Each question raised in #1919 is answerable from the interface block alone: 40-then-50 → 50; does `transfer_from` decrement → yes; allowance vs. balance → a limit, not a reservation
- [ ] CI green

### DIFF
--- a/ecosystem/sep-0041.md
+++ b/ecosystem/sep-0041.md
@@ -178,8 +178,8 @@ address (`spender`) authorizing that spender to transfer or burn up to a
 specified `amount` of the holder's tokens, until a specified ledger number
 (`live_until_ledger`).
 
-Allowances follow the same rules as those of the Stellar Asset contract defined
-in [CAP-46-6]:
+Allowances follow the rules below. These rules match the behavior of the
+Stellar Asset Contract (see [CAP-46-6] for its interface definition):
 
 - Allowances are tracked independently for each `(from, spender)` pair. A
   single `from` may grant distinct allowances to any number of different
--- a/ecosystem/sep-0041.md
+++ b/ecosystem/sep-0041.md
@@ -6,8 +6,8 @@ Title: Soroban Token Interface
 Authors: Jonathan Jove <@jonjove>, Siddharth Suresh <@sisuresh>, Simon Chow <@chowbao>, Leigh McCulloch <@leighmcculloch>
 Status: Draft
 Created: 2023-09-22
-Updated: 2025-08-28
-Version 0.4.1
+Updated: 2026-04-28
+Version 0.4.2
 Discussion: https://discord.com/channels/897514728459468821/1159937045322547250, https://github.com/stellar/stellar-protocol/discussions/1584
 ```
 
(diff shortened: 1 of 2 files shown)


ecosystem/sep-0041.md
```diff
@@ -171,6 +171,46 @@ pub trait TokenInterface {
...
 
+### Allowances
+
+An allowance is a permission granted by a token holder (`from`) to another
+address (`spender`) authorizing that spender to transfer or burn up to a
+specified `amount` of the holder's tokens, until a specified ledger number
+(`live_until_ledger`).
+
+Allowances follow the same rules as those of the Stellar Asset contract defined
+in [CAP-46-6]:
```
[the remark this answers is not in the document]
aristidesstaffieri: wording was tweaked in 4905a3444bce187a19e4417de042f942c4ee032e

ecosystem/sep-0051.md
```diff
@@ -1021,8 +1021,8 @@ Two types of breaking changes can occur in relation to XDR-JSON:
 
 ## Changelog
 
-- `v2.0.1`: Update to `TransactionEnvelope` example to match latest
-  stellar-xdr definitions.
+- `v2.0.1`: Update to `TransactionEnvelope` example to match latest stellar-xdr
```
leighmcculloch: This defines new functionality so probably needs to be a minor version. There's no guarantee every implementer would be following the strict definition already defined.

ecosystem/sep-0041.md
```diff
@@ -6,7 +6,7 @@ Title: Soroban Token Interface
 Authors: Jonathan Jove <@jonjove>, Siddharth Suresh <@sisuresh>, Simon Chow <@chowbao>, Leigh McCulloch <@leighmcculloch>
 Status: Draft
 Created: 2023-09-22
-Updated: 2026-06-08
+Updated: 2026-07-31
 Version 0.5.0
```
leighmcculloch: Can you set version `v0.5.1`?

ecosystem/sep-0041.md
```diff
@@ -51,23 +51,37 @@ indistinguishable.
 
 ```rust
 pub trait TokenInterface {
-    /// Returns the allowance for `spender` to transfer from `from`.
+    /// Returns the amount which `spender` is still allowed to withdraw from
+    /// `from`, or 0 if the allowance has expired.
```
leighmcculloch: I think the use of the word "transfer" in the original doc is preferred, because it uses existing vocab, i.e. the interface has transfer functions, not withdrawal functions.

Suggest removing the statement about expiry as it doesn't clarify anything. The first sentence states that the value is the amount spender is allowed to transfer. If the allowance has expired, the spender cannot transfer any amount. If we are to discuss more of the details, there are other cases it returns 0, if no allowance has ever been specified, and if an allowance has been reset or removed. But I think listing them all is unnecessary.

```suggestion
    /// Returns the amount which `spender` is still allowed to transfer from
    /// `from`.
```

ecosystem/sep-0041.md
```diff
@@ -51,23 +51,37 @@ indistinguishable.
...
     /// - `from` - The address holding the balance of tokens to be drawn from.
     /// - `spender` - The address spending the tokens held by `from`.
     fn allowance(env: Env, from: Address, spender: Address) -> i128;
 
-    /// Set the allowance by `amount` for `spender` to transfer/burn from
-    /// `from`.
+    /// Allows `spender` to withdraw from `from` multiple times, up to a total
+    /// of `amount`, until `live_until_ledger`. If this function is called again
+    /// it overwrites the current allowance with `amount` and
+    /// `live_until_ledger`.
```
leighmcculloch: Suggest keeping the brief first line which is displays in rust docs concisely, and shift the explanation to a new paragraph:
```suggestion
    /// Set the allowance of `spender` to `amount`.
    ///
    /// Allows `spender` to transfer from `from` multiple times, up to a total
    /// of `amount`, until `live_until_ledger`. If this function is called again
    /// it overwrites the current allowance with `amount` and
    /// `live_until_ledger`.
```

leighmcculloch (commented):
Couple minor suggestions inline.

ecosystem/sep-0041.md
aristidesstaffieri: I set this in 4ac0756fe6fb4f359d49bccbe6ef098c7a13d41a

ecosystem/sep-0041.md
aristidesstaffieri: yeah makes sense, I updated this to use transfer and removed the expiry details in 4ac0756fe6fb4f359d49bccbe6ef098c7a13d41a

ecosystem/sep-0041.md
aristidesstaffieri: this was added in 4ac0756fe6fb4f359d49bccbe6ef098c7a13d41a

sisuresh (approved):
LGTM

leighmcculloch:
I'll merge this change in 7 days from today if there are no new feedback.

leighmcculloch (approved):

PR #1930, opened 2026-05-12
Update RWA SEP

ozgunozerk:
## Summary

After @janewang noticed we are missing `recovery_target` from this SEP, I've taken a deeper look to see if there are other discrepancies left. And yes, there were some :) 

## Changes

  - Added `token: Address` to all five hooks (`can_transfer`, `can_create`, `transferred`, `created`, `destroyed`) — required for a single compliance contract to serve multiple RWA tokens with per-token state. Also updated the simplified `can_transfer` snippet in the Architecture Overview.
  - Added the two getters that already exist in the impl: `get_modules_for_hook` and `is_module_registered`.
  - **`IdentityVerifier`**: added `recovery_target(e, old_account) -> Option<Address>`. `RWAToken::recover_balance` calls into this to confirm an identity recovery has already happened before moving balances.
  - **`IdentityRegistryStorage`**: replaced the associated-type `type CountryData: FromVal<Env, Val>` with `Vec<Val>` parameters, and added a note explaining the Soroban SDK constraint (`#[contractclient]` cannot derive on traits with associated types) that drives this choice.
  - Renamed the recovery event `Recovery` → `RecoverySuccess` to match the impl.
  - Dropped the `operator` topic from `AddressFrozen` (impl emits only `user_address` and `is_frozen`).


ozgunozerk:
Prettier check is failing due to sep 23, not this one

janewang (dismissed):

janewang:
Adding @mootz12 and @marcelosalloum for visibility

ecosystem/sep-0057.md
```diff
@@ -682,18 +698,43 @@ pub trait Compliance {
...
+   call can only originate from the token contract whose state is being
+   updated.
+2. Reject calls where `token` is not bound to the compliance contract.
+
+Without both checks, an attacker could call a state-update hook directly with a
+`token` they do not control and corrupt that token's per-module state (e.g.,
+transfer counters, holding-period trackers). Read-only hooks (`can_transfer`,
+`can_create`) cannot corrupt state, but implementations SHOULD apply the same
+checks for consistency and to avoid leaking module-internal decisions to
+untrusted callers.
```
brozorec: I wouldn't suggest adding authorizations on those getter as this data is public anyway and it's unclear how it "avoid leaking module-internal decisions"; not sure the consistency goal is viable with btw. But if there are other arguments, I might be missing, I'd replace SHOULD by MAY.

brozorec (approved):

ecosystem/sep-0057.md
```diff
@@ -6,7 +6,7 @@ Title: T-REX (Token for Regulated EXchanges)
 Author: OpenZeppelin, Boyan Barakov <@brozorec>, Özgün Özerk <@ozgunozerk>, Dennis O'Connell <@droconnel22>
 Status: Draft
 Created: 2025-11-26
-Updated: 2025-12-15
+Updated: 2026-05-12
 Version: 0.1.0
```
leighmcculloch: @ozgunozerk It's common for SEPs to update their version, in this case:
```suggestion
Version: 0.2.0
```
And to also note what changed in the changelog at the very bottom of the SEP:
https://github.com/stellar/stellar-protocol/blob/bcc47c2815abb32306a2a23905571a34c946603f/ecosystem/sep-0057.md?plain=1#L1416-L1418

leighmcculloch (changes requested):
This change updates a draft SEP without changing its version or changelog. Can those be updated?

ecosystem/sep-0057.md
ozgunozerk: Done, thanks!

leighmcculloch (approved):

PR #1933, opened 2026-05-15
SEP-0058: Contract build reproducibility

leighmcculloch:
### What
Propose SEP-0058 (Draft), defining a vocabulary for the build environment information needed to reproduce a Soroban contract's wasm bytes from source.

The vocabulary is two groups of named meta fields:

- **Build environment** — `bldimg` (digest-pinned container image) and repeating `bldopt` (shell-style flags deviating from the baseline `stellar contract build` invocation).
- **Source identification** — `source_repo`+`source_rev`, `tarball_url`, and `tarball_sha256`, in conformant combinations covering git, hosted tarballs, and content-addressed source.

The SEP is storage-agnostic (fields can live in `contractmetav0` per SEP-46, in an on-chain registry, or off-chain). An appendix demonstrates a reproducible build with `docker.io/stellar/stellar-cli`.

### Why
A contract deployed to mainnet is opaque bytes today. Anyone can rebuild from
source and compare hashes, but the build environment required to reproduce a
build (container image, toolchain, flags) isn't well defined or recorded
anywhere, so two independent verifiers have no shared way to reach the same
result.

This SEP is complementary to SEP-55 (attestation-based verification): SEP-55
asks "did a particular trusted CI compile this wasm?", while SEP-0058 provides
what's needed to actually rebuild and compare bytes. A contract can carry meta
supporting both, and verifiers pick whichever path fits their threat model.

Discussion: https://github.com/orgs/stellar/discussions/1923


graydon (approved):

ecosystem/sep-0058.md
```diff
@@ -0,0 +1,401 @@
...
+Define a shared vocabulary for the build environment information needed to
+reproduce a Soroban contract's wasm bytes from source. The vocabulary is a
+small set of named fields with stable meanings; this SEP does not prescribe how
+the fields are produced, stored, or consumed.
+
+Complementary to
+[SEP-55](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0055.md),
+which uses signed CI attestations instead of independent rebuild. The two
+address the same trust question with different trade-offs and can coexist on
+the same contract.
```
tupui, line 25: I would rephrase the bit about "same trust question". Rather it attests the provenance of the build. So right now both SEPs are really complementary and I suggest using a stronger wording that they be used in tandem.

Saw the bit in the rational and think it should read like this here as well.

ecosystem/sep-0058.md
```diff
@@ -0,0 +1,401 @@
...
+The image's default working directory (set via the Dockerfile `WORKDIR`
+instruction) MUST be `/source`, so that a contract mounted at `/source` (e.g.
+`-v "$PWD:/source"`) is the directory the stellar-cli runs in, and relative
+paths like `--manifest-path=contracts/foo/Cargo.toml` resolve against it.
+Verifiers and the build invocation in Appendix A rely on this contract.
+
+Qualities that verifiers SHOULD expect of a trustworthy image source include:
+
+- **CI provenance**: image attestations (e.g. SLSA-style) proving which CI run
+  produced the image, so the image can be traced back to a published pipeline.
```
tupui, line 125: Only mention of SLSA here, I would add a ref at least.

ecosystem/sep-0058.md
```diff
@@ -0,0 +1,401 @@
...
+version in `bldimg`. Producers need only record flags that deviate from this
+baseline; verifiers replay the recorded flags. This lets the vocabulary cover
+any flag the build command supports without the SEP enumerating them, so new
+flags require no spec revision.
+
+Note: a build is unlikely to be reproducible if `--locked` was not included in
+`bldopt`, since without it cargo may re-resolve dependencies against the
+registry instead of using the versions pinned in `Cargo.lock`. It is therefore
+reasonable, when reproducing a build, to always include `--locked` even if it
+was not present in `bldopt`.
```
tupui, line 167: I would use a stronger wording. In this context not sure it would be good to leave any space. Pin all the way.

On bldopt the important component is the version of the CLI. I am specifically concerned about breaking APIs. Which is all good as the CLI's version is pinned.

ecosystem/sep-0058.md
```diff
@@ -0,0 +1,401 @@
...
+MAY carry more than one. Verifiers MAY support any subset of these mechanisms;
+a verifier that, say, only handles `source_repo`+`source_rev` is still
+conformant for the wasms it can verify.
+
+| key              | description                                                                                                                                                                                                                                  | format regex                                   |
+| ---------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ---------------------------------------------- |
+| `source_repo`    | HTTPS URL of the source repository (typically the origin remote). The `github:<user>/<repo>` shorthand defined by SEP-55 is also accepted so consumers can read values produced under SEP-55; the HTTPS URL form is preferred for new wasms. | `^(https?://\S+&#124;github:[^/\s]+/[^/\s]+)$` |
```
tupui, line 179: Why restrict to HTTPS? At least we should allow IPFS in everything IMHO. Same on the forge side. I would rather see CIDs in general. I am afraid about specifying too much the regex and then we end up with platforms only supporting a single way with no incentive to do something else.

See some lighter wording below but that's in the rational part which implementations are not going to consider much.

ecosystem/sep-0058.md
```diff
@@ -0,0 +1,401 @@
...
+developers store and distribute their code in different ways: git, tarballs,
+content-addressable stores. Some can't expose source publicly at all (initially
+private during pre-launch review, permanently restricted to auditors, or hosted
+on something other than git/GitHub). The content-addressed `tarball_sha256`
+form covers these cases: the hash commits to specific source bytes without
+naming a retrieval channel, so an auditor with the bytes can verify while the
+public sees only the hash.
+
+**Source-repo format alignment with SEP-55.** SEP-55 defines `source_repo` as
+`github:<user>/<repo>`; this SEP prefers an HTTPS URL because it generalizes
```
tupui, line 288: Arguable, you have a URN vs URL. With naming you might need a mapping on your side with allowed values. HTTPS here is similarly restrictive as now you restrict to a given protocol. Linked to my other comment.

tupui (commented):
Thanks for putting this together. I left some suggestions.

ecosystem/sep-0058.md
leighmcculloch: This SEP isn't dependent on SEP-55, it doesn't require it, but verifiers can of course combine them. I don't think the proposal needs to create artificial dependence. The main point here is that this proposal does not attempt to replace or conflict with SEP-55.

leighmcculloch:
@tupui Thanks for the feedback. I'll open a change with the minor wording tweaks and links. The PR is a formality and temporary to get the document in the repo. Let's continue discussion at #1923 so that discussion doesn't become fragmented. Can you start sub-threads on the discussion about the scheme support (https, etc) and pinning (--locked) behaviour you are proposing?

tupui:
ok no problem. The lock part is less my alley but will definitely open one for the scheme support part 👍

PR #1946, opened 2026-06-08
Add cap-87 ML-DSA sig verification host functions

jayz22:
Related discussion: https://github.com/orgs/stellar/discussions/1915#discussioncomment-16898752


core/cap-0084.md
```diff
@@ -0,0 +1,485 @@
...
+All signature schemes currently supported by Soroban host functions — Ed25519
+(`verify_sig_ed25519`), ECDSA secp256k1 (`recover_key_ecdsa_secp256k1`) and
+ECDSA secp256r1 (`verify_sig_ecdsa_secp256r1`) — rely on the hardness of the
+elliptic curve discrete logarithm problem, an assumption that does not hold
+against a sufficiently capable quantum computer.
+
+ML-DSA (Module-Lattice-Based Digital Signature Algorithm,
+[FIPS 204]) is one of the two post-quantum signature standards finalized by NIST, and its adoption is underway across the
+signing ecosystem: NSA's CNSA 2.0 suite includes ML-DSA, with software- and
+firmware-signing transitions expected to prefer CNSA 2.0 algorithms by 2025
```
anupsdf: Since we are in 2026, we have to rephrase this from "expected... by 2025". Maybe we could say expected adoption would happen between 2025 and 2030.

core/cap-0084.md
```diff
@@ -0,0 +1,485 @@
...
+
+All signature schemes currently supported by Soroban host functions — Ed25519
+(`verify_sig_ed25519`), ECDSA secp256k1 (`recover_key_ecdsa_secp256k1`) and
+ECDSA secp256r1 (`verify_sig_ecdsa_secp256r1`) — rely on the hardness of the
+elliptic curve discrete logarithm problem, an assumption that does not hold
+against a sufficiently capable quantum computer.
+
+ML-DSA (Module-Lattice-Based Digital Signature Algorithm,
+[FIPS 204]) is one of the two post-quantum signature standards finalized by NIST, and its adoption is underway across the
+signing ecosystem: NSA's CNSA 2.0 suite includes ML-DSA, with software- and
```
anupsdf: How about putting these examples (CNSA 2.0, X.509/PKI, libraries, runtimes etc.) in separate bulletpoints for readability?

core/cap-0084.md
```diff
@@ -0,0 +1,485 @@
...
+  to create highly usable products
+- The Stellar Network should facilitate simplicity and interoperability with
+  other protocols and networks
+
+## Abstract
+
+This CAP adds three host functions to the Soroban environment's exported
+interface, `verify_sig_ml_dsa_44`, `verify_sig_ml_dsa_65` and
+`verify_sig_ml_dsa_87`, implementing the FIPS 204 external verification
+interface `ML-DSA.Verify` for the three standardized parameter sets. Each
```
anupsdf: For better maintainability, can the parameter set be an input so we can provide 1 host function instead of 3? 
SDK can expose 3 if needed. wdyt?

core/cap-0084.md
```diff
@@ -0,0 +1,485 @@
...
+  vector. The total length (1312/1952/2592 bytes) uniquely identifies the
+  parameter set.
+- **Signature** (`sigEncode`, FIPS 204 Algorithm 26): the commitment hash
+  `c_tilde` (32/48/64 bytes), the bit-packed response vector `z`, and the
+  packed hint vector `h`. Total length 2420/3309/4627 bytes.
+- **Context**: an application-chosen domain-separation byte string of length
+  0 to 255, as defined by the FIPS 204 external interface. An empty context
+  is valid and is the default for signers that do not use one.
+- **Message**: an arbitrary byte string. The full message must be provided;
+  see the design rationale for why a pre-hashed interface is not offered.
```
anupsdf: Should we provide a hazmat version that accepts pre-hashed message? To support a use case where message is very very large. Might also be good with less bandwidth over the network.

anupsdf (commented):
Thanks for creating the CAP. I left some questions/comments.

core/cap-0084.md
jayz22: fixed

core/cap-0084.md
jayz22: done

core/cap-0084.md
jayz22: No, the pre-hash version of MLDSA is a different construct than the pure MLDSA and it is different from how other schemes that takes a hash digest works (such as secp256r1).
I've put a detailed explanation in [ffe33dc](https://github.com/stellar/stellar-protocol/pull/1946/commits/ffe33dc962926d83a9198fd2be7ca0e5afd97b6e)

core/cap-0084.md
jayz22: This was explained in the`### One function per parameter set vs. a selector parameter` section. Having 3 distinct functions is better for maintainability.

core/cap-TBD.md
```diff
@@ -0,0 +1,523 @@
...
+CAP: 0087
+Title: Host functions for ML-DSA signature verification
+Working Group:
+    Owner: Jay Geng <@jayz22>
+    Authors: Jay Geng <@jayz22>
+    Consulted: Nicolas Barry <@MonsieurNicolas>, Rocky (Soundness Labs) <@gnosed>
+Status: Draft
+Created: 2026-06-05
+Discussion: https://github.com/orgs/stellar/discussions/1915
+Protocol version: 29
```
anupsdf: You can set the protocol version as TBD since that will be set after CAP is implemented. Here is the README for reference, https://github.com/stellar/stellar-protocol/blob/master/core/README.md#additional-tips

core/cap-0087.md
```diff
@@ -0,0 +1,523 @@
...
+libraries, X.509 toolchains — produce. The context string parameter is
+exposed (rather than fixed to empty) because signatures produced under a
+non-empty context can otherwise never be verified on-chain, and adding the
+parameter later would require a new set of host functions. Passing empty
+bytes yields the common default behavior.
+
+These functions take the full message, whereas
+`verify_sig_ecdsa_secp256r1` takes a 32-byte digest. This reflects a
+difference in the schemes' external interfaces, not in key binding. ECDSA
+verification operates on `H(M)` directly — the hash is external to the
```
anupsdf: Should we spell out hash of the message here instead of `H(M)` for readability?

core/cap-0084.md
anupsdf: Reading `One function per parameter set vs. a selector parameter` section, it sounds like other than metering charge and input length check, rest of the logic will be identical in the 3 functions being proposed. From code maintainability, its error prone to have almost identical logic sitting in three different functions. The user error from invalid-selector error path is not that strong of an argument IMO.

core/cap-TBD.md
jayz22: done.

core/cap-0087.md
jayz22: done.

core/cap-0084.md
jayz22: I think you are mixing implementation with the interface. The three functions will share a common inner implementation, so no code is redundantly copied across (no maintainability burden). However, the three functions expect distinct lengths on each argument. Having them share the same interface just makes things muddy - there will be a new selector argument (5 total args instead of 4) and the documentation will have to spell out all three selector values and their parameter expectations.  
The selector doesn't offer us any maintainability benefit either - if in the future a new variant is added (say ml-dsa-99), we can't extend the selector (it will be a protocol breaking change), so we have to add another function for that specific variant, there will be a two-host function (one is semi-generic with a selector, one is ml-dsa-99 alone). It would just be easier to spell out each variant from the get-go. 
I hope that clarifies it.

core/README.md
```diff
@@ -121,6 +121,7 @@
 | [CAP-0057](cap-0057.md) | State Archival Persistent Entry Eviction | Garand Tyson | Draft |
 | [CAP-0060](cap-0060.md) | Update to Wasmi register machine| Graydon Hoare | Accepted |
 | [CAP-0072](cap-0072.md) | Contract signers for Stellar accounts | Dmytro Kozhevin | Draft |
+| [CAP-0087](cap-0087.md) | Host functions for ML-DSA signature verification | Jay Geng | Draft |
```
anupsdf, line 124: PR looks good to merge. You can rename the file from TBD to cap number otherwise this link from README will fail.

jayz22: done, sorry about that

anupsdf (approved):

PR #1965, opened 2026-06-19
Add SEP-58 bldarg for arguments ahead of bldopt

leighmcculloch:
### What

Add a `bldarg` field to SEP-58. `bldarg` is an ordered list of arguments injected ahead of the existing `bldopt` entries. When no `bldarg` is present it defaults to `contract` then `build`. `bldopt` is unchanged.

### Why

The change is additive. Existing records use `bldopt` against an assumed `contract build` baseline, and keeping `bldopt` leaves those valid. `bldarg` makes that baseline explicit and overridable for builds whose leading arguments differ, and can carry positional arguments or flags. Producers that only deviate by a flag keep using `bldopt` alone, so `contract` and `build` need not be recorded.

Discussion:
- https://github.com/stellar/stellar-protocol/discussions/1923#discussioncomment-17363717


leighmcculloch:
Reworked this to be additive instead of a rename.

`bldopt` stays exactly as it was. New `bldarg` is an ordered list injected ahead of `bldopt`, defaulting to `contract` then `build` when absent. It can carry positional arguments or flags. Existing `bldopt`-only records stay valid, and producers can omit `bldarg` so `contract` and `build` need not be recorded.

---
_Generated by [Claude Code](https://claude.ai/code)_

ecosystem/sep-0058.md
```diff
@@ -152,18 +153,30 @@ Qualities that verifiers SHOULD expect of a trustworthy image source include:
 
 #### Build options
 
-`bldopt` entries describe how the build command was invoked, interpreted
-relative to a baseline default `stellar contract build` invocation for the cli
-version in `bldimg`. Producers need only record flags that deviate from this
-baseline; verifiers replay the recorded flags. This lets the vocabulary cover
-any flag the build command supports without the SEP enumerating them, so new
-flags require no spec revision.
+The build command is the `bldarg` entries followed by the `bldopt` entries,
```
fnando, line 156: Removing the "baseline default `stellar contract build`" framing here orphans a downstream reference: Appendix A still says the wasm carries "a `bldopt` entry for each flag **deviating from the baseline in §1**." With `bldopt` now defined as the literal trailing flags (not deviations from a baseline), that sentence no longer has a §1 baseline to point at. Suggest updating it to something like "a `bldopt` entry for each flag passed."

ecosystem/sep-0058.md
```diff
@@ -152,18 +153,30 @@ Qualities that verifiers SHOULD expect of a trustworthy image source include:
...
+A build is unlikely to be reproducible if `--locked` is not present in
+`bldopt`; without it cargo MAY re-resolve dependencies against the registry
+rather than the versions pinned in `Cargo.lock`.
+
+The recorded fields reconstruct the invocation as the image's entry point
+followed by the `bldarg` entries then the `bldopt` entries, each in the order
+recorded:
+
+```
+<entry point> <bldarg 0> <bldarg 1> ... <bldopt 0> <bldopt 1> ...
```
fnando, line 177: This new invocation model (`<entry point> <bldarg...> <bldopt...>`) is never demonstrated in Appendix A — the worked example relies entirely on the implicit `contract build` default and shows no `bldarg`. Since Appendix A is the concrete reference for producers/verifiers, consider either adding a `bldarg` example or noting that `contract build` is being recorded implicitly (equivalent to omitting `bldarg`).

ecosystem/sep-0058.md
```diff
@@ -93,7 +93,8 @@ the inputs a verifier needs to set up a matching environment before rebuilding.
...
+| `bldopt` | A single shell-style flag passed verbatim as one argument to the build command, as if single-quoted in a shell (e.g. `--profile=release`, `--features=foo,bar`, `--no-default-features`). For flags that take a value, the `=` form keeps the flag and its value in one argument. The entry MAY appear multiple times, once per flag. The order of entries is significant and verifiers MUST replay them in the order recorded, after the `bldarg` entries.                                                                                                                                                                                                                                                                                                                                                                                                                          | `^--[A-Za-z][A-Za-z0-9_-]*(=.+)?$`                                               |
```
fnando: This flips `bldopt` ordering from "not significant" to "significant … MUST replay in the order recorded." That's a semantic tightening of an *existing* field, so it's slightly stronger than the "additive" framing in the PR description. It's defensible (byte-identical rebuild already depends on stored order), and existing records are unaffected because stored order = replay order — but worth calling out explicitly in the changelog so implementers notice the contract change.

fnando (dismissed):
Made a few comments, but looking good! 👍

ecosystem/sep-0058.md
leighmcculloch: Ok yeah, lets revert that part of the change.

ecosystem/sep-0058.md
leighmcculloch: The appendix doesn't say that anymore as part of this PR, and I don't think it is meaningful to say and so was good to remove. In general it is implied that if you don't specify a flag for a command, then whatever default values relating to that flag will take place.

ecosystem/sep-0058.md
```diff
@@ -445,8 +463,53 @@ the recorded `bldimg`, passing every recorded field back through `--meta` so
...
+    cargo build \
+      --release \
+      --target=wasm32v1-none \
+      --manifest-path=contracts/foo/Cargo.toml \
+      --locked
+```
+
+```
+bldimg: docker.io/library/rust@sha256:...
+bldarg: cargo
```
fnando, line 489: One thing I found surprising here: we're saying on the field table that the assumed command when `bldarg` not present is `contract build`. By passing `cargo` here, wouldn't the call be `stellar cargo build`, given that entrypoint is not being overridden?

leighmcculloch: No, the image is the rust image in this example. The rust image has no entrypoint. The entrypoint is defined by the image.

fnando: Ah, got it. It's one fits all stellar-cli image. It's actually supporting other images + build arguments.

fnando (approved):

PR #1979, opened 2026-07-14
Clarify Final Comment Period status in the core/README

anupsdf:
### what 
Clarify Final Comment Period: Acceptance or Rejection in the core/README.

### why
To remove ambiguity around the status of CAPs.

### DIFF
--- a/core/README.md
+++ b/core/README.md
@@ -5,9 +5,8 @@
 ### Primary Workflow
 - **Draft** — A CAP that is currently open for consideration and actively being discussed.
 - **Awaiting Decision** — A mature and ready CAP that is ready for final deliberation by the CAP
-  Core Team. After a maximum of three meetings, a vote will take place that will set the CAP's
-  intended disposition of Acceptance or Rejection for the Final Comment Period, or go back into a
-  **Draft** state.
+  Core Team. After a maximum of three meetings, a vote will take place that will move the CAP
+  to **Final Comment Period** state, or back into a **Draft** state.
 - **Final Comment Period** — A CAP that has entered a Final Comment Period with an
   intended disposition of Acceptance or Rejection. After one week has passed, during which any new concerns should be
   addressed, the CAP will move to **Accepted** or **Rejected** according to its intended
@@ -293,9 +292,9 @@ The protocol meetings will be used to decide on next step:
 - A vote will take place among the CAP Core Team.
   - A unanimous approval from the CAP Core Team will move the CAP to `Final Comment Period` with
     an intended disposition of Acceptance.
-  - Otherwise, the CAP will be given feedback and either move to `Final Comment Period` with an
-    intended disposition of Rejection (if the majority of the CAP Core Team raises concerns) or
-    return to `Draft` (if only a minority of the CAP Core Team raises concerns).
+  - Otherwise, the CAP will be given feedback and either,
+    - move to `Final Comment Period` with an intended disposition of Rejection documenting the reason (if the majority of the CAP Core Team raises concerns), or
+    - return to `Draft` (if only a minority of the CAP Core Team raises concerns).
   - It can take upwards of 3 meetings before a disposition is reached.
 
 ### Final Comment Period -> Accepted or Rejected
--- a/core/README.md
+++ b/core/README.md
@@ -6,11 +6,12 @@
 - **Draft** — A CAP that is currently open for consideration and actively being discussed.
 - **Awaiting Decision** — A mature and ready CAP that is ready for final deliberation by the CAP
   Core Team. After a maximum of three meetings, a vote will take place that will set the CAP's
-  intended FCP disposition (**FCP: Acceptance/Rejection**) or go back into a **Draft** state.
-- **FCP: [Acceptance/Rejection]** — A CAP that has entered a Final Comment Period (FCP) with an
-  intended disposition. After one week has passed, during which any new concerns should be
-  addressed, the CAP will head towards its intended disposition [**Acceptance/Rejection**] or go
-  back into a Draft state.
+  intended disposition of Acceptance or Rejection for the Final Comment Period, or go back into a
+  **Draft** state.
+- **Final Comment Period** — A CAP that has entered a Final Comment Period with an
+  intended disposition of Acceptance or Rejection. After one week has passed, during which any new concerns should be
+  addressed, the CAP will move to **Accepted** or **Rejected** according to its intended
+  disposition, or go back into a **Draft** state.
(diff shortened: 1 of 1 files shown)


MonsieurNicolas:
I would just replace it with "Final Comment Period" as this is how we've been using in the past and how it's referenced in the document anyways

anupsdf:
> I would just replace it with "Final Comment Period" as this is how we've been using in the past and how it's referenced in the document anyways

Removed the disposition from the FCP status. Those will now have to be inferred from the discussion thread.

core/README.md
```diff
@@ -288,16 +288,16 @@ The protocol meetings will be used to decide on next step:
...
+  - Otherwise, the CAP will be given feedback and either move to `Final Comment Period` with an
     intended disposition of Rejection (if the majority of the CAP Core Team raises concerns) or
     return to `Draft` (if only a minority of the CAP Core Team raises concerns).
   - It can take upwards of 3 meetings before a disposition is reached.
 
-### Final Comment Period: [Acceptance/Rejection] -> Accepted or Rejected
+### Final Comment Period -> Accepted or Rejected
 - After a week in Final Comment Period, any major concerns that have not been previously
   addressed can be brought up. If no such concerns remain, the CAP will move to `Accepted` or
   `Rejected` according to its intended disposition.
```
JFWooten4: I propose extending the review period to two weeks for security review of conceptual bases. This supports [Nicolas' proposition](https://groups.google.com/g/stellar-dev/c/ZmlSKP1HoDc/m/yqtPYTOdAQAJ) for more outside contributors in oversight, since more time means less rush cost to dedicate some spare bandwidth. When regulators made a change like this in our industry last year, it cut costs nearly in half for processing burdens.

Also, concerns are meant to be raised during the one-week Final Comment Period, not after it. “After a week in Final Comment Period, any major concerns … can be brought up” is misleading because it implies they are raised once the week has ended. Perhaps:

```suggestion
- During the week-long Final Comment Period, anyone can bring up major concerns
  that have not been previously addressed. If no such concerns remain at the
  end of the period, the CAP will move to `Accepted` or `Rejected` according to
  its intended disposition.
```

anupsdf: Thanks for the suggestion! I have accepted it.

core/README.md
```diff
@@ -6,11 +6,12 @@
 - **Draft** — A CAP that is currently open for consideration and actively being discussed.
 - **Awaiting Decision** — A mature and ready CAP that is ready for final deliberation by the CAP
   Core Team. After a maximum of three meetings, a vote will take place that will set the CAP's
-  intended FCP disposition (**FCP: Acceptance/Rejection**) or go back into a **Draft** state.
-- **FCP: [Acceptance/Rejection]** — A CAP that has entered a Final Comment Period (FCP) with an
-  intended disposition. After one week has passed, during which any new concerns should be
-  addressed, the CAP will head towards its intended disposition [**Acceptance/Rejection**] or go
-  back into a Draft state.
+  intended disposition of Acceptance or Rejection for the Final Comment Period, or go back into a
```
MonsieurNicolas: the reference to "Acceptance or Rejection" is misleading the next state is called "Final Comment Period" or "Draft"

anupsdf: Yeah, its not very clear. The way I understood this workflow is that before rejecting a CAP, we want it to move to `Final Comment Period` for a week. Since, we don't have a sub-status field there is no way to capture the intended disposition of acceptance/rejection when a CAP is in `Final Comment Period`. 

Should we change the workflow so that rejected proposals go from `Awaiting Decision` to `Rejected` instead of parking in FCP for a week?

JFWooten4: I think it's important to give poor disposition proposals the same week to volley final arguments as popular ones. That way, we have a full history of why something failed, and we can move on from it with confidence.

I also think we should add a line explaining the edge case of moving from FCP to Draft, which has historically been around blocking changes. Like: "...back into a **Draft** state, _if, for instance, implementation depends on other new CAPs._"

anupsdf: Another option would be to add a `Notes` (or `Comments` or `Remarks`) field to the CAPs that can capture the intended disposition of acceptance or rejection for CAPs going from `Awaiting Decision` to `Final Comment Period`. This would help in both cases,
1. Capturing the rejection reason when CAPs are in FCP but headed towards Rejected, and 
2. Capturing comments when moving from FCP to Draft. 

Thoughts?

JFWooten4: I could see this as a nice [`<details>`](https://github.com/JFWooten4/stellar-protocol/blob/4823ba0be61f43ec82be816201c9b7ab146730ca/core/cap-0021.md?plain=1#L16-L155) expand box ([example](https://github.com/JFWooten4/stellar-protocol/blob/proposed-details-structure/core/cap-0021.md)). But it could leave the team with a lot of PR triage that would be more permissionlessly done in the discussion. The Public Goods Award has done a nice job with PRs that give time for anyone to suggest new additions and know they'd be considered without prejudice.

The concept is sound; just conceptually large. I think it's worth it for CAPs like 21, which take substantial off-Git discussion analysis to understand the improvements. It would also provide a direct link to the Core Team's approvals.[^1]

[^1]: I like the mailing list as a permanent written archive of voting decisions. But obviously, when they happen in their own thread, they aren't interlinked in any primary header meta discussion link. The notes approach would fix that while also authoritatively linking the indexed meeting specs.


The whole lifecycle could use disposition indications when chatter happens outside of something like a Discussion thread on GitHub.

anupsdf: Thanks for the suggestion! I like the expand box in the example you have cited. We can do that for capturing the rejection reason. 

I have simplified the top Primary Workflow section and added a note to document the reason for rejection when CAP moves from Awaiting Decision to FCP.

core/README.md
```diff
@@ -289,17 +289,20 @@ The protocol meetings will be used to decide on next step:
...
 - A vote will take place among the CAP Core Team.
-  - A unanimous approval from the CAP Core Team will put the CAP in a `Accepted` status.
-  - Otherwise, the CAP will be given feedback and head towards a `FCP: Rejected` status (if the
-    majority of the CAP raises concerns) or a `Draft` status (if only a minority of the CAP
-    raises concerns).
+  - A unanimous approval from the CAP Core Team will move the CAP to `Final Comment Period` with
+    an intended disposition of Acceptance.
+  - Otherwise, the CAP will be given feedback and either,
+    - move to `Final Comment Period` with an intended disposition of Rejection documenting the reason (if the majority of the CAP Core Team raises concerns), or
+    - return to `Draft` (if only a minority of the CAP Core Team raises concerns).
```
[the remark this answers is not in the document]
anupsdf: Done.

MonsieurNicolas (approved):

PR #1994, opened 2026-08-03
SEP-41: permit additional event topics

leighmcculloch:
### What
Clarify that SEP-41 token implementers may emit additional topics beyond those defined, and recommend that any such extension be standardized in its own SEP for interoperability.

### Why
The spec was silent on whether implementers could extend events with extra topics, leaving consumers uncertain about what to expect and discouraging interoperable extensions. Whilst I always took this to be an assumption in the ecosystem for any contract SEP, it's become apparent in conversations that was not their assumption. SEP-41 is an important enough standard this should be documented explicitly.

Note that this is not changed behaviour, but rather capturing existing behaviour already playing out in the ecosystem.

### Discussion

- https://github.com/stellar/stellar-protocol/discussions/1995

### DIFF
--- a/ecosystem/sep-0041.md
+++ b/ecosystem/sep-0041.md
@@ -313,8 +313,7 @@ and a clawback action that emits a clawback event must reduce total supply.
   no separate burn or transfer event is emitted alongside the clawback event.
 - `v0.5.0` - Document the single-value/vec and map data formats for all events,
   generalizing the form already used by `transfer` and `mint`.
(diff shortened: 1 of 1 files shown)


leighmcculloch:
Cross-posting https://github.com/orgs/stellar/discussions/1995#discussioncomment-18140608:

There's been no further discussion on this topic with only positive feedback. 

I will merge the change in 7 days unless there is further discussion.

JakeUrban (approved):

PR #1999, opened 2026-08-21
Add cap 88 initial draft

SirTyson:
Initial draft for MS resolution in closeTime.

### DIFF
--- a/core/cap-0088.md
+++ b/core/cap-0088.md
@@ -1,0 +1,357 @@
+## Preamble
+
+```
+CAP: 0088
+Title: Millisecond-Resolution Close Times
+Working Group:
+    Owner: Garand Tyson <@SirTyson>
+    Authors: Garand Tyson <@SirTyson>
+    Consulted: Marta Lokhova <@marta-lokhova>
+Status: Draft
+Created: 2026-08-21
+Discussion: https://github.com/orgs/stellar/discussions/1998
+Protocol version: TBD
+```
+
+## Simple Summary
+
+This CAP adds a millisecond-resolution close time to each ledger. Today close
+times are whole seconds, which both places a hard one second floor on the
+interval between consecutive ledgers and prevents non whole-second ledger
+targets (e.g. moving from a 5,000 ms to a 4,500 ms target).
+With millisecond resolution, consecutive ledgers may close less than
+one second apart and at arbitrary millisecond intervals. The existing
+whole-second `closeTime` field in `StellarValue` is unchanged, and every
+protocol feature that consumes time (transaction time bounds, `minSeqAge`,
+claimable balance predicates, Soroban ledger timestamps, and upgrade
+scheduling) continues to operate on whole seconds, rounding the close time
+down. Downstream consumers of `closeTime` must update their XDR definitions
+to decode the new union arms, but require no semantic changes and may ignore
+the millisecond-resolution close time entirely.
+
+## Working Group
+
+As specified in the Preamble.
+
+## Motivation
+
+The close time of a ledger must strictly increase from the previous ledger.
+Because close times are currently expressed in whole seconds, the smallest
+possible increment between two ledgers is one second, placing a hard protocol
+floor of one second on ledger close times. Whole-second resolution also
+prevents non-whole-second ledger targets: a target of 2.5 seconds cannot be
+expressed, as the close time that anchors the next ledger trigger timer is
+rounded to a whole second, causing drift and uneven ledger intervals.
+
+The intention of this change is better time resolution for intra-second block
+times, leading towards sub-second block times. With millisecond resolution,
+the protocol itself supports ledgers that close within the same whole second,
+removing close time as a constraint on ledger latency.
+
+### Goals Alignment
+
+This CAP is aligned with the following Stellar Network Goal:
+
+* The Stellar Network should run at scale and at low cost to all participants
+  of the network.
+
+## Abstract
+
+This CAP introduces two new `StellarValue` extension arms,
+`STELLAR_VALUE_SIGNED_MS` and `STELLAR_VALUE_EMPTY_TX_SET_MS`, which extend
+the existing `STELLAR_VALUE_SIGNED` and `STELLAR_VALUE_EMPTY_TX_SET` arms
+with a `closeTimeMs` field holding the close time in milliseconds since the
+Unix epoch, typed as a new `TimePointMilliseconds` typedef (the millisecond
+analogue of `TimePoint`). The existing whole-second `closeTime` field retains
+its current meaning and must equal `closeTimeMs` truncated to whole seconds, so
+consumers that only need whole-second resolution require no changes beyond
+updated XDR definitions capable of decoding the new arms.
+
+Starting with this protocol version, the effective close time of a ledger is
+`closeTimeMs`, and must strictly increase between consecutive ledgers with a
+minimum increment of one millisecond (previously one second). All existing
+time-based protocol features continue to be evaluated against the whole-second
+`closeTime` field, i.e. the effective close time rounded down to the nearest
+whole second.
+
+## Specification
+
+### XDR changes
+
+This patch of XDR changes is based on the XDR files in commit
+`bd7b53f57b6ccd23ca942bbb97c724208e3db90e` of `stellar-xdr`.
+
+```diff mddiffcheck.base=bd7b53f57b6ccd23ca942bbb97c724208e3db90e
+diff --git a/Stellar-ledger.x b/Stellar-ledger.x
+index 9acf1d7..ad6e937 100644
+--- a/Stellar-ledger.x
++++ b/Stellar-ledger.x
+@@ -15,6 +15,11 @@ enum StellarValueType
+     STELLAR_VALUE_BASIC = 0,
+     STELLAR_VALUE_SIGNED = 1,
+     STELLAR_VALUE_EMPTY_TX_SET = 2
++#ifdef MS_CLOSE_TIME
++    ,
++    STELLAR_VALUE_SIGNED_MS = 3,
++    STELLAR_VALUE_EMPTY_TX_SET_MS = 4
++#endif // MS_CLOSE_TIME
+ };
+ 
+ struct LedgerCloseValueSignature
+@@ -52,6 +57,23 @@ struct StellarValue
+             uint32 previousLedgerVersion;
+             LedgerCloseValueSignature lcValueSignature;
+         } proposedValue;
++#ifdef MS_CLOSE_TIME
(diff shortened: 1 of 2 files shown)


anupsdf (dismissed):
LGTM. Added some folks who are experts in the various areas. We can merge this and  make changes in followup PRs if needed.

core/cap-0088.md
```diff
@@ -0,0 +1,329 @@
...
+Status: Draft
+Created: 2026-08-21
+Discussion: https://github.com/orgs/stellar/discussions/1998
+Protocol version: TBD
+```
+
+## Simple Summary
+
+This CAP adds a millisecond component to the ledger close time so that
+consecutive ledgers may close less than one second apart. The existing
```
marta-lokhova: it's not the only reason, is it? with current resolution, we can't do more granular block times, e.g. go from 5000ms to 4500ms

marta-lokhova: ah I see this is discussed in the section below; you might want to briefly mention it here as well

core/cap-0088.md
```diff
@@ -0,0 +1,329 @@
...
+The intention of this change is better time resolution for intra-second block
+times, leading towards sub-second block times. With millisecond resolution,
+the protocol itself supports ledgers that close within the same whole second,
+removing close time as a constraint on ledger latency.
+
+### Goals Alignment
+
+This CAP is aligned with the following Stellar Network Goal:
+
+* The Stellar Network should run at scale and at low cost to all participants
```
marta-lokhova, line 55: not super clear where this goal came from - it's kind of vague. i'd expect to see our scalability goals from the 2026 roadmap here

core/cap-0088.md
```diff
@@ -0,0 +1,329 @@
...
+  of the network.
+
+## Abstract
+
+This CAP introduces two new `StellarValue` extension arms,
+`STELLAR_VALUE_SIGNED_MS` and `STELLAR_VALUE_EMPTY_TX_SET_MS`, which extend
+the existing `STELLAR_VALUE_SIGNED` and `STELLAR_VALUE_EMPTY_TX_SET` arms
+with a `closeTimeMs` field holding the
+millisecond component of the close time in the range `[0, 999]`. The existing
```
marta-lokhova: it's a little strange that to get total ct, the two fields need to be added. from a ux point of view, doesn't this complicate things? `closeTimeMs` is also quite a misleading name - it can be interpreted as total closeTime, causing unexpected bugs. Why not represent closeTimeMs as total time in ms?

core/cap-0088.md
```diff
@@ -0,0 +1,329 @@
...
+Once a node's LCL has reached this CAP's version, the whole-second formats
+remain acceptable only for slots at or below the LCL, which may predate the
+upgrade. For any later slot they are invalid, since no slot after the upgraded
+LCL can run a lower protocol version. See the Protocol Upgrade Transition
+section for more detail.
+
+#### Existing Time-Based Protocol Features
+
+All existing protocol features that consume time continue to be evaluated
+against the whole-second `closeTime` field, which is the effective close time
```
marta-lokhova, line 212: not necessarily a blocker for this cap, but a clarifying question: the validation rules outlined below won't necessarily remain valid once block time on the network changes.

Should we add a warning that these rules will need to re-evaluated down the road? (right now this section reads as if no action is required)

marta-lokhova (changes requested):

core/cap-0088.md
SirTyson: For CAPs, the goals section always reference one of the top level goals defined [in the overview page](https://github.com/stellar/stellar-protocol/tree/master/core#stellar-network-goals). It's usually a very short section, I don't think we want to mix SDF's OKR goals with the generic "network goals" that we use for CAPs.

core/cap-0088.md
SirTyson: I agree the interface is a little funky, but I want to be very sure that we don't break downstream systems. Many folks have expressed concern with any breaking change to closeTime, such that I think it's a strong requirement that the currently existing closeTime struct maintains the exact format it has today.

Given that we have to support a whole second closeTime moving forward, we can store the entire time at ms resolution in the new field, but we'd still need to populate the regular closeTime. IMO that's kinda sketchy, given that consensus and validators would ignore `closeTime` and just use `closeTimeMs`, but most downstream consumers would use `closeTime`. While it's not the most elegant, I'd rather keep the interface as is. This way, everyone is a consumer of `closeTime`, but only a small subset need to be a consumer of `closeTimeMs`. I think this is still the easiest way to give downstream the option to adopt the new time resolution, but not require any breaking changes or introduce additional divergence risks between validators and downstream.

core/cap-0088.md
SirTyson: > not necessarily a blocker for this cap, but a clarifying question: the validation rules outlined below won't necessarily remain valid once block time on the network changes.

Why not? The goal here is to establish some baseline such that the protocol is sound, even with hypothetical sub second ledgers. This does introduce boundary conditions, where multiple ledgers may round and be enforced as the same whole second boundary, but what exactly becomes invalid?

core/cap-0088.md
marta-lokhova: hah, interesting. the goals section seems pretty stale, it still references things like "Stellar network should support off-chain transactions, e.g. Starlight." Anyway, I think this is out of scope for your PR, but we might want to clean this up at some point... (lots of things changed since the last time these were written down)

core/cap-0088.md
marta-lokhova: Why is it problematic to use total ct? Whoever needs ms precision can use ms, and everyone else can just use the rounded-down version. That's a lot less of a footgun than the current interface: forcing downstream to deal with two different fields increases the bug surface area, and the misleading name only adds to the confusion.

core/cap-0088.md
marta-lokhova: i'm not a protocol expert, but it's not at all clear to me why it's safe to leave existing protocol features as-is while the ct granularity is changing at the consensus layer (which impacts what is valid and what is not). 

One example that comes to mind that at a minimum is confusing: user submits a tx with maxTime as rounded-down ct (since there's no ms option). Per your current proposal, overlay is supposed to accept and flood it. But since maxTime < ctMs, the tx is rejected at consensus (however flooding already consumed resources and spammed a bunch of nodes). Are you suggesting that consensus should accept invalid maxTime?

SirTyson: No, the idea is to uniformly treat all "user facing" fields that consume closeTime with whole second precision. In your example, for tx validity enforcement, we round down during consensus validity checking as well, such that overlay and consensus will either both reject or both accept the transaction uniformly.

I agree that there are many assumptions dowstream that 1 ledger == 5 seconds, such that time based fields use 5 seconds as a ledger proxy. Changing close times will inevitably break these assumptions, independent of this specific change.

My goal for this CAP and protocol change is not to make it such that we can upgrade to 400 ms ledgers tomorrow. I think there are many, many issues that will need to be addressed to achieve this in practice. I just want to specify the protocol such that we have consistent rules with time, but I see very little practical benefit (and cost in terms of complexity with downstream) in making the actual interface more complicated. Realistically in the near to medium term future, we might reduce close time or introduce an intra-second close time, but we will not get close to subsecond. If you have true subsecond ledgers, I agree it will make sense to maintain a ms component for TX bounds. If you have a 4.5 second ledger, I think rounding rules + maintaining the current second legacy resolution makes the most sense.

I worry that if we attempt to solve subsecond ledgers here, things will be very complex for consumers (without any gain for the forseeable future), and we'll probably get it wrong anyway, given that the we don't know what subsecond dowstream applications will actually look like yet. I want to do the least complex (but still safe) change to achieve this ms resolution, which I believe the rounding rules achieve.

I think I'm getting a little in the weeds here though, I can add a warning so we can move forward with the CAP.

SirTyson:
@marta-lokhova I've updated the `closeTimeMs` to be the complete time like you sugessted, and also added a new MS timepoint type for clarity. I've also added a warning about potential issues with sub-second ledger times and clarified a few of the sections.

core/cap-0088.md
```diff
@@ -0,0 +1,371 @@
...
+
+Consequently, the ledger that applies the version upgrade still carries a
+whole-second value; the first ledger to carry an `_MS` value is the ledger
+immediately following the upgrade.
+
+A node can only judge the governing protocol version for slots up to its own
+LCL+1, so when validating SCP values for other slots it must approximate the
+rule above based on the protocol version of its own LCL:
+
+* If the node's LCL is below this CAP's version, the whole-second formats are
```
marta-lokhova: are you sure it's correct to fallback to whole-second validation? Doesn't it break this check at consensus time: 
https://github.com/stellar/stellar-core/blob/100cc3816c59357df488b17972aa5e2846ead831/src/herder/HerderSCPDriver.cpp#L280

It seems we may need to relax some validity checks when dealing with very small block times.

core/cap-0088.md
marta-lokhova: Thanks. I agree right now it's too early to see what kind of user assumptions we'd break with much smaller ledgers. As long as we're communicating to the reader that these rules may need to change, I think we're good to proceed.

marta-lokhova (dismissed):
Thanks! Left one question related to validity checks (unsure if this needs resolution at the CAP level, but thought it's worth thinking through).

core/cap-0088.md
SirTyson: Ah sorry, I don't think this was super clear. This is not talking about rounding to a second, which would break the check. Rather, this is about basic message validity. As in, when we accept the `_MS` variant vs the legacy variant of the message. I'll make this more clear, we never round for SCP. I think I'll remove this from the CAP actually, as it's an implementation detail of our own SCP implementation. It's a bit of a tricky upgrade boundary where even though we will only every externalize `_MS` message types post upgrade boundary, our implementation can't always filter out non `_MS` types aggressively, but it's out of scope for the CAP.

SirTyson: Section removed

marta-lokhova (approved):

tacticalnoot:
Garand — CAP-88 is in good shape. One exact Core edge survived a fresh pass.

The CAP says an out-of-sync node must not reject future `_MS` values from its local protocol state, but #5423 `checkCloseTime()` still builds `MAX_TIME_SLIP` using the local LCL’s precision. A pre-upgrade node can therefore reject an otherwise-valid `_MS` future value in the fractional-second tail that a post-upgrade node accepts.

I’d pin one exit vector: **same future value + same wall clock => same context-free time-window decision across pre/post-upgrade observers**. Only predecessor-known protocol semantics should turn “maybe” into valid/invalid.

Would you be up for making that explicit in CAP-88 and the Core regression?

PR #2011, opened 2026-09-03
Add New CAP Proposal workflow

leighmcculloch:
### What

Add a `New CAP Proposal` workflow that mirrors `New SEP Proposal` for PRs adding a new file under `core/`, locking the PR, converting it to a draft, and pointing the author at the CAP process redirecting them to start a discussion first. Add some notes to `core/README.md` to amplify that ideas start as a discussion, and a maintainers assign the CAP number just before merge.

### Why

Maintainers have handled new CAPs this way for a long time. The core/README.md already captures much of this process, the same as SEPs, but core/README.md didn't say cleary who assigns CAP numbers. The workflow is mirrored off the workflow we've been using for SEPs. The changes to the readme don't really introduce any new process just amplify it so it's even more clearer to hopefully both humans and agents when folks use agents to prepare CAPs.

### DIFF
--- a/.github/workflows/new-cap-proposal.yml
+++ b/.github/workflows/new-cap-proposal.yml
@@ -103,7 +103,7 @@ jobs:
               '',
               `> ⚠️ **Never pre-allocate a CAP number in your PR.** Name your file \`cap-TBD.md\` and use \`TBD\` for the protocol version — a maintainer will assign the number on merge, as described in [Creating a CAP Draft](${readmeUrl}#creating-a-cap-draft).`,
               '',
-              `For details on what each CAP status (Draft, FCP, Accepted, Final, …) means, see [CAP Status Terms](${readmeUrl}#cap-status-terms).`,
+              `For details on what each CAP status (Pre-Draft, Draft, FCP, Accepted, Final, …) means, see [CAP Status Terms](${readmeUrl}#cap-status-terms).`,
               '',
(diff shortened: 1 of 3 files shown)


.github/workflows/new-cap-proposal.yml
```diff
@@ -0,0 +1,151 @@
...
+                (c) =>
+                  c.user?.login === 'github-actions[bot]' &&
+                  c.body &&
+                  c.body.includes(marker)
+              )
+            ) {
+              core.info('CAP proposal guidance already posted; nothing to do.');
+              return;
+            }
+
```
anupsdf, line 127: I think you intended to check if the Discussion field is present but missed it in this workflow file. For example, https://github.com/stellar/stellar-protocol/blob/master/core/cap-0088.md refers to https://github.com/orgs/stellar/discussions/1998 

We can add a check. Rest look good to me.

ababove20-jpg (approved):

.github/workflows/new-cap-proposal.yml
leighmcculloch: Thanks. I've updated the workflow so that it still posts the comment on every new PR, but doesn't mark it as a draft if the CAP file contains a discussion link, and posts a slightly different message showing that the first step in the list of steps, opening a discussion, is already complete.

anupsdf (approved):
LGTM! Nice to see the automatic guidance to new PRs.

PR #2013, opened 2026-09-08
SEP-57: Add muxed address event

ozgunozerk:
Adding Muxed Address Event support to SEP-57


ecosystem/sep-0057.md
```diff
@@ -501,6 +501,23 @@ pub struct Transfer {
 }
 ```
 
+When `transfer` targets a `MuxedAddress` with a muxed ID, the event data also
+carries the ID, as in SEP-41. Identity verification, compliance and freezing
+operate on the base address; the ID is only recorded in the event.
+
+```rust
+#[contractevent(topics = ["transfer"])]
+pub struct MuxedTransfer {
```
leighmcculloch, line 510: Heads up that with soroban-sdk v28+ if `to_muxed_id: None` then the result of the `MuxedTransfer` event will be identical to the `Transfer` event and we can start describing events with a single shape.

leighmcculloch (approved):

